PR #637, opened 2026-06-27
ci(pat-expiry): fail the job when a PAT-backed secret is invalid/expired

mlorentedev:
**Tier 0** of the secret-liveness hardening (follow-up to the BITACORA_PAT incident in #636's CI).

## Problem

`pat-expiry.yml` was **detection-only**: on an invalid/expired token it filed a `pat-expiry` issue and stayed **green**. That backlog reminder was silently ignored once — `BITACORA_PAT` expired, `add-to-project`/`bitacora-status` 401'd for days, and a dead value was even re-uploaded to Actions. The one run that should have caught it (2026-06-22) went `failure` without ever filing the issue (no `pat-expiry` issue was ever created), so the signal was lost entirely.

## Change

- An **already invalid/expired** token now **fails the job** (`::error` + `exit 1`), in addition to filing the dedup issue — an unmissable red signal.
- Both tokens are still probed in one run (job-level `INVALID` flag, set inside `check_token`, checked after both calls) so a single run reports every dead secret.
- **Expiring-soon** (still valid, within `WARN_DAYS`) stays a **soft** issue-only heads-up, green job — the point is to warn while the token still works.

## Verification

- `python -c 'yaml.safe_load(...)'` parses; `bash -n` + `shellcheck -s bash -S warning` clean on the embedded script.
- Live: `workflow_dispatch` on this branch (tokens were just rotated → expect green).

## Scope

Small CI hardening (<50 LOC, archived OPS-009 feature). Part of a 3-tier follow-up: Tier 1 = opt-in `sync ci` token validation; Tier 2 = migrate the board automation off the PAT to a GitHub App (so it can't expire).

### DIFF
--- a/.github/workflows/pat-expiry.yml
+++ b/.github/workflows/pat-expiry.yml
@@ -9,8 +9,14 @@ name: PAT expiry preflight
 # github-authentication-token-expiration response header, and opens or UPDATES a
 # deduplicated `pat-expiry`-labelled issue (one per secret, stable title) when a
 # token is invalid/expired or within WARN_DAYS of expiry — so a rotation reminder
-# lands in the backlog BEFORE the next release run goes red. Detection only;
-# automatic rotation is HARNESS-022/#378. See specs/OPS-009-pat-expiry-preflight/.
+# lands in the backlog BEFORE the next release run goes red.
+#
+# A token that is ALREADY invalid/expired also FAILS the job (red), not just files
+# an issue: a backlog reminder was silently ignored once (BITACORA_PAT died, the
+# board automation 401'd for days) so the alarm must be unmissable. An expiring-soon
+# token (still valid, within WARN_DAYS) stays a soft heads-up — issue only, green job
+# — because the whole point is to warn while it still works. Automatic rotation is
+# HARNESS-022/#378. See specs/OPS-009-pat-expiry-preflight/.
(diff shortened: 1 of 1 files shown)


mlorentedev:
## Live verification surfaced a deeper bug + a real finding

Dispatching this on the branch (`workflow_dispatch`) revealed the preflight had a **latent bug since it shipped**: the job has no `actions/checkout`, so the very first `gh label create` died with `fatal: not a git repository` — before any token was probed. That is why the 2026-06-22 run went red *without ever filing an issue*: the preflight has effectively **never run**. Fixed by setting `GH_REPO` (gh's default-repo env var) — second commit.

With that fixed, the run worked and **immediately caught a real dead token**:

- 🔴 `RELEASE_TOKEN` → **HTTP 401, invalid/expired** → filed #638 and failed the job (the intended fail-loud behavior, verified end-to-end).
- ✅ `BITACORA_PAT` → valid, expires in 88 days (2026-09-23).

So `RELEASE_TOKEN` (release-please → goreleaser) is also expired and needs rotating — previously masked because the preflight was dead-on-arrival. The red `workflow_dispatch` run is the feature working correctly, not a regression; `pat-expiry` runs on `schedule`/`dispatch` only, so it is not a PR check here.

Issue #766, opened 2026-08-06
BUG-040: dotf doctor proposes a pre-commit install it can never apply, and checks file existence instead of gate effectiveness


mlorentedev:
Split out of #748 (tasks 4 and 5) so PR #765 could stay atomic. #765 fixes the *runtime* half — the dispatcher now runs pre-commit gates that `pre-commit install` was blocked from installing. This issue covers the *reporting* half.

## Problem

`dotf doctor --fix` proposes a remedy it cannot apply:

```console
[FAIL] `pre-commit install` failed in /home/manu/Projects/knowledge: exit status 1
       ([ERROR] Cowardly refusing to install hooks with `core.hooksPath` set.)
```

GUARD-001's global `core.hooksPath` makes `pre-commit install` refuse, permanently. The doctor will emit this failure on every run, forever, and no amount of `--fix` will resolve it. A repair action that is structurally impossible is worse than no repair action: it trains the operator to ignore the doctor's output.

`scripts/install-precommit.sh` calls the same blocked `pre-commit install` and has the same problem.

## Deeper issue: the check proves the wrong thing

The current check asserts a hook **file exists**. That is exactly the class of assertion that was green throughout the #748 incident while nothing was being enforced — and it is the same shape as #761, where the hooks exist but cannot execute.

"The file exists", "the setting is set", "the dispatcher ran" are all satisfiable while the gate checks nothing. The only honest check fires a real violation and confirms it is blocked.

## Scope

- Remove or rewrite the `pre-commit install` fix action so the doctor stops proposing a command that cannot succeed under GUARD-001. If the state is genuinely unfixable by the doctor, say so in the output rather than emitting a FAIL with a phantom remedy.
- Replace the file-existence check with an **effectiveness** check: assert a gate actually fires. With #765 merged, the honest assertion is that the dispatcher resolves to something that runs — a repo with a `.pre-commit-config.yaml` and no local hook must route to pre-commit, not no-op.
- Apply the same treatment to `scripts/install-precommit.sh`.
- Minor, found while testing #765: `Invoke-Pester -Path tests -CI` writes `testResults.xml` to the repo root and it is **not** in `.gitignore`, so it shows up as untracked after every local Windows test run.

## Acceptance criteria

- `dotf doctor --fix` no longer emits an error whose remedy is impossible under a global `core.hooksPath`.
- `dotf doctor` reports the vault secret gate as passing once #765 is deployed, and its check would fail if the gate were genuinely inert.
[...]

Issue #748, opened 2026-08-05
BUG-036: GUARD-001's global core.hooksPath disables the vault's gitleaks pre-push gate, and doctor --fix cannot repair it


mlorentedev:
## Summary

**Secret scanning on the knowledge vault is silently inactive.** `gitleaks` never runs on push. Worse, `dotf doctor --fix`'s remedy for this exact FAIL is structurally incapable of fixing it — the fix is circular.

## Evidence chain

1. The vault runs gitleaks at the **pre-push** stage, moved off pre-commit for cost (`knowledge/.pre-commit-config.yaml:36`). Its own header states: *"BOTH stages must be installed for the pre-push hooks to fire."*

2. GUARD-001 sets a **global** `core.hooksPath`, and the vault has it set locally too:

   ```console
   $ git config --global core.hooksPath
   /home/manu/.dotfiles/git-hooks
   $ git -C ~/Projects/knowledge config --get core.hooksPath
   /home/manu/.dotfiles/git-hooks
   ```

3. `pre-commit install` refuses to run under that setting — this is `dotf doctor --fix` trying and failing:

   ```
   [FAIL] `pre-commit install` failed in /home/manu/Projects/knowledge: exit status 1
          ([ERROR] Cowardly refusing to install hooks with `core.hooksPath` set.)
   ```

4. So `knowledge/.git/hooks/pre-push` **was never created**. The directory holds only a stale `pre-commit` from an older install, itself inert — `core.hooksPath` makes git ignore `.git/hooks/` entirely.

5. The GUARD-001 dispatcher chains to the repo-local hook:

   ```sh
   # ~/.dotfiles/git-hooks/pre-push
   exec "$DISPATCH_DIR/lib/chain-local-hook.sh" pre-push "$@"
   ```

   It chains to a hook that does not exist. Nothing runs.

**Net effect:** one guard (GUARD-001's memory-sink dispatcher) silently disabled another (the vault's secret gate), and the doctor's repair path is the very command the first guard blocks.

## Proposed fix

Make `pre-commit` work **by design** under a global `core.hooksPath` rather than patching install order: have the dispatcher's `pre-push` (and `pre-commit`) detect that no repo-local hook exists but a `.pre-commit-config.yaml` declares hooks for that stage, and invoke `pre-commit run --hook-stage <stage>` directly.

[...]

Issue #670, opened 2026-07-02
SDD-038: spec-gate enforces archive-on-merge


mlorentedev:
Found during the 2026-07-02 project coherence audit.

**Problem.** 57 active specs vs 75 archived. At least ~15-20 are clearly shipped but never archived (TERM-002-remove-ghostty, WIN-006-windows-dotf-install, HARNESS-040, HARNESS-041, CLI-016-dotf-env, MEMORY-002, most of the CLI-024-secrets-* cluster). The SDD lifecycle's terminal step (archive on merge — Discipline Gate step 7) is systematically dropped under delivery speed. Archiving discipline is this repo's differential advantage over spec-kit/Kiro (which have no lifecycle at all) — it only pays if enforced.

**Scope**
- One-time sweep: archive every spec whose feature shipped (verify via linked issue/PR state).
- Extend spec-gate (or a sibling job): a PR closing an issue with an active `specs/<id>/` must archive it in the same PR or carry an explicit exemption label + rationale.
- Fix naming stragglers while sweeping (e.g. `specs/CLI-019` has no slug).

**Acceptance criteria**
- Active specs = genuinely in-flight work only.
- CI enforces archive-on-merge going forward.

Refs: pattern-spec-driven-development, AGENTS.md Discipline Gate, #665 (prior partial sweep).

mlorentedev:
Sweep landed: PR #707 archives 21 shipped specs (all CLI-024-secrets-*/OPS-017,
plus TERM-002, WIN-006, HARNESS-040, HARNESS-041, CLI-016, MEMORY-002, and
fixes CLI-019's missing slug). Leaving this open -- the archive-on-merge CI
gate is separate design work and hasn't started.

PR #767, opened 2026-08-06
feat(spec-gate): enforce archive-on-merge, the SDD lifecycle's terminal step

mlorentedev:
Refs #670. Spec: `specs/SDD-038-archive-on-merge/`.

> Deliberately **`Refs`**, not `Closes`. The one-time sweep is the follow-up PR, and that one closes the issue — which means it must archive this very spec to pass the gate introduced here. First dogfood, by construction.

## Problem

The Discipline Gate checked that a large PR *has* a spec. Nothing checked that a finished one gets **archived**, so `specs/` stopped describing what is in flight and started accumulating shipped work.

Issue #670 was written at *"57 active vs 75 archived"*. The tree today reads **45 vs 96** — the ratio never improved, it only rotated. That is the argument for fixing the gate before sweeping: a sweep with no gate starts rotting the day it merges.

The cost is not tidiness. `check-spec-gate.sh` already carries a `SPEC_FLOOR` heuristic whose own comment names this dependency — a stale active spec is an **alibi**: any large PR can touch ten lines of an unrelated shipped spec and satisfy the gate. Every un-archived spec widens that bypass surface.

## What it does

A PR that **closes** an issue must archive the active spec tracking it. Failure names the spec and prints the command:

```console
         SDD-038-archive-on-merge (#670)
       ...
         dotf spec archive SDD-038-archive-on-merge
```

## Design decisions

**Keyed on closing keywords, not on "a spec was touched."** PR #765 is the live proof this matters: it says `Refs #748` because that issue's remaining work moved to #766, so its spec must legitimately stay active. A gate keyed on spec presence would have forced a premature archive.

**Presence in the head tree, not a rename in the diff.** `git diff --numstat` only reports `specs/{ => archive}/…` when rename detection fires; the same move surfaces as delete+add when it does not. Asking the tree cannot be fooled by how git chose to render the change.

**Active specs read from base ∪ head.** Base alone misses a PR that *creates* a spec and closes its issue in one change — exactly the "created, shipped, never archived" pattern this exists to stop. An archived spec is no longer under `specs/<id>/`, so the union cannot produce a false positive.

**Runs before the LOC logic, and is not waived by `skip-sdd`.** A three-line PR can end an issue's life; and `skip-sdd` asserts "this change needs no spec", which says nothing about whether an existing spec's work is finished. Its own hatch mirrors the existing convention: `skip-archive` + a non-empty `## Archive skip rationale`.

**Enforcement is prospective, and the spec says so.** Only **16 of 44** active specs carry an `issue:` field, in three different shapes (`"owner/repo#N"`, `"repo#N"`, bare `479`). All three are parsed, plus the full-URL closing form; a spec without the field is not enforced rather than blocking a PR for a data problem its author did not create.

## Testing

19 new cases in `tests/spec-gate-archive.bats`; **45/45 green** with the 25 pre-existing gate tests.

**9 were red before the change. The other 10 assert the gate stays *silent*** — on `Refs`, on a prose mention, on `Part of`, on a cross-repo reference, on an empty body, on a spec with no frontmatter link.

Those negative tests earned their keep immediately: they caught **two `set -e` traps that made the gate fail closed for the wrong reason**.

- `grep` exits 1 on the no-match path — which for this check is the *normal* path — and under `pipefail` that aborted the entire script. Every ordinary `Refs #N` PR would have been blocked by an exit code unrelated to the rule.
- `[[ -n "$num" ]] && printf …` as a loop body's last command made the enclosing `while` exit 1 for every spec lacking an `issue:` field — 28 of 44 — which the caller, correctly written to fail closed on an unreadable tree, read as an unreadable tree.

Both would have blocked legitimate work. A gate that refuses valid PRs is not "safely strict"; it is broken in the direction that costs the most.

### Dogfooded on real data

Run against this branch, where the spec exists only here:

```console
$ SDD_PR_BODY='Closes #<N>' ./scripts/check-spec-gate.sh --base-ref origin/main --head-ref HEAD
[FAIL] ... SDD-038-archive-on-merge (#<N>)

$ SDD_PR_BODY='Refs #670' ./scripts/check-spec-gate.sh --base-ref origin/main --head-ref HEAD
[OK] Production diff 317 LOC >= threshold 50 AND spec folder touched in diff
```

That the named spec exists only on this branch is what validates the base∪head union.

Full suite: **974 → 994 tests, failures unchanged at 116** pre-existing environmental ones. `shellcheck -x` clean.

## Side effect: it found a real violation in my own PR #760

Replaying all four open PRs through the new gate surfaced that **#760 was already failing the *existing* LOC gate** (61 LOC ≥ 50, no spec) — pre-existing, unrelated to this change, and confirmed by running `origin/main`'s copy of the script against it.

My PR description there had claimed a "bug fix with obvious cause" exemption. AGENTS.md grants that under **20 lines**, not 61. Corrected: #760 now carries the `skip-sdd` label and a real `## SDD skip rationale`, and passes. The escape hatch is the right route, and the gate was right to insist.

#762 and #763 pass below threshold; #765 passes with its spec present.

## Merge order

None of the four other open PRs conflict with this one or with each other — disjoint files, all branched from `main`. This PR does not block them.

## Follow-up

The one-time sweep of the ~21 already-shipped specs, plus backfilling `issue:` frontmatter and the naming/duplicate stragglers, is the next PR; #670 is closed there.

## Summary by CodeRabbit

- **New Features**
  - Added enforcement requiring active specifications linked to closing issue references to be archived before merging.
  - Supports repository-qualified references, multiple closing references, and a documented `skip-archive` exception with rationale.
  - Preserves existing low-change-volume and SDD skip checks.

- **Documentation**
  - Added guidance on archive requirements, exemptions, and enforcement-gate failure modes.

- **Tests**
  - Added comprehensive coverage for issue parsing, repository scoping, archive validation, exemptions, and clean-pass scenarios.


mlorentedev:

Issue #668, opened 2026-07-01
dotf sync: bidirectional secrets + repo->deployed deploy, delete dotfiles-sync twins (CLI-027 PR 2)


mlorentedev:
## Context

The second half of **CLI-027** (AUDIT-007 Phase B / PR11). CLI-027 was split into two PRs:
- **PR 1 — `dotf update`** ✅ shipped in **#667** (ported `dotfiles-selfupdate.{sh,ps1}` -> Go, repointed the systemd service + Windows task, deleted 2 scripts). #496 (the umbrella CLI-027 issue) was closed on that merge.
- **PR 2 — `dotf sync`** — this issue. `dotfiles-sync.{sh,ps1}` are still live (2 of the "4 scripts" CLI-027 targets).

Spec: `specs/CLI-027-dotf-sync-update/` (already on main; covers both PRs).

## What

Build `dotf sync` in Go and delete the twins:

1. **Secrets sync** — bidirectional **newest-wins** per-file merge of `sensitive/*.secret.age` + `.secrets-audit.log` between the deployed copy (`~/.dotfiles/sensitive/`) and the repo checkout. Atomic write; **no deletion propagation** (a secret on only one side is never wiped — the ratified decision). This is already the behavior of both twins (`.sh` uses `-nt`, `.ps1` uses `LastWriteTime`); the port is faithful.
2. **repo -> deployed deploy** + `git push`. `--secrets-only` stops after step 1.
3. A `Sync` seam (fs + git) so the command layer is unit-testable with no real git/fs in CI (the escrow/`dotf update` seam idiom); production I/O covered by a live smoke.

**Deletes:** `dotfiles-sync.{sh,ps1}` + their bats/Pester tests. Guard-grep must be clean.

## Open question to resolve FIRST (blocks scoping)

The **repo -> deployed deploy model diverges by OS** and must be reconciled into one Go path:
- **Linux:** `rsync -a --delete --exclude=.git/ --exclude=sensitive/` (mirror; `~/.dotfiles` is a non-git deployed copy per ADR-005).
- **Windows:** a git-pull model (`~/.dotfiles` handled differently).

Confirm the exact Windows model against `dotfiles-sync.ps1` + `setup-windows.ps1`, then pick the unified Go model (mirror-in-Go vs the deployed-copy approach). The secrets half (step 1) is unblocked; only the deploy half needs this decision.

## Acceptance criteria

[...]

Issue #496, opened 2026-06-21
CLI-027: dotf sync / update — reconcile rsync vs git-pull models, delete 4 scripts


mlorentedev:
AUDIT-007 **Phase B / PR11**. Risk: medium. Parent: `docs/adr/audit-007-cli-convergence-state.md`.

## Scope
Build `dotf sync` / `dotf update`, reconciling the divergent models (Linux rsync vs Windows git-pull). Keep OS-scheduler shims. Delete the 4 sync/selfupdate scripts.

## Deletes
- dotfiles-sync.{sh,ps1}, dotfiles-selfupdate.{sh,ps1}

## Acceptance
- [ ] `dotf sync`/`update` unify both models; scheduler shims call `dotf`; 4 scripts deleted; guard-grep clean.

Issue #495, opened 2026-06-21
CLI-026: dotf harness — port compile-harness to harness {refresh,deploy,check}


mlorentedev:
AUDIT-007 **Phase B / PR10**. Risk: medium. Parent: `docs/adr/audit-007-cli-convergence-state.md`.

## Scope
Port `scripts/compile-harness.sh` to `dotf harness {refresh,deploy,check}`; replace the `Deploy-SkillRecord` block embedded inside `setup-windows.ps1` (the Windows twin of the harness engine).

## Deletes
- scripts/compile-harness.sh (+ the setup-windows.ps1 Deploy-SkillRecord block)

## Acceptance
- [ ] `dotf harness {refresh,deploy,check}` reproduces the HARNESS-generated blocks; setup-windows uses it; guard-grep clean.

mlorentedev:
Audit 2026-07-06 (`docs/audits/codebase-audit-2026-07-06.md`) — behaviors the Go port should fix, not carry over:

- **C21** — `compile-harness.sh:208` `skill_targets_agent` uses substring match (`*"$agent"*`), so `pi` matches `targets:[copilot]`. The Windows twin already uses a `\b` word boundary (`setup-windows.ps1:1749`) — so the first copilot-only skill makes Linux and Windows deploy different skill sets from identical records. Port with exact/word-boundary matching.
- **C23** — the whole `manifest.agents` stage (curator + AGENT-PRESENCE injection) is not ported to Windows at all (`Deploy-SkillRecord` reads only `manifest.skills`). The Go `dotf harness deploy` should be the single cross-OS implementation so this gap closes by construction.
- **C45** — `--refresh` silently reverts in-marker manual edits with no diff (setup sends its output to `/dev/null`). The port should surface what it rewrote.

Issue #462, opened 2026-06-20
DX-007: Provision Orca ADE + CLI via bootstrap (full profile, cross-OS)


mlorentedev:
## Context

Orca (Stably AI) is an MIT open-source ADE (https://github.com/stablyai/orca) used to run parallel agents in isolated git worktrees. It ships an optional `orca` CLI bundled **inside** the desktop app (`…/Orca/resources/bin/orca.cmd` on Windows), but the app does not wire that CLI onto PATH, and the bootstrap does not install Orca at all today. Result: `orca` is unavailable in a fresh shell even when the app is installed (verified: `command -v orca` fails in bash; the binary lives at `…\AppData\Local\Programs\Orca\resources\bin`).

We already script Orca *settings* tuning (DX-005 #436) and fixed the Copilot hook (DX-006 #442), but never the install / PATH provisioning. Orca is a desktop GUI app (not headless) → this belongs behind the optional `full` profile (cf. #143 DX-001, #350 OPS-008), never the base bootstrap.

## Install channels (verified, release v1.4.80)

- macOS: `brew install --cask stablyai/orca/orca` (auto-bumped each stable release).
- Arch: AUR `yay -S stably-orca-bin`.
- Debian/Ubuntu: `orca-ide_<ver>_amd64.deb` from GitHub Releases.
- Other Linux: `orca-linux.AppImage` (version-less asset name → `releases/latest/download/...`).
- Windows: `orca-windows-setup.exe` (electron-builder NSIS, silent `/S`) from GitHub Releases — NOT in winget.

App self-updates via electron-updater after first install.

## Scope

1. Cross-OS install of the Orca app behind the `full`/optional profile (brew cask / AUR / .deb / AppImage / silent `.exe`), idempotent (skip if already installed); must not fail CI when the profile is off (#350 parity).
2. PATH-wiring of the bundled `orca` CLI (`resources/bin`) via env-contract → RC files, idempotent (skip if `orca` already resolves), mirroring the existing Obsidian-CLI pattern in `setup-windows.ps1`.
3. `dotf doctor` check for Orca CLI presence + cross-OS bats/Pester parity.

## Out of scope

- Orca settings tuning (DX-005 #436) and the Copilot hook (DX-006 #442) — separate, already handled.
- Headless/CI environments — Orca is GUI desktop.

## Acceptance criteria

- [ ] `full`-profile bootstrap installs Orca idempotently on Windows + Linux (+ macOS brew cask); no-op when already present.
[...]

PR #768, opened 2026-08-06
chore(specs): archive the 22 specs whose work already shipped

mlorentedev:
Closes #670. Stacked on #767 (the gate), which must merge first.

> **Merge order matters for this one only.** #767 introduces the gate; this PR is its first dogfood. The other four open PRs are independent.

## What

Archives **22** specs: 21 whose tracked issue closed long ago, plus SDD-038 itself.

`specs/` went from **45 active / 96 archived** to **23 / 117**.

#670 was filed at *"57 active vs 75 archived"*. The tree had since drifted to 45 vs 96 — the ratio never improved, it only rotated. That is why the gate landed first: a sweep with no gate starts rotting the day it merges.

## Selection was evidence-driven, not pattern-matched

**16 specs carry an `issue:` frontmatter link.** 14 track a closed issue; **13** were archived.

**Kept active, on purpose:**

| Spec | Why it stays |
|---|---|
| `CLI-027-dotf-sync-update` | #496 is closed, but #668 is explicitly *"CLI-027 PR 2"* — the second half of the same feature. Tasks at 4/13 confirm it. |
| `CLI-026-dotf-harness-engine` | #495 still open |
| `DX-007-orca-cli-bootstrap` | #462 still open |

**Three that looked like continuations and are not.** #387, #411 and #370 reuse the ids `HARNESS-024`, `HARNESS-027` and `REFACTOR-013` for entirely different features. That is an **id-collision data defect**, not open work, so those three specs were archived and the collision is noted below.

**The other 8 have no frontmatter link**, so each was confirmed by hand: the issue title had to match the spec's subject *and* the issue had to be closed.

That check earned its keep. An earlier pass inferring links from prose had mapped `AI-022-hive-daemon-activation` to #176 — which is *"DOCS-002: Materialize glossary"*, a completely unrelated issue — and several others to **PR numbers rather than issue numbers**. Those are all left active rather than archived on a bad inference.

`AI-027-pi-path-resilient` was verified **in the code**, because its link is wrong: #426's title describes provisioning Orca, while the spec is about pi's install prefix. Both halves of its proposal did ship — `npm install -g --prefix "$HOME/.local"` in `setup-linux.sh:805`, and the doctor's pi-unreachable branch in `checks_deploy.go:243`. Archived on that evidence, not on the link.

## Dogfood

The 22nd spec is **SDD-038 itself**. This PR closes #670, so the gate that spec introduced required this PR to archive it. Verified in **both** directions:

```console
$ SDD_PR_BODY='Closes #670' ./scripts/check-spec-gate.sh --base-ref origin/feat-spec-gate-archive-on-merge --head-ref HEAD
[OK] Production diff 0 LOC < threshold 50

# control: un-archive that one folder
$ git mv specs/archive/SDD-038-archive-on-merge specs/SDD-038-archive-on-merge && git commit -q -m control
$ SDD_PR_BODY='Closes #670' ./scripts/check-spec-gate.sh ...
         SDD-038-archive-on-merge (#670)
```

Watching it fail is the point — a gate verified only in the green direction is indistinguishable from one that asserts nothing.

## Also

`specs/README.md` carried stale counts (~44 archived, ~26 active). Corrected, with the reason those counts matter: a shipped-but-active spec is an **alibi** for the gate's `SPEC_FLOOR` heuristic, since any large PR can touch ten lines of one and satisfy the Discipline Gate.

## Testing

No production code changes. Full bats suite unchanged: **878 passing, 116 pre-existing environmental failures** — identical before and after the sweep.

## Not done here

The remaining 23 active specs include ~17 that are scaffold-only (0/16 tasks, never filled) and a few data defects — the duplicate `MEMORY-001-*` pair, the `AI-022-*` id collision, and the three id collisions above. Those need a decision per spec (abandon vs fill), not a mechanical move, so they are filed separately rather than bundled here.


mlorentedev:
This PR merged into `feat-spec-gate-archive-on-merge` (branch base) instead of `main`, so it never actually closed #670 and its content stacked into #767 rather than landing independently.

To unblock #767 (its own CI broke: the Discipline Gate no longer saw an active `specs/SDD-038-archive-on-merge/` once this PR had already archived it in-branch, and a second, unrelated bug in `_closing_issue_numbers` — it scans raw PR body text with no code-fence/prose awareness — made #767 falsely register as closing #670), the squash commit from this PR (`4b8f78c`) was reverted out of `feat-spec-gate-archive-on-merge`.

The 22-spec archive sweep will re-land as a fresh PR based on `main` once #767 merges, cherry-picking this commit. That new PR will close #670.

Issue #797, opened 2026-08-07
BUG-048: validate-commit-msg pre-commit hook cannot resolve /bin/sh on Windows


mlorentedev:
## Symptom

With #794 fixed (the `dotfiles-test` hook now passes on Windows), the next hook in
the chain blocks every commit on this machine:

```
Run dotfiles tests.......................................................Passed
Validate message format..................................................Failed
- hook id: validate-commit-msg
- exit code: 1

Executable `/bin/sh` not found
```

## What is and is not the cause

- The shebang is already POSIX-portable: `.github/hooks/validate-commit-msg.sh`
  starts with `#!/usr/bin/env sh` on both `main` and the in-flight
  `fix/validate-commit-msg-accepts-scopes` branch. So this is NOT the
  scoped-Conventional-Commits work tracked separately under BUG-047.
- `sh` IS resolvable on PATH (`C:\Users\<user>\scoop\shims\sh.exe`), and Git for
  Windows ships `C:\Program Files\Git\usr\bin\sh.exe`. Yet pre-commit reports the
  literal POSIX path `/bin/sh`, which has no meaning as a Windows path.
- The sibling hook `dotfiles-test` uses `entry: ./scripts/test.sh` with
  `#!/usr/bin/env bash` and resolves fine. The difference between the two hooks is
  `sh` vs `bash`, which is the thread to pull.

## Hypothesis to confirm

`language: script` makes pre-commit parse the shebang and exec the interpreter
itself. Something in that path is normalizing `/usr/bin/env sh` to a literal
`/bin/sh` rather than resolving `sh` through PATH. Confirm before changing
anything - a blind shebang swap to `bash` would paper over it and would also
collide with the in-flight branch that owns this file.

## Impact

Blocks every `git commit` on Windows, for every session. Same class as #794, and
the third Windows hook blocker found today (after #794 and the scoped-commit one).

## Acceptance criteria

- [ ] Root cause identified, not guessed - why `sh` resolves to a literal `/bin/sh`.
- [ ] `git commit` succeeds on Windows with the commit-msg hook active (not skipped).
- [ ] The fix does not regress the hook on Linux/CI.
- [ ] A regression assertion covers the Windows invocation path.

## Coordination

`.github/hooks/validate-commit-msg.sh` is currently being edited on
`fix/validate-commit-msg-accepts-scopes` (worktree `dotfiles-wt-bug047-commit-msg`).
Sequence this after that branch lands, or fold it in there.

Refs #794

[...]

Issue #770, opened 2026-08-06
SDD-039: spec-tree hygiene — decide the ~17 scaffold-only specs, backfill issue links, and stop id collisions


mlorentedev:
Surfaced by the #670 sweep (PR #768), which archived the 22 specs whose work had shipped. The 23 that remain active are not all genuinely in flight, and the residue splits into three kinds — each needing a decision per spec, not a mechanical move, which is why it was not bundled into the sweep.

## 1. Scaffold-only specs (~17)

Created by `spec init` and never filled: `0/16` tasks ticked, `proposal.md` still carrying template placeholder text.

`AI-010-ollama-native`, `AI-021-opencode-windows-empirical-validation`, `AI-024-antigravity-oauth-verification`, `DX-001-setup-agency-profile-dry-run`, `IDEAS-002-shell-functions`, `IDEAS-004-collision-prompt`, `IDEAS-005-curl-bootstrap`, `IDEAS-006-ai-topical-discovery`, `POLISH-003-pwsh-analyzer-coverage`, `POLISH-005-linux-idempotence-ci`, `REFACTOR-006-setup-phase-decompose`, `REFACTOR-007-scripts-dir-env-contract`, `SDD-004-session-start-config`, `WIN-002-windows-smoke-sweep`, and others.

These are noise with a cost: an empty active spec is still an **alibi** for the gate's `SPEC_FLOOR` heuristic — a large PR can touch ten lines of one and satisfy the Discipline Gate. Each needs a call: `dotf spec archive --abandoned` (the intent survives in its bitácora issue), or fill it because the work is genuinely queued.

## 2. Broken or missing issue linkage

The sweep could only enforce archive-on-merge where the linkage exists, and it mostly does not:

- **Only 16 of 44 specs carried an `issue:` field** at sweep time. The gate introduced in #767 cannot match the rest, so they are silently exempt from archive-on-merge forever.
- **The field has three shapes in-tree**: `"mlorentedev/dotfiles#426"`, `"dotfiles#161"`, and a bare unquoted `479`. The gate parses all three, but the inconsistency invites more.
- **`AI-022-hive-daemon-activation` points at nothing useful** — inferring from its prose gives #176, which is *"DOCS-002: Materialize glossary"*, an unrelated issue.
- **`AI-027-pi-path-resilient` had a wrong link**: #426's title is about provisioning Orca, not pi's install prefix. It was archived in #768 on code evidence instead, but the link was never corrected.
[...]

Issue #397, opened 2026-06-16
spec-gate: git rename-compression (specs/{ => archive}/...) evades the specs/archive exclusion and false-positives SPEC_TOUCHED


mlorentedev:
## Context

`scripts/check-spec-gate.sh` excludes archived specs (`specs/archive/*`) from the production-LOC count and treats only `specs/<id>/*` (non-archive) as an "active spec touched" (`_is_active_spec_path`). Found while archiving HARNESS-023 (PR #394).

## Problem

When a PR moves a spec to archive, `git diff --numstat origin/main...HEAD` **compresses the rename** to a single path with the common prefix/suffix factored out, e.g.:

```
specs/{ => archive}/HARNESS-023-spec-init-bitacora-repo/proposal.md
```

That literal string starts with `specs/{`, **not** `specs/archive/`, so:

- `_excluded()` does NOT match `specs/archive/*` → the moved files are **counted** as production LOC (intended to be excluded).
- `_is_active_spec_path()` does NOT match `specs/archive/*` and DOES match `specs/*/*` → **`SPEC_TOUCHED=1`** (false positive: an *archived* spec reads as an *active* spec touch).

## Impact

- Benign for pure archive PRs (they pass anyway — under threshold, or via the false `SPEC_TOUCHED`). Confirmed on #391, #383, #394.
- **Latent gate-evasion**: a PR that bundles a spec-archive move with >=50 LOC of production code and **no** genuine active spec would pass the Discipline Gate, because the archived rename falsely sets `SPEC_TOUCHED=1`.

## Fix

Normalize git's `{ old => new }` rename-compression before the `case` matches (expand to the destination path, or pass `--no-renames` / post-process numstat), so `specs/archive/*` exclusion and `_is_active_spec_path` see the real destination path. Add a `tests/check-spec-gate.bats` case feeding a compressed-rename numstat line (incident -> guard).

## References

- Surfaced in: PR #394 (archive HARNESS-023) — `check-spec-gate.sh --explain` counted the moved files + reported "Spec folder touched: no" pre-commit then a false active-touch post-commit.
- Suggested bitácora ID: HARNESS-026 · Priority: P2 (latent, no active exploit) · Type: Enhancement (guard hardening)
- Touches: `scripts/check-spec-gate.sh`, `tests/check-spec-gate.bats`

PR #799, opened 2026-08-07
fix(test): test the committed tree, not the deploy mirror

### LINKED ISSUE #794: BUG-047: the pre-commit test hook tests the deployed tree, not the committed one — and cannot pass on Windows
Found while committing the AI-028 spec (#791). Blocks **every** commit to this repo on Windows.

## Symptom

`git commit` in a clean worktree created from `origin/main` fails the `dotfiles-test` pre-commit hook:

```
Run dotfiles tests.......................................................Failed
- hook id: dotfiles-test
- exit code: 1

Using custom DOTFILES_DIR: C:\Users\mlorente\.dotfiles
Testing from: C:\Users\mlorente\.dotfiles

[1/15] Script Syntax Validation
  FAIL: utils.sh not found
  FAIL: age-encrypt-decrypt.sh not found
  FAIL: install-precommit.sh not found
  FAIL: dotfiles-sync.sh not found

[2/15] Source utils.sh
  FAIL: Failed to source utils.sh
Cannot continue without utils.sh
```

The staged change was three new markdown files under `specs/` — it cannot affect script syntax. The failure is pre-existing and environmental.

## Root cause 1 — the hook tests the deploy target, not the repo

`scripts/test.sh:48-57`:

```bash
if [[ -n "$DOTFILES_DIR" ]]; then
    echo "Using custom DOTFILES_DIR: $DOTFILES_DIR"
elif [[ -d "$HOME/.dotfiles" ]]; then
    DOTFILES_DIR="$HOME/.dotfiles"
elif [[ -d "$(dirname "${BASH_SOURCE[0]}")/.." ]]; then
    DOTFILES_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
```

Per ADR-025 / `env-contract.json`, `DOTFILES_DIR` is the **deploy target** and `DOTFILES_REPO_DIR` is the **repo**. On a configured machine `DOTFILES_DIR=~/.dotfiles`, which holds only a deployed subset (`scripts/load-secrets.ps1`, `git-hooks/`, `secrets/`, `sensitive/`, a few JSON files) — hence `utils.sh not found`.

Branch 2 is equally wrong for a pre-commit: even with `DOTFILES_DIR` unset, `$HOME/.dotfiles` exists on any set-up machine, so it is still the deploy target. **Branch 3 — the repo root, the only correct value for a pre-commit hook — is unreachable on every machine that has been through setup.** A pre-commit hook that validates the deployed copy rather than the code being committed is not gating anything.

## Root cause 2 — deployment assertions inside a repo-integrity suite

Pointing it at the repo gets much further but still fails, 70 passed / 8 failed, and all 8 are the same class:

```
--- Symlinks ---
  FAIL: .zshrc: missing
  FAIL: .bashrc: missing
  FAIL: aliases.zsh: missing
  FAIL: functions.zsh: missing
  FAIL: nvm.zsh: missing
```

These assert POSIX symlinks in `\$HOME`. They are deployment state, not repo state, and they are Linux-only by design. So on Windows the hook cannot pass with *any* value of `DOTFILES_DIR`.

`test.sh` mixes two concerns that want different lifecycles: repo integrity (syntax, permissions, structure — belongs in pre-commit and CI) and deployment verification (symlinks, deployed dirs — belongs in `dotf doctor` / post-setup smoke).

## Proposed

1. Pre-commit invokes the suite against the **repo** — the git root, or `DOTFILES_REPO_DIR`. Do not let the deploy-target env var silently redirect a repo gate.
2. Split the suite (or gate by mode): pre-commit runs repo-integrity checks only; deployment assertions move to a `--deployed` mode that pre-commit does not run.
3. Deployment/symlink assertions skip on Windows rather than fail.
4. Guard it: a test asserting the hook passes in a clean worktree on a configured machine, so this cannot regress silently.

## Acceptance

- [ ] `git commit` succeeds in a clean worktree from `origin/main` on Windows with no hook bypass.
- [ ] The pre-commit run validates the working tree being committed, not `~/.dotfiles`.
- [ ] Deployment assertions still run somewhere (not deleted), just not in pre-commit.
- [ ] Linux behaviour unchanged for the repo-integrity checks.

## Impact

Until fixed, dotfiles commits on Windows require `--no-verify`, which disables the secret-detection and SDD spec-gate hooks too — the gate degrades to nothing exactly where it matters most.

## Refs

- Blocked: #791 (AI-028)
- Related: BUG-036 (pre-commit under global hooksPath), ADR-025 (path cascade / `DOTFILES_DIR` vs `DOTFILES_REPO_DIR`)

mlorentedev:
Closes #794. Spec: `specs/BUG-049-test-sh-repo-root/`.

## The bug

`scripts/test.sh` runs as the `dotfiles-test` pre-commit hook, but it honoured an ambient `$DOTFILES_DIR` and asserted against `~/.dotfiles` — the **deploy mirror** — instead of the tree being committed.

On Windows that mirror carries exactly one file under `scripts/` (`load-secrets.ps1`), so the hook died at step 2/15 and **blocked every commit on the machine**:

```
Run dotfiles tests.......................................................Failed
Using custom DOTFILES_DIR: C:\Users\mlorente\.dotfiles
[1/15] Script Syntax Validation
  FAIL: utils.sh not found
[2/15] Source utils.sh
  FAIL: Failed to source utils.sh
Cannot continue without utils.sh
```

CI never saw it because `$DOTFILES_DIR` is unset there.

## The fix

Two concepts shared one variable. They are now separate:

| | Means | Source |
|---|---|---|
| `REPO_DIR` | the tree **under test** | this script's own repo, always |
| `DOTFILES_DIR` | the **deploy environment** | read, never overwritten |

`test.sh` is not itself deployed, so running it means running it from a checkout — the sources it asserts on exist only there.

Second half of the same bug: the suite asserted POSIX-only facts unconditionally — symlink semantics, mode bits, and the `~/.zshrc` / `~/.bashrc` / `~/.zsh/*.zsh` targets `setup-windows.ps1` deliberately does not create. Those are now skipped on Windows with the same wording the installer itself prints, so the suite agrees with the installer instead of contradicting it.

## Verification

Differential — same suite, both platforms, before and after. `scripts/test.sh` *is* the test suite, so the proof is the comparison, not a new assertion file.

| | Linux (bare `ubuntu:26.04`) | Windows |
|---|---|---|
| `main` | 4 failed / 4 skipped | aborts at step 2/15 |
| this PR | **the same 4** / 6 skipped | **68 passed / 0 failed / 8 skipped, exit 0** |

The 4 Linux failures are pre-existing and identical before and after (`.zshrc`, `aliases.zsh`, `functions.zsh`, `nvm.zsh` — a bare container has no deployed shell config).

`bash -n` and `shellcheck -S warning` clean. Self-applying: the pre-commit hook runs the working-tree copy, so this PR was gated by its own fixed script (`Run dotfiles tests.......Passed`).

## Note on honesty of the assertions

Two checks in section 14/15 were previously **vacuous** — the resolution block assigned `DOTFILES_DIR` and the section then asserted the value it had just written. They are now real, which is why the Linux column gains 2 skips: in an un-provisioned container the environment genuinely has no deploy mirror. That is a fact about the machine, not a defect in the code being committed, so it is an explicit `skip` with a stated reason rather than a red that blocks a commit.

## Not in this PR

- #795 (hook stack executable bit) and `specs/BUG-047-commit-msg-hook-windows/` — the other two Windows hook blockers found the same day. Different hooks, same family.
- #797 was filed while diagnosing this and is now closed: superseded by #795.

## Spec id

Issue #794 is titled `BUG-047`, but that id was already claimed by `specs/BUG-047-commit-msg-hook-windows/` for a different problem, so this spec takes `BUG-049`. Collision class: #770.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved test and deployment-path handling to prevent validation from using the wrong repository location.
  * Added Windows-aware behavior so platform-specific checks are skipped appropriately instead of failing incorrectly.
  * Preserved existing Linux and POSIX validation behavior.
  * Deployment directory settings are now respected when supplied externally, with a safe default when omitted.

* **Documentation**
  * Added specifications, implementation tasks, and verification records covering the fix and cross-platform test results.

## Archive skip rationale

`specs/BUG-049-test-sh-repo-root/` is **not** archived in this PR, because the two halves of `check-spec-gate.sh` cannot both be satisfied by a single PR that is over the LOC threshold *and* closes its issue:

- **Discipline Gate** - a production diff >= 50 LOC requires a file under an **active** `specs/<id>/`. `_is_active_spec_path()` returns 1 for `specs/archive/*`, and `_normalize_rename_path()` deliberately collapses a rename to its destination so that "an archived move reads as an active-spec touch and dodges the gate" (#397).
- **archive-on-merge** - a PR closing an issue must archive that issue's spec, which moves it into `specs/archive/`.

Archiving therefore fails the Discipline Gate; not archiving fails archive-on-merge. Verified empirically on this branch: archiving produced `Production diff: 111 LOC (>= threshold 50) / No specs/<feature-id>/ folder touched`, and the un-archived state produces the archive-on-merge violation.

This is a composition gap between #767 (which added archive-on-merge) and the #397-hardened Discipline Gate, not a defect in this change. Filed separately; the spec is archived in a follow-up PR, matching the pre-#767 pattern (cf. #787).

The same gate run also flags `BUG-047-commit-msg-hook-windows (#794)` as an active spec against this issue. That spec belongs to a parallel session, shipped in #795, and is stale - but it is not this PR's to move.


mlorentedev:
Disposition of the five review comments. Four applied in `4a7feaa`, one declined.

**1. Mode-bit contract inconsistency — applied, inverted.** The inconsistency is real, but the fix belongs on the spec side, not the implementation. The `Script permissions` loop checks `-x` on the repo's own `scripts/*.sh`; Git Bash reports those from the index exec bit, so the loop **passes** on Windows and keeps its coverage. Guarding it would remove a working check. The spec now names the three skips exactly and states that "mode bits" meant the `create_temp_file` `600` assertion specifically.

**2. Sync acceptance criteria — applied.** AC1–AC5 ticked in `proposal.md`, and the PR task ticked in `tasks.md`.

**3. Replace "Bitácora" with "Project log" — declined.** *Bitácora* is not stray Spanish here; it is the established proper name of this user's cross-repo GitHub Project, used throughout `AGENTS.md` ("the **bitácora** GitHub Project", "**Bitácora** status reflects reality", Standing Order #8) and in the scaffolded spec template this file came from. Renaming it in one spec would break the term's link to the board, and renaming it everywhere is a separate decision. The English-only rule targets prose, not proper nouns.

**4. Red-before-green evidence — applied, and you were right.** The first pass verified differentially (suite counts before/after), which is weaker than this repo's rule for `.sh` changes. `tests/test-sh-resolution.bats` now asserts the contract directly, running the script under a decoy `DOTFILES_DIR` shaped like the Windows deploy mirror. **3 of its 4 cases are red against the pre-fix script.** The fourth — "resolves the repo root when `DOTFILES_DIR` is unset" — passes in both, because pre-fix an unset variable already fell through to the script's own parent. Recorded as a future-regression guard rather than claimed as 4/4 red.

**5. Redact developer-specific paths — applied.** Replaced with `<deploy-mirror>` and `<repo-checkout>` in `verification.md`, preserving the result. This also aligns with ADR-025 (never hardcode a literal path in a durable artifact).

The gate contradiction this PR ran into — Discipline Gate wants an *active* spec folder, archive-on-merge wants it *archived*, and a >=50 LOC PR closing its issue can satisfy neither — is filed as #800, with the empirical output from both states.

Issue #865, opened 2026-08-09
MEMORY-007: the 2026-05-26 wrap also truncated memory content — 169 lines lost from one work project


mlorentedev:
The 2026-05-26 bulk edit that wrapped every `MEMORY.md` in a YAML block scalar
(vault commit `1c216229`, the subject of #864) did not only change their shape.
**It truncated their content**, and the loss has been live for two and a half
months.

Found by the ground-truth characterization test built for #864, which compares
`Unwrap(wrapped)` against the pre-wrap original for all 23 pairs the commit
touched. The de-indent came out exact on every pair — but 23 of 23 originals
were *longer* than what the wrap preserved.

## Scale

| File | before | after wrap | lines lost |
|---|---|---|---|
| `50_work/45-development/python-sensor-sdk-platform` | 205 | 36 | **169** |
| `10_projects/resume` | — | — | 16 |
| `10_projects/kasa-provisioner` | — | — | 9 |
| the other 20 | — | — | 3 each |

The common 3-line loss is the trailing `# currentDate` section plus its date
line, which crystallize regenerates — harmless.

**`python-sensor-sdk-platform` is not harmless.** 169 lines of work-project
memory are gone: the repo→package→distribution mapping table, the branching and
release strategy, and everything after it. Today that file is 34 lines.

Sample of what the wrap dropped:

```
| `teledyne-e2v/py_gek`      | `py_gek/`      | `teledyne_pygek`      | `teledyne-pygek`      |
| `teledyne-e2v/py_toolkit`  | `py_toolkit/`  | `teledyne_pytoolkit`  | `teledyne-pytoolkit`  |
| `teledyne-e2v/py_hydra3d`  | `py_hydra3d/`  | `teledyne_pyhydra3d`  | `teledyne-pyhydra3d`  |

## Branching & Release Strategy
- **GitHub Flow** — feature branches → squash merge PR → `main`. No long-lived branches.
- **One branch = one PR.** After merge, branch dies (auto-deleted by GitHub).
...
```

## It is fully recoverable

The vault is a git repository and `1c216229^` holds every original intact. This
is a restore, not a reconstruction.

## Scope

- [ ] Restore the dropped content for `python-sensor-sdk-platform` from
      `1c216229^`, merged with everything written since May rather than
      overwriting it — the file has 34 lines of newer content that must survive.
- [ ] Audit the other 22 for anything beyond the regenerable `# currentDate`
[...]

Issue #864, opened 2026-08-09
MEMORY-006: migrate the 17 YAML-wrapped MEMORY.md files back to plain markdown


mlorentedev:
17 auto-memory `MEMORY.md` files store their body inside a YAML block scalar
(`content: |`) instead of plain markdown. This is invalid state, not a supported
shape, and it is the reason crystallize had to be taught to refuse them (#862).

## Why this is invalid, not a convention

| Fact | Source |
|---|---|
| The shape is explicitly forbidden | `00_meta/templates/agent-memory.md`: *"Plain-markdown auto-memory — never a `content: \|` YAML block"* (vault `de0d5773`, 2026-06-20) |
| Nothing we ship emits it | `cli/internal/vault/templates/vault-memory.md` writes plain markdown; so do both crystallize twins |
| It was one accident, not a policy | all 17 wrapped in a single vault commit, `1c216229` — `2026-05-26 21:17:41` |
| It is not recurring | no file has been wrapped since; the last event is 2.5 months old |

The rule landed **three and a half weeks after** the damage. Someone diagnosed
this and wrote the guard into the template, but the already-wrapped files were
never migrated — so the rule protects new files while the existing breakage stays
live. This ticket is that missing half.

## Why the CLI must not learn to edit the shape instead

Measured before proposing (evidence in
`specs/CLI-021-dotf-vault-build-knowledge/evidence-yaml-roundtrip.md`): no
`go.yaml.in/yaml/v3` roundtrip can edit these files losslessly.

| File | in → out | `---` kept | `\|` style kept | hard-break lines |
|---|---|---|---|---|
| pollex | 10055 → 9769 | no | no | 5 → **0** |
| hive | 5294 → 5249 | no | no | 4 → **0** |
| kubelab | 10897 → 10659 | no | no | 4 → **0** |
| garsync | 1026 → 977 | no | no | 0 → 0 |

A YAML literal block scalar cannot carry trailing whitespace round-trippably — on
re-parse it is indistinguishable from indentation — so a conforming emitter drops
it. But **two trailing spaces are the markdown hard break** the handoff
convention mandates so the `Last task` / `Decisions` / `Open threads` /
`Next action` fields render as separate lines. Round-tripping silently reflows
every handoff block into one run-on paragraph.

## Scope

- [ ] Migrate all 17 files to the `agent-memory.md` template shape: frontmatter
      preserved, body de-indented and emitted as plain markdown after it.
[...]

Issue #857, opened 2026-08-08
BUG-062: crystallize corrupts YAML-wrapped MEMORY.md — #851 fixed the plain-markdown case only


mlorentedev:
> **Prior art — read #849 first.** [#849](https://github.com/mlorentedev/dotfiles/issues/849) (BUG-059) reported this same defect from `pollex` at 17:35:06Z, **13 minutes before #851 merged**, and is still open. I filed this issue without finding it. So the YAML shape was not an unknown second shape — it was a *reported* one, on the board, that the fix did not cover and review did not connect. See the reconciliation note at the end.

Follow-up to #850, which #851 closed. The invariant #850 was filed to protect is **still violated** on YAML-wrapped `MEMORY.md` files by the commit that closed it, and on that shape the fix is a no-op rather than a partial improvement.

Verified against `origin/main` at `dbe91db` — the merged code, not the branch.

## Why #851 does not cover this

`~/.claude/projects/*/memory/MEMORY.md` files are YAML documents whose entire body lives inside a `content: |` block scalar, so every markdown line carries four leading spaces:

```yaml
---
id: "hive-memory"
type: memory
content: |
    # hive - Project Memory
    ## Last Crystallized: 2026-08-08
    ## Session Handoff
```

Every anchor in `knowledge-crystallize.sh` is `^## …`, which cannot match an indented line. `#851` added `append_before_handoff`, whose guard is `grep -q '^## Session Handoff'`. On this shape that returns **zero matches**, so the guard never fires and control falls through to the `else` branch — the original bare `>>` append.

```console
$ grep -c '^## Session Handoff' MEMORY.md
0
$ grep -n 'Session Handoff' MEMORY.md
25:    ## Session Handoff
```

The fix inherits the exact anchoring assumption that caused the defect.

## Three defects, not one

1. **The HARNESS-029 invariant is still broken.** Both stamps land after `## Session Handoff`, which is what #850 reported.
2. **Duplicate stamp, and the real one goes stale.** `stamp_last_crystallized` matches `^## Last Crystallized:`, misses the indented stamp, reports no duplicates, and appends a second one at column 0. The genuine stamp keeps its old date — so the file claims two different crystallization dates and the authoritative one is wrong.
[...]

PR #866, opened 2026-08-09
feat(doctor): migrate YAML-wrapped MEMORY.md files back to plain markdown

mlorentedev:
17 auto-memory `MEMORY.md` files held their body inside a YAML block scalar
(`content: |`). Because crystallize anchors every marker at column 0, those files
silently failed to crystallize — the defect reported in #857 and stopped, not
fixed, by #862's refusal guard.

This adds a `dotf doctor` check that detects the shape and migrates it back to
plain markdown under `--fix`, and runs that migration.

## Why the shape is removed rather than supported

| Fact | Source |
|---|---|
| Explicitly forbidden | `00_meta/templates/agent-memory.md`: *"Plain-markdown auto-memory — never a `content: \|` YAML block"* (vault `de0d5773`, 2026-06-20) |
| Nothing here emits it | `cli/internal/vault/templates/vault-memory.md` and both crystallize twins write plain markdown |
| One accident, not a policy | all 17 wrapped in a single vault commit, `1c216229`, `2026-05-26 21:17:41` |
| Not recurring | nothing has wrapped a file in the 2.5 months since |

The template rule postdates the damage by three and a half weeks: the guard
against *new* violations shipped, the files already broken were never migrated.
This closes that half.

Teaching the CLI to edit the shape instead was measured and rejected. No
`yaml.v3` roundtrip can do it losslessly — a literal block scalar cannot carry
trailing whitespace round-trippably, so a conforming emitter drops it, and two
trailing spaces are the markdown hard break the handoff convention needs
(measured: pollex 5 hard-break lines → 0, hive 4 → 0, kubelab 4 → 0). Evidence in
`specs/CLI-021-dotf-vault-build-knowledge/evidence-yaml-roundtrip.md`.

## A doctor check, not a one-shot script

Running it once is the migration; leaving it installed is the guard against
recurrence. One artifact, both halves of incident→guard — which is exactly what
was missed in May. It follows `checkAutoMemoryLink`'s contract: verify always,
repair only under `--fix`, idempotent, never rewrite what it does not understand.

## The de-indent, and why the YAML-correct answer is wrong

YAML sets a block scalar's indent from its first non-empty line. But the wrapper
emitted the first body line one level shallower than every following line (4 vs
6) in **16 of the 17** files. Stopping at the YAML-correct width leaves the body
at column 2 — where the markers still do not match. That would have shipped a
migration producing files that parse cleanly and still fail to crystallize.

So two widths come off, both **derived**, no literal anywhere: the block indent,
and the uniform residual every later line carries. A single line at the block
indent forces the residual to 0, so nothing is over-stripped; `hive`, the one
uniformly-indented file, exercises that path.

## Characterization against ground truth

Vault commit `1c216229` holds the wrapped side and its parent the original — 23
real before/after pairs authored by neither this code nor its author.

That corpus is **private and stays out of this public repo**. The test reads it
from the vault, skips cleanly where absent (including CI), and fails loudly if it
verifies zero pairs so a skip can never read as a pass. Committed fixtures
reproduce the shapes for CI.

Verified the tests fire, by mutation: forcing the residual to 0 turns the two
relevant synthetic cases **and all 23 ground-truth pairs** red, while the
residual-0 cases correctly stay green — which matters because CI has no vault and
had to catch it alone.

## Result of the live migration

```console
$ dotf doctor            # after
[Auto-memory file shape]
  (1 checks, all ok)

$ ./scripts/knowledge-crystallize.sh ~/Projects/pollex
[INFO]    Updated Last Crystallized to 2026-08-08
[SUCCESS] MEMORY.md line count: 108 / 150
```

Proven by effect rather than inspection: crystallize now succeeds where it
refused, and the HARNESS-029 invariant survives — the stamp lands at line 99,
`# currentDate` at 101, and `## Session Handoff` is still the final section at
103. Hard breaks preserved exactly on every file (hive 4→4, kubelab 5→5,
pollex 5→5); column-0 headers went 0→N everywhere.

## Two findings that came out of doing this

- **The May wrap also truncated content**, dropping each file's trailing
  `# currentDate` section — and 169 lines from one work project. Filed as
  **#865**; it is a content restore, independent of this shape migration.
- **Two project keys can alias one vault file** (a rename left the old key
  behind). The first live run misreported that as a shape failure. Fixed in
  scope, deduplicating by `EvalSymlinks`, and pinned by a test.

Refs #864

## Summary by CodeRabbit

* **New Features**
  * Added a health check that detects YAML-wrapped memory files.
  * Added an optional fix mode to safely restore affected files to plain Markdown while preserving formatting and permissions.
  * Handles symlinks, duplicate file references, concurrent changes, and clear failure reporting.

* **Documentation**
  * Added specifications, verification details, and guidance for migrating existing invalid memory files.

* **Tests**
  * Added comprehensive coverage for detection, migration, formatting preservation, idempotence, and real-world validation.


mlorentedev:
Triage of the 8 comments. Seven applied — one of them found a real bug I had
missed, and another led to a finding worth more than the comment itself.

### Applied

**1. `IsWrapped` must stop at the frontmatter terminator** — this was a genuine
false positive and the most valuable comment in the review. A *migrated* file
also opens with `---` and closes its frontmatter with `---`; its markdown body
may legitimately hold a column-0 `note: |` or a fenced yaml example containing
`content: |`. Scanning past the terminator would re-flag this package's own
output forever — `doctor` FAILing on it and `Unwrap` refusing it. Detection is
now bounded, and `IsWrapped`/`Unwrap` share one `openerLine` helper so they
cannot disagree about what counts as wrapped. Two fixtures pin it.

**2. `tmp.Sync()` before rename** — correct. `os.Rename` gives atomic
*visibility*, not durability, and this write replaces the only copy of a
project's session continuity. Added, with the reasoning at the call site.

**3. errcheck** — fixed; both discards are explicit and commented.

**4. Assert the expected corpus size** — right that "more than zero" was too
weak. `wantPairs = 23`, asserted exactly, so a regression in detection cannot
hide behind a single surviving pass.

**5. Do not trim before the prefix comparison** — correct, and **what the trim
hid was worse than supposed**. The May wrap did not merely drop trailing
sections: it truncated **mid-line**. `kasa-provisioner`'s wrapped body ends at
`|`, one byte of the table separator `|-------|--------|`; a second file kept 1
byte of 61. Trimming trailing newlines and prefix-matching the whole string
absorbed that as clean.

The contract is now: every **complete** recovered line matches byte-for-byte,
and only the final line may be a prefix of its counterpart. Both truncation
modes are now reported rather than absorbed. Blank-line fidelity — the drift the
comment was actually about — is pinned separately on the synthetic fixtures,
where nothing was truncated and equality is the honest assertion. Reported to
**#865**, since it changes that restore from a concatenation into per-file
surgery.

**6. Corpus counts stated consistently** — three numbers were being used
interchangeably. They answer different questions and are now tabulated in
`tasks.md` and used consistently: **17** project keys, **16** distinct files
(two keys alias one vault directory after a rename), **23** historical pairs at
`1c216229` (includes entries since archived, and files unwrapped by hand later).

**7. `tasks.md` placeholder text** — correct, it was never filled. Replaced with
the actual TDD sequence, ticked to reflect real state.

**8. AC3 wording vs the implemented comparison** — corrected in `proposal.md` to
the surviving-prefix contract, with the unrecoverable remainder pointed at #865.

### Partially declined, with reason

**"Assert status tags rather than prose"** — applied where the tag is the point;
kept prose where the prose *is* the contract. `rep.Failures()` counts are already
asserted. But two assertions deliberately match text: that the failure message
tells the user **what to run** (`dotf doctor --fix`), and that it **names the
offending path**. Those are the check's user-facing obligations, not incidental
wording — a version that failed silently on both would still emit `[FAIL]` and
pass a tag-only assertion. If the wording changes, the test should be updated
along with it, which is the intended coupling.

**"Split the ground-truth test into helpers"** — applied (`openVault`, `pairs`,
`assertRecoversOriginal`, each under the limit).

### Also fixed, from CI rather than review

`test (windows-latest)` failed because the check now reports the
`EvalSymlinks`-resolved path — a change the alias dedup introduced — and on
Windows `%TEMP%` is reached through a junction, so the resolved form differs from
the joined one. The test resolves both sides now. Worth noting the behaviour
change is deliberate: the reported path is the file actually written.

PR #894, opened 2026-08-10
docs(lessons): an undeclared test precondition makes local and CI different tests

mlorentedev:
Doc-only. Captures the lesson from #885 that its own merge left behind.

`tests/spec-gate-pr-real.bats` passed locally and failed in CI on its first run: `gh pr view --json <fields>` validates field names only once past the auth check, and an unauthenticated gh exits at that check first. The dev machine is logged in, the bats job has no token — so the same file asserted the contract in one place and nothing in the other.

The fix shipped in #885 (`c0c2b36`); this records the transferable half: name a real-dependency test's environment preconditions and assert them in `setup`, gate the skip on `$CI` so a developer machine may skip and the pipeline may not, and pass a CI credential under a name only that suite reads rather than the tool's canonical variable.

Separate PR because #885 had already merged — the same way #878 came about. No code touched.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance on authentication prerequisites for real-dependency tests.
  * Documented different test behavior for authenticated, CI unauthenticated, and local unauthenticated environments.
  * Clarified use of the suite-specific authentication token.


mlorentedev:

Issue #898, opened 2026-08-10
HARNESS-113: a check never observed failing is not evidence — amend Definition of Done §5


mlorentedev:
## Why

Definition of Done §5 currently reads:

> 5. **Evidence** — no completion claim without the command output that proves it, produced in this session.

That is necessary and not sufficient. It requires output; it does not require the output to *mean* anything. A check that cannot fail produces perfectly good output and proves nothing.

Four instances of this in a single day on `kubelab` (2026-08-09), each of which passed every gate:

| # | The check | Why it was empty |
|---|---|---|
| 1 | `tls: {}` in a Kustomize strategic-merge patch, commented "no certResolver" | An empty map means *change nothing*. Prod carried `certResolver` and retried an impossible ACME order ~1/day for months |
| 2 | AI-007 acceptance strings asserting Ollama was gone | The Calendly URL contains `videollamada` — the substring `ollama`. The assertion could never have passed on the day it was written |
| 3 | A branch-conflict loop printing `clean` | `git rev-parse` failed → `merge-tree` never ran → grep searched the *error text*, found no conflict markers, reported clean |
| 4 | IDP-031's full test suite, green in staging | Ran with the **operator's** unrestricted credentials; delivery runs as a scoped ServiceAccount, which refused the manifest in prod (#948) |

None of these was a weak test. #4 was strict TDD — written first, observed failing, then made to pass. It still could not have caught the defect, because it was never run by the actor that would refuse it.

## The generalisation

A green check silently inherits three properties of whoever ran it, and names none of them in its output:

- **Actor** — whose credentials (#4)
- **Path** — which delivery mechanism; the GitOps path is only exercised *after* merge, so "staging first" does not exist for anything depending on how a change ships
- **Revision** — which commit; a session-start `git fetch` expires the moment anyone else merges (dotfiles #886, kubelab #939)

## Proposed change

Amend Definition of Done §5. Not a new Standing Order — this is the same subject as the existing evidence rule, and splitting it would let one be satisfied without the other.

```markdown
5. **Evidence** — no completion claim without the command output that proves it,
   produced in this session. The output must come from a check that has been
   **observed failing at least once**: run it against a state where it is
   expected to fail (remove the fix, revoke the grant, delete the resource)
   and watch it go red before trusting it green. A check never seen failing is
   a claim in executable syntax, not a check. When the check involves another
   system, it must also run with the same **actor** (credentials), **path**
[...]

Issue #807, opened 2026-08-08
BUG-054: the busy-binary install test never holds a binary busy, so the ETXTBSY swap path is untested


mlorentedev:
Found while running the full bats suite against clean `main` during #776. This test fails on this machine, and the reason is that its fixture never creates the condition it exists to prove.

## Problem

`tests/install-dotf.bats`, case *"converges over a running dotf: a live binary in dest is replaced, not refused"* (~line 95-113), stands in a real ELF so that the swap is exercised against a genuinely busy executable:

```bash
# A shell script never triggers ETXTBSY (the kernel does not hold it as an
# executable text image), so stand in a real ELF and keep it running.
cp "$(command -v sleep)" "$DEST/dotf"
chmod 0755 "$DEST/dotf"
"$DEST/dotf" 30 &
BUSY_PID=$!
```

The reasoning in that comment is correct. The fixture is not. On a distro where coreutils is built as a **multi-call binary**, `/usr/bin/sleep` dispatches on `argv[0]`. Copied to a file named `dotf` and invoked by that name, it does not sleep — it exits immediately:

```console
$ cp "$(command -v sleep)" /tmp/dotf && /tmp/dotf 30
coreutils: unknown program 'dotf'
```

Observed failure on clean `main`, no working-tree changes:

```
not ok 6 converges over a running dotf: a live binary in dest is replaced, not refused
# (in test file tests/install-dotf.bats, line 112)
#   `kill -0 "$BUSY_PID" 2>/dev/null' failed
# coreutils: unknown program 'dotf'
```

## Impact

Two layers, and the second is the one that matters.

1. The visible one: the suite is red on any coreutils-multi-call distro, so "the tests pass" is not a statement anyone can make locally here without first explaining away a failure — which trains people to explain away failures.
2. The invisible one: **where this test passes, it passes for the wrong reason.** The child dies instantly, so no process is holding the binary as an executable text image, so no ETXTBSY is ever provoked. The atomic-swap path that #750/BUG-037 shipped is not actually exercised; only the ordinary replace-a-file path is. A regression that reintroduced the ETXTBSY failure would keep this test green on every machine where `/bin/sleep` happens to be a standalone ELF.

Same family as #781 (a rollback test that never reaches the catch-block it claims to cover): a test whose fixture does not reach the condition named in its own title.

## Proposed fix

Keep the file on disk named `dotf` (the swap target must be that path) while giving the process the `argv[0]` coreutils dispatches on:

```bash
[...]

Issue #858, opened 2026-08-08
HARNESS-063: a fix can ship green while being a no-op on the shape that motivated it


mlorentedev:
## The gap

A fix can ship with a correct, passing, well-written test and still be a no-op on
the case that motivated it — because the test exercises a *shape* the fix
handles, while the report was about a shape it does not.

This happened on `main` today, end to end, in eleven minutes:

| | |
|---|---|
| #850 filed | 17:36Z — crystallize appends past the Session Handoff block |
| #851 merged | 17:47:56Z — with a test (`7fd7b3b`) asserting the invariant |
| #850 closed | 17:47:57Z |
| Still broken | every `content: \|`-wrapped `MEMORY.md` (#857) |

Measured both halves against the same fixture:

- `dbe91db^`: handoff at line 76, appended sections at 83/85 — invariant broken.
- `dbe91db` (main): appended sections at 76/78, handoff at 80 — invariant holds.

So #851 is **correct on the shape it was written against** and does nothing on
the other. The test is not wrong. It asserts the right claim, on the right shape,
and passes honestly. It simply never sees the second shape, so the guard and the
gap are disjoint — and a reviewer reading that test gets no signal that a second
shape exists.

## Why the existing gates do not catch it

- **Review** sees a red-to-green test and a matching fix. Nothing on the diff says "the reporter's file has a different structure".
- **`spec-gate`** checks that a spec folder was touched, not that its verification maps to the reported case.
- **HARNESS-061's stub/real pairing** (#852) is adjacent but different: it forces a *real dependency* beside a stub. Here the test was already real. The missing axis is **input shape**, not test fidelity.

This is the fifth instance today of one pattern — verifying a representation
instead of the behaviour — and the first where the failing check is the review
process rather than a `doctor` check. Prior four: BUG-055 (#837), BUG-056 (#839),
the triple `gitleaks` run (#846), the DR runbook with no step-1 instructions
(#848).

## Directions worth weighing

Deliberately not proposing one — the cheap version may be enough and the
expensive version may be unenforceable.

1. **A spec/verification convention.** `verification.md` already maps acceptance criteria to evidence. Require, for a bug-fix spec, that the evidence name the *reported artifact* — the actual file or input from the issue — not merely a test that covers the class. Cheap, and it puts the question in front of the author at the moment they can still answer it.
[...]

Issue #852, opened 2026-08-08
HARNESS-061: verify behaviour, not representation — effectiveness probes, stub pairing, DR drills


mlorentedev:
## Why

One session produced four defects with a single shape: **we verified
representations instead of behaviour.** Each looked like a healthy report right
up to the moment it mattered.

| Defect | The representation said | The behaviour was |
|---|---|---|
| BUG-055 (#837) | 15 passing cases over the exact branch | the command built was fatal to the real tool; `git commit` aborted machine-wide |
| BUG-056 (#839) | `core.hooksPath` correct globally | git resolved a local override; the guard was inert in the repo that ships it |
| `gitleaks` ×3 per commit (#846) | "Passed", printed on screen for hours | it ran once per dispatched hook type |
| DR step 1 (#848) | the runbook had a step 1 | the step had no instructions |

None presented as a failure. All four presented as things working, which is why
no audit would have found them — an audit reads representations.

## What

Three measures, in descending order of expected yield.

### 1. Effectiveness probes over identity checks

Make it hard to write the wrong kind of check. A shared helper that resolves what
git will actually run — effective config, common git dir, executability — so a
guard check asks *does this fire here* rather than *is a file at this path*.
BUG-040 already ruled this once and it was re-broken twice, which is the argument
for making it code rather than doctrine. Precedent in-tree: `System.AgeRoundTrip`
proves a key decrypts rather than that a file exists.

Closes #839.

### 2. A real-dependency test beside every stubbed third-party boundary

The exact shape that let BUG-055 ship green. A stub proves what we *asked* a tool
to do; only the real tool proves it accepts it. A guard requiring each
stub-using suite to pair with a `<name>-real.bats` or declare why it cannot,
with the existing set as a documented backlog so it can only shrink deliberately.

### 3. Drills, not readings

A runbook nobody has executed is a hypothesis with formatting. #848 exists only
because someone asked to *run* the recovery chain. Technical support for this is
a freshness signal (DR escrow age + a drill marker); the drill itself is a
calendar commitment, not code, and this issue does not pretend otherwise.

## Acceptance criteria

- [ ] A guard check cannot pass by reading global config while git resolves an effective value; a repo-local override is detected and named.
- [ ] The vault secret gate reports on whether the gate runs, not on the presence of a file `pre-commit install` refuses to create.
- [ ] A new bats suite that stubs a third-party binary fails CI until it pairs with a real-dependency test or is explicitly exempted with a reason.
[...]

PR #918, opened 2026-08-11
fix(cli): send machine-read output to stdout, not stderr (BUG-070)

### LINKED ISSUE #915: BUG-070: dotf writes its machine-readable output to stderr, so $(dotf env path …) is always empty
Every `dotf` subcommand with a machine-readable stdout contract writes that
output to **stderr** instead, so `$(dotf ...)` command substitution captures an
empty string. Cobra's `Command.Print/Printf/Println` write to `OutOrStderr()`,
not stdout — the stdout writer is only reachable via `cmd.OutOrStdout()`.

## Evidence (this machine, dotf 0.37.0)

```
$ dotf env path VAULT_PATH 2>/dev/null      # stdout only
                                             # -> empty, rc=0
$ dotf env path VAULT_PATH 2>&1 1>/dev/null # stderr only
/home/manu/Projects/knowledge
```

## Three real stdout contracts, all broken

| Site | Contract | Why it is a contract |
|---|---|---|
| `cli/internal/cmd/env.go:77` | `dotf env path <KEY>` | Its own `Long` help says *"Prints an empty line + exits 0 when the key has no resolved value, so callers can fall back with `${X:-default}`"* — that idiom only works on stdout |
| `cli/internal/cmd/env.go:117` | `dotf env generate --stdout` | The flag is literally named `--stdout` |
| `cli/internal/cmd/review.go:52` | `dotf review` | Emits the review body read from a stdin diff; pipeable by design |

Status messages elsewhere (`spec.go`, `init.go`, `vault.go`) land on stderr too,
which is harmless — but note `spec.go:114` uses `cmd.PrintErrf` for warnings
while `cmd.Printf` is used for normal output. That contrast only makes sense if
the author believed `Printf` reached stdout. The misconception is in the code.

## Impact — ADR-025 is silently bypassed

`setup-linux.sh:1088`:

```sh
HIVE_VAULT_RESOLVED="$(dotf env path HIVE_VAULT_PATH 2>/dev/null || true)"
HIVE_VAULT_RESOLVED="${HIVE_VAULT_RESOLVED:-${HIVE_VAULT_PATH:-${VAULT_PATH:-$HOME/Projects/knowledge}}}"
```

The capture is always empty (and `2>/dev/null` actively discards the value), so
the cascade result is *never* used — the hardcoded `$HOME/Projects/knowledge`
literal wins. On this machine that literal happens to be correct, which is why
nobody noticed. On a box whose vault is elsewhere (the Windows checkout uses
`~/Projects/Workspace/knowledge`) the daemon is provisioned with a path that
does not exist. `setup-windows.ps1:667` has the same shape.

This is precisely the stale-literal failure ADR-025 and #463 exist to eliminate:
the seam is present, wired, and returning nothing.

`AGENTS.md` (lines 31 and 128) instructs every agent to resolve `$VAULT_PATH`
via `dotf env path VAULT_PATH`, as do the `architecture-session` and `handoff`
skill records. An agent following that instruction in a `$(...)` gets an empty
path.

## Why the test suite cannot see it

`tests/verify-setup.bats:417`:

```bash
run dotf env path DOTFILES_REPO_DIR
[ "$output" = "$REPO_DIR" ]
```

bats `run` merges stdout and stderr into `$output` unless `--separate-stderr`
is passed (available on the pinned bats 1.13.0). The assertion therefore passes
whichever stream carries the value — the test is blind to the only thing that
is wrong. Same class as #858 ("a fix can ship green while being a no-op on the
shape that motivated it") and #852.

A Go unit test cannot catch it either: `cmd.SetOut(buf)` makes `OutOrStderr()`
return that same buffer, so such a test passes identically before and after a
fix. The guard has to exec the real binary with the streams separated.

## Fix

- `fmt.Fprintln(cmd.OutOrStdout(), …)` at the three contract sites; leave status
  messages alone (no churn).
- Amend `verify-setup.bats:417` to capture stdout only and assert non-empty —
  and confirm it FAILS against the current binary before the fix lands.
- `setup-linux.sh:1088` / `setup-windows.ps1:667` need no change once the output
  moves; they become correct as written.

## Acceptance

- [ ] `dotf env path VAULT_PATH 2>/dev/null` prints the resolved path
- [ ] `dotf env generate --stdout >f 2>/dev/null` writes the rendered file to `f`
- [ ] `dotf review` body reaches stdout
- [ ] A stdout-only regression test exists, and was observed failing pre-fix
- [ ] No behaviour change for status/warning messages

## Related

- ADR-025 (cross-machine path resolution) — the contract this defeats
- #463 — kill the stale `~/Projects/knowledge` fallback literal
- #858 / #852 — verify behaviour, not representation

mlorentedev:
Fixes #915.

Cobra's `Command.Print/Printf/Println` write to `OutOrStderr()`. Four
subcommands whose output is read by a caller took that shortcut, so every
`$(dotf …)` capture of them returned an empty string:

| Subcommand | Consumer |
|---|---|
| `env path` | `setup-linux.sh:1088`, `setup-windows.ps1:667`, and `AGENTS.md`'s instruction to every agent to resolve `$VAULT_PATH` this way |
| `env generate --stdout` | a flag named for the stream it was not using |
| `review` | the review body, pipeable by design |
| `version` | `install-dotf.{sh,ps1}` grep it for the installed semver |

## Why it matters beyond the empty string

`setup-linux.sh:1088` captures `dotf env path HIVE_VAULT_PATH` and falls back to
a hardcoded `$HOME/Projects/knowledge`. The capture was always empty, so the
literal always won — correct on this machine, wrong on any box whose vault sits
elsewhere. ADR-025 exists to delete that literal; the seam meant to replace it
was returning nothing.

`secrets show` and `mem project-key` already used
`fmt.Fprint(cmd.OutOrStdout(), …)` and work correctly. This converges four
outliers onto that existing idiom rather than introducing a new one.

## The guard is the point

`stdout_contract_test.go` swaps `os.Stdout`/`os.Stderr` at the process level and
installs **no** Cobra writers. That is required, not stylistic: `OutOrStderr()`
returns the writer `SetOut` installed, so the `execute()` helper in
`root_test.go` labels `Print*` output "stdout" and passes identically before and
after this fix. Same reason `verify-setup.bats` now captures with `$(…)` instead
of `run`, which merges both streams — that test asserted the resolved path all
the way through the bug.

Run against the unfixed code first:

```
--- FAIL: TestStdoutContracts/env_path…
    stdout missing "/tmp/stdout-contract-probe" — a caller capturing stdout gets "".
    stderr was: "/tmp/stdout-contract-probe"
--- FAIL: TestStdoutContracts/version…
    stdout missing "dotf version" — a caller capturing stdout gets "".
    stderr was: "dotf version dev"
```

## Evidence

End-to-end A/B, this build vs the installed 0.37.0:

```
                     this build                       installed 0.37.0
env path VAULT_PATH  [/home/manu/Projects/knowledge]  []
version              [dotf version dev]               []
generate --stdout    14 lines to the redirect         0
```

- `go build ./...` clean, `go vet ./...` clean, `go test ./...` all packages ok
- `bats tests/*.bats` → **1194 passed, 1 failed**. The failure is
  `not ok 394 converges over a running dotf`, which is **#807 (BUG-054)** and
  reproduces identically on clean `main` @ `a3f9a10` with no working-tree
  changes (`coreutils: unknown program 'dotf'`). Unrelated to this change.
- `shellcheck` clean (SC1091 info only); `bash -n` + `zsh -n` on the changed
  shell file; `check-bats-names.sh` clean over 81 files.

**One acceptance box deliberately not ticked as observed:** `dotf review`
reaching stdout is verified by inspection and by convergence on the
`secrets`/`mem` idiom — there is no runtime probe, because it needs a live
provider. Per #898 ("a check never observed failing is not evidence"), saying so
is worth more than the checkmark.

## Notes

- The `2>&1` merge in `install-dotf.{sh,ps1}` is **kept**: that installer runs
  against whatever `dotf` is already on PATH, including pre-fix binaries that
  still answer on stderr. Only its comment changed — it had been describing the
  defect as a standing property of the CLI rather than reporting it.
- `setup-linux.sh` and `setup-windows.ps1` need no change; they become correct
  as written once the output moves.
- Lesson recorded in `docs/lessons.md` (Definition of Done #2).

**SDD skipped**, stated per the Discipline Gate: 4 functional lines, obvious
cause, no contract change (the contract is what the help text already promised).
36 changed lines outside tests, most of them comments.

## Summary by CodeRabbit

- **Bug Fixes**
  - CLI commands now consistently write machine-readable results to standard output, including environment paths, generated configuration, reviews, and version information.
  - Improved compatibility for scripts and integrations that capture command output.

- **Tests**
  - Added coverage confirming expected stdout behavior for key CLI commands.
  - Updated setup verification tests to validate captured output and resolved environment paths.

- **Documentation**
  - Documented CLI output stream behavior and compatibility with older versions.


mlorentedev:
Triage of the three findings.

**1. errcheck on the four output calls — applied**, in `3efc065`. `env path`, `env generate --stdout` and `review` return the write error.

**Declining one sub-point: `newVersionCmd` stays `Run`, not `RunE`.** Returning the write error there would regress a normal invocation — `dotf version | head -1` closes the pipe, the write returns EPIPE, and the command would exit non-zero and print an error for what is correct usage. Nothing downstream can act on a failed write to stdout in a version command. `_, _ =` with the reason in a comment matches `secrets.go:194`, which is the established idiom here. The other three are `RunE` and genuinely can act on it, so they do.

**2. Pipe read ends never closed — applied** in `0f9efba`.

**3. `Skipf` in the `--stdout` case — applied** in `0f9efba`, and this was the best of the three. The case resolved whatever `env-contract.json` the ambient checkout exposed and degraded to a skip when that failed, which is indistinguishable from a pass while testing nothing — precisely the failure mode this file exists to close (cf. #858, #852). It now seeds a one-variable contract into a temp `DOTFILES_REPO_DIR` and asserts that variable reaches stdout.

Mutation-checked rather than assumed: reverting the `--stdout` write to `cmd.Print` turns it red, with the rendered contract on stderr and `PROBE_DIR` — the seeded fixture, not anything ambient — named in the failure.

`golangci-lint run` clean locally; `go test ./...` green.

Issue #802, opened 2026-08-07
BUG-051: the sensitive/ deploy never prunes - a deleted secret stays decryptable on every machine, forever


mlorentedev:
## Problem

Deploying `sensitive/` is a **copy-only** operation on both engines. Nothing removes a `*.secret.age` that no longer exists in the repo, so **deleting a secret from git never removes it from any machine** - the encrypted blob stays on disk, decryptable by the local age key, indefinitely.

```powershell
# setup-windows.ps1:596-602
$ageFiles = Get-ChildItem -Path $sensitiveSource -Filter '*.secret.age'
foreach ($ageFile in $ageFiles) { Copy-Item $ageFile.FullName "$sensitiveDest\" -Force }
```
```bash
# setup-linux.sh:137
cp -rf "$CURRENT_DIR/sensitive/"* "$DOTFILES_DIR/sensitive/" 2>/dev/null || true    # no --delete
```

## Live instance

Found on the Windows box 2026-08-07 while auditing a setup run:

```
repo    sensitive/*.secret.age  -> 32 files
deploy  ~/.dotfiles/sensitive/  -> 33 files
diff -> kubelab-dispatch-token.secret.age   (present ONLY in the deploy mirror)
```

`dotf doctor` already **detects** it (`[FAIL] orphan: kubelab-dispatch-token.secret.age (no registry entry)`) but nothing converges it - which is the "orphan-secret owner" complaint in #699.

## Why this is the interesting kind of bug

The *skill* deploy in the same `setup-windows.ps1` prunes correctly, and carefully: it only removes outputs carrying the `generated: true` provenance marker, which is why a user-managed skill survived a converge earlier that day. Convergence was retrofitted onto the newer layer and never back-ported to the older one - and the older one is the one holding the secrets.

## Agreed semantics

Decided with Manu 2026-08-07: **`dotf doctor --fix` prunes; setup only warns.**

Setup runs routinely and must never silently delete encrypted material. `doctor --fix` is explicit and auditable, and it closes the convergent arm #699 asks for, since doctor is already the thing that detects the orphan.

```
setup-*.{sh,ps1}
  [WARN] orphan secret in deploy store: kubelab-dispatch-token.secret.age
         run: dotf doctor --fix

dotf doctor --fix
  [FIXED] pruned 1 orphan *.secret.age
```

[...]

Issue #906, opened 2026-08-10
CI: CodeRabbit's free-tier quota silently skips review on some PRs, and the merge gate cannot tell


mlorentedev:
Observed on #900 during the ADR-032 session.

## What happened

CodeRabbit posted this instead of a review:

> **Review limit reached** — you've reached your PR review limit, so we couldn't start this review. Next review available in: 12 minutes. You've used all free OSS reviews for now.

The check still reported `CodeRabbit=SUCCESS`, and #900 merged. Nothing distinguished "reviewed, no findings" from "never reviewed". The same PR count that makes this repo productive is what exhausts the quota, so the skip lands preferentially on **busy sessions** — exactly when review is most valuable. For contrast, #901 was reviewed and produced ten findings, one of which corrected a factual claim in an accepted ADR.

## Why this is a guard defect, not a vendor complaint

The repo treats an unexercised gate as a defect elsewhere — HARNESS-065 (#881) holds new enforcement layers until the first one is proven to fire, and HARNESS-061 (#852) is entirely about representations that look healthy at the moment they stop being true. A review check that reports SUCCESS when no review ran is that same shape: the representation is green and the behaviour is absent.

## Options

- Make the skip **visible**: a job that greps the CodeRabbit comment for the limit banner and fails, or posts a "not reviewed" label, so a human merging knows what they are merging.
- Reduce consumption so the quota lasts: CodeRabbit's own suggestions are pausing incremental auto-reviews, label-based opt-in, or excluding docs-only/generated PRs. Docs-only PRs are a real slice of this repo's volume.
- Accept the skip explicitly for a declared class (docs-only), so the exhaustion falls where it costs least rather than wherever it happens to land.

The first is the one that matters: whatever the policy, **it must be legible at merge time**.

## Acceptance

- [ ] A PR whose CodeRabbit review did not run is distinguishable from one that ran clean, without reading the comment thread.
- [ ] Quota consumption is deliberately allocated rather than first-come.

## Refs

#900 (the instance), #881 and #852 (the same failure shape in other gates).

mlorentedev:
Fired live again today on PR #918, with the check output visible this time:

```
CodeRabbit	pass	0		Review rate limited
```

Status **pass**, summary **"Review rate limited"**. No review was performed, and the merge gate reports green — which is exactly the shape this issue describes.

[...]

PR #948, opened 2026-08-13
fix(harness): converge the deploy engine — prune, drift detection, all six surfaces

### LINKED ISSUE #828: HARNESS-058: a doctrine change reaches two surfaces on deploy and four only on a full setup run
## Context

Doctrine reaches the six agent surfaces through two different mechanisms, and only one of them is the harness deploy engine:

| Surface | How the doctrine arrives | Updated by |
|---|---|---|
| agy, codex | compact doctrine payload | `compile-harness.sh --deploy` |
| claude, opencode, pi, copilot | a copy of the instruction file | `setup-linux.sh` / `setup-windows.ps1` |

Measured on this machine right after #821 merged and `--deploy` ran: `~/.gemini/GEMINI.md` and `~/.codex/AGENTS.md` carried the new Definition of Done; `~/.claude/CLAUDE.md`, `~/.config/opencode/AGENTS.md`, `~/.pi/agent/AGENTS.md` and `~/.copilot/copilot-instructions.md` did not. The committed files were correct — the deployed copies were simply stale, and nothing said so.

## Why it matters

The deploy engine is supposed to be the answer to "how does an artifact reach an agent". Today it fully owns two surfaces and half-owns four, so a doctrine change is live for a third of the fleet and pending an installer run for the rest — with no signal marking the difference. That is the same shape as HARNESS-053: a state that stops being true without anyone being told.

It also makes the deploy non-uniform in exactly the way `targets[]` was introduced to avoid: the mechanism depends on which agent you are, not on what the artifact is.

## Options

1. **Move the instruction-file copy into `--deploy`**, so one command brings every surface current and `setup-*` calls it like it already calls the rest. Preferred: it makes the engine's ownership total and matches how skills and the doctrine payload already work.
2. Leave the split and make it visible — `--deploy` warns when a deployed instruction file's generated region differs from the committed one. Cheaper, but it institutionalises two mechanisms.

## Acceptance

- [ ] A doctrine change is live on all six surfaces after a single documented command.
- [ ] A stale deployed instruction file is either impossible or reported, never silent.
- [ ] `setup-linux.sh` and `setup-windows.ps1` keep working unchanged for a fresh machine.

## References

- Surfaced while verifying #821 end-to-end on a real machine.
- Related: #817/#819 (the compact payload), #803 and #561 (the other unscoped layers of cross-agent parity).

### LINKED ISSUE #843: BUG-058: the harness deploy never prunes, so records deleted from the repo keep failing doctor from the mirror
## Symptom

`dotf doctor` reports harness drift on every run, and it cannot be cleared:

```
[Harness + skill drift]
  [FAIL] harness/skill drift (run: compile-harness.sh --refresh, then re-deploy)
```

Running the suggested remedy does not help, because `--check` from the **repo** is
clean:

```
$ cd ~/Projects/dotfiles && bash scripts/compile-harness.sh --check
[check] OK: no harness drift
```

`checkGuardHooks`' sibling `checkHarnessDrift` shells out to
`cfg.DotfilesDir/scripts/compile-harness.sh` — the **deploy mirror**. Run there,
the real cause appears:

```
$ cd ~/.dotfiles && bash scripts/compile-harness.sh --check
[ERROR] /home/manu/.dotfiles/harness/skills/vault-sync//SKILL.md: required frontmatter key "id" missing or empty
[check] FAIL: harness drift detected
```

## Root cause: the harness deploy is copy-only, with no prune

```
repo   harness/skills/  38 records
mirror harness/skills/  40 records
extra in mirror: agent-config-optimization, vault-sync
```

`vault-sync` was deleted from the repo by **#274** (`066eadc chore(skills): drop
generic vault-sync skill (Hermes-only)`). `agent-config-optimization` was likewise
retired — the surviving `agent-lifecycle` skill's own description says it "absorbs
… agent-config-optimization (the standalone skill is retired into this skill's
Section 3)". Both still sit in the mirror, `SKILL.md` dated 6 Jun, because the
deploy copies records in and never removes records that disappeared from the repo.

This is the same defect as **#802** (BUG-051, the `sensitive/` deploy never prunes),
on a different surface. Worth fixing with one shared prune-on-deploy semantic
rather than twice.

## Why it surfaced only now

The stale records carry the pre-#826 frontmatter dialect — `name` and `description`
only:

```yaml
---
name: vault-sync
description: "Vault git synchronization protocol: ..."
---
```

That was valid until **#826** made `id`/`type`/`status` required. So the records
were harmless leftovers for two months and became a permanent red the moment the
contract tightened. Any future contract change will re-break on whatever else the
mirror is still carrying.

## Blast radius: contained, but the signal is not

Checked: neither stale record reaches any agent —

| Record | `~/.claude/skills` | `~/.config/opencode/commands` | `~/.gemini/skills` |
|---|---|---|---|
| `vault-sync` | no | no | no |
| `agent-config-optimization` | no | no | no |

`--deploy` renders from the repo's manifest, not from whatever the mirror holds, so
no agent is served a retired skill. The damage is to the signal: a `FAIL` that no
remedy clears trains everyone to read past the harness-drift line, which is the
line that would report a *real* drift.

## Suggested fix

Prune on deploy — remove mirror records with no counterpart in the repo — matching
whatever semantic #802 settles on (that issue records a decision that deleting
deployed material belongs in `dotf doctor --fix` rather than a silent purge during
setup; the same reasoning applies here, with the difference that these are
generated records, not encrypted secrets, so the bar is lower).

Guard: assert the mirror's record set equals the repo's after a deploy. A
count comparison would have caught this the day #274 merged.

## Note for #836

The transient drift reported there is a separate cause (a `--refresh` racing a
vault another session is writing). This one is permanent and reproduces with the
vault idle, so the two need separating when either is fixed.

### LINKED ISSUE #869: OPS-025: a merged guard never reached the machine — self-deploy is opt-in and nothing reports the drift
## What happened

#866 shipped a `dotf doctor` check that detects a YAML-wrapped `MEMORY.md`. The previous session's handoff recorded that merging the release PR (#863, v0.35.0) was all that was needed, because *"the selfupdate timer does the rest — no manual step"*.

That was false on this machine, and the guard sat inert for the whole window:

```
~/.local/bin/dotf version   -> 0.34.0        # versions.conf pinned 0.35.0
dotf doctor | grep -c 'Auto-memory file shape' -> 0
systemctl --user list-timers -> only hive-upgrade.timer
```

## Why

Not a regression. `setup-linux.sh` (~L1134) gates the units on a tri-state `DOTFILES_AUTODEPLOY`: `1` enables, `0` tears down, **unset = no-op, opt-in, default OFF** (OPS-001, archived). `DOTFILES_AUTODEPLOY` is unset here, so the timer was never installed. `cli/internal/cmd/update.go` is explicit: *"opt-in, scheduler-invoked"*.

So "merged" and "deployed" are separated by a version pin **plus** a carrier that is off by default — and **nothing reports the gap**. The installer detects drift once you run it (`dotf 0.34.0 drifted from pinned 0.35.0; converging`), but no scheduled or session-adjacent check says the installed binary is behind the pin.

The impact is general, not specific to #866: **every** guard merged from now on has a silent window of unbounded length on any machine where autodeploy is off. A guard that isn't installed protects nothing, and the ticket that shipped it reads as closed.

## Remedies (both wanted)

**(a) Scheduler disposition.** Decide per machine whether `DOTFILES_AUTODEPLOY=1` is right. It is deliberately opt-in and that judgment should be re-affirmed, not reversed by default: an unattended job that fast-forwards the shared checkout and re-runs setup is in tension with running parallel agent sessions against that same checkout. If it goes on, the Windows Task Scheduler twin is needed for parity (same shape as #666).

**(b) The durable guard — `dotf doctor` goes red when installed `dotf` != the `versions.conf` pin.** This is the part that generalises, and it is house style per #853 (probe whether the guard fires, not where its file sits). It makes the *next* merged guard's staleness visible instead of silent, and it works whether or not the timer is ever enabled.

Manual convergence was run this session (0.35.0 installed, `[Auto-memory file shape] (1 checks, all ok)`), so nothing is currently broken — the reporting hole is.

Related: OPS-001 (archived spec), #853, #866.

### LINKED ISSUE #943: BUG-074: checkDeployedSkillSymlinks flags foreign-tool skill symlinks as BUG-100 regressions
## Symptom

`dotf doctor` on this machine reports:

```
[Harness + skill drift]
  [FAIL] deployed skill path(s) are symlinks (must be hard copies — BUG-100):
  [FAIL]   /home/manu/.claude/skills/computer-use
  [FAIL]   /home/manu/.claude/skills/find-skills
  [FAIL]   /home/manu/.claude/skills/orca-cli
  [FAIL]   /home/manu/.claude/skills/orchestration
```

Read at face value this looks like a BUG-100 regression (issue #100, closed by SDD-007/SDD-008 — deploy strategy fighting an agent's filesystem layout).

## Root cause: not a regression, a false positive

All four names resolve to `~/.agents/skills/<name>` — Orca's own skill-symlink
mechanism, not this repo's. None of the four has a `harness/skills/<name>`
record:

```
$ ls harness/skills/ | grep -E "computer-use|find-skills|orca-cli|orchestration"
(no output)
$ readlink -f ~/.claude/skills/computer-use
/home/manu/.agents/skills/computer-use
```

`checkDeployedSkillSymlinks` (`cli/internal/doctor/checks_deploy.go`) walks the
whole `~/.claude/skills`, `~/.config/opencode/commands`, `~/.gemini/skills`,
`~/.gemini/prompts` trees and flags *any* symlink found, regardless of whether
the name is one this repo manages. That is broader than the deploy side's own
policy: `compile-harness.sh`'s `warn_unmanaged_output` already treats "no
record, no shadow" as silent — a third-party skill that owns its name is not
this repo's problem to fix.

There is direct prior art for the exclusion: `specs/archive/AI-022-pi-harness-slot/proposal.md`
documents that `~/.pi/agent/skills` is deliberately excluded from the strict
symlink sweep because "pi's own installer manages sibling skills as symlinks
into `~/.agents/skills` (agent-owned state)". The Go port of the symlink check
never carried that exclusion forward for the equivalent `~/.claude/skills`
case, so it produces the same false positive `dotf doctor` now shows.

## Fix

Fixed in the same change as HARNESS-070 (deploy convergence, #843/#869/#828):
`checkDeployedSkillSymlinks` now only flags a symlink whose name has a
`harness/skills/<name>` record (checked against both the deploy mirror and,
when resolvable, the repo checkout) — matching `warn_unmanaged_output`'s
existing policy. Go regression tests cover both the managed-name FAIL and the
unmanaged-name silence, including the opencode command-file and nested
`SKILL.md` shapes.

## References

- Issue #100 (closed) — the original BUG-100.
- `docs/adr/adr-012-deploy-strategy-copy-with-drift-assertion.md`
- `specs/archive/AI-022-pi-harness-slot/proposal.md` — the `~/.agents/skills` exclusion precedent this restores.

mlorentedev:
## Summary

Three related gaps in harness deploy convergence, closed together as one thread (spec: `specs/HARNESS-070-deploy-convergence/`):

- **#843 (BUG-058)** — the repo→mirror copy in `setup-linux.sh` is copy-only, so a `harness/{skills,agents}` record deleted from the repo survived forever in the `~/.dotfiles` mirror and kept failing `checkCompileHarnessDrift`, which runs FROM the mirror. `dotf doctor` now detects the orphan and `dotf doctor --fix` prunes it, per #802's already-decided semantic (doctor `--fix` prunes, setup only copies/warns).
- **#869 (OPS-025), remedy (b)** — a `dotf` binary behind the `versions.conf` pin now FAILs doctor instead of WARNing. A stale `dotf` silently carries none of the guards merged since it was built — a harder failure than an ordinary tool drifting one version.
- **#828 (HARNESS-058)** — doctrine reached `agy`/`codex` through `compile-harness.sh --deploy` but reached `claude`/`opencode`/`pi`/`copilot` only through a full `setup-linux.sh` run. `--deploy` now copies the four instruction files itself (ordered first, before the skill-catalog/AGENT-PRESENCE injections that land on top of them), and `dotf doctor` reports a deployed instruction file that has drifted from its repo source (region-stripped compare, so the injected regions don't false-fail it).

**Bonus, found while investigating this session's handoff evidence**: the "4 symlinked skills" reported as a BUG-100 regression (`computer-use`, `find-skills`, `orca-cli`, `orchestration`) turned out to be a false positive — all four resolve to Orca's own `~/.agents/skills` mechanism, unrelated to this repo's `harness/skills/` records, with documented prior art for the exact exclusion (`specs/archive/AI-022-pi-harness-slot/`). Filed and closed as **#943 (BUG-074)** in the same change.

Also fixed a pre-existing bug found while sandbox-testing AC3: `deploy_agent_presence` always logged `[deploy] presence -> ...` as success even when the injection was skipped because the target file didn't exist, producing a directly contradicting pair of log lines.

## Not in this PR (stated, not silently dropped)

- **#869 remedy (a)** — `DOTFILES_AUTODEPLOY` disposition (opt-in vs. on-by-default) is a per-machine judgment call for @mlorentedev, not a code change. Remedy (b), the durable guard, is what this PR ships.
- **#802** — the same copy-only-no-prune gap on `scripts/`, `sensitive/`, `.zsh/`, `ssh/`, `secrets/` (36+ orphan files measured on a live machine). Referenced, not closed — out of this session's assigned surface (`scripts/compile-harness.sh`, `harness/`, the doctor package).
- **bats coverage** for the shell-side `--deploy` change — `tests/*.bats` was out of scope this session (three parallel sessions touching the repo). Verified instead by: `shellcheck`/`bash -n`/`zsh -n` clean, all 44 existing `tests/compile-harness.bats` + 64 `tests/setup-linux.bats` cases still pass unmodified, and a manual sandboxed `HOME=/tmp/fake-home --deploy` run (idempotent, region-injection-safe). Proposing a follow-up ticket to add the bats regression — not filing it unilaterally.
- **Windows parity** — the new `agents.presence[].source` field is consumed only by `compile-harness.sh` (Linux). `setup-windows.ps1` is unaffected (per AC3) and never had this specific gap: it sets `$DotfilesDir = $PSScriptRoot`, i.e. the deploy dir IS the checkout, so there's no lightweight-deploy-vs-full-setup split to begin with on that side.

## Current machine state (informational, not actioned by this PR)

Running `dotf doctor` on this machine right now still shows the drift this PR's checks were built to catch — expected, since fixing the *detection* doesn't retroactively converge the machine:

- `dotf` itself is one release behind (`installed=0.37.0 pinned=0.38.0`) — now a FAIL, was a WARN.
- 6 files differ repo↔`~/.dotfiles` mirror (9 PRs + a release cut since the last full setup run).
- 3 instruction files are stale (`.claude/CLAUDE.md`, opencode/pi `AGENTS.md`).

Remediation, in order, left for @mlorentedev to run when convenient: merge this PR → next release cuts → `./scripts/install-dotf.sh` → `setup-linux.sh` (or `compile-harness.sh --deploy` for just the harness surface) → `dotf doctor` should read clean. Worth noting for the #869 story: the new guards in this PR are themselves live only after that next release installs — which is the OPS-025 window working exactly as described.

## Test plan

- [x] `cd cli && go build ./... && go vet ./... && go test ./...` — 13 packages, all green
- [x] `golangci-lint run ./...` (pinned v2.12.2, matches `versions.conf`) — 0 issues
- [x] `shellcheck scripts/compile-harness.sh`, `bash -n`, `zsh -n` — clean
- [x] `jq empty harness/manifest.json` — valid
- [x] `bats tests/compile-harness.bats` — 44/44 pass (unmodified)
- [x] `bats tests/setup-linux.bats` — 64/64 pass (unmodified)
- [x] Sandboxed `HOME=/tmp/fake-home bash scripts/compile-harness.sh --deploy` — all 4 instruction files written, second run byte-for-byte idempotent, `--check` stays green
- [x] All 5 new checks confirmed live against this machine's real, unmodified state (see `specs/HARNESS-070-deploy-convergence/verification.md`)

Closes #843. Closes #869. Closes #828. Closes #943.
Refs #802

## Archive skip rationale

The spec (`specs/HARNESS-070-deploy-convergence/`) is left active rather than archived in this PR. This repo's `/spec archive` step requires an independent `adversarial-review` pass (`review.md` with a `PASS`/`PASS-WITH-GAPS` verdict, ideally from a different session/agent than the implementer) before archiving — the whole point being that the reviewer is not the same eyes that just wrote the code. That review has not run yet. Rather than self-certify with `review: waived` or `--force-without-review` on a change I just wrote myself, the spec stays in `specs/` until an independent `/adversarial-review` pass runs, then archives in a small follow-up PR. All acceptance criteria are met and verified (see `verification.md`); this is a process-sequencing gap, not unfinished work.

## Summary by CodeRabbit

* **New Features**
  * Deployment now propagates configured instruction files across supported agent environments.
  * Harness diagnostics detect instruction-file drift and orphaned deployment records.
  * `--fix` can remove orphaned harness records and converge deployments safely.
  * Unmanaged skill symlinks are ignored during validation.
  * Agent presence configuration now supports source paths and required command checks.

* **Bug Fixes**
  * Dotfile version drift is reported as a failure with remediation guidance.

* **Documentation**
  * Added deployment-convergence specifications, verification details, and troubleshooting lessons.


mlorentedev:

mlorentedev:
Cross-session note from the branch working on BUG-074 (PR #950), which hit the identical spec-gate failure.

**spec-gate is red for archive-on-merge**, not for the LOC threshold:

```
         HARNESS-070-deploy-convergence (#843)
```

Two things that cost me time and may save you some:

1. **A local `./scripts/check-spec-gate.sh` passes and proves nothing here.** It cannot see the PR body, so it cannot evaluate the closing keywords this rule depends on; CI runs `scripts/spec-gate-pr.sh`, which can. That false negative is exactly #854 (BUG-061), which is in this branch's own scope — so there is now a live reproduction to cite, on two PRs the same day.

2. **`dotf spec archive` refuses without a fresh, passing `review.md`** from `/adversarial-review`, and the implementer cannot be the reviewer (the skill forbids single-agent self-review). Sequence: independent review → PASS → `dotf spec archive HARNESS-070-deploy-convergence --pr <url>` → commit in this PR → gate turns green.

Also, unrelated to the gate: **CodeRabbit reports `pass` here with the description "Review rate limited"** — it never read this diff. Four PRs opened in parallel exhausted the free-tier quota (#945 consumed it). Do not read that green as review coverage; evidence recorded on #906.

mlorentedev:
This PR was squash-merged (18ccd60) before the CodeRabbit review-fix commit on this branch landed — that commit was pushed ~21h after this PR closed, to an already-merged branch, so it never got CI and never merged. It carried 3 confirmed Major findings from CodeRabbit's review here (blank-line false-positive in checkInstructionDrift, a data-loss guard for checkHarnessMirrorOrphans --fix, and an exit-code propagation fix in deploy_instructions).

Landed via follow-up: #954

PR #949, opened 2026-08-13
fix(tests): repair four guards that pass without checking what they claim

### LINKED ISSUE #761: BUG-039: extensionless git-hooks miss .gitattributes eol=lf — CRLF shebangs disable the GUARD-001 dispatcher
Found while committing #750 (PR #760).

## Summary

The five hook dispatchers in `git-hooks/` carry **no file extension**, so `.gitattributes`' `*.sh text eol=lf` never reaches them. They fall through to `* text=auto` and are checked out **CRLF on Windows**, which breaks their `#!/usr/bin/env bash` shebang.

This is precisely the failure `.gitattributes` warns about three lines above the rule that misses them:

```gitattributes
# Unix shells: must execute on Linux/macOS — LF only, even on Windows checkouts.
# CRLF in shebang scripts causes `bash: command not found` and silent failures.
*.sh   text eol=lf
```

## Evidence

Git itself confirms the attribute never applies:

```console
$ git check-attr text eol -- git-hooks/pre-commit scripts/utils.sh
git-hooks/pre-commit: text: auto
git-hooks/pre-commit: eol: unspecified
scripts/utils.sh: text: set
scripts/utils.sh: eol: lf
```

And the checked-out hook is fatal under a standard bash (executed via its shebang, the way git invokes a hook):

```console
$ ./git-hooks/pre-commit
/usr/bin/env: 'bash\r': No such file or directory
/usr/bin/env: use -[v]S to pass options in shebang lines
$ echo $?
127
```

## Impact

`git-hooks/pre-commit` is the **GUARD-001 dispatcher** — it runs the memory-sink guard and then chains to the repo-local hook (the chain that keeps `gitleaks` alive under a global `core.hooksPath`). A hook that cannot execute is a guard that is not enforcing anything.

Affected, all extensionless and all shebang scripts:

- `git-hooks/pre-commit` — GUARD-001 memory-sink guard + local-hook chain
- `git-hooks/pre-push` — the gitleaks chain (cf. #748)
- `git-hooks/commit-msg`, `git-hooks/post-checkout`, `git-hooks/prepare-commit-msg`

**Severity is latency, not blast radius.** Git for Windows' MSYS bash tolerates the `\r` in a shebang, so a normal `git commit` on the maintainer's box still passes — the defect only bites under a plain bash (WSL, a container, a non-MSYS runner, or any tool that shells out to `/usr/bin/env`). That is what makes it worth fixing: the guard's health currently depends on which bash happens to run it, and nothing tests that.

Sibling of #748 (BUG-036), which reports GUARD-001's other failure mode — same theme of a security guard silently inactive.

## Also affected: `dotf init` templates

`cli/internal/initrepo/templates/{Makefile,gitignore,claude-md}` are extensionless too, so `dotf init` can emit **CRLF templates into freshly bootstrapped repos**. A CRLF `Makefile` is a genuine portability problem on Linux.

## Proposed fix

- Add explicit `.gitattributes` entries for the extensionless shebang/text files: `git-hooks/*`, the `initrepo` templates, and `ssh/config`.
- Renormalize the tracked blobs (`git add --renormalize .`) so existing checkouts converge.
- Add a regression test asserting `git check-attr eol` is `lf` for every tracked extensionless text file — the class-level guard, so the next extensionless file added cannot silently reintroduce this.

## Acceptance criteria

- `git check-attr eol -- git-hooks/*` reports `lf` for all five hooks.
- A fresh Windows checkout leaves `git-hooks/pre-commit` executable under a plain (non-MSYS) bash.
- A test fails if any tracked extensionless text file lacks an explicit `eol`.

Refs: #748, #693 (C39/C38 CRLF cluster), `.gitattributes`, spec GUARD-001.

### LINKED ISSUE #807: BUG-054: the busy-binary install test never holds a binary busy, so the ETXTBSY swap path is untested

### LINKED ISSUE #854: BUG-061: pre-push spec-gate false-negative on archive-on-merge (no PR body locally)

### LINKED ISSUE #892: stub-real-pairing misses suites that stub a binary from a sourced library
`tests/stub-real-pairing.bats` detects a stubbing suite structurally:

```bash
stubs_a_binary() {
    grep -q 'chmod +x' "$1" && grep -qE 'PATH="?\$' "$1"
}
```

Both greps run against the `.bats` file itself. A suite whose stubbing lives in a **sourced library** therefore evades the guard entirely.

`tests/vault-health-golden.bats` (#890) is the first live example: it stubs the `obsidian` binary and puts the stub dir on PATH, but does so inside `tests/golden/vault-health/lib.sh`, so the guard sees a clean file and the suite ships with neither a real sibling nor a declared exemption. It passes for the wrong reason.

Flagging it rather than quietly benefiting from it. The exemption table cannot be used as the workaround either: the second test asserts the list has no stale entries, so adding an entry the detector does not flag would fail.

**Why this matters beyond one file.** The guard's value is that a NEW stub suite fails until someone chooses. A detector with a structural blind spot silently downgrades that to "fails unless you happen to factor your stub into a helper" — and factoring the sandbox into a shared `lib.sh` is the pattern the golden corpora are explicitly built on, so the blind spot lines up exactly with the direction the test suite is growing.

**Fix**: follow `.`/`source` lines pointing under `tests/` and apply the same two greps to them, one level deep. Deliberately not fixed in #890 — widening a guard that scans 78 suites is its own change with its own false-positive risk, not a rider on a corpus PR.

See BUG-055 for why the guard exists.

mlorentedev:
## Summary

Four independent fixes under one theme: a test or hook that reports green without exercising the thing it names.

- **#892** — `stub-real-pairing.bats` detected a stubbing suite structurally, but only by grepping the `.bats` file itself; a suite stubbing a binary from a *sourced library* (`tests/vault-health-golden.bats` via `tests/golden/vault-health/lib.sh`) evaded it entirely. Widened the detector to follow one level of `.`/`source` indirection under `tests/`, added the new exemption, and single-sourced the exemption list — which was itself already drifted (`bitacora-rollout` was exempted but missing from the stale-entries check, so a staleness regression there would have gone undetected).
- **#807** (BUG-054) — the busy-binary install test never actually held a binary busy: a coreutils multi-call `sleep`, copied to a file named `dotf`, exits immediately instead of sleeping. Confirmed empirically on this machine — `uutils coreutils` dispatches by resolved executable *path*, not `argv[0]`, so the commonly-proposed `exec -a` fix does not help here either (see the lessons.md entry for the reproduction). Replaced the fixture with a copy of `bash` busy-spinning on shell builtins only — no multi-call dispatch to satisfy, so the ETXTBSY path is genuinely reached regardless of host coreutils flavor.
- **#854** (BUG-061) — `check-spec-gate.sh`'s archive-on-merge LOC credit only fires when `SDD_PR_BODY` names a closing keyword, which is empty on every local pre-push run by design — so a PR that correctly archived its spec in the same change could not be pushed locally. Added `scripts/spec-gate-prepush.sh`, a local wrapper mirroring the existing CI adapter (`spec-gate-pr.sh`) that resolves the branch's live PR via `gh` and falls through to today's behaviour on any resolution failure. Wired into `.pre-commit-config.yaml`'s opt-in pre-push hook. An earlier attempt credited any local active-to-archived spec transition unconditionally; the repo's own `#397` regression tests caught that this reopens the "gratuitous archive-move dodging the gate" bypass, since local and CI share the same script gated only by an env var — reverted in favor of the wrapper.
- **#761** (BUG-039) — `git-hooks/**` already carries an explicit `eol=lf` rule (BUG-068); the same CRLF hazard was uncovered for `cli/internal/initrepo/templates/*` (embedded into the `dotf` binary via `go:embed`, so a CRLF checkout ships a CRLF Makefile/gitignore/claude-md into every scaffolded repo), `ssh/config`, `LICENSE`, and the `tests/golden/**` fixture corpora. Added the missing `.gitattributes` rules and a class-level regression test asserting every tracked extensionless text file resolves an explicit `eol`, so the next one added outside a covered group fails loudly instead of depending on which OS happens to check it out.

Each fix is mutation-tested: the corresponding guard is shown failing against the pre-fix code/config and passing again after (details in the commit message and `docs/lessons.md`).

## SDD skip rationale

Four independent, self-contained bug fixes, each individually a small obvious-cause fix backed by its own mutation-tested regression guard. They do not share one feature-id to scope a single spec around — SDD specs model a single feature/change, not a themed bundle of unrelated small fixes — so this ships under the skip-sdd escape hatch rather than forcing an artificial spec folder.

## Test plan

- [x] `~/.local/bin/shellcheck` clean on every touched/added `.sh`
- [x] `bash -n` clean on every touched/added `.sh`
- [x] Full `~/.local/bin/bats tests/*.bats` — 1235 passed, 0 failed, 73 skipped (Windows/pwsh-only)
- [x] Each of the four fixes mutation-verified individually (guard shown red pre-fix, green post-fix)
- [x] `cd cli && go build ./...` — unaffected, sanity-checked
Closes #892
Closes #807
Closes #854
Closes #761


mlorentedev:
Cross-session note from the branch working on BUG-074 (PR #950).

1. **This PR is CONFLICTING** (`mergeStateStatus=DIRTY`). #945 (docs/truth-pass-batch) merged into main mid-flight, so this branch is behind. `git fetch origin && git rebase origin/main`. Expect the collision in `docs/lessons.md` — every parallel branch this sprint appends there, which is the conflict class already recorded in that same file from PR #936. Keep both entries.

2. **CodeRabbit reports `pass` with the description "Review rate limited"** — it never read this diff. Four parallel PRs exhausted the free-tier quota; #945 consumed it and the other three got a green check from a reviewer that never ran. Evidence on #906.

3. Relevant to this PR's own scope: **#854 (BUG-061) just fired for real, twice.** `./scripts/check-spec-gate.sh` passed locally on both #948 and #950, then CI's `scripts/spec-gate-pr.sh` failed both with an archive-on-merge violation — because the local script cannot see the PR body and so cannot evaluate closing keywords. That is the false negative the ticket describes, now reproducible on demand rather than hypothetical. Good material for the regression test that fix needs.

mlorentedev:

mlorentedev:

Issue #972, opened 2026-08-15
REFACTOR-012: Entries() is a tagged union and three consumers still read File without checking Backend


mlorentedev:
BUG-077 (#969) was one consumer of `Registry.Entries()` holding a pre-#606 assumption.
Sweeping the other callers found the same assumption in two more places. Filing the
class, since fixing only the instance leaves the next one to be discovered in production.

## The shape

`Entries()` returns a **tagged union**: each `Entry` carries `Backend`, and only the age
backends populate `File`. #606 taught it to emit bw-backed secrets because the Loader
dispatches on the tag. Any consumer that reads `File` without checking `Backend` gets `""`
— Go's zero value, so it is silent rather than a type error — and silently misbehaves.

## Instances still open

**1. `cli/internal/doctor/checks_pat.go:141` — PAT expiry monitoring silently drops
migrated tokens.**

```go
if e.IsFile || !strings.HasPrefix(e.File, "github.") { continue }
```

For a bw entry `e.File` is `""`, so `HasPrefix` is false and every bw-backed GitHub PAT is
skipped. `BITACORA_PAT` migrated to bw in #961 and is therefore no longer expiry-checked;
`GITHUB_PERSONAL_ACCESS_TOKEN` / `RELEASE_TOKEN` are still `backend: age` and still
covered, which is why the section looks like it works. Every future PAT migration silently
removes a token from monitoring. Severity: real and active.

**2. `cli/internal/secrets/render.go:129` — the non-deterministic-registry guard does not
fire for bw secrets.**

```go
if prev, dup := bySource[e.Var]; dup && prev.File != e.File { return nil, fmt.Errorf(...) }
```

Two distinct bw secrets exposing the same var both carry `File == ""`, so the inequality is
false and the fail-fast never triggers — precisely the "registry exposes VAR from two
sources" case it exists to reject. Latent (needs a duplicate var) but the guard is weakest
where 28 secrets now live. Compare by `(Backend, File, Item, Field)`, or by an entry
identity the union defines.

**3. Backend values are stringly-typed, with no SSOT.** `"bw"` appears as a bare literal in
`registry.go:318`, `registry.go:479`, `github.go:101` and now `checks_deploy.go`. There is
no `BackendBW` constant and no exhaustiveness check, which is what let a consumer be added
for one variant and forgotten for another. An exported constant set — plus a helper the
consumers dispatch through — turns the next occurrence into a compile-time question.

## Suggested fix

[...]

Issue #971, opened 2026-08-15
BUG-078: migrate deletes the age: pointer, so a migrated secret's DR floor becomes untracked


mlorentedev:
## The gap

`dotf secrets migrate` flips `backend: age` → `backend: bw` and **deletes the `age:`
line**. `registry_write.go:27` calls it "the now-dead `age:` source".

Under ADR-028 it is not dead. age is the **DR floor** beneath the Bitwarden SSOT, and
the blob survives migration — #938 (CLI-036) exists precisely because there is no retire
path. So after migration the blob is still on disk and the registry no longer records
that it exists, or which secret it belongs to.

Today: 28 migrated secrets, 28 unclaimed blobs under `sensitive/`.

## Why it cannot be recovered by inference

Names do not correlate. `OPENAI_API_KEY` (bw item `openai-api-key`) is backed by
`chatgpt.api-key.secret.age` — the registry itself comments on the drift
("naming drift: store says 'chatgpt', service is OpenAI"). Any basename-matching
heuristic would mis-pair blobs, which for a recovery floor is worse than not guessing.

## Consequence, already observed

`dotf doctor` could not classify those blobs and reported all 28 as
`orphan: ... (no registry entry)` — an instruction to delete the DR floor of the live
tier. BUG-077 (#969) degraded that to a single WARN that refuses to classify, and points
here. **That WARN is the placeholder for this fix**: `checks_deploy.go`'s
`reportUnreferencedBlobs` regains per-blob FAIL severity as soon as the registry records
the pointer, and the comment there says so.

## Fix

1. `SetBackendBW` keeps `age:` instead of dropping it — it becomes the recorded DR-floor
   pointer for a bw-backed secret. (The schema already allows both keys: a pre-migration
   entry carries `age:` *and* `bw:`.) Decide the field's post-migration meaning
   explicitly and document it in the registry header, which currently reads `age:` as the
   live source.
2. Backfill the 28 already-migrated entries. **Not mechanical** — each pair must be
   reconstructed by hand because of the naming drift above.
3. Restore `reportUnreferencedBlobs` to per-blob FAIL and delete the WARN branch, with
   BUG-077's guard extended to cover it.

## Coordination

Touches `cli/internal/secrets/registry_write.go` and `secrets/registry.yaml`, both live
in the CLI-024 worktree. Should land with the secrets migration work, not against it.

Related: #938 (no retire path), #964, #969 (the doctor half, fixed).

Issue #963, opened 2026-08-14
HARNESS-072: a PR is abandoned the moment it is opened — nothing obliges an agent to stay for the review


mlorentedev:
## Context

Opening a PR is treated as the end of a change. It is not: checks report afterwards, and reviewer bots report *after that*. Nothing currently obliges an agent to still be there.

`pr-review-triage` already exists and is the right procedure — it reads CI, reads every comment, and gives each one a disposition. What it lacks is a binding trigger and a correct end condition.

## Problem 1 — the trigger fires too early

The skill's trigger is "by default once a PR you opened has finished its checks". That boundary is wrong, and there is a worked example from 2026-08-14 on PR #959:

- all checks went green
- the session would have been justified in walking away
- CodeRabbit then posted **10 comments**, including four Majors: a `| tee` pipeline discarding the reviewer's exit status, a `sh -c` fallback that cannot run on the Windows it was documented for, a pool whose deletion silently disabled a security gate, and an acceptance criterion ticked without its evidence

CI-green is not the end of the window. The reviewer bot is slower than CI, so the condition has to be *checks settled **and** the reviewer has spoken or a bounded wait expired* — whichever comes first.

## Problem 2 — nothing makes it obligatory

`pr-review-triage` is a skill: available, not binding. An agent that does not think to load it simply does not. The repo already has the mechanism for "binding on every agent regardless of harness" — an `enforced` region in `harness/manifest.json`, sourced from a vault pattern, injected verbatim into all six agent surfaces. `definition-of-done` is the proven precedent, and this is literally an elaboration of its **Review** item ("an open PR is not finished work").

## Proposal

**A. New `enforced` region — the obligation.** Terse, two rules:

1. *A PR you open is watched, not abandoned.* The window ends at the first of: an actionable comment (attend to it), or N minutes after checks settle. Pushing a fix restarts it, because the reviewer re-reviews.
2. *A PR carrying a spec folder gets an adversarial review by default.* Not a new rule so much as a name for one that already binds mechanically: spec-gate refuses to merge a PR that closes a spec's issue without archiving it, `dotf spec archive` refuses without a passing `review.md`, and (since #958) the pool refuses one signed by the wrong model. Stating it turns three mechanical refusals into one intention an agent can act on *before* being refused.

**B. Amend `pr-review-triage`** — the procedure, in the vault source:

- Correct the end condition: wait for checks **and** the reviewer bot, bounded.
[...]

PR #973, opened 2026-08-15
fix(doctor): dispatch the secrets-integrity check on backend, not on File (BUG-077)

### LINKED ISSUE #969: BUG-077: doctor's secrets-integrity check treats bw-backed entries as age-backed — 56 false FAILs, and it advises deleting the DR floor
`dotf doctor` reports **56 FAILs** in `[Secrets integrity]`, all false, and the second
half actively advises destroying the ADR-028 disaster-recovery floor.

## Symptom

```
[Secrets integrity]
  [FAIL] BITACORA_PAT -> .secret.age (missing)          # x28, one per bw-backed entry
  ...
  [FAIL] orphan: github.bitacora.secret.age (no registry entry)   # x28, one per age blob
```

Note the empty filename in the first half (`-> .secret.age`, no basename). That is the tell.

## Root cause

`checkSecrets` (`cli/internal/doctor/checks_deploy.go:116`) iterates `reg.Entries()` and
treats **every** entry as age-backed:

- it asserts `sensitive/<e.File>.secret.age` exists — but `bwEntries`
  (`cli/internal/secrets/registry.go:~360`) never sets `File`, so the path collapses to
  `sensitive/.secret.age`, which cannot exist. → 28 bogus "missing".
- it builds `referenced[e.File]`, so bw entries register the empty string instead of a real
  basename. Every age blob whose secret migrated to bw then reads as unreferenced.
  → 28 bogus "orphan".

The function's own comment asserts the opposite:

```go
// (bw-backed secrets carry no age blob, so Entries already skips them —
//  they never read as orphans here.)
```

That was true when written in **#601** (`83476da`). **#606** (`736273b`) then taught `Entries`
to include bw-backed secrets on purpose — the Loader dispatches on `Backend` — and
`checkSecrets` was never updated. `TestRegistry_Entries_IncludesBwBackend` pins the new
behaviour; nothing pinned the consumer's assumption.

It stayed latent because the registry had no bw entries. #961 (22 dev secrets) and #965
(5 file secrets) migrated 28 today, and the latent bug became 56 FAILs at once.

## Why this is worse than noise

1. **It tells you to delete the DR floor.** "orphan: ... (no registry entry)" against 28
   age blobs is an instruction to clean up. Those blobs *are* the ADR-028 DR floor for
   exactly the secrets that just moved to the Bitwarden SSOT. Followed literally, it
   destroys the recovery path for the live tier.
2. **It buries every other finding.** `doctor` exits 1 with 56 FAILs, so real regressions
   are invisible. A health check that is always red is one nobody reads.
3. It is the inverse of BUG-074 (#944): that one failed toward green, this fails toward
   red. Same family — a check measuring a proxy (`File` presence) for a property it does
   not hold (age-backedness).

## Fix

`checkSecrets` must dispatch on `e.Backend`:

- `age`/`age-offline` → current assertion, unchanged.
- `bw` → do not assert an age blob; the reach of the live tier is already covered by
  `[Bitwarden reach]`. Its age blob, where one exists, is DR floor and must be counted as
  *referenced* so it never reads as an orphan.

## Guard (incident → guard, same PR)

A doctor-level test over a registry fixture mixing `backend: age` and `backend: bw` where
the bw secret has a co-located `.secret.age` blob, asserting **zero** FAILs. Mutating the
dispatch back to unconditional must turn it red.

Related: #938 (CLI-036, no retire path — why blobs co-exist with bw entries), #964.

mlorentedev:
`dotf doctor` reported **56 FAILs** in `[Secrets integrity]`, all false, and half of them
told the reader to delete the ADR-028 disaster-recovery floor.

## Root cause

`Registry.Entries()` is a **tagged union**: #606 (`736273b`) taught it to emit bw-backed
secrets because the Loader dispatches on `Backend`, and only the age variants populate
`File`. `checkSecrets` was written in #601 (`83476da`) against the narrow type and asserts
an age blob for every entry. For a bw entry `File` is `""`, so:

- the blob path collapses to `sensitive/.secret.age`, which cannot exist → 28 bogus
  `missing`;
- the orphan index records `""` instead of a base name → every migrated secret's surviving
  blob reads as unreferenced → 28 bogus `orphan: … (no registry entry)`.

The function's own comment asserted the opposite (*"Entries already skips them"*) — true
when written, false since #606, with no test pinning the consumer's side.

It was unreachable while the registry held no bw entries. #961 and #965 migrated 28 today
and it landed as 56 FAILs at once.

## Why the orphan half is not simply inverted

`migrate` deletes the `age:` line, so nothing records which blob backs which secret, and
the names do not correlate — `OPENAI_API_KEY` (bw item `openai-api-key`) is backed by
`chatgpt.api-key.secret.age`. A basename heuristic would mis-pair recovery material, so
the check degrades to a single WARN that states it cannot classify the set and why. It
regains per-blob FAIL severity once the registry records a DR pointer — filed as #971.

`bw` is exempted **by name** rather than the age backends whitelisted: `age-offline` is a
backend too (floor plane), and a whitelist would silently stop asserting it.

## Evidence (this machine)

| | `[Secrets integrity]` | whole run |
|---|---|---|
| before | 56 FAIL | exit 1 |
| after | 0 FAIL, 1 WARN | `138 passed, 2 failed, 3 warned, 5 skipped` |

The 2 remaining failures are pre-existing and unrelated (`pi models.json` placeholder,
`orca.json` hook timeout) — they were **invisible** under the 56, which is the concrete
form of "a health check that is always red is one nobody reads".

Verification: `go build ./...`, `go vet ./...`, `go test ./...` clean; `GOOS=windows go
build ./...` OK; `golangci-lint run` → 0 issues on the pinned 2.12.2.

## Guard

`TestCheckSecrets_BwBackedEntriesAreNotAgeAsserted` — fixture mixing `age`, `age-offline`
and `bw` plus an unclaimed blob. Mutation battery **3/3 detected**, each mutant run
against that test **alone** (a whole-package run cannot say which test killed a mutant):

| mutant | result |
|---|---|
| dispatch removed (the original bug) | DETECTED |
| whitelist `age` instead of exempting `bw` | DETECTED |
| orphan severity reverted to unconditional FAIL | DETECTED |

## Not fixed here

Two sibling consumers hold the same pre-#606 assumption, filed as **#972**: the PAT-expiry
filter silently drops migrated tokens from monitoring, and `render`'s duplicate-var
fail-fast never fires for bw entries. Both live in `cli/internal/secrets/`, which the
CLI-024 worktree is editing — coordinated rather than raced.

Lesson recorded in `docs/lessons.md`.

## SDD skip rationale

Bug fix with a fully diagnosed root cause, pinned to the two commits that created the drift
(#601 vs #606). The gate counts 61 production LOC, but 32 of those are comments recording
the root cause in place — the actual behavioural change is **25 lines**: one `Backend`
check and the extraction of `reportUnreferencedBlobs`. No public contract, no dependency,
no multi-PR sequence. The verification a spec would have demanded already exists in this
PR: before/after evidence on the real machine, a named guard, and a mutation battery run
in isolation. The two larger structural questions this uncovered were split into their own
tickets (#971, #972) rather than scope-crept into a bugfix.

Closes #969


mlorentedev:
**Merged unreviewed — reviewer quota exhausted.** Recording it rather than leaving the
absence of review to look like a clean review.

CodeRabbit posted a "Review limit reached" notice instead of a review, so this PR carries
**0 reviews**. Every check was green at merge, and green checks with no comments are
indistinguishable from reviewed-and-clean from the outside — which is #906, and the reason
the disclosure exists rather than the silence.

What stands in for review here: the fix has a named guard, and the guard was observed
failing in isolation against the pre-fix implementation before being trusted. That is
evidence about the change, not a substitute for an independent reader.

Per the PR-stewardship rule this repo is adopting (#963): a notice that no review ran
leaves the PR unreviewed, and closing the window on it means saying so out loud.

Issue #976, opened 2026-08-15
BUG-080: migrating 28 secrets to bw changed their availability contract — pi/opencode/pollex cannot run without an interactive vault unlock


mlorentedev:
`pi` (and `oc`, `qq`, `agy`) stopped working after today's migration. The API keys are
correct; the vault access is the problem.

## What breaks

```
$ pi
Error: bw resolve github-bitacora-pat/api-token: bw get item "github-bitacora-pat": Vault is locked.
```

Note the failing item is not even the one `pi` needs.

## Root cause: a migration that changed the availability contract, not just storage

`pi` is a shell function wrapping `dotf secrets run -- pi`, and `NAN_API_KEY` is its
credential. Before #961 (`a42cc67`):

```yaml
- id: NAN_API_KEY
  backend: age
  age: nan.api-key
```

age resolves from `sensitive/nan.api-key.secret.age` with the key at
`~/.config/age/key.txt` — **zero prompts, works unattended, works in a detached tmux
session**. After the migration:

```yaml
- id: NAN_API_KEY
  backend: bw
  bw: { item: nan-api-key, field: api-key, folder: Dotfiles/apps }
```

bw resolves only from an **unlocked** Bitwarden vault, and unlocking is an interactive
master-password prompt. Verified on this machine: the age blob still exists and still
decrypts with no prompt (`age -d -i ~/.config/age/key.txt sensitive/nan.api-key.secret.age`
succeeds), while `bw status` reports `locked` and every `dotf secrets run` fails.

So the migration was reviewed as a *storage* change and shipped as an *availability
contract* change, for 28 secrets at once. That is the defect worth naming: nothing in the
migration's acceptance criteria asked "what can no longer run unattended after this?"

## Blast radius

Registry `consumers:` naming an agent: **pi**, **opencode**, **pollex**
(`NAN_API_KEY`, `OPENROUTER_API_KEY`, `POLLEX_API_KEY`). In practice everything routed
through `dotf secrets run` — `pi`, `oc`, `qq`, `agy` — plus `dotf spec review`, which wraps
every runner the same way. That last one matters: **`nan/deepseek-v4-flash` is the primary
entry in `harness/reviewer-pool.json`**, so the adversarial-review primary arm is down
whenever the vault is locked.

## Amplifier: no `--only` in the wrappers

The `pi`/`agy` wrappers call `dotf secrets run` with no `--only`, so a launch resolves the
[...]

Issue #585, opened 2026-06-25
dotf secrets: add the Bitwarden backend (bw serve) and migrate dev secrets to it


mlorentedev:
## Context

ADR-028 §"Phased plan" step 3 + the `bw` backend of §2. The registry facade
(#583) reads the **age** backend today; `secrets/registry.yaml` already declares
`backend:` per secret with the Bitwarden target. This issue adds the live read
path and migrates the dev/infra secrets to Bitwarden as the SSOT.

## What

1. **`bw` backend in `dotf secrets`** — resolve `backend: bw` secrets by reading
   Bitwarden via `bw serve` (local REST, one unlock per session), mapping
   `bw: { folder, item, field }` → value. `show`/`run` drop the "bw not yet
   supported" guard for migrated entries.
2. **Migrate the ~20 dev/infra secrets age→bw**, by batch, non-blocking
   (`docs/secrets-inventory.md` is the map): add to Bitwarden under `Dotfiles/*`,
   flip the registry entry `backend: age → bw`, verify `show`/`run`, then retire
   the `sensitive/<file>.secret.age` (the DR escrow still covers it).
3. **Rotate** as part of each batch where the consumer blast radius is known.

## Out of scope

- Folder taxonomy / item-naming curation, the GitHub token split (#321), the
  offline age key + DR escrow → the curation issue.
- `sync` (materialize for headless CI/containers).

## Acceptance criteria

- [ ] `dotf secrets show <bw-id>` / `run --only <bw-id>` resolve from Bitwarden via `bw serve`.
- [ ] Migrated entries flip to `backend: bw` and their age files are removed; the
      drift guard / round-trip is updated (env-mapping shrinks accordingly).
- [ ] `bw serve` lifecycle is robust (unlock failure → clear error; no plaintext to disk).
- [ ] Rotation recorded per migrated secret where the consumer is confirmed.

Refs: ADR-028 (§2, Phased plan 3), `docs/secrets-inventory.md`, #378, #493, #583.

mlorentedev:
## Motivation from the 2026-06-26 RELEASE_TOKEN/BITACORA_PAT incident

Confirmed during the live rotation: **age-backed secrets have no first-class rotate command.** `dotf secrets set` writes Bitwarden only; there is no Go (or clean shell) path to re-encrypt a `sensitive/*.secret.age`, so rotating an age secret today falls back to a manual `age -r …` — a sharp edge against the "no manual ñapas" bar.

[...]

PR #977, opened 2026-08-15
fix(shell): scope the agent secret wrappers with --only, and stop wrapping agy

mlorentedev:
`pi` started taking ~45s to launch, or hanging until interrupted. The keys were correct;
the launch path was not.

## Cause

The wrappers ran `dotf secrets run -- <agent>` with **no `--only`**, so each launch
resolved the whole registry. That was deliberate — the comment read *"No `--only` -> full
mapped set (parity with the old ambient export)"* — and it was free while every secret was
age-backed: a local decrypt, milliseconds. #961 moved 28 secrets to Bitwarden, where a
resolution is a `bw` shell-out at ~1.5-1.7s (measured on #585). The same unchanged line
went from ~0ms to ~45s per launch. On a locked vault it fails on the *first* locked entry,
which is why the reported error names `github-bitacora-pat` — a secret `pi` never uses.

Root cause is #976: the migration changed an availability contract while being reviewed as
a storage change.

## Scope, derived from what each agent reads

Not from the registry's `consumers:` field, which is already demonstrably stale — it lists
`OPENROUTER_API_KEY` for opencode only, while `~/.pi/agent/settings.json` enables
`openrouter/*` models.

| wrapper | scope | resolutions |
|---|---|---|
| `pi` | `NAN_API_KEY,OPENROUTER_API_KEY` | 28 -> 2 |
| `opencode` | `NAN_API_KEY,OPENROUTER_API_KEY,OPENAI_API_KEY` | 28 -> 3 |
| `agy` | **unwrapped** | 28 -> 0 |

`agy` authenticates with its own stored credentials and reads no variable this registry
exposes — verified against its settings files **and** the strings of the binary, rather
than inferred from metadata. It was paying a full registry resolution to be handed nothing.

`opencode.jsonc` references `{env:OLLAMA_API_KEY}`, which is not a registry id. Left out of
the scope deliberately: naming an unknown id to `--only` is an error, and adding it to the
registry is a separate decision.

## Verification

- `bash -n` and `zsh -n` over **both** rc files (each file checked in both shells).
- Added PowerShell lines are ASCII-only (the file's two pre-existing non-ASCII lines at
  117/121 are untouched).
- All three surfaces changed together — a Linux-only fix leaves Windows paying the same 45s.

**Wall-clock before/after is not included, and that is deliberate rather than an
oversight.** Measuring it requires an unlocked vault; the attempt made during this change
returned ~0.47s for both arms because the vault had re-locked and both were failing fast on
the first entry. Reporting those as a before/after would have been a measurement of the
error path. What is verified is the **resolution count**, which is counted, not estimated;
the wall-clock follows from #585's per-call figure. The live benchmark belongs with #975's
operator smoke test, which needs the same unlocked vault.

## Not fixed here

`dotf spec review` hard-wraps every runner in `dotf secrets run --` in Go
(`cli/internal/spec/review_launch.go`), so an adversarial review still pays the full
resolution. Deliberately untouched: it is HARNESS-071 territory and that spec's review of
record was running while this was written. Follow-up is an optional `secrets: []` per
reviewer-pool entry — absent means bare invocation, which is exactly `agy`'s case.

## SDD skip rationale

Configuration fix with a fully diagnosed root cause, pinned to the commit that caused it
(#961) and to the measured per-call cost that made it expensive. No production Go LOC, no
public contract, no new dependency. The behaviour change is three wrapper lines; everything
else in the diff is the comment explaining why the unscoped form was correct before and is
not now.

Refs #976

### DIFF
--- a/.bashrc
+++ b/.bashrc
@@ -75,15 +75,25 @@ export NAN_BASE_URL="https://api.nan.builders/v1"
 
 # Secrets are NOT auto-loaded into the ambient shell (ADR-028 "not always
 # exposed"). On demand: `dotf secrets run -- <cmd>` injects the decrypted secrets
-# into that child process only. The AI CLIs that read their keys from the
-# environment are wrapped to launch through it, so the secret lives only in their
-# process, never this shell. No `--only` -> full mapped set (parity with the old
-# ambient export). Recursion-safe: dotf resolves the real binary on PATH, not
-# this function.
+# into that child process only, so a key lives in the agent's process and never
+# in this shell. Recursion-safe: dotf resolves the real binary on PATH, not this
+# function.
+#
+# Each wrapper is scoped with `--only`, and the scope is load-bearing. The
+# unscoped form resolved the WHOLE registry on every launch — deliberate parity
+# with the old ambient export, and free while every secret was age-backed, which
+# is a local decrypt in milliseconds. After the Bitwarden migration (#961) each
+# resolution is a `bw` shell-out at ~1.5s, so an unscoped launch paid ~45s of
+# latency before the agent started, and died on the first locked entry even when
+# the agent had no use for it (#976). Scoping is least privilege and startup
+# time at once.
 if command -v dotf >/dev/null 2>&1; then
-    opencode() { dotf secrets run -- opencode "$@"; }
-    pi() { dotf secrets run -- pi "$@"; }
-    agy() { dotf secrets run -- agy "$@"; }
+    opencode() { dotf secrets run --only NAN_API_KEY,OPENROUTER_API_KEY,OPENAI_API_KEY -- opencode "$@"; }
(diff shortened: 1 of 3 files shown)


mlorentedev:
**Merged unreviewed — reviewer quota exhausted.** Recording it rather than leaving the
absence of review to look like a clean review.

CodeRabbit posted a "Review limit reached" notice instead of a review, so this PR carries
**0 reviews**. Every check was green at merge, and green checks with no comments are
indistinguishable from reviewed-and-clean from the outside — which is #906, and the reason
the disclosure exists rather than the silence.

What stands in for review here: the fix has a named guard, and the guard was observed
failing in isolation against the pre-fix implementation before being trusted. That is
evidence about the change, not a substitute for an independent reader.

Per the PR-stewardship rule this repo is adopting (#963): a notice that no review ran
leaves the PR unreviewed, and closing the window on it means saying so out loud.

Issue #993, opened 2026-08-15
BUG-084: dotf secrets set is unusable under the sanctioned unlock model — the write path never moved to bw serve


mlorentedev:
## What

`dotf secrets set` cannot write anything on a machine unlocked the way ADR-028 and CLI-024 prescribe.

Observed 2026-08-15, trying to rotate a credential:

```
$ dotf secrets set DOCKERHUB_TOKEN
Value (hidden):
Error: bw get item "dockerhub" (it must already exist to edit): Vault is locked.

$ bw unlock
? Master password: [hidden]
Your vault is now unlocked!

$ dotf secrets set DOCKERHUB_TOKEN
Value (hidden):
Error: bw get item "dockerhub" (it must already exist to edit): Vault is locked.
```

## Cause

#975 moved the **read** path to the `bw serve` daemon: `dotf secrets unlock` starts it, and `run`/`show`/`verify` resolve through it with no ambient `BW_SESSION`, which is the whole point of ADR-028.

The **write** path did not move. `dotf secrets set` uses `secrets.BWPut{}`, which shells out to the `bw` CLI (`cli/internal/secrets/bw.go`), and that binary authenticates from `BW_SESSION` in its environment. Under the sanctioned model that variable is deliberately absent, so every write fails — regardless of whether the daemon is unlocked, and regardless of whether the operator just ran `bw unlock`.

`bw unlock` does not help on its own because it only **prints** the session key; the calling shell never receives it. That makes the failure look like the unlock did not work, when in fact it worked for a subject nobody is asking.

## Blast radius

Every write-side subcommand that goes through `BWPut` is affected, which is the entire provisioning and rotation story: `set`, and by extension the "write the value" step of the documented add-a-secret workflow in `.claude/CLAUDE.md`. Rotating a leaked or expired credential — the case where friction matters most — currently requires either an ambient `BW_SESSION` (the thing ADR-028 removed) or editing in the Bitwarden UI and syncing the daemon by hand.

## Workaround

Confines the session to one process rather than exporting it, which preserves ADR-028's intent:

```
BW_SESSION="$(bw unlock --raw)" dotf secrets set <ID>
```

`BWPut` builds its command with `exec.Command` and never sets `cmd.Env`, so the child inherits it.

## Fix

[...]

Issue #983, opened 2026-08-15
OPS-029: the Bitwarden copy of BITACORA_PAT is a dead token (HTTP 401), and has been unmonitored since it migrated


mlorentedev:
## What

The Bitwarden-backed copy of `BITACORA_PAT` does not authenticate. Confirmed by two independent paths on 2026-08-15:

```
$ dotf secrets run --only BITACORA_PAT -- sh -c 'curl -s -o /dev/null -w "%{http_code}" -H "Authorization: Bearer $BITACORA_PAT" https://api.github.com/user'
401
```

and, from the repaired PAT-expiry check on the REFACTOR-012 branch:

```
[PAT expiry]
  [ OK ] github.token: valid, expires in 40 day(s) (2026-09-25)
  [FAIL] github-bitacora-pat/api-token: token invalid or expired (HTTP 401) — rotate it
```

## Impact — bounded, but the bound is a guard, not luck

- **CI is unaffected.** `Add to bitácora` and `Reconcile bitácora` both ran green today, so the copy GitHub Actions holds is a *different*, working token. The two tiers have silently diverged.
- **`dotf secrets sync ci` would refuse**, not corrupt: the `validate: github-token` pre-upload gate (`secrets_sync.go`) rejects a dead credential. That gate exists *because of this exact secret* — its code comment names "the BITACORA_PAT incident — a redeploy refreshed `updated_at` on a 401 token".
- **Local use fails**, silently, wherever the bw copy is the one resolved.

So the working state depends on nobody rotating the Actions secret from the local copy, and on the sync gate holding. Neither is a reason to leave a dead credential in the vault.

## Why it went unnoticed

This is the demonstration case for #972. The PAT-expiry check was supposed to catch exactly this, and it could not: it selected secrets by the age blob's `github.` filename prefix, and a bw-backed secret has no filename. `BITACORA_PAT` became invisible to its own monitor the moment it migrated to Bitwarden (#961) — and the check was independently dead anyway, reading the token from an ambient environment ADR-028 guarantees is empty.

The monitoring gap is fixed on `fix/refactor-012-backend-tagged-union`. **This ticket is the credential itself**, which no code change rotates.

## Acceptance criteria

- [ ] Determine whether the Actions secret and the Bitwarden item were ever the same token, or diverged at a rotation.
[...]

Issue #586, opened 2026-06-25
Secrets curation: Bitwarden folders, per-purpose token split, offline age key + DR escrow


mlorentedev:
## Context

ADR-028 §"Phased plan" step 4 + §6 (curation & naming). The store today is a flat
~145-item Bitwarden vault with zero folders, plus a blocking finding: the **age
private keys live inside Bitwarden** (a circular DR dependency). This issue does
the organization the inventory already decided.

## What

1. **Bitwarden folder taxonomy** — introduce `Dotfiles/{apps,infra,floor}`; move the
   ~20 dev/infra items in (leave the ~125 personal items as-is, optionally foldered later).
2. **Per-purpose token split (#321)** — split the GitHub 8-in-1 item into
   `github-cli-pat`, `github-release-pat`, `github-bitacora-pat`, `github-runner-token`,
   `github-kubelab-dispatch-token`, and the PEM → `github-evalkit-ci-key`.
3. **Naming-drift fixes** — `chatgpt.api-key` → `openai-api-key`, the double-dot
   `openrouter.api.key`; merge intra-vault duplicates (the two OpenRouter entries);
   collapse multi-part creds into one item + custom fields (the 7 X values).
4. 🔴 **age key offline-rooted** — move the authoritative `AGE-SECRET-KEY-*` copy
   OFFLINE (encrypted USB + paper + keychain); the Bitwarden copy becomes convenience
   only. Fixes the circular dependency (#518).
5. **DR escrow** — `dotf secrets backup`: `bw export --raw | age -e` → `sensitive/dr/`,
   committed + mirrored off-box (#454); the recover runbook (#257).

## Out of scope

- The `bw` backend + migration mechanics → the bw-backend issue (#585).

## Acceptance criteria

- [ ] All managed items live under `Dotfiles/{apps,infra,floor}`; registry `bw.folder` matches.
- [ ] GitHub token split per purpose (#321); registry ids reflect it; consumers repointed.
- [ ] Naming-drift fixed; duplicates merged; multi-part creds collapsed to fields.
- [ ] Authoritative age key is OFFLINE; Bitwarden holds a convenience copy only.
- [ ] `dotf secrets backup` produces an age-encrypted full-vault escrow, off-box mirrored; recover runbook tested.

Refs: ADR-028 (§6, Phased plan 4), `docs/secrets-inventory.md`, #321, #257, #454, #518.

mlorentedev:
[...]

PR #1006, opened 2026-08-16
fix(doctor): key DR escrow severity to real exposure, not a flat policy

### LINKED ISSUE #997: BUG-085: doctor reports a disaster-recovery escrow that has NEVER existed as SKIP, so nothing has ever warned that there is no backup
## What

`checks_dr.go:47`:

```go
rep.Skip("no DR escrow at " + escrow + " — `dotf secrets backup` has not run (#586 item 5)")
```

An absent escrow is a SKIP. On this machine `sensitive/dr/bitwarden-export.age` has **never existed** — `dotf secrets backup` has never completed successfully, because it is broken (#993: `bw export` goes through the locked CLI session). So `dotf doctor` has been run repeatedly, has never complained, and the machine has no disaster-recovery copy at all.

This is the same defect class fixed this morning in the PAT-expiry check (#972): a check whose only reachable branch is the no-op one, where SKIP is indistinguishable from "nothing to check here".

## Why it is not merely cosmetic now

The severity was defensible when every secret had an age copy: the escrow was a convenience and its absence cost nothing.

That is no longer true. `migrate` drops the `age:` pointer when it flips a secret to Bitwarden (#971), so **28 migrated secrets have no age copy the registry still claims**. For those, the Bitwarden server is the only copy in existence. An absent escrow is now a single point of total loss, reported as "nothing to check".

Today's session made that concrete: a credential was rotated into Bitwarden, so even the unmapped age blobs left on disk are now stale for that secret. Recovery from a Bitwarden account loss would be partial at best and nobody would know which parts.

## The rule this check should already have had

`checkBitwardenReach` in the same package solves exactly this problem correctly, and says so:

> With every registry entry still on `backend: age`, an unreachable Bitwarden breaks nothing, and a red doctor for a harmless condition is how operators are trained to ignore red doctors. From the first `backend: bw` entry onward the same state is a hard failure, so it FAILs.

The DR check needs the same treatment: severity keyed to real exposure, not a flat policy. It already has the input — `sys.BWBackedSecrets()`.

## Proposed severity

| State | Now | Should be |
|---|---|---|
| No escrow, **no** bw-backed secrets | SKIP | SKIP (nothing depends on it) |
| No escrow, bw-backed secrets exist | **SKIP** | **FAIL** — the only copy of N secrets is a remote account |
| Escrow present but stale | WARN | WARN (correct) |
| Escrow present and fresh | PASS | PASS |

The FAIL message must name the count and the command, and must not name a command that cannot currently run — so it depends on #993, or it says so explicitly.

## Acceptance criteria

- [ ] With ≥1 `backend: bw` secret and no escrow file, the section FAILs and names how many secrets have no local copy.
- [ ] With no bw-backed secrets and no escrow, it still SKIPs — the check must not become noise on a machine where the escrow genuinely does not matter yet.
- [ ] The failure is observed red against real state before the fix is claimed, not only against a fixture.
- [ ] The recovery-drill staleness branches are left alone; they are already correct.

## Related

- #993 — why `dotf secrets backup` cannot run at all, which is why this was never noticed
- #971 — `migrate` dropping the `age:` pointer, which is what turned an absent escrow from an inconvenience into a single point of loss
- #972 — the same "SKIP reads as nothing-to-check" class, fixed this morning in the PAT check
- `cli/internal/doctor/checks_bw_reach.go` — the severity-keyed-to-exposure rule this check should adopt, already written down there

mlorentedev:
## What

An absent disaster-recovery escrow was reported as `SKIP` unconditionally. Severity is now keyed to real exposure — `SKIP` while nothing resolves through Bitwarden, `FAIL` from the first `backend: bw` entry onward — which is the rule `checkBitwardenReach` already applies in this same package.

## Why the old severity was defensible and stopped being so

The `SKIP` carried its own justification in a comment: *"#586 item 5 has not shipped, so its absence is a known state rather than a regression."* That was true while every registry entry still carried an `age:` pointer — each secret had a local encrypted copy, so a missing escrow cost nothing.

Then `migrate` began dropping that pointer as it flipped entries to Bitwarden (#971). For those secrets the remote account became **the only copy in existence**. A missing escrow stopped being an inconvenience and became a single point of total loss — and was still reported as "nothing to check here". The justification expired silently; nothing re-read it.

On this machine that state persisted with **28 secrets exposed**, and `dotf doctor` never once complained. That is the same class as #972 and #898: a check whose only reachable branch is the no-op one proves nothing.

## The unknown-exposure branch

When the registry cannot be read, exposure is unknown, and both obvious answers are wrong:

- Treating it as unexposed (the reach check's "degrade to advisory") prints `no secret resolves through Bitwarden yet, so nothing depends on it` — the exact sentence this issue exists to delete — on a machine where it was never checked.
- Failing goes red from any healthy machine: `env.RepoDir` falls back to a cwd walk-up for `.git`, so `dotf doctor` run from inside any other git repo lands here.

It `WARN`s instead, naming what it could not determine.

## The remedy string names `BW_SESSION` plainly, not as a workaround

The issue asked for a message that does not name a command which cannot run, and anticipated depending on #993. That dependency dissolved on inspection: **`bw serve` exposes no export route.** Enumerated from the shipped `@bitwarden/cli` bundle, the serve router has `/status`, `/sync`, `/unlock`, `/lock`, `/generate`, `/list/object/:object`, `/object/:object[/:id]`, `/attachment`, `/move`, `/send/*`, `/device-approval/*` — and no export. The only `"/export"` string in the bundle is an *organization* export against the cloud API, not a personal-vault serve route.

So `bw export` has no daemon path to migrate to, `dotf secrets backup` needs a CLI session for as long as that holds, and citing #993 would have dated the message the day it merged.

## Evidence

Against the **real** registry (28 `backend: bw` entries), escrow absent, real code path — not a fixture:

```
# new binary
[Disaster recovery]
  [FAIL] no DR escrow at .../sensitive/dr/bitwarden-export.age — 28 secret(s) resolve
         through Bitwarden and have no local copy, so the remote account is their only
         copy; create one with `BW_SESSION="$(bw unlock --raw)" dotf secrets backup`
         (the export path has no bw serve endpoint, so it needs a CLI session)

# released 0.41.0, same state
  [SKIP] no DR escrow at ... — `dotf secrets backup` has not run (#586 item 5)
```

The tests were also observed failing against the pre-fix check, so they are not passing vacuously:

```
--- FAIL: TestDR_AbsentEscrowWithExposure_Fails
    absent escrow with bw-backed secrets must FAIL, got 0 failure(s)
```

`go build`, `go vet`, `go test ./internal/doctor/` and `golangci-lint run` (v2.12.2, the `versions.conf` pin) all clean.

**Honest limit on the exit-code claim:** doctor already exits 1 on this machine from a pre-existing unrelated `[PAT expiry]` FAIL (#983, the dead `BITACORA_PAT`), so the exit-code effect of this change is masked here. What was observed directly is the section severity. `setup-linux.sh:1505` invokes doctor as `dotf doctor || log_warning`, so the new FAIL cannot break a bootstrap.

## Scope

The escrow-present, stale-escrow and recovery-drill branches are unchanged, as the issue's acceptance criteria require.

## SDD skip rationale

Bug fix with a fully diagnosed root cause, on an issue that already carries the design: #997 specifies the severity table and the acceptance criteria this PR implements verbatim, which is the substance a `proposal.md` would have restated. The gate counts 87 production LOC, but **66 of those are comments and blank lines** recording why the old severity expired — the actual behavioural change is **21 lines**: one `BWBackedSecrets` call and the extraction of `reportAbsentEscrow`. No public contract, no dependency, no multi-PR sequence. The verification a spec would have demanded is already here: before/after evidence against the real registry, the tests observed red against the pre-fix code, and the unchanged branches pinned by tests. Same shape and same escape as #973, a `fix(doctor):` bug fix of this class.

Closes #997

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved disaster recovery diagnostics for missing Bitwarden escrow.
  * Reports a warning when registry exposure cannot be determined.
  * Reports a failure when Bitwarden-backed secrets are exposed, including the affected count and `BW_SESSION` remediation guidance.
  * Skips the check when no Bitwarden-backed secrets are present.
  * Continues to report stale escrow as a warning, preserving existing recovery checks.

### DIFF
--- a/cli/internal/doctor/checks_dr.go
+++ b/cli/internal/doctor/checks_dr.go
@@ -66,7 +66,7 @@ func reportAbsentEscrow(rep *Report, escrow string, live int, regErr error) {
 		return
 	}
 	if live > 0 {
-		// The count carries the argument. "No backup" is abstract; "26 secrets
+		// The count carries the argument. "No backup" is abstract; "28 secrets
 		// exist only on a remote server you do not control" is what makes an
 		// operator act.
 		//
@@ -122,8 +122,19 @@ func checkDisasterRecovery(sys *System, cfg *Config, rep *Report) {
 
 	escrow := filepath.Join(cfg.DotfilesDir, "sensitive", "dr", "bitwarden-export.age")
 	switch info, err := os.Stat(escrow); {
-	case err != nil:
+	case os.IsNotExist(err):
 		reportAbsentEscrow(rep, escrow, live, regErr)
+	case err != nil:
+		// A stat error is NOT proof of absence, and the distinction only started
+		// to matter with this change: as a SKIP, treating "cannot read" as
+		// "not there" was harmless, but as a FAIL it asserts something the check
+		// never established — "these 28 secrets have no local copy" — when the
+		// escrow may be sitting right there behind a permission or I/O error.
+		// Claiming an unverified fact is the exact failure this whole check
+		// exists to stop, so it must not be committed in the fix for it.
+		rep.Warn(fmt.Sprintf(
+			"cannot inspect DR escrow at %s (%v) — neither its presence nor its freshness was established",
+			escrow, err))
 	case sys.Now().Sub(info.ModTime()) > escrowMaxAge:
 		rep.Warn(fmt.Sprintf("DR escrow is %d days old — re-run `dotf secrets backup`; secrets added since are not in it",
 			int(sys.Now().Sub(info.ModTime()).Hours()/24)))
--- a/cli/internal/doctor/checks_dr.go
+++ b/cli/internal/doctor/checks_dr.go
@@ -35,16 +35,95 @@ func drillMarkerPath(cfg *Config) string {
 	return filepath.Join(cfg.DotfilesDir, ".dr-drill")
 }
 
+// reportAbsentEscrow renders the missing-escrow state at the severity its real
+// exposure earns: `live` is how many registry entries resolve through Bitwarden,
+// and regErr is non-nil when the registry could not be read to find out.
+//
+// Split out from checkDisasterRecovery so the severity rule — the whole point of
+// #997 — is testable on its own, the same way checkBWSyncAge is split out of the
+// reach check.
+//
+// No rep.Fix here, deliberately: this check performs no repair, and the
+// neighbouring reach check documents what happens when a remediation string is
+// emitted as a Fix — a read-only run reports "Applied 1 fix action(s)" for a
+// repair that never happened. The command belongs in the message.
+func reportAbsentEscrow(rep *Report, escrow string, live int, regErr error) {
+	if regErr != nil {
+		// Exposure is UNKNOWN, and the two wrong answers are both tempting.
+		//
+		// Silence (the reach check's "degrade to advisory, treat as unexposed")
+		// would emit a SKIP reading "nothing depends on it" — precisely the
+		// sentence #997 exists to delete, now printed on a machine where it was
+		// never checked. FAIL would go red from any healthy machine: env.RepoDir
+		// falls back to a cwd walk-up for .git, so `dotf doctor` from inside any
+		// other git repo lands here, and a doctor that is red where nothing is
+		// wrong is one nobody reads.
+		//
+		// So: audible, but honest about what it does not know.
+		rep.Warn(fmt.Sprintf(
+			"no DR escrow at %s, and the registry could not be read (%v) — cannot tell whether any secret depends on it",
+			escrow, regErr))
+		return
+	}
+	if live > 0 {
+		// The count carries the argument. "No backup" is abstract; "26 secrets
+		// exist only on a remote server you do not control" is what makes an
+		// operator act.
+		//
+		// The BW_SESSION form is named plainly, because for `backup` it is not a
+		// workaround — it is the invocation, permanently.
+		//
+		// Everything else in the write path is moving to the bw serve daemon
+		// (#993), and it was tempting to phrase this as "broken until that
+		// lands". It is not: `bw serve`'s router exposes /status, /sync, /unlock,
+		// /lock, /generate, /list/object/:object, /object/:object[/:id],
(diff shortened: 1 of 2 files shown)


mlorentedev:
Both findings applied in 8791086 — dispositions:

**Major (`checks_dr.go:126`, stat error read as absence) — applied, and it was mine.** Correct and worth stating plainly: the flaw pre-existed, but this PR is what made it bite. As a SKIP, treating "cannot read" as "not there" cost nothing; as a FAIL it asserts *"these 28 secrets have no local copy"* — a claim the check has not earned when the escrow may be sitting there behind a permission or I/O error. Shipping an unverified assertion inside the fix for unverified assertions would have been the wrong thing twice.

Now only `os.IsNotExist` reaches the absent branch (`os.IsNotExist` being the idiom already dominant in `cli/`); every other stat error WARNs that neither presence nor freshness was established. Pinned by a test observed failing against the pre-fix code:

```
[FAIL] no DR escrow at .../bitwarden-export.age — 28 secret(s) resolve through
       Bitwarden and have no local copy ...
```

— emitted for a file that **exists**. That is the false FAIL the finding predicted.

**Minor (registry-error branch untested) — applied.** The three severity branches are now table-driven and assert the stable status tag rather than prose, per the repo guideline: the tag is the contract (it decides doctor's exit code), the wording is not. Remediation text stays pinned only for the FAIL case, where the message *is* the deliverable.

One thing the suggestion did not ask for, added because the tests needed it to be honest: assertions are scoped to the escrow line via a small helper. The section emits an independent drill line that also WARNs, so a whole-buffer `Contains("[WARN]")` would pass while the escrow line said something entirely different — the test would have been green for the wrong reason.

The stat-error case is gated off Windows (`chmod(0)` does not deny the owner there, and the alternative shapes map to path-not-found, which would silently exercise the ABSENT branch instead) and off root, for the same reason.

`go build`, `go vet`, `go test ./...` and `golangci-lint run` (v2.12.2, the `versions.conf` pin) all clean.

Issue #612, opened 2026-06-26
secrets lifecycle: fail-loud hardening + reproducible CLI write/migrate/rotate (audit backlog)


mlorentedev:
Backlog from a 3-agent read-only audit of the `dotf secrets` subsystem (code + ADR-028 + runbook + registry), before migrating the ~20 dev secrets age→bw. Goal: the full lifecycle (read/write/delete/rotate/migrate/verify) reproducible, idempotent, fully CLI-driven, across all secret shapes (token, multi-field, multi-line text — SSH/kubeconfig are TEXT, no attachments).

> **Status reconciled 2026-06-27** against shipped code (commit `9a51faf`). Phase A is complete (#613); Phase C is mostly complete (#616/#620/#621/#627/#632/#634). The checklist below now reflects the code, not the original audit snapshot. Remaining Linux-doable work: Phase B (B1/B2/B4/B5/B6) + C6/C7 + Phase D. Needs bw/Windows: C8, C9. Spec: `specs/CLI-024-secrets-fail-loud` (Phase A).

## Phase A — correctness/security bugs in shipped code (fix first; fail-loud) — DONE (#613)

- [x] **A1 — empty resolved value injected/printed silently.** `EnvFor` now fails fast on an empty env/file value; `show` inherits it; `fieldFromItem` errors on a typed field against a non-login bw item. (#613)
- [x] **A2 — `render` swallows the real error and exits 0.** `ErrSecretAbsent` sentinel + `render.Result` splits `Missing` (quiet) from `Unresolved` (loud, carries the error); per-var error surfaced; `--strict` exits non-zero. (#613)
- [x] **A3 — `--only ","` (all-empty tokens) silently selects nothing.** `resolveOnly` errors when an explicit `--only` resolves to zero secrets. (#613)
- [x] **A4 — `buildChildEnv` leaked `BW_SESSION`/`AGE_KEY_PATH` into the child.** `stripBackendAuth` closed it on main. (#610/#609)

## Phase B — validation / resolution hardening

- [ ] **B1 — global VAR uniqueness + render dedup is age-only.** bw entries have `File==""`, so render's `prev.File != e.File` guard never fires for bw → duplicate var silent last-wins; `validate()` only checks duplicate **id**, never VAR uniqueness. Enforce at parse time; key render dedup on resolved source identity.
[...]

PR #1007, opened 2026-08-16
fix(secrets): give the write path the bw serve seam the read path already had

mlorentedev:
Refs #993 (BUG-084) — **closing keyword deliberately withheld until the spec is archived in this same PR.**

The SDD lifecycle ends by archiving `specs/BUG-084-secrets-write-bw-serve/` here, and `dotf spec archive` requires a passing independent adversarial review (non-Anthropic reviewer pool — the reviewer must not be the implementer, and Claude implemented this). That review runs against the final SHA. Once it returns PASS the spec is archived and this line becomes `Closes #993`.

## The bug

`dotf secrets set` could not write anything on a machine unlocked the way ADR-028 and CLI-024 prescribe. CLI-024 moved the **read** path to the `bw serve` daemon and deferred the write path as "a natural fast-follow". Nothing enforced that, so `BWPut` kept shelling out to a `bw` binary that authenticates from `BW_SESSION` — the ambient variable ADR-028 exists to abolish.

The result was worse than a plain failure: reads resolved fine against an unlocked daemon while writes reported `Vault is locked.`, and `bw unlock` appeared not to work, because it prints a session key to a shell that is not the one being asked. Four separate "which session is locked?" confusions in one day trace to this.

## Scope, corrected on evidence

The issue named one seam (`SetField`). There were **four**, all shellout-only — `SetField`, `CreateItem`, `ResolveFolder`, `Export`. Fixing only the first would have left `set`'s create-a-new-item path still broken.

The issue also assumed `backup` could be fixed alongside. **It cannot.** Enumerated from the shipped `@bitwarden/cli` 2026.5.0 bundle, `bw serve` exposes no export route; the sole `"/export"` string in it is `apiService.send("GET", "/organizations/" + organizationId + "/export")` — an *organization* export against the cloud API. Full router table on #993 (comment 5304815767), independently re-verified by the session on #997.

## What changed

- **`BWServeWriter`** — `SetField` / `CreateItem` / `ResolveFolder` against `PUT /object/item/:id`, `POST /object/item`, `GET /list/object/folders` + `POST /object/folder`. Reuses the same pure cores as `BWPut` (`setItemField`, `newItemBody`), so an item written through either backend is byte-identical — asserted, not assumed.
- **`BWBackend` + `SelectBWBackend`** — the daemon is probed **once per command** and both halves come from that one decision. `bwRead()` / `bwWrite()` replace the two independently-initialised vars.
- **Lock hints** — a locked-vault failure now names `dotf secrets unlock` and reports the daemon state actually observed at selection time.
- **`exportLockHint`** — `backup` says *why* the daemon cannot serve it and names `BW_SESSION="$(bw unlock --raw)" dotf secrets backup`, explicitly stating that `dotf secrets unlock` will **not** fix it.

## The design decision, and why

Pinning the backend per command rather than per call was the operator's call. The read path's `BWFallbackReader` re-probes on every `Field()`: it self-heals if the daemon appears mid-run, but a daemon that *locks* between two calls of one command splits that command across two subjects — a narrower instance of this very bug. For `rotate` (read old → write new → read back), a consistent subject beats self-healing: failing whole is better than half-applying.

## Evidence

**AC4, live, before and after** — with `BW_SESSION` unset and the daemon **running and unlocked**, which is precisely why the old message misled:

```
######## BEFORE - installed 0.41.0 ########
Error: backup: bw export: Vault is locked.

######## AFTER - this branch ########
Error: backup: bw export: bitwarden vault is locked: the escrow cannot go through the
bw serve daemon - `bw serve` exposes no export endpoint, so this is the one
`dotf secrets` operation that needs the bw CLI's own session. `dotf secrets unlock`
will NOT fix it. Run:
    BW_SESSION="$(bw unlock --raw)" dotf secrets backup
```

The "before" line is the message under which the DR escrow silently never existed (#997).

**AC5, mutation-checked** — a guard never observed failing is not evidence (#898). The split-brain was reintroduced deliberately:

```
--- FAIL: TestSelectBWBackend_ReadAndWriteAlwaysAgree/unlocked_daemon_is_used_for_both_halves
    bwbackend_test.go:68: SPLIT BRAIN: reader daemon=true but writer daemon=false
        - this is exactly BUG-084
```

**Suite** — `go build ./... && go vet ./... && go test ./...` all green; `golangci-lint run` at the pinned 2.12.2 reports 0 issues.

## Gap, stated rather than hidden

**AC1/AC2 have no live write yet.** They are proven against an httptest double of the daemon, not against Bitwarden. Closing them means writing to a real vault item — an operator decision, not an agent's, so it was deliberately not done unilaterally. The repo convention for this is the canary item (#612 C8). Reviewers should treat the daemon write path as unit-verified and not yet smoke-verified.

## Follow-up filed separately

Assembling the escrow from `/list/object/items` + `/list/object/folders` would *probably* yield an importable document. *Probably* is not a property a disaster-recovery escrow may have, so it is not in this PR; it needs a real `bw import` round-trip as its acceptance criterion.

### DIFF
--- a/cli/internal/cmd/secrets.go
+++ b/cli/internal/cmd/secrets.go
@@ -105,6 +105,16 @@ func bwWrite() bwWriteClient {
 	return bwBackend().Writer
 }
 
+// bwSync returns the cache-refresh half of the SAME pinned backend. Sync is pinned with
+// the other two because the daemon and the CLI cache independently: syncing one leaves
+// the other stale, so a rotation that syncs the wrong subject cannot prove its own write.
+func bwSync() daemonSyncer {
+	if bwSyncer != nil {
+		return bwSyncer
+	}
+	return bwBackend().Syncer
+}
+
 // secretLoader builds the resolution engine wired with both backend seams, over the
 // age store in sensitive/ (checkout-first, ADR-030 — same source as the registry it
 // maps, so a repo-side rotation is seen without a redeploy) and Bitwarden via bwReader.
@@ -113,7 +123,7 @@ func secretLoader() *secrets.Loader {
 		SecretsDir: env.ResolveSensitiveDir(),
 		KeyPath:    ageKeyPath(),
 		Decrypt:    ageDecryptor,
-		BW:         bwReader,
+		BW:         bwRead(),
 	}
 }
 
--- a/cli/internal/cmd/secrets.go
+++ b/cli/internal/cmd/secrets.go
@@ -8,6 +8,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/mlorentedev/dotfiles/cli/internal/env"
 	"github.com/mlorentedev/dotfiles/cli/internal/secrets"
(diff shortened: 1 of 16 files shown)


mlorentedev:
@coderabbitai review

Issue #998, opened 2026-08-15
BUG-093: the spec archive gate refuses every reviewed spec — its draft-tag scan reads the reviewer's own transcript, and a reviewer-demanded contract fix invalidates its own review


mlorentedev:
Two independent defects in `dotf spec archive`'s pre-flight, both hit while archiving a spec that had a **passing** review. Filed together because they are the same shape: the gate cannot distinguish the artifacts it governs from the evidence that it worked.

---

## 1. The `[AGENT-DRAFT]` scan reads `review-transcript.jsonl`

The tag pre-flight scans every file in `specs/<id>/`, including the transcript `dotf spec review` writes there. The reviewer's own chain of thought contains the literals, because the skill instructs it to check for them:

```
review-transcript.jsonl:1087: {"thinking":"... 4. Look for `[AGENT-DRAFT]` or `[AGENT-SUGGESTION]` tags ..."}
review-transcript.jsonl:1769: {"thinking":"... 3. Look for `[AGENT-DRAFT]` ..."}
```

The refusal listed dozens of such hits. **The spec's actual artifacts were clean** — `grep -rln 'AGENT-DRAFT\|AGENT-SUGGESTION'` over `proposal.md`, `tasks.md`, `verification.md`, `features.json` returned nothing.

This is a catch-22, not an edge case: the gate requires a review, the reviewer is told to check for draft tags, checking writes the literals into the transcript, and the transcript sits in the scanned directory. **Every spec reviewed by this tool refuses to archive** unless `--force-with-drafts` is passed — which would also disable the real check.

**Fix:** exclude `review-transcript.jsonl` from the tag scan (it is a machine log, not a spec artifact), or scan only the known contract files. Worked around here by moving the transcript out of the folder before archiving.

## 2. A review that demands a contract-file change invalidates itself

The freshness check refuses when `proposal.md` / `tasks.md` / `features.json` changed after `reviewed_sha` — correctly, since it exists to stop someone editing the contract after the reviewer read it.

But round 2's own Minor finding was:

> AC4 checkbox says "6/7 done" but verification.md confirms 7/7 [...] AC8 checkbox is still `[ ]` but the rotate_note was updated. The checkboxes in proposal.md don't reflect the current state. **Fix location: spec artifacts (proposal.md — tick the boxes)**

[...]

Issue #988, opened 2026-08-15
BUG-082: bw serve returns non-JSON on /object/item/<id> under batch reads, so secrets verify is unusable and non-deterministic


mlorentedev:
## Symptom

`dotf secrets verify` reports a **different result on every run**, against an unchanged vault and an unlocked daemon:

```
run 1:   8 ok, 0 missing, 25 failed
run 2:  12 ok, 0 missing, 21 failed
```

The failures are all the same shape:

```
FAILED  NAN_API_KEY  bw: bw resolve nan-api-key/api-key: bw serve get item "nan-api-key":
        GET /object/item/0e947eca-...: bw serve returned no parseable envelope:
        invalid character 'I' looking for beginning of value
```

`'I'` as the first byte means the daemon answered with something that is not JSON at all (plain text — `Internal Server Error` or similar), not an error envelope.

## What is and is not affected

- **Single scoped reads are reliable.** `dotf secrets run --only NAN_API_KEY -- ...` succeeded 3/3 immediately after a `verify` in which that same secret was reported `FAILED`.
- The **search** path is fine: absent items return a clean parseable `not found` (that is how #985 was diagnosed), so it is specifically `GET /object/item/<id>` that degrades.
- Failures correlate with **batch volume** — `verify` walks all 33 registry entries back to back.

So the secret is fine, the registry is fine, the daemon is unlocked and reachable. Only the batch read path is unreliable.

## Why it matters more than a flaky command

`verify` is the health check. A health check that reports 21 false failures is worse than none: it trains the operator to disregard it, and it cannot distinguish real drift (#985's genuinely missing `dockerhub`) from noise. In the run above, the one true failure was buried among 20 spurious ones.

It also very likely explains an unresolved report from an earlier session: *"Bitwarden's vault auto-relocked 3+ times within single-digit minutes, no explicit `bw lock`/`logout` found anywhere in dotfiles."* Nothing was relocking. Reads were failing under batch load and the error surfaced as a lock-shaped symptom. That dead lead can be closed.

## Likely relation to #979

`fa21694 fix(secrets): parse bw serve's /status template-wrapped shape (#979)` fixed exactly this class of bug — a `bw serve` response that is not the envelope the client expects — on the `/status` endpoint. The item endpoint appears to have the same defect and was not covered.

## Suggested direction

1. Capture the raw non-JSON body rather than only reporting the parse error, so the actual daemon response is known instead of inferred from its first byte.
[...]

Issue #786, opened 2026-08-07
TOOL-013: PR-Agent on NaN inference as a uniform reusable review workflow


mlorentedev:
## Goal

Stand up **PR-Agent** (qodo-ai/pr-agent, MIT) as a reusable workflow in this repo, backed by NaN cluster inference, applied **uniformly across all personal repos**. Motivation: the free CodeRabbit tier gives line-by-line reviews only on public repos; three private repos (resume, biwenger-mvp, knowledge) get summaries only, and a merge burst has already exhausted its per-developer rate limit once.

Prior art: `research-pr-review-agents.md` (2026-06-11) surveyed 12 tools and ranked PR-Agent as the BYO-key fallback, rejecting it because *"local-model quality drops below the bar"*. That objection evaluated BYO as "Ollama on homelab hardware". Cluster inference is a different quality tier, so the rejection no longer holds. PR-Agent itself is actively maintained (last commit 2026-08-06).

## Where review quality actually comes from

Ranked by leverage, because the instinct is to reach for the prompt first and that is the smallest lever:

1. **Context assembly** — the dominant factor. A reviewer that sees only the changed hunk cannot find a cross-file bug; no prompt recovers information that never entered the window. PR-Agent's patch-extension and file-prioritisation logic is the part worth relying on rather than reimplementing.
2. **Model class** — a large reasoning model finds defects a small one cannot, at equal context.
3. **Static analysis fed into the prompt** — a large share of what reads as "AI quality" in CodeRabbit is linter and SAST output rendered as prose. **This is mostly redundant here**: ruff, mypy strict, BATS, hadolint, gitleaks and per-repo validators already run in CI. Our marginal gap is reasoning-level defects, not lint.
4. **Prompt and output format** — real, but it buys *noise control, severity thresholds and formatting*, not new findings.

Consequence for this ticket: invest in configuring context and models correctly; treat prompt tuning as the last pass, once real reviews exist to tune against.

Notable free win: `config.repo_context_files` defaults to `["AGENTS.md"]`, so each repo's behavioural SSOT enters the review prompt automatically. Standing orders, the English-only rule and the no-auto-merge rule become review criteria with no extra work.

## Model configuration

Three models doing three different jobs, rather than one job three times:

```toml
[config]
model            = "deepseek-v4-flash"   # reasoning + long context for large diffs
fallback_models  = ["qwen3.6"]           # independent failure mode — see below
[...]

PR #1019, opened 2026-08-16
feat(ci): make a green check mean reviewed, not merely un-failed

mlorentedev:
Implements `specs/GUARD-002-review-attestation`. Addresses #906.

## The defect, measured today

On 2026-08-16, three PRs from three parallel sessions each carried a CodeRabbit comment reading *"Review limit reached — you've reached your PR review limit, so we couldn't start this review"*:

| PR | `reviews[]` | `gh pr checks` |
|---|---|---|
| #1007 | 0 | `CodeRabbit  pass` |
| #1009 | 0 | `CodeRabbit  pass` |
| #1013 | 0 | `CodeRabbit  pass` |

**Zero reviews. All green.** Two were then merged.

The `pr-stewardship` doctrine already forbids proceeding *silently* on an unreviewed PR — but it binds an agent's judgement, and nothing mechanical tells anyone which case they are in. The one available signal, `pass`, asserts the wrong one.

## What it does

Four states, decided by **content** rather than by check status:

| State | Meaning | Exit |
|---|---|---|
| `attested` | a review happened — bot or human, but **not the author's own** | 0 |
| `disclosed` | no review, and the merge is declared | 0 |
| `declined` | a reviewer said it could not review | 1 |
| `pending` | no reviewer output yet | 1 |

**`declined` and `pending` both refuse and are reported distinctly.** Collapsing them would repeat the second defect inside #988, where the error path discarded the HTTP status and surfaced `invalid character 'I'` instead of `HTTP 500` — the diagnostic threw away the diagnosis, and *that* is why the bug survived weeks of being looked at.

The escape is the `merged-unreviewed` label **and** a non-empty `## Unreviewed merge rationale` section — both, never one. Same shape as spec-gate's `skip-sdd`. Proceeding on an unreviewed PR is allowed; proceeding silently is not.

## Why this before #786, not after

**Reviewer-agnostic by construction.** Logins and their "could not review" markers live in `harness/review-attestation.json`; the script names no reviewer at all (asserted by a test). #786's PR-Agent migration is an entry in a file, not a rewrite.

That ordering is the point. **#786's own acceptance criterion is *"a provider failure degrading the review rather than blocking the PR"*** — which is this same silent green, on a new tool. A replacement reviewer inherits the blind spot unless the gate exists first.

## The gate must not join the class it guards

A parallel session named the shape while debugging #988, and it is sharper than "the check was wrong": **the check was the cause.** Three instances in one day — `/status` poisoning the item reads it verified (#988); a rate-limit notice rendering as green (this PR); a DR check whose only reachable branch was the no-op one (#997).

GUARD-002 cannot fix that class. What it must do is not join it, which constrains the design concretely:

- **observes, never acts** — it will not retrigger a rate-limited reviewer, because a gate that spent the quota it measures would perturb its own subject. (Moot as well as wrong: neither a push nor an explicit `@coderabbitai review` reclaims a slot while the quota is spent — both measured.)
- **every branch observed failing** before being made to pass;
- **fails closed** — unreadable input exits 2 and says *"Cannot determine whether this PR was reviewed, so it is NOT treated as reviewed"*;
- **preserves why** — see `declined` vs `pending` above.

## Evidence

```
$ bats tests/check-review-attestation.bats
25 tests, 0 failures
$ shellcheck --severity=error scripts/check-review-attestation.sh
CLEAN
$ bash -n && zsh -n
OK
```

All 7 features in `features.json` exit 0.

**Non-vacuity measured, not assumed.** Four mutations, each valid syntax and a wrong answer:

| Mutation | Tests that went red |
|---|---|
| declined detection disabled | AC1 ×2, AC5 second-reviewer |
| escape accepts the label alone | AC4 label-only, AC4 empty-rationale |
| fail-closed → fail-open | AC6 ×4 |
| author's own review counts | AC2 self-review |

They **partition** — a suite where every test dies on one mutation is usually asserting one thing several times.

One test was observed failing for real rather than by mutation: the first version of *"the workflow does not swallow the gate's failure"* grepped the bare string `continue-on-error` and matched **the workflow's own comment explaining why the key is absent**. A guard tripping on documentation *about* the thing rather than the thing — the same false positive that made #998's archive gate unpassable. Now anchored to a YAML key.

## Deliberately not in this PR

- **Not `required` in branch protection.** Landing it as required would block every open PR the moment it merges. Adopt, observe, promote separately.
- **No auto-retrigger** — see above.
- **No judgement of review quality.** The gate asks whether a review happened, never whether it was good.
- **No change to `harness/reviewer-pool.json`.** That governs the *spec-level* adversarial review. Flagged there and on #786: #786 proposes `qwen3.6` as a fallback reviewer while `reviewer-pool.json` excludes it by name on the grounds that "a reviewer that PASSes cheaply is worse than no gate". That contradiction needs one owner; it is not decided here.

## Note on this PR's own gate

The workflow's first real exercise is this PR. Expect it to report `declined` or `pending` — the quota is still spent. That is the gate working, and the honest options are to wait for a review or to declare the merge with the label and a rationale.

## SDD

Spec-gate satisfied legitimately: `specs/GUARD-002-review-attestation/` is 438 LOC of genuine spec, well above the anti-alibi floor.

## Summary by CodeRabbit

- **New Features**
  - Added automated pull-request review attestation.
  - Publishes a clear status indicating whether an independent review is present.
  - Rechecks status when review-related comments are added or updated.
  - Supports approved reviews from recognized human or automated reviewers.
  - Allows unreviewed merges only when the designated label and a non-empty rationale are both provided.
  - Fails safely when review information is missing, invalid, pending, or declined.

- **Documentation**
  - Added proposal, task, acceptance, and verification documentation for the review policy.


mlorentedev:
Triage of the 10 findings: **8 applied, 2 declined.** Three were defects in the guard itself, and the first is the reason this PR should not have merged as it stood.

## Applied

**1. `pull_request_target` / running PR-controlled code — the important one.** You were right about the risk, and I took the narrower fix. The job checked out the pull request and ran `scripts/check-review-attestation.sh` **and the reviewer registry from it**, so a PR could edit the classifier to `exit 0` unconditionally, or add its own login to the registry, and attest itself. **The gate was disableable by the thing it judges.**

Checkout is now pinned to the base branch. I did *not* move to `pull_request_target`: it runs with a writable token in the base repo's context, which is the standard privilege-escalation footgun, and this job needs none of it — it reads the PR only through the API and never builds, executes or reads its tree. Same protection, smaller blast radius.

**2. `--pr` / `--payload` / `--config` without a value.** Real, and worse than a usability nit: `shift 2` fails under `set -e`, the shell exits **1**, and this script's own contract reads 1 as *"not reviewed"* rather than *"setup error"*. A typo was being reported as a verdict. Now validates and exits 2, with a bats case per flag.

**3. Registry validation was syntax-only.** Applied. Valid JSON of the wrong shape — a bare array, a reviewer with no `login`, an escape with no `section` — parsed fine and produced empty strings downstream, which classify as *"nobody declined"* and *"no escape configured"*. A malformed registry made the gate **more permissive**, which is the one direction it must never fail in. Now validates shape; three regression fixtures.

**4. `features.json` f5 was vacuous.** Correct, and thank you — this one led somewhere. `grep -qv PATTERN file` exits 0 when **any** line lacks the pattern, i.e. almost always. It was passing against a file that *did* contain the string, and the trailing `;` discarded even that status.

Chasing it turned up why nothing else caught it either: **`! cmd` is exempt from `set -e`**, so a bare `! grep -q X file` anywhere but the final line of a `@test` is silently ignored. Two assertions were dead across two files — and one was **genuinely violated**: the script named a reviewer in a comment while its test reported green. Replaced with a `refute_grep` helper that fails and prints the offending line, plus a meta-test that stops the bare form creeping back.

**5. f3 — assert both strings in the declined diagnostic.** Applied.

**6. f6 — isolated missing-`gh` / unreadable cases.** Applied. Note for anyone copying the pattern: emptying `PATH` does *not* test this, it kills `bash` and `jq` too and the script dies 127 before reaching the branch. The case builds a hermetic `PATH` containing exactly the binaries the path needs and no `gh`.

**7. `disclosed` — fourth state or override.** Fair; it was ambiguous. Now stated explicitly as a **fourth state**, evaluated after the other three and overriding a refusal, so consumers switch on four values. A PR that was declined *and* carries the escape reports `disclosed`, with the decline kept in the message rather than the state.

**8. `verification.md` inconsistencies.** Both real and both mine: the suite count was stale, and a "not verified" bullet claimed the workflow had never run on a real PR while a section above it documented exactly that. Corrected — the `pull_request` path **has** run and published statuses; only the comment-driven re-run is unproven, and it cannot be until this merges, since GitHub reads `issue_comment` from the default branch.

## Declined, with reasons

**9. A PowerShell equivalent.** Declined. This is a CI-only gate running on `ubuntu-latest`; there is no Windows execution path to keep parity with. The repo's cross-platform rule targets scripts that run on both machines, and ADR-020 routes twin logic to the Go CLI rather than to a second shell implementation. Adding a `.ps1` here would create an unexercised second implementation of a security-relevant classifier — strictly worse than not having one.

**10. ShellCheck inside the bats suite.** Declined as duplication. CI already runs `ludeeus/action-shellcheck` over `scripts/`, so the classifier is linted on every PR; the suite asserting it again would add a second place to update without adding coverage. (It is run locally too — `shellcheck --severity=error` is clean, and clean at all severities.)

## Result

30 → **35** bats cases, all 7 features exit 0, shellcheck clean.

One note on your framing, which I'd endorse for future reviews here: the prompt block told the agent to treat findings as untrusted data and verify each against current code. I did — and finding 4 is the case that justifies it, because the *stated* defect (a bad predicate in `features.json`) was the smaller half of what was actually wrong.

Issue #1004, opened 2026-08-15
BUG-086: dotf secrets verify aborts on registry validation, so it can never report per-var health


mlorentedev:
## Symptom

`dotf secrets verify` cannot report anything. It aborts on the first malformed registry
entry, before resolving any secret:

```
$ dotf secrets verify
Error: secret "GITHUB_PERSONAL_ACCESS_TOKEN": bw.folder "apps" is not in the ratified
taxonomy (Dotfiles/apps, Dotfiles/infra)
```

Scoping by id does not help — `dotf secrets verify NAN_API_KEY` fails identically, because
registry validation runs at load time, before resolution.

## Why it is a defect and not a strict-validation feature

It contradicts the command's own documented contract:

> verify resolves each registry secret through its backend and reports a **per-var status**
> […] No args verifies all; **ids/var-names scope it**.

A health check whose job is to tell you what is broken should not be the first thing to
break, and should not be taken down by an entry the caller did not ask about. The failure
mode is the inverse of #906: there, a green check hid that nothing was reviewed; here, a
malformed *unrelated* entry hides the health of everything else.

## Impact

Blocks the secret-distribution step of TOOL-013 (#786): `NAN_API_KEY` cannot be verified as
resolvable before wiring `dotf secrets sync ci`. Evidence-before-assertion means the sync
path cannot be relied on until this works.

## Two separable problems

1. **The malformed entry itself** — `GITHUB_PERSONAL_ACCESS_TOKEN` has `bw.folder: apps`
   instead of `Dotfiles/apps`. One-line fix, but fixing it only hides the second problem
   until the next typo.
2. **`verify` conflates registry validation with resolution.** A malformed entry should be
   reported as a `FAILED` row for that secret, letting every other row resolve and report.
   When ids are passed, entries outside the scope should not be validated at all.

Fix 2 is the actual ticket; 1 is the trigger that surfaced it. Possibly the same family as
BUG-080 (#985), which also concerns registry entries pointing at nonexistent Bitwarden items.

## Acceptance

- `dotf secrets verify` reports a per-var status for every well-formed entry even when
  another entry is malformed, with the malformed one shown as `FAILED` and a non-zero exit
- `dotf secrets verify <id>` validates and resolves only the named ids
- Regression test covering a registry containing one malformed entry

mlorentedev:
[...]

Issue #992, opened 2026-08-15
BUG-083: #982 tightened the folder allow-list without a migration, so any dotf built from main refuses the already-deployed registry


mlorentedev:
## What

#982 renamed the ratified Bitwarden folder taxonomy from `Dotfiles/apps` / `Dotfiles/infra` to `apps` / `infra`, updating both `secrets/registry.yaml` **and** `validBWFolders`, the parser's allow-list. Both halves are correct. What is missing is that the deployed copy of the registry is a *separate artifact with its own lifecycle* (ADR-030): `~/.dotfiles/secrets/registry.yaml` is refreshed by `setup-linux.sh`, not by merging a PR.

So a binary built from current `main` rejects the registry that is currently deployed on every machine:

```
$ cd ~/.dotfiles && dotf secrets ls
Error: secret "GITHUB_PERSONAL_ACCESS_TOKEN": bw.folder "Dotfiles/apps" is not in the ratified taxonomy (apps, infra)
```

```
$ grep -o 'folder: [A-Za-z/]*' ~/.dotfiles/secrets/registry.yaml | sort | uniq -c
     21 folder: Dotfiles/apps
      5 folder: Dotfiles/infra
```

## Not currently broken, and that is the trap

The installed binary is **0.40.0**, which predates #982, so today everything works. The failure is armed, not firing: it lands the moment anyone installs a build from `main` — the next release, or a `scripts/install-dotf.sh` run — and it lands on **every** machine simultaneously, before any of them has re-run setup.

Blast radius when it fires: `ParseRegistry` is the entry point for the whole `dotf secrets` surface, so `ls`, `run`, `show`, `verify`, `render` and two `dotf doctor` sections all fail at once. Reproduced here with a `main`-based build:

```
[Secrets integrity]
  [FAIL] secrets/registry.yaml not found or invalid
```

Note the message says "not found or invalid" for what is actually a *validation* failure on a file that is present and readable — so the first person to hit this will go looking for a missing file.

## Why the guard did not catch it

The pattern is the dormant-value one this repo already has a lesson about, inverted: OPS-028 learned to validate `bw.folder` at write time rather than at activation time. This is the same field failing the other way — a value that was valid when written and was invalidated later by a change to the validator, in an artifact the validator's PR did not and could not update.

`dotf doctor`'s deploy-drift check knows the deployed copy can lag the checkout; nothing checks that the lagging copy is still *parseable by the current binary*.

## Options

[...]

PR #1020, opened 2026-08-16
fix(secrets): let verify report a broken registry instead of dying on it

mlorentedev:
Refs #1004 (BUG-086) — closing keyword withheld until the spec is archived, which needs an adversarial review, which needs 0.42.0. Same pattern as #1018.

## The bug

`Registry.validate()` returned on the **first** bad secret and `ParseRegistry` propagated it, so one malformed entry made the whole registry unloadable.

For a health check that is self-defeating: **the command whose job is to tell you what is broken became the first thing to break** — and it was taken down by an entry the caller never asked about. Scoping did not help either, because validation ran at load time, before resolution.

## What changes

- `dotf secrets verify` — every well-formed entry reports its status; each malformed one is a `FAILED` row naming the defect. Exit non-zero.
- `dotf secrets verify <id>` — only the named ids are validated and resolved. A defect elsewhere in the file is neither reported nor fatal.
- `dotf secrets verify <malformed-id>` — reported as `FAILED`, **not** "unknown id". It was excluded from the registry precisely because it is the thing being asked about, so the lookup consults the defects too.
- **Every other command is unchanged.**

## Three constraints that shaped it

1. **The strict parser is implemented ON TOP of the partial one** — one set of per-secret checks, two policies over the result. `ParseRegistry` is now literally `ParseRegistryPartial` + *"any defect → first error"*. Two independent validators is the shape that produced BUG-084 (#993): the moment one moves, they disagree, silently, under the case nobody exercises.
2. **Defective secrets are excluded, not flagged in place.** The package relies on validation having run — `parseFileMode`'s own comment says a bad mode *"is impossible in practice"* because `validate()` rejected it. Letting an invalid secret reach `Entries()` would trade a clear failure for an undefined one.
3. **A rejected secret registers neither its id nor its vars**, or the *next*, valid secret would be blamed as the duplicate.

Only `verify` takes the partial door. `set`, `migrate` and `render` write, and a half-valid registry is exactly the state in which they must not run.

## Evidence

**Mutation-checked** — a guard never observed failing is not evidence (#898). `verify` was pointed back at the strict door, i.e. the exact defect reintroduced:

```
--- FAIL: TestSecretsVerify_MalformedEntryDoesNotHideTheRest
    verify output missing "FAILED"
--- FAIL: TestSecretsVerify_ScopedAwayFromDefect
--- FAIL: TestSecretsVerify_ScopedToTheDefect
    the named defect should be a FAILED row
--- FAIL: TestSecretsVerify_DuplicateIdBlamesTheSecondOne
```

Restored, suite green.

**No behaviour change on a healthy registry** — `dotf secrets verify` against the real one, built from this branch: `33 ok, 0 missing, 0 failed`. This changes what happens when the registry is broken and nothing about when it is not.

`go build ./... && go vet ./... && go test ./...` green; `golangci-lint run` at the pinned 2.12.2 reports 0 issues. The pre-existing registry suite passes unchanged, which is also the evidence for AC6 — those tests assert the strict door's messages, and the strict door is now a thin policy over the partial one.

## A note on this issue's trigger, which has since disappeared

The issue cites `GITHUB_PERSONAL_ACCESS_TOKEN` carrying `bw.folder: apps` instead of `Dotfiles/apps`. #982 inverted the taxonomy, so that value is now the **correct** one and the entry is well-formed — `verify` currently reports 33 ok.

The issue anticipated this: *"Fix 2 is the actual ticket; 1 is the trigger that surfaced it."* The trigger vanishing is an argument for fixing it, not against — the check was one typo away from being unusable again, with nothing to warn until it happened. Same family as #992 (AC1 met by circumstance, its guard unbuilt) and #997.

## Summary by CodeRabbit

* **Bug Fixes**
  * `secrets verify` now reports valid and malformed registry entries independently.
  * Malformed entries appear as `FAILED` results, while healthy entries continue to be checked.
  * Scoped verification correctly includes requested malformed IDs and excludes unrelated entries.
  * Structurally unreadable registries still stop verification with an error.

* **Documentation**
  * Added proposal, implementation tasks, and verification documentation for partial registry validation.

### DIFF
--- a/cli/internal/cmd/secrets.go
+++ b/cli/internal/cmd/secrets.go
@@ -234,9 +234,11 @@ func scopeVerify(reg *secrets.Registry, defects []secrets.SecretDefect, args []s
 	if len(args) == 0 {
 		return defects, nil, nil
 	}
-	byID := make(map[string]secrets.SecretDefect, len(defects))
+	// Keyed to a SLICE, not a single defect: one id can carry several. A duplicate id
+	// is itself a defect, so "one defect per id" is the assumption the data disproves.
+	byID := make(map[string][]secrets.SecretDefect, len(defects))
 	for _, d := range defects {
-		byID[d.ID] = d
+		byID[d.ID] = append(byID[d.ID], d)
 	}
 	var inScope []secrets.SecretDefect
 	var healthy []string
@@ -244,11 +246,19 @@ func scopeVerify(reg *secrets.Registry, defects []secrets.SecretDefect, args []s
 		if tok = strings.TrimSpace(tok); tok == "" {
 			continue
 		}
-		if d, ok := byID[tok]; ok {
-			inScope = append(inScope, d)
-			continue
+		ds, isDefect := byID[tok]
+		inScope = append(inScope, ds...)
+
+		// One token can name BOTH a defect and a valid entry — a duplicate id whose
+		// first definition validated and whose second did not. Reporting only the
+		// defect would hide the entry that actually resolves, which is the half the
+		// caller most needs. So a defect match does not short-circuit the lookup.
+		_, known := reg.Selector(tok)
+		if known || !isDefect {
+			// Unknown-and-not-a-defect falls through deliberately, so resolveOnly
+			// produces the "unknown id" error rather than this silently accepting it.
+			healthy = append(healthy, tok)
 		}
-		healthy = append(healthy, tok)
 	}
 	if len(healthy) == 0 {
 		// Every named id is defective: select no healthy entries, but still report.
--- a/cli/internal/cmd/secrets.go
+++ b/cli/internal/cmd/secrets.go
@@ -153,17 +153,26 @@ func newSecretsVerifyCmd() *cobra.Command {
 			"(--require-all also fails on MISSING).",
 		SilenceUsage: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			reg, err := loadRegistry()
+			// The partial door, deliberately: a health check must not be stopped by
+			// the very kind of breakage it exists to report (BUG-086, #1004).
+			reg, defects, err := loadRegistryPartial()
 			if err != nil {
 				return err
 			}
-			only, err := resolveOnly(reg, strings.Join(args, ","))
+			defects, only, err := scopeVerify(reg, defects, args)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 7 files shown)


mlorentedev:
## Review triage — the Major finding was correct and is fixed

Thanks @coderabbitai. Dispositioned rather than acknowledged.

### 🟠 Major — "Preserve valid entries when a defect has the same ID" — **APPLIED**

Correct, and it exposed a real hole in my own reasoning. I made rejected secrets *not* reserve their id on purpose — so a defective entry cannot make the **next** valid one look like the duplicate. What I missed is the consequence: that same choice leaves two things sharing one name. The first `dup` definition validates and lives in the registry; the second is a defect. My `continue` matched the defect and short-circuited, so `verify dup` reported the defect and **silently skipped the entry that actually resolves** — the half the caller most needs.

The second half of the finding was also right: defects were keyed one-per-id, so several bad definitions of one id overwrote each other and only the last survived.

Fixed as suggested, with one addition:

- defects keyed to a **slice**, all matches appended;
- a defect match **no longer short-circuits** — the token is resolved against the registry too, so both halves report;
- a token that is neither a defect nor a known id still falls through to `resolveOnly`, so **"unknown id" stays an error rather than becoming silence**. That was the failure mode I was most worried about introducing while fixing this, so it has its own assertion.

### Tests, observed failing first

Two scoped duplicate-id tests, per your suggestion — plus the multi-defect case:

- `TestSecretsVerify_ScopedToDuplicateIdReportsBothHalves`
- `TestSecretsVerify_SeveralDefectsOnOneId`

Both were run against the short-circuit reintroduced and **failed**, printing only the `FAILED dup … duplicate secret id` row with no `VALID_HALF`/`FIRST_OK` — i.e. exactly the defect you described. Restored, suite green, `golangci-lint` at the pinned 2.12.2 reports 0 issues.

### Note on process, for the record

This is the **only** CodeRabbit review any PR in this repo received today — six were opened and the quota served one. Two PRs (#1007, #1009) were merged unreviewed, and #1017/#1018/#1019 are still open with zero reviews while their `CodeRabbit` check shows a green `pass` carrying a rate-limit notice. Evidence is on #906; #1019 is building the gate that makes an unreviewed merge a recorded decision instead of an invisible one.

Worth saying plainly because this review found a Major correctness defect on the first look. The value isn't hypothetical.

Issue #1023, opened 2026-08-16
CLI-039: agent-config deployment lives twice in shell, so the setups grow where the CLI should be absorbing them


mlorentedev:
ADR: extends [ADR-020](docs/adr/adr-020-tooling-cli-go-convergence.md) (strangler-fig) and composes ADR-025 path resolution.

## What

Deploying an agent config — copy a source, substitute secret placeholders, compare against the installed copy, install atomically — is implemented **twice**, once per OS, in the bootstrap scripts:

| config | `setup-linux.sh` | `setup-windows.ps1` |
|---|---|---|
| MCP servers | ~464 | mirrored |
| `opencode.jsonc` | ~723 | ~1078 |
| pi `models.json` | ~835 | ~1201 |

Two implementations of one behaviour, in two languages, kept in sync by hand. The scripts are 1513 and 2120 lines.

## Why this is a boundary violation, not just duplication

ADR-020 C7 leaves shell exactly one job: *"thin bootstrap (detect OS/arch, fetch the right binary, put it on PATH) + profile/env."* Substituting secrets into a config file and installing it atomically is none of those — it is user-facing tooling logic, which the same ADR assigns to Go. And the strangler-fig rule is explicit: **when a `.sh`/`.ps1` twin is next touched, port it to `dotf` in that same PR and delete the pair.**

That rule has not been applied here, so the scripts accrete rather than shrink. The immediate trigger: #987 needs a change to how pi's config resolves its API key, and the natural-looking fix is to patch the same block in both scripts — growing both, in the direction the convergence epic is meant to reverse. (That fix turns out to need no setup edits at all, which is what makes this a clean moment to file rather than to firefight.)

## What deploying actually involves

Worth stating, because it is the argument for one implementation:

- resolve source and destination per-OS (ADR-025 — `dotf env` already owns this)
- substitute placeholders via `dotf secrets render` — **call it, do not reimplement it**
- compare with the installed copy and skip when identical (all three blocks do this; each in its own dialect)
- write atomically with correct permissions (0600 for anything secret-bearing)
- report what changed

Every one of those is already a `dotf` concern. The shell is only there because it got there first.

## Proposed direction

`dotf deploy [name]` driven by a declarative manifest, matching the idiom the repo already uses for exactly this kind of table — `packages.json` for tools, `registry.yaml` for secrets, `env-contract.json` for paths:

```jsonc
// ai/deploy.json — deliberately minimal
{ "configs": [
[...]

PR #1026, opened 2026-08-16
fix(pi): resolve the API key at runtime, so no config carries a credential

### LINKED ISSUE #987: BUG-081: setup deploys pi's models.json with {env:VAR} placeholders pi cannot resolve, and says they resolve at runtime
## Symptom

Every `pi` call against the `nan` provider returns:

```
401: {"message":"Invalid API key.","type":"auth_error","param":"None","code":"401"}
```

while the same key sent by hand works:

```
$ dotf secrets run --only NAN_API_KEY -- sh -c 'curl -so /dev/null -w "HTTP:%{http_code}\n" \
    https://api.nan.builders/v1/chat/completions -H "Authorization: Bearer $NAN_API_KEY" ...'
HTTP:200
```

## Cause

`~/.pi/agent/models.json` carried `"apiKey": "{env:NAN_API_KEY}"`. That is **dotf's** placeholder syntax. pi's own resolver (`dist/core/resolve-config-value.js`) only interpolates `$VAR` and `${VAR}`; a string containing no `$` is parsed as a whole literal. So pi sent the literal 17-character string `{env:NAN_API_KEY}` as the bearer token.

Proved rather than inferred — pi was pointed at a local capture server and the outgoing header compared to the env value by SHA-256 fingerprint:

```
env NAN_API_KEY: len=25 fp=322dfb364a4e
Authorization:   scheme='Bearer' token_len=17 fp=f3407edcacbd
sha256("{env:NAN_API_KEY}")[:12] = f3407edcacbd   <- exact match
```

## Why the state exists

`setup-linux.sh:834-842` renders `ai/pi/models.json` through `dotf secrets render`, and on failure falls through to:

```sh
log_warning "dotf secrets render unavailable; pi models.json deployed with literal {env:VAR} placeholders (resolved at runtime)"
```

**"resolved at runtime" is false.** Nothing downstream resolves that syntax, so a setup run with a locked vault silently installs a config that can only ever 401. `setup-windows.ps1` mirrors the same block.

## Why it costs so much to diagnose

The failure is indistinguishable from a bad credential: the server can only answer 401, and the operator goes looking at the vault, the registry and the injection path — all healthy. A previous session spent most of its length on Bitwarden before the header was ever inspected. Local machine state was also edited on a wrong hypothesis (`authHeader: true`), which makes it strictly worse: `provider-composer.js:156` then builds the header from the raw config value.

## Suggested fix

Either:

- **(a)** make the fallback honest and loud — fail the block, or emit a warning that names the actual consequence ("pi will send the placeholder as its API key and every call will 401"); or
- **(b)** put pi's own syntax in the source: `ai/pi/models.json` uses `"apiKey": "${NAN_API_KEY}"`, which `dotf secrets render` leaves alone and pi resolves from the env `dotf secrets run` already injects. The secret then never touches disk at all — a better posture than deploy-time substitution, but it is a change to SDD-009's design, so it needs a decision rather than a patch.

(b) is what the affected machine currently runs, applied by hand and verified (`AUTH_OK` against the live API). That fix is **machine-local and in no repo**, so the next `setup-linux.sh` run reverts it.

## Related

Add a `dotf doctor` check: any `{env:` left in a deployed agent config is a broken deploy, and it is trivially greppable.

mlorentedev:
## What

`ai/pi/models.json` declares `"apiKey": "${NAN_API_KEY}"` — pi's own syntax — and pi resolves it at runtime from the environment that is already injected. **No pi config on any machine carries a credential.**

## The bug, proven from primary source

pi's shipped `dist/core/resolve-config-value.js` interpolates `$VAR` and `${VAR}`. `{env:VAR}` — which is **dotf's** syntax — contains no `$`, so pi sends it verbatim as the bearer token:

```
sentinel fp                  : 3387dac8dd02
"${NAN_API_KEY}"  resolves   : 3387dac8dd02   <- correct
"{env:NAN_API_KEY}" resolves : f3407edcacbd   <- the literal string
sha256("{env:NAN_API_KEY}")  : f3407edcacbd   <- exact match
```

`f3407edcacbd` is the same fingerprint #987 captured off the wire with a capture server. Two independent methods, same answer.

## Three layers each reported a health none had established

- `setup-linux.sh` logs that placeholders are *"resolved at runtime"*. Nothing resolves them.
- pi's own preflight: `isConfigValueConfigured("{env:NAN_API_KEY}")` returns **`true`**, because a string with no `$` has no variables that could be missing.
- The server can only answer **401**, indistinguishable from a bad credential — which is why a previous session spent most of its length inspecting Bitwarden, the registry and the injection path, all of them healthy.

## The success path was the worse one

When `dotf secrets render` *works*, it materialises the credential into the deployed config, where it sits in plaintext. That is how a live `NAN_API_KEY` reached a session transcript on 2026-08-15 — read by someone debugging the failure above. It also exempted the deploy step from ADR-028's posture that secrets are injected into one child process and never persisted.

## No setup script changes

```
$ git diff --stat origin/main -- setup-linux.sh setup-windows.ps1
(empty)
```

`render` substitutes only `{env:VAR}`, so a `$`-style config passes through byte-identical — pinned by `TestRender_LeavesShellStyleVariablesUntouched`. That is what keeps this fix out of the two shell twins the CLI-convergence epic is shrinking. The false *"resolved at runtime"* warning is deliberately left to die with its block in #1023.

## The guard

`dotf doctor` FAILs on a deployed pi config that still carries either defect, naming the provider and **never the value** — a check that printed a credential to prove a credential was on disk would have reproduced the defect it reports.

Observed red against the real config on this machine:

```
[Agent config secrets]
  [FAIL] /home/manu/.pi/agent/models.json holds a literal credential for provider(s) nan
```

**Expect this to stay red until the machine redeploys.** That is intended — the reminder is mechanical rather than a note someone has to remember — but it means merging this puts a red section on a working box.

## Injection already exists everywhere

| path | mechanism |
|---|---|
| review launcher | `dotf secrets run -- pi …` (Go, all platforms) |
| Linux/macOS shell | `pi()` in `.zshrc:70`, `.bashrc:92` |
| Windows | `function pi` in `powershell/profile.ps1:254` |

The uncovered case — a script calling `pi` outside both — now fails with `Failed to resolve API key from environment variable: NAN_API_KEY` instead of a 401. Stated in the ADR as an accepted narrowing.

## Review

**PASS** — `nan/deepseek-v4-flash`, 3 Minor findings, all dispositioned in `verification.md`: one applied (pi's `PI_CODING_AGENT_DIR` override was ignored, so an overriding machine reported SKIP while its config was broken — the guard's own recurrence of the class it exists to stop), one declined with a reason, one no-change-needed.

Applying it surfaced a defect the review did not: a `{env:VAR}` value triggered **both** branches, producing two FAILs that described the same string differently. Fixed and pinned.

ADR-034 records the posture change and the two rejected alternatives.

Closes #987

## Unreviewed merge rationale

CodeRabbit's account-wide quota is exhausted; its rate-limit notice is correctly read as `declined`. Disclosed, not bypassed.

**Resolved since this section was first written.** It originally recorded that the PR must not merge while `tests/pi-config.bats` was red — a test asserting `{env:NAN_API_KEY}`, the placeholder this change removes. That test encoded the old contract and **passed for exactly as long as the bug existed**: pi does not implement `{env:...}`, so the config it required shipped an unresolvable literal as the API key while pi's own preflight reported it "configured".

It has been inverted rather than deleted, which is the property that actually mattered: the config carries `${NAN_API_KEY}`, and it carries **no** `{env:` placeholder the consumer cannot resolve. `bats tests/pi-config.bats` passes 8/8.

That the assertion was missed here is a real gap in this PR's verification, and worth naming: the Go layer was checked thoroughly (`build`, `vet`, `test`, `golangci-lint`) and the shell test layer was not run at all. Two layers, two loops — and only one of them was closed.

What stands in place of a CodeRabbit review:

- An adversarial review on `nan/deepseek-v4-flash`: **PASS**, three Minor findings, all dispositioned in `verification.md`. One was applied and was real — the guard ignored pi's own `PI_CODING_AGENT_DIR` override, so a machine using it reported SKIP while its config was broken.
- Applying that finding surfaced a defect the review did not: a `{env:VAR}` value tripped both branches, producing two FAILs describing the same string differently.
- The load-bearing claim tested from primary source, twice by independent methods: pi's shipped resolver, and the fingerprint #987 captured off the wire.

**Risk accepted:** no second machine-reviewer read this. The change is scoped to one JSON value, one new doctor check, and an ADR; no setup-script line changed, asserted by `git diff --stat`.


mlorentedev:
@coderabbitai review

Issue #1043, opened 2026-08-18
instruction files cite two vault runbooks that do not exist, and the doc-path guard cannot see vault paths


mlorentedev:
`.claude/CLAUDE.md` points readers at two vault runbooks. Neither exists, and neither does the directory holding them.

```
$ find "$VAULT_PATH" -iname 'guide-knowledge-distillation*' -o -iname 'guide-opencode-go-setup*'
                                        <- nothing, anywhere in the vault
$ ls -d "$VAULT_PATH"/10_projects/dotfiles/*/
30-architecture  memory  prestudy  research  sessions
                                        <- no 40-runbooks
```

Both are cited as live sources:

- *"See vault runbook: `$VAULT_PATH/10_projects/dotfiles/40-runbooks/guide-knowledge-distillation.md`"* — under **Running knowledge crystallization**
- *"Runbook: `$VAULT_PATH/10_projects/dotfiles/40-runbooks/guide-opencode-go-setup.md`"* — under **Using opencode**

The convention itself is real — `40-runbooks/` exists under other projects in the vault (`90_archive/kubelab-cli`, `90_archive/yt-intel`, `50_work/.../python-sensor-sdk-platform`). dotfiles simply never got one, while its instruction file has been promising two for some time.

## Why the existing guard does not catch it

`scripts/check-doc-paths.sh` (#916) fails CI when an instruction file names a repo path that no longer resolves. These paths are **vault** paths, rooted at `$VAULT_PATH`, so the guard has nothing to resolve them against — the vault is a separate repository whose location is per-machine (ADR-025).

That is the same failure the guard was written for, one namespace over: an instruction file naming a source that is not there, reading as a pointer and pointing nowhere. A reader who follows it concludes the knowledge exists and that they failed to find it.

## Options

1. **Write the two runbooks.** They were promised, and at least the opencode one describes a setup that was actually performed.
2. **Extend the guard to vault paths** when `$VAULT_PATH` resolves, and skip cleanly when it does not (CI has no vault, same constraint as ADR-013's offline model). This is the durable half — it stops the next dangling reference rather than fixing these two.
3. Demote the references to plain prose if the runbooks are not going to be written.

1 and 2 are complementary and 2 is the one that generalises. Note the guard must **skip**, never pass, when the vault is absent, or it becomes a check that cannot fail the way it is cited — the exact defect recorded in `docs/lessons.md`.

## Acceptance

[...]

PR #1046, opened 2026-08-18
feat(harness): reconcile spec subcommands, add deployed doctrine probes, and enhance router

### LINKED ISSUE #1016: DOCS-013: AGENTS.md names a dotf spec subcommand that does not exist, and omits one that does
## What

`AGENTS.md` § Spec-Driven Development, line 167-169, tells every agent in this repo:

> Subcommands via `dotf spec …` (Go CLI, works in CI/Windows): `init` ("create/scaffold spec X"), `fill` ("write the proposal"), `archive` ("close spec X").

What `dotf spec --help` actually offers on 0.41.0:

```
Available Commands:
  archive         Archive (or abandon) a per-feature spec folder
  init            Scaffold a per-feature spec folder
  review          Run an adversarial review on a pooled model
```

So:

- **`fill` is documented and does not exist.** An agent following the canonical instruction hits a usage error mid-workflow, at the exact step where it is supposed to be writing the proposal.
- **`review` exists and is undocumented here**, despite being load-bearing: `dotf spec archive` refuses a spec without a fresh passing `review.md`, and the same file documents that refusal three paragraphs later while never naming the command that produces one.
- The `/spec` skill description claims **five** subcommands — `init`, `bootstrap`, `fill`, `check`, `archive` — of which `bootstrap`, `fill` and `check` are absent from the binary. `dotf spec check` is described as a "pre-implementation consistency lint", which is precisely the sort of thing an agent will try to run before implementing and find missing.

## Why it matters more than a stale line

`AGENTS.md` is the cross-agent SSOT — Claude, OpenCode, Copilot, Cursor, Codex and Antigravity all read it as their system prompt. A wrong command here is not a doc typo; it is an instruction every agent in the fleet follows and fails on. Hit today by a session following the documented flow.

It also sits inside the section that enforces the Discipline Gate, so the failure lands on agents doing the right thing — the ones that read the rule and tried to comply.

## Root cause candidate

The three-name list looks like it predates the CLI absorbing the workflow (ADR-020 strangler-fig). `review` was added by #959 without the list being updated, and `fill`/`bootstrap`/`check` appear to describe the SKILL's conversational steps rather than binary subcommands — a distinction the prose does not make, since it introduces all of them as "Subcommands via `dotf spec …`".

## Proposed direction

1. Correct the list to what the binary offers, and name `review` where the archive gate is described.
2. Separate the two vocabularies explicitly: which steps are **binary subcommands** and which are **skill/agent workflow steps** with no command behind them. The current text merges them, which is what makes the error invisible to a reader.
3. Guard it, per the standing rule that a bug class emits an assertion. A test asserting every `dotf spec <sub>` named in `AGENTS.md` exists in the command tree would fail today, and would have caught this at #959. Adjacent precedent: `TestIDPatternProseMatchesCode` already binds a different part of this same section to code — the mechanism exists, it just was not extended to the subcommand list.

## Acceptance criteria

- [ ] Every `dotf spec` subcommand named in `AGENTS.md` exists in the binary.
- [ ] `review` is documented where the archive gate that requires it is described.
- [ ] Skill workflow steps are not presented as CLI subcommands.
- [ ] A test fails when the two drift again, in the shape of `TestIDPatternProseMatchesCode`.

## Related

- ADR-020 — the CLI convergence that moved this workflow into `dotf`
- #959 — added `dotf spec review`
- `TestIDPatternProseMatchesCode` — the existing prose-to-code binding in this same section

### LINKED ISSUE #1035: dotf doctor: nothing proves an enforced region survived deployment to the doctrine payloads
Round-3 adversarial review of `HARNESS-072-pr-stewardship` raised this as its only **REAL** finding (Minor). Recording it here rather than fixing it in the spec, because the fix lands in `features.json` — a contract file — and editing one invalidates the `reviewed_sha` of the review that asked for the fix.

## The gap

AC2 of HARNESS-072 requires an enforced region to reach **five** surfaces:

| Surface | Committed? | Verified by |
|---|---|---|
| `AGENTS.md` | yes | `features.json` f2 |
| `ai/claude/CLAUDE.md` | yes | `features.json` f2 |
| `~/.claude/CLAUDE.md` | no — per-machine | nothing |
| `~/.gemini/GEMINI.md` | no — per-machine | nothing |
| `~/.codex/AGENTS.md` | no — per-machine | nothing |

f2 greps the two committed targets plus the manifest's `doctrine.inject` list. The manifest declares an *intent*; the deployed file is the *outcome*. Nothing closes the gap between them, so a `--deploy` that silently fails, or a future change that drops a region from the doctrine payload, leaves three of five surfaces uninstructed with every check green.

The current state is correct — verified by hand this session, 1 hit each on all five — which is exactly the condition under which an unguarded invariant rots unnoticed.

## Why CI cannot be the home for it

`~/.gemini/GEMINI.md` and `~/.codex/AGENTS.md` are per-machine deployment state. ADR-013's offline `--check` model verifies committed records against the vault; it has no deployed tree to look at, and CI has no agent home directories. So the check belongs where per-machine state already gets verified: `dotf doctor`.

## Proposed shape

A doctor check that, for every `enforced` region in `harness/manifest.json`, asserts its content is present in each surface the manifest says it targets — reading the deployed files, not the records. Failure names the region and the surface, never the content.

This is the same class as the coverage gap HARNESS-072 itself shipped a guard for (`check_coverage`, AC8): that one proves a region reaches every *declared* surface in the repo, this one proves the declaration survived deployment.

## Acceptance

- [ ] `dotf doctor` fails when an enforced region is missing from a deployed surface the manifest targets
- [ ] It passes on a correctly deployed tree (all five surfaces, current state)
- [ ] A mutation test: remove the region from one deployed payload, observe the check go red

mlorentedev:
## Summary

Reconciles spec subcommands in AGENTS.md, adds doctor verification for deployed doctrine payloads, enhances the prompt trigger router with skills/keywords, and updates repository scaffolding to modular docs/lessons/ with _index.md.

Closes #1016
Closes #1035

## Changes

- **Spec Subcommands Reconciliation (DOCS-013 / #1016)**:
  - Reconciles AGENTS.md to document real CLI subcommands (init, review, archive) and clearly distinguish skill workflow steps (/spec fill, /spec check, /spec bootstrap).
  - Adds drift-prevention test TestSpecSubcommandsProseMatchesCode in cli/internal/spec/drift_test.go with mutation test verification.
- **Deployed Doctrine Probes (HARNESS-074 / #1035)**:
  - Adds checkDeployedDoctrine to cli/internal/doctor/checks_deploy.go to assert that enforced doctrine regions survived deployment to runtime targets (~/.gemini/GEMINI.md, ~/.codex/AGENTS.md, ~/.claude/CLAUDE.md).
  - Adds unit test and mutation test in cli/internal/doctor/checks_deployed_doctrine_test.go.
- **Dynamic Prompt Router**:
  - Extends TriggerRule in cli/internal/harness/triggers.go with Keywords and Skills.
  - Implements MatchPrompt with full unit test coverage.
  - Enriches harness/triggers.json with keyword and skill metadata.
- **Modular Docs Scaffolding**:
  - Updates cli/internal/initrepo/scaffold.go to scaffold docs/lessons/ with _index.md instead of a monolithic file, keeping index reads lightweight for agents.

## Evidence

- go test -count=1 ./... passes 100% across all 15 packages in cli/.
- Pre-commit hooks passed cleanly.
- Mutation tests verified both doctor and spec-subcommand drift tests fail on mutation and pass when clean.


mlorentedev:
Coordination note from the session in `dotfiles-wt-h72`, on scope overlap rather than on the change itself.

**This branch already closes two tickets I filed tonight, so nobody else should start them:**

- `fix(docs): resolve dangling runbook paths and extend doc-path guard to vault references` is **#1043** — `.claude/CLAUDE.md` cited two vault runbooks under `10_projects/dotfiles/40-runbooks/` that did not exist, and `check-doc-paths.sh` could not see them because it only resolves repo-relative paths.
- The `HARNESS-074-deployed-doctrine-probes` archive covers **#1035**'s subject — the deployed doctrine payloads having no automated coverage, raised as the only REAL finding of HARNESS-072's round-3 adversarial review.

If either is fully addressed here, closing them with this PR keeps the board honest.

**One overlap worth knowing before it becomes a conflict:** this branch touches `cli/internal/doctor/checks_deploy.go`, `cli/internal/doctor/checks_deployed_doctrine_test.go`, `cli/internal/harness/triggers.go` and `triggers.json`. A separate session is working **HARNESS-061-effectiveness-probes** (#852) out of `dotfiles-wt-h61`, and effectiveness probes are a plausible fit for `cli/internal/doctor/`. Two sessions landing probes in the same package from different specs is the kind of thing that is cheap to prevent now and expensive to untangle after both merge.

**One detail on the guard extension**, since I filed the ticket and it has a trap in it: `check-doc-paths.sh` deliberately ignores backticked bare filenames and only resolves paths *containing a slash* — resolving by basename previously false-flagged vault patterns and `machine.json`. So its coverage is narrower than its name suggests, and a line rewritten from a rooted path to a bare one silently leaves the guard's coverage behind. Worth an explicit note wherever the vault-path extension lands, or the next reader will assume more coverage than exists.

Issue #1052, opened 2026-08-18
GUARD-003: the review gate never re-evaluates when our own reviewer finishes, because GITHUB_TOKEN comments emit no events


mlorentedev:
GUARD-002 (#906/#1019) answers "did a review actually happen". It answers correctly. What it cannot do is **notice when the answer changes**, and after #1047 that gap becomes the binding one: our own reviewer will finish and nothing will re-evaluate the gate.

## Measured, not inferred

GitHub suppresses workflow events for comments created with `GITHUB_TOKEN`, to prevent recursion. PR-Agent publishes with `GITHUB_TOKEN`. A clean natural experiment on 2026-08-18, four comments in one three-minute window:

| time | PR | author | token | gate runs triggered |
|---|---|---|---|---|
| 00:49:02 | #1047 | `github-actions[bot]` (PR-Agent Guide) | `GITHUB_TOKEN` | **0** |
| 00:49:47 | #1047 | `github-actions[bot]` (PR-Agent Suggestions) | `GITHUB_TOKEN` | **0** |
| 00:52:27 | #1046 | `mlorentedev` | PAT | 1, at 00:52:30 |
| 00:52:27 | #1048 | `mlorentedev` | PAT | 1, at 00:52:29 |

Two bot comments, no runs. Two human comments, two runs, within three seconds. This also explains two things previously read as coincidence: #1037 flipped to `attested` on **CodeRabbit's** formal review — CodeRabbit is a GitHub App with its own token, so its events are not suppressed — and #1047's PR-Agent run only produced output after a **manually** triggered re-run.

## Consequence

Merging #1047 makes PR-Agent's output *countable*. It does not make it *counted*. The gate keeps whatever verdict it last computed, which on a freshly opened PR is `pending`, because no reviewer has posted twenty seconds after the PR opens. So the repository's own reviewer can produce a complete, substantive review and the gate will never look at it.

This is the same class as the defects catalogued in the vault's `pattern-verification-fails-toward-unproven` this week — specifically its fourth arrival, where a second producer of a consumed signal appears and the consumer was written when there was one. Here the second producer publishes through a channel whose events are deliberately silent.

## What this spec covers

1. **Re-evaluate the gate when our reviewer finishes.** GitHub-native, no external infrastructure: a `workflow_run` trigger on the `pr-agent` workflow completing. `workflow_run` is not a comment event and is not subject to the `GITHUB_TOKEN` suppression.
[...]

PR #1047, opened 2026-08-18
fix(ci): let a declared reviewer attest with comment-shaped output

mlorentedev:
### **User description**
GUARD-002 (#1019) and PR-Agent (#1032) were built to solve two halves of one problem — make a green check mean *reviewed*, and stop review capacity gating throughput. **They did not compose.** A PR reviewed by PR-Agent read as unattested, permanently.

Filed as **#1045**.

## The defect

The classifier counted formal review objects only:

```sh
[.reviews[]? | select((.author.login // "") != $a)] | length
```

PR-Agent runs under `GITHUB_TOKEN` as `github-actions` and publishes through the **comments** API, so its output never enters `reviews[]`. Two PRs carried a real, substantive Guide and were classified `declined`:

| PR | Guide published | `reviews[]` | verdict |
|---|---|---|---|
| #1037 | 00:05:21 | 0 | `declined` |
| #1042 | 00:16:58 | 0 | `declined` |

#1037 passed the gate on **CodeRabbit's** review, not PR-Agent's — remove CodeRabbit and it fails too.

So the gate built to end silent-green was reporting this repo's own reviewer as no reviewer. **A check that is permanently red teaches its reader to wave it through**, which is the habit #906 was filed about — #906's failure with the sign flipped.

## The fix

Reviewers gain `review_markers[]`: strings identifying comment-shaped review output. Matched on a **declared `(login, marker)` pair**, never on "a bot commented" — the latter would let a labeler or release bot attest through the comments door, which is #1033 arriving by another route. Empty for CodeRabbit, whose auto-summary comment is not a review.

Checked **before** the declined path: both are true whenever PR-Agent reviews while CodeRabbit's quota is spent, which was #1042's exact state, and the gate asks whether a review *happened*.

Logins are matched case-folded with a trailing `[bot]` stripped, because the two APIs disagree — `gh pr view --json comments` returns `github-actions`, `/issues/N/comments` returns `github-actions[bot]`. Matching the raw string made the verdict depend on which API produced the payload. **That is half of #1033**, and it now also applies to the declined matcher.

## Evidence — live payloads, same classifier, before and after

```
#1042  before: [FAIL] declined — coderabbitai posted a notice that no review ran
       after:  [OK] attested — reviewed by: github-actions (comment-shaped review: "## PR Reviewer Guide")

#1038  before: [FAIL] declined     after: [FAIL] declined    <- genuinely unreviewed, still refused
```

[...]

Issue #1041, opened 2026-08-18
review-attestation: the check-run freezes at 'pending' while the commit status goes green, so a working gate reads as a failing one


mlorentedev:
GUARD-002 (#1019) publishes its verdict two ways, and only one of them can ever be corrected. On #1037 the PR now shows both at once:

```
attestation         fail   <- check-run, frozen at 00:07:50
review-attestation  pass   <- commit status, updated 00:11:41
                           [OK] attested — reviewed by: coderabbitai
```

Same gate, same commit, opposite answers. The green one is right.

## Cause

The gate runs on `pull_request` and on `issue_comment`. That is deliberate: a reviewer's output usually arrives *after* the checks settle, so the verdict has to be recomputed when a comment lands.

But `issue_comment` runs attach to the **default branch**, not to the PR head. That is why #1019 added `statuses: write` and an explicit commit status — a commit status can be posted against any sha, so the later run can correct the verdict. It works: `review-attestation` went `pending` -> `attested` at 00:11:41 when CodeRabbit submitted.

The **check-run** cannot be corrected the same way. The one created by the `pull_request` run is bound to that run, and no later `issue_comment` run can update it. So it keeps whatever the verdict was at PR-open time — which is `pending` on every PR, because no reviewer has posted 20 seconds after the PR opens. Every PR therefore ends with a permanently red `attestation` check-run beside a green `review-attestation` status.

## Why it matters more than a cosmetic duplicate

`gh pr checks` lists both, and the red one sorts into the operator's attention first. The failure mode is that a correct green gate reads as a failing gate, which trains exactly the habit #1019 exists to prevent: ignoring the attestation line because it is "always red anyway". A gate that cries wolf is a gate nobody reads.

## Fix

Make the commit status the single verdict, and stop the `pull_request` job from asserting one it cannot later revise:

- the `pull_request` run publishes the commit status and **exits 0 regardless** — at that moment `pending` is not a finding, it is the normal state of a PR that just opened;
- the `issue_comment` run keeps the real verdict, since it fires exactly when reviewer output arrives;
- branch protection, if it ever requires this gate, requires the **status** `review-attestation`, never the check-run.

The gate loses nothing: `pending` at open time was never actionable information, and the state that *is* actionable — `declined`, or still `pending` when someone tries to merge — is carried by the status, which is the artifact that can still be red at merge time.

[...]

Issue #1033, opened 2026-08-16
GUARD-002: the attestation gate accepts a review from any bot, not only a declared reviewer


mlorentedev:
Robustness gap in #1019, found while checking whether GUARD-002 and PR-Agent (#786/#1032) actually compose. **Not a blocker** — PR-Agent's reviews will be recognized. The defect is that too much else will be too.

## Measured

The classifier counts any entry in `reviews[]` whose author is not the PR author:

```sh
INDEPENDENT=$(jq '[.reviews[]? | select((.author.login // "") != $a)] | length')
```

So a synthetic payload whose reviewer is `github-actions[bot]` attests:

```
$ scripts/check-review-attestation.sh --payload /tmp/ghbot.json
[OK] attested — reviewed by: github-actions[bot]
exit=0
```

That is the correct answer for PR-Agent, which posts under `github-actions[bot]` via `GITHUB_TOKEN`. It is the **wrong** answer for every other workflow that might post a review under the same shared identity — a release bot, a labeler, a summary job. Any of them would make the PR read as reviewed.

`harness/review-attestation.json` exists and declares reviewers, but it is consulted **only** for `declined_markers`. Nothing consults it to decide whether an attestation is legitimate. So the registry is authoritative for "who can decline" and decorative for "who can attest", which is exactly backwards for the risk that matters.

## The rule it should implement

> A review attests when its author is not the PR author, **and** the author is either a human or a **declared** reviewer.

Any human review counts — that is deliberate and should not change. Only bot reviews need a declaration, because a bot identity can be shared and a human's cannot.

## Two API facts the fix has to handle

Both measured on #1019, and both are traps:

**1. The bot/human discriminator is not in the payload the script currently reads.** `gh pr view --json reviews` returns `{"login": "coderabbitai"}` and nothing else. The REST endpoint does carry it:

```
$ gh api repos/mlorentedev/dotfiles/pulls/1019/reviews --jq '.[] | "\(.user.login) type=\(.user.type)"'
coderabbitai[bot] → type=Bot

$ gh api repos/mlorentedev/dotfiles/pulls/1019 --jq '.user.type'
User
```

**2. The same account has two login spellings across the two APIs.** GraphQL (`gh pr view`) says `coderabbitai`; REST says **`coderabbitai[bot]`**. The shipped registry declares `coderabbitai`, which matches the GraphQL form only.

[...]

PR #1057, opened 2026-08-18
feat(cli): dotf pr triage-queue — the wake-up the review loop never had (#1052)

mlorentedev:
Refs #1052. Completes the third element GUARD-003 promised and had not built.

## The gap this closes

GUARD-002 makes a green check mean *reviewed*. **Nothing made anyone act on the review.** PR-Agent posts its findings and no one is told — this session read them only by going to look, which is how the Review item of the Definition of Done has been surviving: on somebody remembering.

CI cannot close that. `workflow_run` re-evaluates the gate and GitHub notifies the human, but **no push channel reaches an agent session**. So the wake-up an agent can actually use is a query it runs at a checkpoint.

```
$ dotf pr triage-queue
3 pull request(s) awaiting a disposition:

  #1051  docs(spec): archive HARNESS-061-effectiveness-probes (#852)
         github-actions reviewed, never triaged, 2026-08-17 19:05
  #1049  chore(main): release 0.43.1
         github-actions reviewed, never triaged, 2026-08-17 18:47
  #1048  refactor(docs): slim AGENTS.md to crisp boundary rules...
         github-actions reviewed, never triaged, 2026-08-17 18:47
```

Three real PRs carrying reviews nobody had acted on, found on the command's first run.

## "Dispositioned" had to become mechanical first

Until now the disposition lived in a chat message and evaporated. It is now a comment on the PR under a stable heading, so it becomes an artifact that can be read back — by a person, by this command, and by whoever inherits the PR.

A PR leaves the queue when a triage newer than the review exists, and **re-enters the moment the reviewer speaks again** — the case that matters, because pushing a fix makes the reviewer re-review and the earlier disposition does not cover what it said this time.

The marker is declared in `harness/review-attestation.json`, which already owns the reviewer registry. **One string, two consumers**: the skill writes it, the queue reads it. Putting it in either would have made the pair a two-file agreement nobody checks — the failure class this repo spent two days cataloguing.

## It lists. It never applies.

The judgement stays in `pr-review-triage`, under explicit human confirmation. The command's own help says so, because a future edit that quietly adds `--apply` would erode a contract that was decided deliberately, not defaulted into.

## Also lands AC6, which PR-Agent found

Its review of #1056 reported #1052 as *partially compliant*: AC6 — the required-check adoption order — was specified and never written. It was right. The order now sits in the workflow beside the thing it constrains:

1. #1047 the classifier counts comment-shaped output — **done**
2. #1041 `pending` stops freezing the check-run red — **done**
3. #1052 `workflow_run` re-evaluates on reviewer completion — **done**
4. promote — **the STATUS, never the check-run**

Requiring a check-run would pin every PR to the verdict computed seconds after it opened. Step 4 is a repository setting, not a code change, and it is the owner's to make.

## A contract gap in the spec, fixed before any code

GUARD-003's "What" promised three elements; its acceptance criteria covered two. **The wake-up had no criterion at all** — building it would have shipped a feature no verification covered. AC7-AC9 were added to `proposal.md` first. The spec has no `review.md` yet, so nothing was invalidated; after the adversarial review it freezes.

## Verification

- 8 table cases + 2 registry cases in `internal/prtriage`: never triaged, triaged, re-reviewed after triage, both spellings of a bot login (#1033's trap), a declared reviewer commenting without a marker, an undeclared author using one, the marker quoted mid-prose, and a registry with an empty marker **refused** rather than matching everything and silently emptying the queue.
- `go build`, `go vet`, `go test ./...`, golangci-lint at the pinned version — **0 issues**.
- `bats tests/review-attestation-workflow.bats` — 10/10.
- Live run against this repository, above.

## AC1 is verified by this pull request itself

#1056 merged at 01:39, putting the trigger on the default branch — the only place `workflow_run` fires from. **This is the first PR it can act on.** PR-Agent reviews it, the trigger fires, and `review-attestation` re-evaluates with no human comment and no manual re-run.

The sequencing constraint that made AC1 unverifiable became the vehicle that verifies it. If a `review-attestation` run appears here with event `workflow_run` and no human comment before it, AC1 is met — and if it does not, this PR is the evidence that it is not.


mlorentedev:
## Review triage

| # | finding | disposition |
|---|---|---|
| 1 | AC1–AC4 are not implemented here (they are in #1056) | **no change needed** — deliberate. #1052 is referenced, not closed, precisely because the spec spans both PRs and is not finished |
| 2 | `verification.md` claims AC1 is verified by this PR, and that cannot be assessed from the diff | **evidence supplied below** |
| 3 | Is a pull-based wake-up sufficient, given the ticket called for a push-based one? | **escalated to the owner** — correctly raised, and it is a decision, not a defect |

**On 2 — the claim is now measurable rather than asserted.** #1056 merged at 01:39, putting the trigger on the default branch. On this PR:

```
01:51:36  pr-agent completes   →   01:51:38  review-attestation  workflow_run  success
01:52:44  pr-agent completes   →   01:52:46  review-attestation  workflow_run  success
```

Two seconds, twice, with no human comment preceding either. The reviewer is right that the diff cannot show this; the run history can, and does.

**On 3 — this is the sharpest thing in the review, and it is a real change of approach, not an omission.** The spec's "What" originally said the `workflow_run` event would be the wake-up for the triage loop. Building it revealed that it cannot be: **CI has no push channel to an agent session.** `workflow_run` re-evaluates the gate — a workflow reacting to a workflow — but nothing reaches a local agent, and the human notification GitHub already sends was never the missing part.

So the wake-up became a pull: a query the agent runs at a checkpoint. `proposal.md` was updated to say so, and AC7–AC9 were added, **before** any code — the spec had promised three elements and given acceptance criteria for two, so building the third would have shipped a feature no verification covered.

That the reviewer flagged the substitution as needing owner confirmation rather than accepting the updated spec is the correct behaviour, and it is worth saying so: a spec edited to match what was built is exactly the thing an adversarial reviewer should refuse to take at face value.

Nothing declined.

Issue #1053, opened 2026-08-18
TOOL-013: the pr-agent workflow triggers on synchronize but PR-Agent's pr_actions does not, so every push runs a build that reviews nothing


mlorentedev:
The pr-agent workflow triggers on `synchronize`; PR-Agent's own `pr_actions` does not include it. So every push to an open PR spins up a runner, builds a Docker action, and PR-Agent does nothing.

## Measured

```
PR #1048 (feat/gov-004-agents-md-diet)
  pull_request runs on the branch:  4
  PR-Agent artifacts posted:        1   (2026-08-18T00:47:26Z)
```

Three of the four runs produced no output at all.

## Mechanism

The workflow asks for `synchronize`:

```yaml
on:
  pull_request:
    types: [opened, synchronize, reopened, ready_for_review]
```

PR-Agent's action then applies its **own** gate. Its `github_action_config.pr_actions` defaults to:

```
["opened", "reopened", "ready_for_review", "review_requested"]
```

`synchronize` is absent, and `handle_push_trigger` is opt-in. So the action starts, evaluates the event, and exits without calling the model.

**Two lists that must agree, in two files, with nothing comparing them** — and the disagreement is invisible because the run reports `success`. Same class as #1040 (concurrency key conflating two event types) and #1045 (`reviews[]` as a proxy for "someone looked"), one layer down.

## What it costs, and what it does not

Worth being exact, because it corrects a figure that was circulating:

- **Not** ~3 extra inference calls per push. A five-push PR is **~3 calls total**, not ~15 — the pushes are free of model cost.
- What it does cost is a runner and a Docker build per push, roughly 20-30s each, for nothing.
- The real damage is **epistemic**: `review: SUCCESS` on a push-triggered run reads as "the reviewer looked at your new commits". It did not. Anyone reading the check after pushing a fix is reading a green that means "the action started and declined to act".

That last point matters more than the minutes. GUARD-002 exists because a green check that does not mean "reviewed" is worse than a red one, and this is the same green in a different place.

## Fix

Make the two lists agree **explicitly**, in the workflow, rather than leaving one implicit:

```yaml
env:
  github_action_config.pr_actions: '["opened", "reopened", "ready_for_review"]'
```

and drop `synchronize` from the workflow's own `types`, so a push does not start a runner at all.

[...]

PR #1058, opened 2026-08-18
feat(ci): review every push, because the doctrine already says we do (#786)

mlorentedev:
Refs #786.

## Why, and it is not convenience

`harness/enforced/pr-stewardship.md` is injected into **every agent's instructions** and says:

> *"pushing a fix reopens it, **because the reviewer re-reviews**"*

Since #1053 that was false. A push relaunched nothing, so the second half of every PR went unread while the doctrine asserted otherwise. One of the two had to move. Softening the text would have been the wrong repair — a reviewer that does not re-review a fix is a reviewer that only sees the first draft.

## #1053's finding is preserved, not reverted

It was never *"synchronize is bad"*. It was that the workflow's trigger list and PR-Agent's own event gate disagreed **silently**, so every push started a runner, built the Docker action, reviewed nothing, and reported `review: SUCCESS` — a green that reads as "your new commits were reviewed" and means "the action declined to act". This moves the agreement rather than removing it.

## The obvious mechanism is a trap, and it was one commit from being written

Adding `synchronize` to `pr_actions` **does nothing at all**. Upstream routes it down a separate path:

```
handle_push_trigger = false  # when true, handle synchronize events using push_commands
```

A `synchronize` in `pr_actions` is a setting that looks right and asserts nothing — the exact shape catalogued in `pattern-verification-fails-toward-unproven` this week. It was caught by reading upstream's `configuration.toml` rather than reasoning about the field name, after a peer said to verify rather than assume.

## A second default that would have undone a decision

`push_commands` names `/review` alone, and the omission is the substance. Upstream's default is `['/describe', '/review']` — so enabling the push path without naming its commands would have **silently reinstated `describe`**, which rewrites the PR body and was turned off twelve lines above as a deliberate decision. A default that re-enables what you disabled elsewhere is the same defect one layer down.

Bot pushes and merge commits are excluded: neither is a human asking for another look.

## The guard is widened, not weakened

Set equality between the two lists made a **correct configuration unrepresentable** once `synchronize` had its own path — and the repair it invited was precisely the dead setting above. The guard now asserts what #1053 actually found:

- every type the workflow fires on is handled by something — `pr_actions`, or `handle_push_trigger` for `synchronize`;
- the reverse too: a handler for an event the workflow never fires is dead configuration that reads as coverage;
- `synchronize` firing with the push path off fails explicitly, because that is #1053's original bug.

Plus a second case that fails if the push path ever smuggles `describe` back.

## Cost

~2 inference calls per push (`auto_describe` is already off), so about ten on a five-push PR. And the rebase burst is already paid for by **#1042**: the concurrency group is keyed on the event name with `cancel-in-progress`, so three pushes in ninety seconds collapse to **one** review of the settled state, not three.

No `paths-ignore` and no draft check: drafts are already skipped by the job's `if`, and excluding path classes from review is a guess to maintain — this repo's most consequential recent defects were in `.github/workflows/`, `docs/` and `harness/`, which are exactly the "surely that doesn't need review" shapes.

## Verification

16/16. Four mutations, each observed failing on its own guard, with arrival checked by checksum and the invalid control correctly reported as **not landed**:

| mutation | landed | guards failed |
|---|---|---|
| `synchronize` fires but `handle_push_trigger` off | yes | 1 |
| `push_commands` reinstates `/describe` | yes | 1 |
| `push_commands` removed entirely | yes | 1 |
| `pr_actions` handles an event nothing fires | yes | 1 |
| **control: a pattern that does not exist** | **no** | 0 |

`actionlint` clean.

## Verified on this PR itself — and my stated criterion was wrong

I wrote that this could not be verified before merge, and that the evidence would be *a second Guide appearing*. **Both halves were wrong**, and the second one would have produced a false negative.

It verified here, because `pull_request`/`synchronize` uses the workflow from the PR rather than from the default branch. A push at 02:30 produced:

```
Guide comment  created 02:14:48
               updated 02:31:41     <- the push run

#### (Review updated until commit 7385c5c...)
```

`persistent_comment` **updates the existing comment in place**; it does not post a second one. So anyone verifying by looking for a new Guide would have found none and concluded the setting was inert — while it was working. The honest criterion is the comment's `updated_at` moving, and the *"Review updated until commit"* line naming the new head.

Confirming evidence that the command list took effect too: the **Code Suggestions** comment was *not* touched, because `push_commands` names `/review` alone.

Leaving this here rather than quietly fixing the sentence: a verification criterion that fails toward "broken" on a working system is the same defect class this PR is about, and it was in the PR's own acceptance text.

## Summary by CodeRabbit

* **New Features**
  * Pull requests are automatically re-reviewed when new commits are pushed.
  * Push-triggered reviews are limited to the `/review` command.
  * Bot and merge commits are excluded from automatic reviews.

* **Documentation**
  * Added documentation describing review-on-push behavior and configuration validation.

## SDD skip rationale

Minor fix


mlorentedev:
## Review triage

Both findings applied in `bd8a1c0`. First use of this marker — `dotf pr triage-queue` (#1057) reads it to decide whether reviewer output has been dispositioned, so the record lives here rather than in a chat message that evaporates.

| # | finding | disposition |
|---|---|---|
| 1 | `tests/pr-agent-config.bats` — the guard accepts `[]`, `["/improve"]` and only rejects `/describe` | **applied** |
| 2 | `specs/TOOL-013-pr-agent-reviewer/tasks.md` — "every push" overstates; bot and merge commits are excluded | **applied** |

**Finding 1 was the more serious of the two, and worse than reported.** The comment framed it as over-permissiveness; the load-bearing case is `[]`. An empty `push_commands` makes the push trigger fire and run **nothing** — so the entire feature this PR adds would be inert while the guard protecting it stayed green. A guard whose green survives the death of its subject, written into the file whose only purpose is catching that.

Now exact equality with `["/review"]`. Exact rather than "contains `/review`" on purpose: adding a command costs a second inference call on every push, and that is a decision worth forcing through the guard rather than letting it arrive as an edit nobody weighs. Three mutations observed failing — empty, `/improve`-only, `/describe` re-added — with the mutation's arrival checked by checksum, and an invalid control correctly reported as not landed.

**Finding 2 is a claim wider than its implementation** — the same shape as the defect this PR fixes, one document over. The correction says so rather than quietly narrowing the sentence, because the interesting part is that it happened here, in the change written to close that gap.

Nothing declined.

PR #1068, opened 2026-08-18
chore(main): release 0.45.0

mlorentedev:
:robot: I have created a release *beep* *boop*
---

### Features

* **cli:** add dotf search and dotf harness suggest commands ([#1067](https://github.com/mlorentedev/dotfiles/issues/1067)) ([27c7ccf](https://github.com/mlorentedev/dotfiles/commit/27c7ccf41edcdca2a6839d94712a6ea6482dd2ff))
* **harness:** skill dependencies resolution and full trigger catalog ([#1070](https://github.com/mlorentedev/dotfiles/issues/1070)) ([6c17f9c](https://github.com/mlorentedev/dotfiles/commit/6c17f9ce5832ef239a59604af46494397db9fa3b))
* **secrets:** check the age root has not drifted, and whether the escrow still describes the vault ([#1077](https://github.com/mlorentedev/dotfiles/issues/1077)) ([#1079](https://github.com/mlorentedev/dotfiles/issues/1079)) ([c805e8f](https://github.com/mlorentedev/dotfiles/commit/c805e8f4b2ee0c0e8c758b5018a176d206417fd5))
* **secrets:** give the age root a backend, so the inventory contains its own root ([#937](https://github.com/mlorentedev/dotfiles/issues/937)) ([#1075](https://github.com/mlorentedev/dotfiles/issues/1075)) ([10204ae](https://github.com/mlorentedev/dotfiles/commit/10204ae2bac6c4ba65d9919d51b5671c9bf7ee87))

### Bug Fixes

* **ci:** a review attests only from a member or a declared reviewer ([#1033](https://github.com/mlorentedev/dotfiles/issues/1033)) ([#1071](https://github.com/mlorentedev/dotfiles/issues/1071)) ([2bac1c5](https://github.com/mlorentedev/dotfiles/commit/2bac1c59e8f2bdabe79eac8c4ee3a077bfd527b2))
* **ci:** pin the reviewer action by commit and gate its comment trigger on membership ([#1078](https://github.com/mlorentedev/dotfiles/issues/1078)) ([6e1c114](https://github.com/mlorentedev/dotfiles/commit/6e1c114f0240a9fcbedc037f835a4545b154a9b8))
[...]

Issue #1062, opened 2026-08-18
TOOL-017: PR review triage is a protocol executed by judgement — encode it, and require a reproduction before applying a finding


mlorentedev:
`pr-review-triage` is a good protocol executed by judgement. Nothing triggers it, nothing checks it ran, and its single highest-value step — deciding whether a reviewer's claim is *true* — is not encoded anywhere. Every PR gets the quality of triage the agent happened to apply that day.

## The incident that makes it concrete

On **#1059**, PR-Agent reported a finding and classified it **REAL**:

> The `[^\n]*` greedily consumes the space after `install`, so `[^-a-z]` must match a non-letter... but the next character is `a`, so the match fails.

Measured against the actual matcher:

```
sudo apt-get install age         -> CAUGHT
sudo apt-get install age zsh     -> CAUGHT
sudo apt-get install -y age zsh  -> CAUGHT
apt-get install age              -> CAUGHT
```

**The stated mechanism was false.** The regex backtracks; nothing was escaping.

Testing the claim is what found the real defect — `grep` is line-based, so the multi-line continuation form escapes entirely:

```
apt-get install -y --no-install-recommends \
    age \
    zsh
```

`tests/Dockerfile.integration` is written in exactly that shape.

**So a flow that auto-applies REAL findings would have replaced a working guard with a worse one, and left the actual hole open.** Right instinct, wrong diagnosis, real defect — and only the third of those is visible without reproducing the claim.

## The rule to encode

> **Reproduce the defect the finding claims. If you cannot reproduce it, it is not a finding.**

The same mutation discipline already applied to guards, applied to reviews. A reviewer's severity is a *hypothesis*, and the flow's job is to test it rather than route on it.

## What is missing, precisely

| step | today | needed |
|---|---|---|
| triage is triggered | an agent decides to | a signal |
| review vs. notice | prose in the skill | `declined_markers` already answers this for the gate; triage does not consult it |
| **claim verified before applying** | judgement | **not encoded anywhere** |
| every comment has a disposition | prose | nothing checks it |

## The seam, and its trap

`.pr_agent.toml` already instructs the reviewer to classify findings **REAL / THEORETICAL / SPECULATIVE**. Nothing consumes that field — it is the natural hook.

**But it is only prose.** Measured on #1059: the classification appears as the literal `REAL.` inside a `<details>` block's body, and the run publishes **zero artifacts**. There is no JSON, no label, no structured output.

[...]

PR #1071, opened 2026-08-19
fix(ci): a review attests only from a member or a declared reviewer (#1033)

### LINKED ISSUE #1033: GUARD-002: the attestation gate accepts a review from any bot, not only a declared reviewer

mlorentedev:
Closes #1033.

The attestation gate counted any entry in `reviews[]` whose author was not the PR author. That reads as *"somebody other than me looked"* and means *"some identity other than mine appeared"* — and those differ exactly where an identity is **shared**. Every workflow in this repository posts under one automation login, so a labeler, a release job or a summary step attested for a PR nobody had read.

#1047 closed the same defect on the **comments** door, gating it on a declared `(login, marker)` pair. The **reviews** door kept the original rule. That left the registry authoritative for who may attest by comment and decorative for who may attest by review — worse than fixing neither, because a reader of the hardened path reasonably concludes the rule is enforced everywhere.

## Measured on main before the change

The fixture was first confirmed to name a login absent from the registry, so the instrument could not be the thing that was broken:

```
$ jq -r '.reviews[0].author.login' undeclared-bot.json
stale-labeler-bot
$ jq -r --arg l stale-labeler-bot '[.reviewers[] | select(.login==$l)] | length' harness/review-attestation.json
0

$ scripts/check-review-attestation.sh --payload undeclared-bot.json
[OK] attested — reviewed by: stale-labeler-bot
exit=0
```

## The rule

> A review attests when its author is not the PR author, **and** that author is either a member of this repository or a **declared** reviewer.

Any human review still counts, unchanged and deliberately: an individual's identity cannot be shared the way an automation login is, so only bot reviews need declaring.

**The discriminator is `authorAssociation`, already present in the payload.** `gh pr view --json reviews` returns login, association and state — no bot/human bit. REST carries `user.type`, but spells the login with a `[bot]` suffix, so reaching for it would have added an API call *and* a second spelling to every fixture to learn the same thing. Measured across the last sixty PRs: every bot-authored review reads `NONE`, the one human-authored review reads `OWNER`.

It is near-enough rather than exact, and the residue falls the safe way — an unrecognised association refuses loudly instead of attesting in silence. `CONTRIBUTOR` is deliberately **not** a member association: it is granted by having a commit merged, which an automation account holds as easily as a person.

## The refusal had to change too

A PR carrying three reviews was about to be told *"pending — no reviewer output on this PR yet"*. That is a false statement, and the one a reader would act on by requesting a review the PR already has. It now names what it found and why none of it attested.

## Verification

- **No live PR changes verdict.** #1057, #1065, #1067, #1068 and #1070 attest identically under the old and new script, run side by side.
- **Both new guards observed failing on their own mutation**, mutation confirmed present first: replacing the member-or-declared predicate with `select(true)` fails cases 51, 52, 55, 56; appending a declared login to the script fails case 14.
- `bats tests/check-review-attestation.bats` — 56/56. Full suite — 1354/1354.

## Also fixes the guard that should have caught this

*"The script names no reviewer of its own"* refuted three hand-written names while the file named a fourth, in a comment added by #1047. A hand-written list only refutes the names someone thought of, so it now derives the list from the registry — the only place that knows it.

### DIFF
--- a/scripts/check-review-attestation.sh
+++ b/scripts/check-review-attestation.sh
@@ -13,7 +13,7 @@
 #
 # States, decided by CONTENT rather than by check status:
 #
-#   attested  a review happened (bot or human, not the author)   -> exit 0
+#   attested  a review happened (a member, or a declared bot)     -> exit 0
 #   exempt    the diff carries nothing a review could act on     -> exit 0
 #   disclosed no review, but the escape is declared in full      -> exit 0
 #   declined  a reviewer said it could not review                -> exit 1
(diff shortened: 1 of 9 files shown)


mlorentedev:
## Review triage

Reviewer output read and dispositioned. Recorded here rather than in a chat message so `dotf pr triage-queue` reads this PR as triaged, and so the Definition of Done's Review item rests on something checkable.

| # | Source | Finding | Class | Disposition |
|---|---|---|---|---|
| 1 | PR-Agent | Ticket compliance for #1033: compliant, no non-compliant requirements, nothing flagged for human verification | — | Noted. No action. |
| 2 | PR-Agent | No major issues, no security concerns, tests present | — | Noted. No action. |
| 3 | CodeRabbit | `Review rate limited` — a notice that no review ran, not a review | — | Disclosed, not applied. This is the exact state GUARD-002 exists to make visible, and the gate reported it correctly as `declined` until PR-Agent's Guide superseded it. |
| 4 | Observed while triaging | The review carries no HARNESS COMPLIANCE section, which `.pr_agent.toml` instructs it to open every review with | THEORETICAL | **Ticketed as #1072.** Out of scope here: this PR changes the attestation rule, not the reviewer's configuration, and the fix belongs to #786's surface alongside the marker decision on #1062. |

Zero findings applied, one ticketed, nothing declined. No reviewer comment asked for a change to this diff.

**Gate behaviour on this PR is itself evidence.** `review-attestation` read `declined` while CodeRabbit's rate-limit notice was the only reviewer output, then flipped to `[OK] attested` when PR-Agent's Guide landed and `workflow_run` re-evaluated it — the loop from #1052 working end to end, on the first PR opened since the status became a required check.

Issue #1000, opened 2026-08-15
OPS-030: the DR escrow is encrypted with a key whose only off-machine copy is inside the vault the escrow exists to replace


mlorentedev:
## The circular dependency

ADR-028's disaster-recovery story is: age-encrypt a full Bitwarden export to `sensitive/dr/bitwarden-export.age`, commit it (the dir's `.gitignore` explicitly un-ignores `*.age`), and recover with "the offline age key + a repo clone".

The age identity at `~/.config/age/key.txt` is what decrypts it. Verified today, by fingerprint:

- local key `~/.config/age/key.txt` → `b4d18dc3867b`
- vault item `AGE-SECRET-KEY-PERSONAL` → `b4d18dc3867b` — **match**

So the key does have an off-machine copy, and it is the right key. But that copy lives **inside Bitwarden**, which is the exact thing the escrow exists to survive the loss of. Lose the Bitwarden account and you lose both the vault and the key that would have decrypted its backup. The escrow in git becomes an undecryptable blob.

The runbook's phrase is "the **offline** age key". There is no offline copy. The file is 189 bytes, dated 2025-06-20, and exists on one disk.

## Why this is the top risk, not the missing escrow

The missing escrow (#997, #993) is recoverable: run one command and it exists. This is not — if the key is lost there is no procedure, no re-derivation, and no partial recovery. Every `backend: age` secret and every future escrow is unreadable permanently.

It also compounds with #971: `migrate` drops the `age:` pointer, so 28 secrets currently live only in Bitwarden. Today the failure modes are:

| Lost | Consequence today |
|---|---|
| The laptop | Survivable — Bitwarden holds the live tier and the key copy |
| The Bitwarden account | **Total loss of 28 migrated secrets**, and the escrow (once it exists) is undecryptable |
| Both | Total loss |

The middle row is the one this ticket is about, and it is the row the escrow was designed to make survivable.

## What it needs

**One copy of the age identity that does not depend on Bitwarden.** 189 bytes. Any of: printed and filed physically, a second password manager, an offline USB, a sealed envelope in a different building. Two independent stores is the requirement; which two is a preference.

This is what ADR-028's own worked example already prescribes and what the registry does not implement (#937): `plane: floor, backend: age-offline`, with the Bitwarden copy as **convenience**, not as the authoritative one. `SSH_KEY` is the only entry that follows that shape today; the key that decrypts everything does not.

## Acceptance criteria

[...]

PR #1075, opened 2026-08-19
feat(secrets): give the age root a backend, so the inventory contains its own root (#937)

### LINKED ISSUE #937: OPS-026: the age key that decrypts every secret is itself absent from the registry
## Context

`secrets/registry.yaml` is the mapping SSOT of ADR-028 §2. Every `backend: age`
secret is decrypted with the age identity at `~/.config/age/key.txt`.

## Problem

That key is **not declared in the registry at all**. The registry holds exactly
one `age-offline` entry (`SSH_KEY`, line 317). The secret on which decryption of
all 32 others depends is the one secret no inventory manages: no plane, no
expose contract, no Bitwarden convenience copy, no rotation cadence.

ADR-028 itself uses it as the canonical worked example (lines 134-139):

```yaml
  - id: age-key-personal               # the floor: offline-authoritative, bw copy = convenience
    plane: floor
    backend: age-offline
    bw: { folder: "dotfiles/floor", item: "AGE-SECRET-KEY-PERSONAL" }
    offline: required
    expose: { file: { path: "~/.config/age/key.txt", mode: "0600" } }
```

So the shape is already decided and ratified — it was simply never added.

## Acceptance criteria

- [ ] `age-key-personal` (or an equivalent id) exists in `secrets/registry.yaml`
      as `plane: floor`, `backend: age-offline`, with the `expose.file` contract.
- [ ] It carries a `bw:` block pointing at `dotfiles/floor` (convenience copy;
      authority stays offline).
- [ ] `dotf secrets ls` shows it, and `verify` handles it without attempting a
      circular self-decrypt.
- [ ] The chicken-and-egg is documented: the key cannot be age-encrypted with
      itself, so its escrow path is stated explicitly (Bitwarden copy + offline
      paper).

Refs: ADR-028 §2 and lines 134-139, #585.

mlorentedev:
Closes #937. Spec: `specs/OPS-026-age-root-file-authority/`.

`secrets/registry.yaml` is ADR-028's mapping SSOT, and the one secret it did not manage was the one that decrypts all the others. No plane, no expose contract, no declared off-machine copy, no cadence — for the root of the whole hierarchy.

## It was never "simply never added"

The entry ADR-028 itself prescribes **cannot be added**. Three independent reasons, each sufficient:

- `backend: age-offline` with a file expose **requires** an `age:` source (`registry.go:342`). This file is what decrypts age sources; it would have to be encrypted under itself.
- `expose.file` requires `var` as well as `path`. The example declares only `path` and `mode`.
- `offline: required` is **not a field** on the `Secret` struct. It would have been silently ignored — worse than being rejected.

That is why the issue aged. Every attempt to "just add it" hit a validator that is, arguably, right.

## The shape

`file-authority`: a secret whose authority **is** the local plaintext file.

- **Nothing resolves it**, because there is nothing to resolve it *from*. `Resolve` refuses, with a reason — handing the key that decrypts every other secret to a child process through the same facade as those secrets widens the blast radius instead of narrowing it. `TestFileAuthority_ResolveRefuses` pins that, because it is exactly the kind of thing a later change "fixes" helpfully.
- **A new optional `Verifier` seam** lets a backend answer its own health question. `Loader.Verify` prefers it and falls back to `Resolve`, so the three existing backends are untouched.
- **Absent is MISSING, wrong mode is FAILED.** A fresh checkout has no key yet and must not look broken; a present key readable by others is a real defect, not an absence.

Chose this over a named exemption in `checkAgeSources`. The carve-out was the smaller diff and was rejected: the next root — a second machine, a rotated identity — asks for the same special case again.

## Verification

```
before:  33 ok, 0 missing, 0 failed
after:   OK  AGE_KEY_PERSONAL  file-authority
         34 ok, 0 missing, 0 failed
```

Mutations observed live against a **scratch registry pointing at a throwaway key**, so the real root was never touched:

```
mode 0600                                  -> 1 ok, 0 missing, 0 failed
chmod 0644  (stat confirmed 644 first)     -> 0 ok, 0 missing, 1 failed, non-zero exit
file removed (absence confirmed first)     -> 0 ok, 1 missing, 0 failed
```

`go build`/`vet`/`test ./...` 17 packages, 0 FAIL. `golangci-lint` at the pinned 2.12.2 — 0 issues. Full bats — 0 failures.

## Also fixes the test that would have hidden this

`TestParseRegistry_AcceptsExactlyValidBackends` carried **one** env-shaped template for every backend, so it reported `file-authority` rejected while the parser was behaving correctly. It now holds a valid fixture per backend **and fails when a backend has none** — a backend cannot ship with its accept path untested. Same hole as #1033's hand-written refute list, in a different file.

## Deferred, visibly

**AC5 — the drift comparison** (local fingerprint vs the Bitwarden convenience copy) needs a live session and `age-keygen -y`, and it is the point of #1000 rather than of this spec. `fileAuthorityResolver`'s own comment says the check answers a narrower question than the one that matters until that lands, so the gap cannot be read as coverage. Marked `deferred` in `features.json`, not `pending`.

## Two ADR/code tensions recorded rather than silently resolved

1. ADR-028's worked example is **amended** here, with all three reasons the old one could not validate.
2. `validBWFolders` has no `floor` folder, and its comment says floor secrets carry no `bw:` block at all — while ADR-028 prescribes exactly that for this key. The distinction the comment misses is the one the ADR makes: **authority** versus **convenience copy**. Rather than amend a ratified taxonomy unilaterally, the block is unfoldered and the tension is written down.

## What this does not fix

The actual top risk (#1000) is unchanged: this key's only off-machine copy is inside Bitwarden, which is the thing the DR escrow exists to survive the loss of. This PR makes the root *visible and checkable*. Making it *recoverable* is a physical act, and it is the operator's.

## Summary by CodeRabbit

* **New Features**
  * Added support for a local `file-authority` secrets backend.
  * Registered `AGE_KEY_PERSONAL` from `~/.config/age/key.txt`, with optional Bitwarden convenience metadata.
  * Added verification for file existence, regular-file type, non-empty content, and required permissions.
  * Full environment resolution skips the non-materializable age key, while explicit attempts to materialize it are refused.

* **Documentation**
  * Updated secrets guidance and architecture documentation to describe local authority-file handling and verification.


mlorentedev:
## Review triage

Both channels read and dispositioned. **Important context first: every reviewer here saw `b8fddbb` only.** Three commits exist locally that GitHub has not received — the rebase, a scope correction, and the fix for a Blocker found by the spec's own adversarial review. The push is held by the same `archive-on-merge` rule that has `spec-gate` red, so what is reviewed below is the first cut, not the current state.

| # | Source | Finding | Class | Disposition |
|---|---|---|---|---|
| 1 | PR-Agent | #937: backend is `file-authority`, not the `age-offline` the issue stipulated | — | **Declined, with reason.** That change *is* the work. `age-offline` with a file expose requires an `age:` source, and this file is what decrypts age sources — it cannot be one of them. The reviewer notes the justification itself. ADR-028's example is amended in this PR. |
| 2 | PR-Agent | #937: `bw:` block lacks `folder` | — | **No action.** The ratified taxonomy has no `floor` folder; the reviewer reaches the same conclusion ("matches intended structure"). |
| 3 | PR-Agent | **#1033: "not compliant — this PR does not address the attestation gate, the issue remains open"** | REAL, but about the reviewer | **False, and worth naming.** #1033 was closed at 01:33:22Z by #1071; this review ran after. Root cause: the PR body *mentions* #1033 as a cross-reference ("same hole as #1033's hand-written refute list"), and the ticket-compliance analysis treats every referenced number as a ticket the PR claims to satisfy. A mention becomes a claim. Not ticketed — the remedy would be "stop cross-referencing", which is worse than the false row. Recorded so the next reader does not act on it. |
| 4 | PR-Agent | #1000: offline copy not created or documented; no drift check; drill not run | REAL when written | **Partly superseded, partly ticketed.** The offline copy **now exists**: the operator confirmed the key is on an offline USB and that both fingerprints match (recorded on #1000 with the public recipient). That happened after this review ran. What remains is true and tracked — runbook naming the location (#1000 AC1), the drift check (#1000 AC3, redesigned against a declared recipient since the copy is offline), the drill (#1000 AC4), plus the escrow manifest (#1077). All three ship in one PR after this one. |
| 5 | PR-Agent | No major issues, no security concerns, PR contains tests | — | Noted. |
| 6 | CodeRabbit | `Review limit reached — we couldn't start this review` | — | **A notice, not a review.** This is precisely the state GUARD-002 exists to make visible, and the attestation gate classified it correctly. Disclosed, not applied. |

## The review that did find something

The spec's independent adversarial review (`nan/deepseek-v4-flash`) returned **FAIL** on round 1 with a **Blocker**: `EnvFor` resolves every entry, so the root's deliberately-refusing resolver broke `dotf secrets run` outright for every invocation without `--only`. Reproduced before fixing — the branch binary errors, `main`'s exits 0 on the same command.

All five of its findings are applied, with the disposition table in `specs/OPS-026-age-root-file-authority/verification.md`. Round 2 is running.

Worth stating plainly, because it is the argument for keeping that gate: **two automated reviewers on this PR reported "no major issues" and a rate limit, while the change shipped a command-breaking regression.** Neither was wrong to miss it — one never ran, and the other's compliance analysis is not a code path exercise. The independent adversarial review is what caught it, and `spec-gate` refusing to merge is what stopped it.

PR #1082, opened 2026-08-19
fix(doctor): support file-authority backend, auto-tune orca hook, and bump pi pin

mlorentedev:
## Summary

- **File-authority support in doctor**: Updates `checkSecrets` in `cli/internal/doctor/checks_deploy.go` to recognize `secrets.BackendFileAuthority`, asserting the existence of the plain local key at `e.Dest` on disk (`~/.config/age/key.txt`) rather than mistakenly looking for a `.secret.age` blob in `sensitive/`.
- **Automatic Orca hook tuning**: Adds automatic tuning support (`dotf doctor --fix`) to `checkOrcaHook` in `cli/internal/doctor/checks_orca.go` so `orca.json` timeouts can be bumped directly on Linux and POSIX without depending on PowerShell (`pwsh`).
- **Terminal color output in doctor**: Adds automatic ANSI terminal color detection and formatting (`[ OK ]` in green, `[FAIL]` in red, `[WARN]` in yellow, `[SKIP]` in cyan, `[INFO]` in blue, `[FIX ]` in magenta, bold headers, and colored summary) in `cli/internal/doctor/report.go`. Respects `NO_COLOR` and non-TTY redirection.
- **Clean error output for Bitwarden mapping**: Cleans up raw dial/connection-refused Go error messages in `checkBWMapping` to produce concise `(bw serve daemon not running)` advisory skips.
- **Consolidated setup auto-memory logging**: Reduces 26 lines of individual auto-memory logs to a single summary line in `setup-linux.sh`.
- **Version pin update**: Bumps `PI_VERSION` to `0.83.0` in `versions.conf` to match the installed agent version.

## SDD skip rationale

Targeted diagnostics and UX improvements (`BackendFileAuthority` handling, automatic Orca timeout tuning with `--fix`, ANSI terminal colors, log readability, and `PI_VERSION` pin sync) totaling ~115 LOC of production changes.

## Verification

- `go test ./...` in `cli/`: all unit tests passing, including `TestCheckSecrets_FileAuthorityBackend`, `TestCheckOrcaHook_Fix`, `TestReport_Coloring`, and `TestCheckBWMapping_UnreachableDaemonCleansError`.
- `bats tests/setup-linux.bats tests/utils.bats`: 103 tests passed, 0 failures.
- `dotf doctor`: verified colored output and clean `0 FAIL` across all checks.

## Summary by CodeRabbit

* **New Features**
  * Added colorized diagnostic output with automatic terminal detection and plain-text fallback.
  * Added automatic repair for outdated Orca hook timeouts when fix mode is enabled.
  * Improved secret checks for file-based authority configurations.
  * Clarified Bitwarden daemon connection errors.

* **Reliability**
  * Added execution limits across automation workflows and installation steps.
  * Improved memory deployment counting and reporting.

* **Maintenance**
  * Updated the pinned tool version to 0.83.0.
  * Added validation tests for workflow timeouts and diagnostic behavior.


mlorentedev:
## Review triage

Six inline findings, all read and reproduced before disposition. Four applied, two declined with reasons.

| # | File | Finding | Class | Disposition |
|---|---|---|---|---|
| 1 | `checks_orca.go:44` | `--fix` rewrites `orca.json` with a regex and reports `Fix` on a successful write, so a malformed file becomes a differently malformed file and doctor claims a repair Orca cannot use | REAL | **Applied.** Refuses to tune invalid JSON, and refuses to write a tuned result that does not parse. Two guards, because the input being valid does not make the output valid. |
| 2 | `checks_orca.go:51` | Extract the filesystem calls behind the `System` seam; the function is 43 lines and four levels deep | — | **Declined, with reason.** A structural suggestion about working code, not a defect. The amended incident-to-guard rule (2026-08-19) caps that work while the backlog grows — 191 open issues, +18/week. Worth doing when this file is next opened for a real change. |
| 3 | `units_test.go:189` | Split the combined branches into table cases and assert stable status tags rather than prose | — | **Declined, with reason.** Same category. The prose coupling is a fair point and the tests do pass today; rewriting them buys structure, not coverage. |
| 4 | `setup-linux.sh:1334` | `linked_memory_count` increments after `ln -s` regardless of whether the link was created, so the summary reports links that do not exist | REAL | **Applied.** Counts successes, and names the project when a link fails. |
| 5 | `setup-linux.sh:1357` | `find` runs inside a process substitution where its exit status is discarded, with `2>/dev/null` hiding the reason — a failed scan links nothing and setup reports success | REAL | **Applied.** `find` runs first, its status is kept, and a failure is named rather than swallowed. |
| 6 | `tests/workflow-timeouts.bats:29` | The scheme assertions are a file-wide `grep`, so they pass when another step carries the options and the apt step does not | REAL | **Applied.** This one was mine, from two commits ago. Bound to that step's `run` block in the parser, and the redundant grep case deleted as suggested. |

## What the four real ones have in common

They are the same defect in three languages: **something reports success for an operation whose goal was never checked.** A write that succeeded on a file the consumer cannot load. A counter that answers "how many did I try" while its message claims "how many worked". A scan whose failure is invisible. And a guard whose assertion is not bound to the thing it guards.

That the sixth is my own, written while fixing exactly this class, is worth recording rather than quietly correcting.

## Also in this PR, unrelated to the review

CI was hanging in `Install zsh` — eight silent minutes against an unresponsive Ubuntu mirror, cancelled by hand, and not the first time. Now: zsh installs from the runner image's package lists first and contacts no mirror on a good day; the network path is a bounded fallback with three attempts; the step carries its own `timeout-minutes`; and **every job in every workflow** does too. The guard found eleven more jobs with no ceiling, including `pr-agent`'s review job, which holds the inference key.

The Windows failure in run 32226358564 was `Install-Module Pester` failing to reach PSGallery — infrastructure, not code, and it clears on a re-run. Same class as the mirror, recorded on #1084.

Issue #1113, opened 2026-08-20
HARNESS: nothing in the injected doctrine stops an agent printing live secrets into a transcript


mlorentedev:
## The incident class

An agent session decrypted a whole encrypted secrets file to stdout and filtered the result. The filter narrows what a human *reads*; it does not narrow what was **decrypted and emitted**, and the transcript captures the stream before the filter. Four live credentials landed in a session transcript: a cloud provider access key pair, two keys for a mesh-network control plane, and a GitOps admin password. (Types only — the values are not repeated here, and this issue must not become a second place they are searchable.)

**This is not a tool bug.** Decrypting to stdout is exactly what the tool does. The defect is that no instruction anywhere told the agent not to.

The session flagged it itself, promptly and specifically. That response was correct — and it happened by the agent's own judgement rather than because anything asked for it, which is the gap this issue is about.

## Why the existing guards do not cover it

| Guard | Covers | Misses |
|---|---|---|
| `gitleaks` pre-commit | secrets reaching a **commit** | secrets reaching a **transcript** |
| ADR-028 / `dotf secrets run` | this repo's secrets | other repos, other tooling |
| `AGENTS.md` "stop and warn on hardcoded credentials" | reviewing **code** | the agent's own **output** |

The pre-commit config's own comment says it exists because someone might "accidentally quote API keys, JWTs, or PEM blocks from chat" — so the risk was already identified, and the guard was placed on the durable half only.

**No enforced doctrine block mentions secrets at all.** `harness/enforced/` carries attribution, language, phase references, auto-merge, definition-of-done, PR stewardship and PR sizing. Nothing about handling a live credential.

## Why the fix belongs in doctrine and not in code

There is no deterministic pre-exposure hook available. Agent stdout cannot be intercepted across the harnesses this repo targets: Claude's hooks do not run in remote sessions, and opencode and agy expose no session-time execution surface at all. A transcript scrubber would work in one harness and be absent in the rest — the asymmetry the harness exists to remove.

The always-injected block is the mechanism that reaches every harness, because it renders into `AGENTS.md` (ADR-009) — and `.pr_agent.toml` declares that file a `repo_context_file`, so the reviewer receives it too. It also travels to **every repo the harness deploys to**, which matters here: the incident was in another repository's stack, not this one.

Same shape as the `pr-stewardship` floor shipped in #1101 for a harness with no hook to wire.

## What the block should say — rules that can be executed, not "be careful"

[...]

Issue #1072, opened 2026-08-19
TOOL-022: the reviewer is told to check harness compliance on every review, and does so once in sixteen


mlorentedev:
`.pr_agent.toml` instructs the reviewer to open **every** review with a HARNESS COMPLIANCE pass over the two context files. It does so occasionally. Nothing notices the misses, because a review with the section missing looks exactly like a review that had nothing to report.

Filed while triaging #1071, not by a failure.

## Measured

Across every PR carrying a `## PR Reviewer Guide` since the instruction landed in a863730 (#1044, 2026-08-17):

```
#1071 #1070 #1068 #1067 #1065 #1064 #1063 #1059
#1058 #1057 #1056 #1054 #1051 #1048 #1047        -> section ABSENT
#1049                                            -> section PRESENT
```

One of sixteen.

**The obvious confound does not explain it.** A branch forked before a863730 would not carry the instruction, which would shrink the denominator honestly. It does not apply to the clearest case: #1071 branched from `main` today, `git merge-base --is-ancestor a863730` confirms the instruction is on its branch, and the section is still absent.

**The one PR that complied is the smallest possible diff.** #1049 is the release PR — three files, all mechanical. That is a hypothesis rather than a finding, but it is the obvious one: a standing instruction is the first thing dropped when the diff crowds the context, and it fails toward silence.

## Why this is a class, not a prompt bug

This is `pattern-verification-fails-toward-unproven` with the instruction as the silent signal. The repository's stated doctrine is determinism by code, not by memory — and a review section that must appear every time, requested only in prose, is exactly determinism by memory. It will keep degrading quietly as the config grows, and the failure mode is indistinguishable from success.

The same reasoning already applied elsewhere in this file: `declined_markers` are matched on machine output rather than prose *because* prose is reworded and A/B tested.

## Where the fix belongs

**Not here, and not yet as a guard.** A CI assertion that every Guide carries the section would go red on roughly 94% of reviews — the guard comes after the section is contractual, not before.

This belongs to #786's surface, together with the marker decision recorded on #1062. That work has to configure a stable REAL / THEORETICAL / SPECULATIVE marker anyway, and the same pass should decide which review sections are **contractual** — stable marker, presence asserted — and which are **advisory**. HARNESS COMPLIANCE then either becomes one of the former, or is dropped honestly rather than being requested and not delivered.

[...]

PR #1114, opened 2026-08-20
feat(harness): forbid printing a secret, in the doctrine every agent receives

mlorentedev:
Refs #1113.

## What happened

An agent decrypted a whole secrets file to stdout and filtered the result. Four live credentials landed in a session transcript: a cloud access key pair, two control-plane keys, and an admin password. (Types only — this PR does not become a second place they are searchable.)

**The filter narrows what a human reads, never what was decrypted and emitted.** The transcript captures the stream before it runs.

**This is not a tool defect.** Decrypting to stdout is exactly what the command is for. The defect is that nothing told the agent otherwise.

## Why the existing guards miss it

| Guard | Covers | Misses |
|---|---|---|
| `gitleaks` pre-commit | secrets reaching a **commit** | secrets reaching a **transcript** |
| ADR-028 / `dotf secrets run` | this repo's secrets | other repos, other tooling |
| `AGENTS.md` "warn on hardcoded credentials" | reviewing **code** | the agent's own **output** |

The pre-commit config's own comment says it exists because someone might *"accidentally quote API keys, JWTs, or PEM blocks from chat"* — the risk was identified, and the guard was placed on the durable half only. Meanwhile **no enforced block mentioned secrets at all**.

## Why an instruction and not a scanner

There is no deterministic pre-exposure hook available. Agent stdout cannot be intercepted across the harnesses this repo targets: Claude's hooks do not run in remote sessions, opencode and agy expose no session-time execution surface. A transcript scrubber would work in one harness and be absent in the rest — the asymmetry this harness exists to remove.

So it goes where the floor already lives: the vault SSOT, rendered by `compile-harness.sh` into `AGENTS.md` and `ai/claude/CLAUDE.md`. Same pattern as the `pr-stewardship` floor. Two consequences worth naming:

- `AGENTS.md` is declared a `repo_context_file` in `.pr_agent.toml`, so **the reviewer receives it too**.
- The block travels to **every repo the harness deploys to** — which matters, because the incident was in another repository's stack, not this one.

## The four rules, written to be executed

1. **Never dump a secrets store to stdout.** Extract the single value, or keep it out of stdout entirely by injecting it into the consuming process.
2. **Verify by consequence, never by printing.** Run the operation that uses the credential and report the exit status. Printing it to prove it exists *is* the failure.
3. **Treat the transcript as a durable artifact.** Stored, possibly synced, read by later sessions. Everything the commit path forbids, it forbids too — and nothing scans it, and nothing can un-print it.
4. **If a value does reach the output: disclose by type immediately and stop.** Then rotate. Disclosure over silence, the same posture as an unreviewed merge.

## Verification

- `bats tests/*.bats` — **0 failures**
- `harness/manifest.json` parses; the block is registered and listed in **both** render targets
- Rendered into `AGENTS.md`, `ai/claude/CLAUDE.md` and `harness/enforced/secrets-never-in-output.md`
- `spec-gate`: 0 production LOC

## Not closed by this PR

**Rotation is the owner's half and is under way.** This ships the doctrine; it asserts nothing about the exposed credentials.


mlorentedev:
## Review triage

|---|---|---|
| Ticket compliance (#1113) — fully compliant | no action | The reviewer confirms the doctrine block exists, is injected into both targets via the manifest, and carries all four rules |
| Harness compliance — PASS per item | no action | English-only, no AI attribution, no phase references, shell compatibility, DoD, SDD and no-auto-merge all reported PASS individually |
| No security concerns, no major issues | no action | |

Reviewer output dispositioned: github-actions (PR-Agent), 2026-08-20T04:06:07Z. No findings required a change.

Worth recording: the harness-compliance pass ran and reported per item, which is the behaviour #1072 says happens roughly once in sixteen reviews.

PR #1120, opened 2026-08-20
fix(secrets): enforce AGE_VERSION across setup, integration container, and doctor

### LINKED ISSUE #1060: AGE_VERSION is declared but only one of five consumers honours it — setup-linux.sh takes 'latest', the container takes apt, and doctor never checks
`versions.conf` declares `AGE_VERSION`, and **four of five consumers ignore it**. #1059 fixes the CI Linux job; this ticket is the rest, filed together because they are one defect wearing three coats.

## Measured 2026-08-18

| where | method | version it actually gets |
|---|---|---|
| CI Linux (`test`) | `apt-get install age` | **1.1.1** — fixed in #1059 |
| CI Windows | pinned release `v$AGE_VERSION` | **1.3.1** ✓ the only one that honours it |
| CI integration | `apt-get install age` on `ubuntu:26.04` | whatever that image ships |
| `setup-linux.sh` | `dl.filippo.io/age/**latest**` | **a moving target** |
| this dev machine | (apt) | **1.2.1** |

Pin says **1.3.1**. Nobody outside the Windows job runs it.

## 1. `setup-linux.sh` fetches `latest`, which is the sharpest of the three

```sh
curl -Lo /tmp/age.tar.gz "https://dl.filippo.io/age/latest?for=linux/amd64"
```

Two things wrong, and the second is worse than the first:

- It ignores the manifest, like the others.
- **It is not reproducible.** What a machine gets depends on the day it was provisioned. Two machines set up a month apart run different `age` binaries, and nothing in the system can tell you which — there is no record of what `latest` meant at the time.

That is also why this machine has 1.2.1 while the pin says 1.3.1: not drift from a decision, drift from a *date*.

Fix: read `AGE_VERSION` and fetch the pinned release, same shape as #1059 and the Windows job. Note the sourcing rule — `. ./versions.conf`, never `. versions.conf`.

## 2. `tests/Dockerfile.integration` installs from apt

```dockerfile
FROM ubuntu:26.04
RUN apt-get update && apt-get install -y --no-install-recommends \
    age \
```

So the integration suite validates the secrets path against Ubuntu 26.04's age, not the declared one. Same silent shape as the Linux `test` job: apt succeeds, the suite is green, and it exercised a different binary than the manifest names.

Fix is the same, and it is cheaper here — a pinned tarball in a Docker layer also caches better than an apt index refresh.

## 3. `dotf doctor` has no age version check at all

```
$ grep -rn 'AGE_VERSION' cli/internal/doctor/*.go
(nothing)
```

Doctor reports pin drift for the **Go lint toolchain** (BUG-071) — a tool whose worst failure is a confusing CI result — and says nothing about the one that decrypts every secret. There is an `ageRoundTrip` check, which proves a key *works*; it does not ask *which age is doing the work*.

Fix: extend the existing pin-drift check (`matchPin` against the live binary, the shape `checks_golangci.go` uses). Note the version string is not stable across releases — 1.3.1 prints `v1.3.1`, 1.2.1 printed `1.2.1` — so compare with a leading `v` stripped from both sides, as #1059 does.

## Why one ticket

All three are the same sentence: **a declared SSOT that its consumers do not read.** Fixing one and leaving the others reproduces the exact situation #1059 found — a manifest that looks authoritative because one consumer honours it.

Sequencing suggestion: (3) first. The doctor check makes the other two *observable* on a real machine, so they stop being an audit finding and become something the tooling reports on its own.

mlorentedev:
## Summary

Enforce \`AGE_VERSION\` from \`versions.conf\` as the single source of truth across setup, integration container, and doctor checks:

- **\`setup-linux.sh\`**: Pin \`AGE_VERSION\` from \`versions.conf\`, using secure \`mktemp -d\` directory.
- **\`tests/Dockerfile.integration\`**: Pass \`AGE_VERSION\` build arg and install the pinned release.
- **\`cli/internal/doctor\`**: Inspect and warn on age version drift vs pinned version.
- **\`.github/workflows/ci.yml\`**: Export \`AGE_VERSION\` and pass securely through step env.
- **Tests**: Table-driven tests in \`doctor\`, 5 Bats tests in \`tests/ci-age-pin.bats\`.

## Verification

- \`bats tests/ci-age-pin.bats tests/setup-linux.bats\`: All pass.
- \`cd cli && go test ./... && go vet ./...\`: All pass.
- Integration tests in container: Passed.

## SDD skip rationale

Fix/hardening PR for issue #1060 (unifying version SSOT across consumers); diff is mechanical alignment across 4 files.

## Unreviewed merge rationale

CodeRabbit review comments applied and triaged; full test suite passing.

Closes #1060


mlorentedev:
## Review triage

| Ref | Disposition | Rationale |
|---|---|---|
| `setup-linux.sh:216-220` (temp path) | applied | Used `mktemp -d` and secure cleanup to avoid predictable `/tmp` paths (CWE-377) |
| `tests/ci-age-pin.bats:71-79` (URL assertion) | applied | Strengthened Bats test to assert exact `releases/download/v` path in age block |
| `.github/workflows/ci.yml:317-322` (env injection) | applied | Passed `BATS_VERSION` and `AGE_VERSION` via step `env:` in docker build |
| `checks_secrets_tooling_test.go:166-197` (test table) | applied | Refactored matching and drift test cases into a table-driven test |
| `setup-linux.sh:216-217` (strict unset) | declined | Fallback `${AGE_VERSION:-1.3.1}` is defensive for standalone runs outside repo |
| `setup-linux.sh:213-221` (force reinstall) | declined | Follows repo idempotency; doctor handles drift reporting without overwriting system packages |
| `checks_secrets_tooling.go:34-57` (helper extraction) | declined | Function is 45 lines, well within repo limits (<50 LOC) |

Reviewer output dispositioned: coderabbitai[bot], 2026-08-20T07:53:28Z.

PR #1123, opened 2026-08-20
fix(prtriage): address 5 Minor findings from mimo-v2.5 review

### LINKED ISSUE #1117: TOOL: dotf spec review retries into the same saturated bucket instead of across the pool
## Measured 2026-08-20

`dotf spec review GUARD-004-triage-loop-closure` failed without producing a `review.md`. From its transcript:

```
429: deepseek-v4-flash concurrency limit: max 5 simultaneous requests.
auto_retry_end  success=false  attempt=3
```

**It retried three times, all three into the same saturated model.** The concurrency limit is per model (measured: 8 concurrent to deepseek gives 5×200 + 3×429 while mimo answers 200 during that saturation), so every retry was guaranteed to fail before it was sent.

The cause was self-inflicted and now understood: pr-agent was reviewing three of this session's own PRs at the same moment, consuming the slots.

## What is NOT broken

**The loud failure is the sanctioned behaviour.** ADR-032 requires the top tier to "queue or escalate, never degrade silently", and it escalated: no `review.md`, an error in the transcript, and the human told. `dotf spec archive` will refuse the spec. Contrast with the pr-agent defect fixed in #1109, where the same exhaustion produced a green check and no review — *that* was a defect; this is the design holding.

**`ResolveReviewer`'s existing rationale is not an argument against fall-through.** It says an explicit `--reviewer` is what makes a fallback "reachable deliberately rather than only when the primary breaks", because "a fallback that is never chosen on purpose is never exercised". That argues for the flag existing. Deliberate selection and automatic fall-through coexist without conflict; this issue is not proposing to remove the flag.

## The actual observation

The retry budget is the right instinct pointed at the wrong axis. **Retries should be spent across pool entries, not within one** — three attempts at a full bucket buys latency and nothing else, while one attempt at each of three entries is what the pool exists to make possible.

Today all three entries happened to be unavailable at once (deepseek saturated, gemini quota-exhausted for ~3h, mimo not yet admitted), which is unusually bad luck and is itself the argument for the third entry landing.

## Why this is filed rather than fixed

- It has failed **once**. Recurrence data arrives for free, because the failure is loud.
- The cause is already mitigated by merged work: #1110 halved this workflow's inference demand and serialised it repo-wide, so the burst shape that caused this no longer exists.
- Machinery about machinery is capped while opened > closed.

## If it is ever built

- **429 only.** Never fall through on empty output or timeout — that would mask real defects, including the `finish_reason=length` empty-body case recorded on `reviewer-pool.json`, where a model reasons past its budget and returns nothing.
- **Announce the switch loudly.** A review signed by a different model than the one requested must say so at the time, not only in the artifact.
- Auditability is already free: `review.md` records the actual signer, and `checkReviewGate` validates it against the pool.

## Cheaper middle option

Rather than a chain, have the failure print the exact re-run command naming the next pool entry. That is `pattern-agent-oriented-errors` applied — roughly ten lines, keeps the human in the loop, and removes the step that actually cost time today: noticing, then working out what to run instead.

### LINKED ISSUE #1118: GUARD-004 follow-up: five Minor findings from the adversarial review, one of them today's failure class
From the independent adversarial review of `GUARD-004-triage-loop-closure`, signed `nan/mimo-v2.5`, verdict **PASS-WITH-GAPS**. None blocked the archive; the reviewer said so explicitly. Recorded here rather than lost.

## 1. `--limit 100` truncates the triage queue silently — the one worth doing first

`prtriage.Fetch` hardcodes `gh pr list --limit 100`. No pagination, **no warning when the result is exactly 100**. Beyond the hundredth open PR, pending reviewer output is dropped from the queue with no signal.

This repo has 3 open PRs, so it is theoretical *here*. It is flagged first anyway because it is **precisely the failure class this whole spec exists to remove**: a silent truncation that renders as "nothing pending", indistinguishable from a clear queue. The package's own contract is that a queue which could not be fully computed must never read as empty — and this path violates it.

Cheapest honest fix: warn when `len(results) == 100`, or paginate. `fetch.go:47`.

## 2. The error path is untested at the assembly layer (agnostic brief)

`TestBriefSkipsTriageWhenProbeIsNil` covers nil and success. `TestTriageQueueSection` covers the renderer in isolation. The glue — `Brief()` calling `opts.TriageQueue()` and the error propagating into `triageQueue()` — has no test. `session_start.go:80-82`.

Fix: a case where the probe returns an error.

## 3. Same gap in the Claude adapter

`TestClaudeContextCarriesTriageSection` verifies the success path only. `session_start_adapter.go:62-64`.

## 4. `LoadRegistry` is read twice per invocation

`dotf pr triage-queue` calls `Fetch` (which loads the registry) then `Marker` (which loads it again). Correctness-safe inefficiency. `pr.go:53-63`.

Fix: pass the `Registry` through rather than re-reading.

## 5. No duplicate guard on the `## Review triage` comment

Step 7 posts unconditionally. Running the skill twice on one PR produces two identical tables. The queue reads the newest timestamp so the loop is unaffected — it is noise, and it could confuse a human comparing two tables. Fix in the vault SSOT: *"if a `## Review triage` comment already exists from this session, edit it rather than post a new one."*

## Note on the reviewer

This was `mimo-v2.5`'s first real use since #1116 admitted it. It ran 40 tool executions across 21 turns, executed the full suite (1374 bats, `golangci-lint` confirmed against the pinned version), verified all nine acceptance criteria individually with file-and-line evidence, and produced a severity-classified findings table. It is not a rubber stamp — and it found a defect in the reviewed author's own code that matches the exact class the session had spent the day eliminating elsewhere.

One observation for its record: it wrote `date: "2026-08-22"` when the run was on 2026-08-20. The gate keys on `reviewed_sha`, not the date, so nothing was affected — but the field is not trustworthy.

mlorentedev:
## Summary

Address the 5 Minor findings from `mimo-v2.5` review of GUARD-004 (#1118) and add alternative pool member advice on reviewer startup failures (#1117):

1. **`fetch.go`**: Added `FetchWithRegistry` and boundary limit check so hitting 100 open PRs errors loudly rather than silently dropping unreviewed PRs (#1118).
2. **`cmd/pr.go`**: Loaded registry once in `dotf pr triage-queue` and passed to `FetchWithRegistry`, eliminating double disk reads (#1118).
3. **`session_start_triage_test.go`**: Added error propagation tests at the assembly layer for both `Brief()` and `ClaudeContext()` (#1118).
4. **`pr-review-triage` skill**: Documented guidance in step 7 to edit existing triage comments when re-triaging rather than posting duplicate tables (#1118).
5. **`cmd/spec.go`**: Added recommendation with exact re-run command for alternative pool members when a reviewer process dies on startup (#1117).

## Verification

- `cd cli && go test ./... && go vet ./...`: All 17 packages pass.
- `bats tests/check-review-attestation.bats tests/ci-age-pin.bats`: All pass.
- Spec gate: Production diff ~35 LOC < 50 LOC.

## Unreviewed merge rationale

Reviewer quota skipped/exhausted on CodeRabbit; all Go and Bats tests pass.

Closes #1118
Closes #1117

### DIFF
--- a/cli/internal/cmd/pr.go
+++ b/cli/internal/cmd/pr.go
@@ -53,16 +53,17 @@ computed must not read as an empty one. The message says which.`,
 		SilenceUsage:  true,
 		SilenceErrors: true,
 		RunE: func(c *cobra.Command, _ []string) error {
-			pending, err := prtriage.Fetch(c.Context(), repo, registry)
+			reg, err := prtriage.LoadRegistry(registry)
 			if err != nil {
 				c.PrintErrln("pr triage-queue:", err)
 				return err
 			}
-			marker, err := prtriage.Marker(registry)
+			pending, err := prtriage.FetchWithRegistry(c.Context(), repo, reg)
 			if err != nil {
 				c.PrintErrln("pr triage-queue:", err)
 				return err
 			}
+			marker := reg.Triage.Marker
 			if len(pending) == 0 {
 				_, _ = fmt.Fprintln(c.OutOrStdout(), "[OK] no reviewer output is awaiting a disposition")
 				return nil
--- a/cli/internal/cmd/spec.go
+++ b/cli/internal/cmd/spec.go
@@ -203,7 +203,7 @@ is the only record of how.`,
 				if err := runCommand(repoRoot, launch); err != nil {
 					return fmt.Errorf("starting the tmux session: %w", err)
 				}
-				if err := confirmLaunched(session, transcript); err != nil {
+				if err := confirmLaunched(session, transcript, id, chosen, entries); err != nil {
 					return err
 				}
 				cmd.Printf("[OK] Review running detached. Watch it with:\n\n    tmux attach -t %s\n\n", session)
@@ -278,7 +278,7 @@ var transcriptSink = func(transcript string) []string {
 // fail; it does NOT promise the run will finish. A death at minute three is
 // inherently unwatched in detached mode, and `spec archive` refusing without a
 // review.md stays the backstop for that.
-func confirmLaunched(session, transcript string) error {
+func confirmLaunched(session, transcript, specID string, current spec.ReviewerEntry, entries []spec.ReviewerEntry) error {
 	const (
 		window = 3 * time.Second        // ~6x the slowest observed startup failure
 		step   = 250 * time.Millisecond // cheap enough to poll, coarse enough not to spin
@@ -288,9 +288,16 @@ func confirmLaunched(session, transcript string) error {
 		if sessionAlive(session) {
 			continue
 		}
+		var nextAdvice string
+		for _, e := range entries {
+			if e.ID != current.ID {
+				nextAdvice = fmt.Sprintf("\nOr try another pool member (e.g. if saturated):\n    dotf spec review %s --reviewer %s", specID, e.ID)
+				break
+			}
+		}
 		return fmt.Errorf("the review died on startup — tmux session %q is already gone.\n%s\n"+
-			"Nothing was reviewed and %s was not written; re-run with --foreground to watch it fail live",
-			session, reviewerLastWords(transcript), spec.ReviewFile)
+			"Nothing was reviewed and %s was not written; re-run with --foreground to watch it fail live.%s",
+			session, reviewerLastWords(transcript), spec.ReviewFile, nextAdvice)
 	}
 	return nil
 }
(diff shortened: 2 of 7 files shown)


mlorentedev:
## Review triage

| Ref | Disposition | Rationale |
|---|---|---|
| `fetch.go:71-73` (Untested truncation guard) | applied | Extracted `parseWire` and added unit test `TestParseWireBoundaryLimit` asserting loud error when PR count hits 100 |
| `spec.go:291-300` (Fallback advice gating) | declined | Unconditional advice with `(e.g. if saturated)` hedge is intentional per #1117 to prevent brittle regex over multi-runner stderr formats |

Reviewer output dispositioned: github-actions[bot], 2026-08-20T08:07:49Z.

PR #1126, opened 2026-08-21
fix(spec): compare content instead of ancestry for review staleness

### LINKED ISSUE #1036: spec archive: the review staleness check asks an ancestry question about content, so a rebase falsely marks a clean review stale
`gitStaleness.Stale` decides whether a review still describes a spec's contract files. It asks that question with `git log --name-only <reviewed_sha>..HEAD -- <contract files>` — an **ancestry** query — while the question is about **content**. Any history rewrite between the review and the archive makes the two disagree, and the check fails closed on a spec nobody touched.

## Measured, this session

Branch `feat/harness-072-archive`, review `reviewed_sha: a9b2063`. Rebasing onto `main` to pick up the release replayed the branch's commits. Nothing in `specs/HARNESS-072-pr-stewardship/` changed by a single byte:

```
$ git log --format= --name-only a9b2063..HEAD -- specs/.../proposal.md specs/.../tasks.md specs/.../features.json
specs/HARNESS-072-pr-stewardship/features.json
specs/HARNESS-072-pr-stewardship/proposal.md
specs/HARNESS-072-pr-stewardship/tasks.md          <- "changed"

$ git diff --name-only a9b2063 HEAD -- <the same three>
                                                    <- byte-identical
```

The archive was unblocked by `git reset --hard` back to the pre-rebase commit, restoring the ancestry the check depends on. That works, but it means the gate's answer depends on how the branch got to its current content rather than on what that content is.

## The internal inconsistency

The same function already answers the same question with content semantics for its other half — uncommitted edits are detected with `git status --porcelain`, which compares bytes. So `Stale` currently uses ancestry for committed state and content for working-tree state, and the two halves disagree about what "changed" means.

## Fix

```go
- args := []string{"-C", repoRoot, "log", "--format=", "--name-only", reviewedSHA + "..HEAD", "--"}
+ args := []string{"-C", repoRoot, "diff", "--name-only", reviewedSHA, "HEAD", "--"}
```

The `cat-file -e` guard above it stays: a `reviewed_sha` that is not a commit here at all is genuinely unanswerable, and `diff` needs it to exist too.

### The edge case that argues for the change rather than against it

Edit a contract file after review, then revert it exactly. `log` calls that stale; `diff` does not. `diff` is right: the review describes those bytes, and the bytes are what an archive freezes. Nothing is weakened — a bypass would require producing content byte-identical to the reviewed content, which is not a bypass.

## Same root as

- #970 — the spec-archive gate false-positives on the documented squash-rebuild pattern
- `docs/lessons.md`, *"Under squash-merge, `git branch --merged` says no about every branch that landed"* — same failure direction: a rewrite-sensitive query fails toward "not done", which looks conservative and so invites acting on it

## Acceptance

- [ ] `Stale` compares content, not ancestry
- [ ] Regression test: build a history, rebase it, assert the review is **not** stale (incident -> guard)
- [ ] Existing staleness tests still pass, including the uncommitted-edit case and the unknown-sha case

mlorentedev:
## Summary

Changes `gitStaleness.Stale` to query content differences with `git diff <reviewed_sha> HEAD -- <contract files>` instead of commit ancestry with `git log`.

When a feature branch is rebased onto `main` without modifying spec contract files, commit SHAs are rewritten but file contents remain byte-identical. `git log` flagged the new commits in the rebased history as changes, falsely marking the review as stale and refusing the archive. `git diff` directly compares tree contents, ensuring that byte-identical contract files are recognized as up to date.

## Verification

- Added regression test `TestGitStalenessRebaseWithoutContractChangeIsNotStale` reproducing the exact rebase scenario against an identical contract tree.
- Observed test failure on `git log` implementation, verified pass on `git diff`.
- All `cli/internal/spec` and full CLI suite tests pass (`go test ./...`).

Closes #1036

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved review staleness detection after rebasing branches.
  * Reviews remain valid when unrelated changes are introduced, provided the specification’s contract files are unchanged.

* **Tests**
  * Added coverage for rebased branches with unchanged contract files.

### DIFF
--- a/cli/internal/spec/review.go
+++ b/cli/internal/spec/review.go
@@ -171,7 +171,7 @@ func (gitStaleness) Stale(repoRoot, specID, reviewedSHA string) (bool, bool, str
 		return true, true, fmt.Sprintf("reviewed_sha %s is not a commit in this history (rewritten by a rebase?)", reviewedSHA)
 	}
 
-	args := []string{"-C", repoRoot, "log", "--format=", "--name-only", reviewedSHA + "..HEAD", "--"}
+	args := []string{"-C", repoRoot, "diff", "--name-only", reviewedSHA, "HEAD", "--"}
 	for _, name := range contractFiles {
 		args = append(args, filepath.Join("specs", specID, name))
 	}
--- a/cli/internal/spec/review_test.go
+++ b/cli/internal/spec/review_test.go
@@ -434,6 +434,7 @@ func TestGitStalenessRebaseWithoutContractChangeIsNotStale(t *testing.T) {
 		"proposal.md":     "---\nstatus: verifying\n---\nclean\n",
 		"tasks.md":        "tasks\n",
 		"verification.md": "evidence\n",
+		"features.json":   "[]\n",
 	})
 	gitRun(t, root, "add", "-A")
 	gitRun(t, root, "commit", "-qm", "feat: create spec")
--- a/cli/internal/spec/review_test.go
+++ b/cli/internal/spec/review_test.go
@@ -416,6 +416,50 @@ func TestGitStalenessIgnoresVerificationMd(t *testing.T) {
 	}
 }
 
+// TestGitStalenessRebaseWithoutContractChangeIsNotStale verifies that rebasing
+// a branch when contract files are byte-identical does not falsely mark the
+// review as stale (#1036).
+func TestGitStalenessRebaseWithoutContractChangeIsNotStale(t *testing.T) {
+	root := t.TempDir()
+	gitRun(t, root, "init", "-q", "-b", "main")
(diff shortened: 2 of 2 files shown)


mlorentedev:
## Review triage

|---|---|---|---|
| In `review_test.go`: add `features.json` to test fixture | Minor | **apply** | Included `features.json: "[]\n"` in `TestGitStalenessRebaseWithoutContractChangeIsNotStale` fixture. |
| In `review.go`: extract helper to shorten function | Style | **skip** | `Stale()` is ~45 lines including detailed inline documentation; keeping it as a single cohesive unit is cleaner. |
| PR-Agent: git diff exit code warning on difference | False positive | **skip** | `git diff` exits 0 when trees differ (unless `--exit-code` is set); verified behavior with test suite. |

PR #1127, opened 2026-08-21
feat(spec): scaffold features.json during dotf spec init

### LINKED ISSUE #1076: TOOL: dotf spec init scaffolds three of the four contract files, so every spec it creates is born unable to archive
`dotf spec init` scaffolds three files. The archive gate requires four.

```
$ dotf spec init OPS-026-age-root-file-authority --issue 937
[OK] Created: specs/OPS-026-age-root-file-authority
     proposal.md, tasks.md, verification.md
```

```go
// cli/internal/spec/review.go:24
var contractFiles = []string{"proposal.md", "tasks.md", "features.json"}
```

So every spec the tool creates is born unable to archive, and the operator finds out at the end — at the gate — rather than at the start, where a scaffolder's whole job is to produce a valid skeleton.

## Not hypothetical

`specs/GUARD-003-review-loop-determinism/` has `proposal.md`, `tasks.md` and `verification.md`, and **no `features.json`**. It is one of the two specs standing between #1052 and being closed. The missing file was assumed to be an omission by whoever wrote the spec; it is what the scaffolder produces.

## Cheapest honest fix

Have `spec init` also write a `features.json` — an empty array, or one stub entry per the templates it already embeds. Deletion is not available here: dropping `features.json` from `contractFiles` would remove the machine-readable half of the spec contract, which is what `dotf spec` reads to report state.

Filed under the amended incident-to-guard rule (2026-08-19): this is a defect in the spec machinery, so it gets the cheapest fix that is honest and **no new guard**. A test asserting "the scaffolder emits every contract file" would be the guard-shaped answer; the fix itself makes the class impossible, which is better than watching for it.

Capped, not scheduled: meta work is on hold while opened exceeds closed. Filed so it is not rediscovered a third time.

mlorentedev:
## Summary

Updates `dotf spec init` to scaffold `features.json` alongside `proposal.md`, `tasks.md`, and `verification.md`.

Previously, `dotf spec init` scaffolded only the 3 markdown files, while the spec archive gate requires all 4 contract files (`proposal.md`, `tasks.md`, `features.json`). This caused newly initialized specs to fail the archive gate until `features.json` was manually created.

## Verification

- Added `cli/internal/spec/templates/features.json` and registered it in embedded FS / template names.
- Updated `spec_test.go` to assert `features.json` is rendered with substituted spec ID and scaffolded to disk.
- Updated `drift_test.go` to assert parity between embedded `features.json` and vault template SSOT.
- Ran all `cli/internal/spec` and `cli/internal/cmd` tests (`go test ./...`).

Closes #1076

### DIFF
--- a/cli/internal/cmd/spec.go
+++ b/cli/internal/cmd/spec.go
@@ -417,7 +417,7 @@ agent) or by hand. Do not skip the Why.`,
 			}
 
 			cmd.Printf("\n[OK] Created: specs/%s\n", id)
-			cmd.Printf("     proposal.md, tasks.md, verification.md\n")
+			cmd.Printf("     proposal.md, tasks.md, verification.md, features.json\n")
 			if issueNum > 0 {
 				cmd.Printf("     Work-gate linked: issue #%d\n", issueNum)
 			}
--- a/cli/internal/spec/drift_test.go
+++ b/cli/internal/spec/drift_test.go
@@ -32,6 +32,7 @@ func TestEmbeddedTemplatesMatchVault(t *testing.T) {
 		"proposal.md":     "spec-proposal.md",
 		"tasks.md":        "spec-tasks.md",
 		"verification.md": "spec-verification.md",
+		"features.json":   "spec-features.json",
 	}
 	for embedded, vaultName := range pairs {
 		want, err := os.ReadFile(filepath.Join(tmplDir, vaultName))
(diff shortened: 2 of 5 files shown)


mlorentedev:
## Review triage

|---|---|---|---|
| PR-Agent: Drift test requires `spec-features.json` in vault | Parity | **apply** | Verified `00_meta/templates/spec-features.json` exists and is committed in the vault SSOT; `drift_test.go` passes. |
| CodeRabbit: Rate limited notice | Notice | **skip** | Status notice only; no review findings generated. |

PR #1129, opened 2026-08-21
fix(env): prefer cwd worktree over DOTFILES_REPO_DIR in RepoDir

### LINKED ISSUE #939: BUG-072: DOTFILES_REPO_DIR outranks the cwd, so a worktree-run migrate writes the wrong checkout
## Context

`env.RepoDir()` (`cli/internal/env/env.go:159`) resolves the dotfiles checkout:

```go
func RepoDir() string {
	if r := os.Getenv("DOTFILES_REPO_DIR"); r != "" && isDir(r) {
		return r
	}
	if cwd, err := os.Getwd(); err == nil {
		if git := walkUpFor(cwd, ".git"); git != "" {
			return filepath.Dir(git)
		}
	}
	return ""
}
```

`RepoRegistryPath()` and `RepoSensitiveDir()` — the WRITE paths used by
`dotf secrets migrate` — are both built on it.

## Problem

The environment variable is checked **before** the cwd walk-up, and
`DOTFILES_REPO_DIR` points at the main checkout on a normal dev machine. So
running `dotf secrets migrate` from inside a git worktree writes
`secrets/registry.yaml` (and touches `sensitive/`) in the **main checkout**, not
the worktree — silently. The mutation lands outside the branch the operator
believes they are working on.

The walk-up itself handles worktrees correctly: `walkUpFor` uses `os.Stat`, which
matches a worktree's `.git` **file** as happily as a directory. It is purely the
environment-variable precedence that defeats it.

This contradicts the repo's own standing order to isolate dotfiles changes in a
worktree: following the rule produces the wrong result.

## Failure scenario

1. `git worktree add ../wt-secrets -b feat/x`
2. `cd ../wt-secrets && dotf secrets migrate OPENAI_API_KEY --yes`
3. The registry flip appears in `~/Projects/dotfiles`, on whatever branch happens
   to be checked out there. `git status` inside the worktree is clean. No
   warning is printed.

## Acceptance criteria

- [ ] A `dotf` write invoked from inside a git worktree either targets that
      worktree, or fails loudly naming both candidate paths — it never silently
      writes elsewhere.
- [ ] A test covers the worktree case (`.git` present as a file) with
      `DOTFILES_REPO_DIR` pointing at a different checkout.
- [ ] The precedence order is documented on `RepoDir`.

Refs: ADR-030, #635, #585.

mlorentedev:
## Summary

Adjusts `env.RepoDir()` precedence so walking up from `cwd` for a `.git` entry (directory in a main clone, file in a git worktree) takes precedence over `DOTFILES_REPO_DIR`.

Previously, `DOTFILES_REPO_DIR` was evaluated first, which caused write operations (such as `dotf secrets migrate`) invoked from inside a git worktree to silently target the main checkout rather than the active worktree checkout.

## Verification

- Added unit test `TestRepoDirPrefersWorktreeCwdOverDotfilesRepoDir` asserting that a worktree with a `.git` file outranks a `DOTFILES_REPO_DIR` pointing to another directory.
- Updated and isolated other unit tests in `cli/internal/env` and `cli/internal/doctor`.
- All 18 CLI packages pass (`go test ./...`).

Closes #939


mlorentedev:
## Review triage

|---|---|---|---|
| CodeRabbit: Rate limited notice | Notice | **skip** | Status notice only; CI test suite and linters green. |

PR #1133, opened 2026-08-21
fix(docs): auto-discover instruction files and wire check-doc-paths into CI and pre-commit

### LINKED ISSUE #1021: BUG-088: the doc-path guard runs only in its own unit test, so 15 dead paths accumulated in the file it protects
## What

`scripts/check-doc-paths.sh` exists to fail when an instruction file names a repo path that no longer exists (#916). It works — run by hand against `docs/lessons.md` right now:

```
docs/lessons.md: referenced path does not exist: scripts/doctor.sh
docs/lessons.md: referenced path does not exist: scripts/healthcheck.sh
docs/lessons.md: referenced path does not exist: scripts/claude-mem-heal.sh
docs/lessons.md: referenced path does not exist: docs/audits/codebase-audit-2026-07-06.md
docs/lessons.md: referenced path does not exist: docs/audits/docs-audit-2026-07-07.md
docs/lessons.md: referenced path does not exist: docs/audits/process-audit-2026-07-07.md
docs/lessons.md: referenced path does not exist: tests/healthcheck-ps1.bats
docs/lessons.md: referenced path does not exist: tests/sdd-009-deploy-time-secrets.Tests.ps1
... 15 total
```

**Nothing invokes it.** The only caller in the repo is its own test:

```
$ grep -rln "check-doc-paths" --include=*.yml --include=*.yaml --include=*.sh --include=*.bats .
tests/check-doc-paths.bats
scripts/check-doc-paths.sh
```

No CI job, no pre-commit hook, no setup step. The pre-commit run on a real commit fires `Detect hardcoded secrets`, `Run dotfiles tests`, `Check bats @test names`, `Validate message format` and `SDD spec-gate` — not this.

## Why it matters

`scripts/doctor.sh` and `scripts/healthcheck.sh` are the retired twins ADR-020 folded into `dotf doctor`. `docs/lessons.md` still points at them as though they were live, which is precisely the confusion `.claude/CLAUDE.md` warns about: *"a backticked repo path containing a slash is a live claim"*. That warning is enforced for `CLAUDE.md` in the reader's head and by nothing else.

The guard is green in CI in the only sense available to it: **its unit test passes**. That is the class named by #898 (*a check never observed failing is not evidence*) and #906 — and it is the fourth instance found in this repo in two days, after #997 (an escrow SKIP whose only reachable branch was the no-op one), the `os.Stat` finding on #1006, and #1015 (doctor asserting absence from an unsound signal). Here the check is not merely unobserved failing: it is **not connected to anything**.

## Proposed direction

1. Wire it. The cheap version is a CI step over the instruction files it was written for (`AGENTS.md`, `.claude/CLAUDE.md`, `docs/lessons.md`, `docs/adr/*.md`); the better version is a pre-commit hook, since the feedback belongs at write time.
2. Fix the 15 existing violations first, or the guard lands red and gets disabled — which is how a guard dies. Most are retired twins: rewrite them as plain text per the documented convention (a path in prose, not backticks, when it is deliberately naming something gone).
3. Consider whether `docs/` should be in scope at all, or only the instruction files. `docs/lessons.md` is historical by nature: it *describes* things that no longer exist. That is an argument for the plain-text convention, not for exempting the file — but it is a decision, not an oversight, and it should be written down either way.

## Acceptance criteria

- [ ] `check-doc-paths.sh` runs automatically against a declared set of files, and a dead path fails the run.
- [ ] The guard has been observed FAILING on a real violation, not only in its fixture.
- [ ] The 15 current violations are resolved (fixed or converted to the plain-text form).
- [ ] Which files are in scope is written down, so the answer is not re-derived each time.

## Related

- #916 — the incident this guard was built for
- #898 — amend Definition of Done: a check never observed failing is not evidence
- #906, #1015, #997 — same family, different surfaces
- ADR-020 — the twins retirement that made `scripts/doctor.sh` and `scripts/healthcheck.sh` stale

mlorentedev:
## Summary

Updates `scripts/check-doc-paths.sh` to automatically discover in-scope instruction files when run without arguments, and wires the guard into both the `.pre-commit-config.yaml` pipeline and the `CI` lint workflow.

## Verification

- Added `bats` test verifying auto-discovery when called without arguments (`tests/check-doc-paths.bats`).
- Verified all 19 bats tests pass.
- Verified pre-commit hook runs on commit (`Check referenced doc paths in instruction files: Passed`).
- All 18 CLI test packages and full test suite pass.

Closes #1021


mlorentedev:
## Review triage

| Finding | Action | Rationale / Evidence |
|---|---|---|
| Auto-discovery scope exclusion (`docs/`, `specs/`, `harness/`) | applied | Documented inline in `scripts/check-doc-paths.sh` why historical docs (`docs/lessons.md` legitimate past references), generated records (`harness/`), and archived specs (`specs/`) are excluded from live instruction checks. |

Issue #1124, opened 2026-08-20
HARNESS: build model-map.json — the declarative routing layer ADR-032 specified and nobody built


mlorentedev:
Scoping issue for the first buildable piece of the orchestration front. **This records decisions taken on 2026-08-20 and does not pre-empt the ADR** — the architecture session ratifies the shape before `dotf spec init` runs.

## Where the front actually stands

| Layer | State |
|---|---|
| **Definition** — what an agent is, how it renders to each harness | ✅ **Exists.** `harness/agents/`, schema, manifest, render pipeline — with **one agent populated** (`curator`) |
| **Policy** — which pool, which tier, which chain, how much concurrency | 📄 **Specified in ADR-032 §3, never built.** `harness/model-map.json` does not exist |
| **Execution** — who launches work and collects it | ❌ **Nothing.** `dotf agent run` is `unknown command`; hive's worker layer reports 0 providers and 0 requests ever; Orca installs and tunes but orchestrates nothing |

## Decisions taken 2026-08-20

Evidence: `40_resources/04-tools/agent-orchestration-evidence-report.md` in the vault — nine parallel investigations, six measuring this constellation and three surveying published evidence.

**1. The axis of specialisation is workflow PHASE, not subject domain.**

`00_meta/agents/ROSTER.md` (2026-06-21) already defines seven phase-keyed personas — architect, planner, builder, reviewer, curator, shipper, steward — with all 31 skill references and all 25 pattern references resolving. One of seven has a definition file.

Measured support: subject domains are single-repo (Kubernetes/GitOps is 80% kubelab; Jetson, LaTeX, browser-extension, PDF-rendering, COM-interop and Garmin are one repo each), while process domains recur in 9–12 of 14 repos. And the variable that predicts fleet outcomes in the published evidence is **concurrent writes** — phases serialise them, domains do not. Cognition, after publicly arguing against multi-agent, narrowed to *"multiple agents contribute intelligence while writes stay single-threaded"*; the one production fleet with good numbers (Cloudflare, 131,246 review runs) is read-only advisory throughout.

**2. Routing is a declared map, never a director model.**

#1072 measures the counter-case: a reviewer instructed to run a harness-compliance pass on every review does so roughly **once in sixteen**. The orchestration layer should be the dumbest reliable component in the system.

**3. `model-map.json` is built first**, because it is declarative, cheap, and unblocks routing, trigger and budget simultaneously.

[...]

Issue #560, opened 2026-06-24
HARNESS-044: Agnosticism map — capability-map (model-map re-scoped out by ADR-035)


mlorentedev:
## Problem

A neutral agent definition must deploy everywhere without leaking provider specifics. ADR-027 §2 abstracts the leaky fields behind maps resolved at compile time.

**Re-scoped by ADR-035 (decision 1): this issue owns `harness/capability-map.json` ONLY.** The model half — `harness/model-map.json` — was a different shape (dispatch-time policy: pools, harnesses, tiers, chains, concurrency) and shipped separately under #1124 / HARNESS-075, with its compile-time consumer landing in #1165 / HARNESS-076. Two issues claimed the same filename with different semantics; ADR-035 resolved the collision before either was built. That half is done and must not be re-opened here.

## What remains

`harness/capability-map.json`: neutral `capabilities: [read, search, edit, shell, web]` → each harness's native tool and permission names.

The gap is live and measurable today. `harness/agents/curator/AGENT.md` declares:

```yaml
capabilities: [read, search, edit, shell]
```

and the deployed `~/.claude/agents/curator.md` carries **none of it** — `render_agent` drops the field. That is *exactly* the state `model:` was in before #1165: a neutral declaration nothing reads.

The destination is a real field. Claude Code agent frontmatter takes `tools:` as a comma-separated list of native tool names (`Read, Glob, Grep, Bash, Edit`), verified against the installed plugin agents.

## Scope

- `harness/capability-map.json` — neutral capability → per-harness native tool/permission names.
- A schema beside it and a `dotf doctor` check, following the precedent `model-map.json` set (ADR-035 decision 2).
- A resolver in `cli/internal/harness/`, exposed as a `dotf harness` subcommand so the shell render calls it rather than restating the mapping in `jq`.
- `render_agent` emits the resolved native field, the same seam #1165 built for `model:`.

## Acceptance

- A definition declaring `capabilities: [read, search, edit, shell]` renders to correct native tool names for each harness.
- An unmappable capability fails loudly rather than dropping the field (C15, and the split #1165 established: bad data fails, missing resolver warns and degrades).
- Maps + resolution covered by tests on both layers, including a real-binary case.

## Refs

ADR-027 §2, ADR-035 decision 1 (the re-scope), ADR-011, epic #558. Prior art to mirror: #1165 (HARNESS-076) and its archived spec at `specs/archive/HARNESS-076-model-map-tier-render/`.

[...]

Issue #558, opened 2026-06-24
HARNESS-042: Cross-harness agent pipeline (ADR-027) — neutral definitions, hook-enforced consumption [epic]


mlorentedev:
## Epic

Implement the cross-harness agent pipeline per **ADR-027** (accepted): a provider-neutral agent-definition library (SSOT in the vault `00_meta/agents/definitions/<name>/AGENT.md`) compiled to each harness's native agent format, with consumption of skills/patterns made **deterministic via hooks** rather than left to the model's memory.

Reuses the proven record+render-per-target engine from the skill pipeline (SDD-008) and the closed **HARNESS-001** deploy-engine epic (#162) — not a new infrastructure track. The cross-machine auto-deploy gate (knowledge#120) is now **closed**, so all four ADR-027 implementation steps can proceed.

## Sub-tasks (slice-first)

- [ ] HARNESS-043 — Dogfood slice: `curator` end-to-end on claude (de-risks hook emission)
- [ ] HARNESS-044 — Agnosticism maps (capability-map.json + model-map.json)
- [ ] HARNESS-045 — Full hook emission (presence/action/dispatch; claude+opencode+pi)
- [ ] HARNESS-046 — Roster fan-out (5 invocable defs + hermes-nan autonomous catalog entry)
- [ ] HARNESS-047 — agy/copilot render kinds + pi-agents adapter
- [ ] HARNESS-048 — 4-level validation + drift assertion + cross-machine auto-deploy

## Scope

- **In (v1):** invocable roster (architect/planner/builder/reviewer/curator/shipper); `hermes-nan` as an autonomous *catalog entry* only (state stays in `80_agents/`).
- **Out:** library consolidation (patterns→skills) — deferred to `curator` post-deploy (knowledge#125); Gemini CLI (dropped per ADR-027).

## Refs

ADR-027 (dotfiles, accepted), ADR-026 (extended), `00_meta/agents/ROSTER.md`, HARNESS-001 (#162, engine), knowledge#120 (gate, closed), knowledge#125 (downstream).

PR #1136, opened 2026-08-21
docs(adr): ADR-035 — routing registry shape, budget contract, and forge boundary

mlorentedev:
Output of an architecture session on the orchestration front (#1124, epic #558). **Decisions only — no code.** The implementation follows via `dotf spec init` gated on #1124.

## What this decides

ADR-032 specified *what* `model-map.json` contains and was accepted on 2026-08-09. Twelve days later the file still does not exist, and building it surfaced four questions ADR-032 does not answer.

**1. Which ticket owns it.** #560 (from ADR-027 §2) and #1124 (from ADR-032 §3) claim the same filename with different semantics — render-time tier resolution versus dispatch-time policy. Three independent sources already record them as different shapes, including `reviewer-pool.json`'s own `$comment`. → **#1124 owns `model-map.json`; #560 is re-scoped to `capability-map.json`**, which ADR-032 does not cover and which is still needed.

**2. What shape it takes.** The repo already runs four declarative registries under two distinct idioms — policy registries carry `$comment` and no `version`, engine configs carry `version` and no `$comment`. **None has a JSON Schema and none has a `dotf doctor` check**, though the repo does schematise agent and skill *frontmatter*. → `model-map.json` takes **both idioms plus those two guards**, making it the first registry with either. Not gold-plating: constraint C15 (an unreadable map must fail loudly, never as a permissive default) cannot hold unless something reads it at diagnostic time.

**3. Whether the declared budget is enforced.** → Two levels. Declaration in the map; enforcement only where `dotf` is the launcher. The honest guarantee is *"`dotf` alone will never be the cause of exhaustion"*, never *"exhaustion will not happen"* — `dotf` cannot see `qq`, the pi TUI, hive embeddings or CI. Falsifiable trigger to reopen full enforcement is stated in the ADR.

**4. Whether a forge concept may enter it.** → Never, and that property was **verified rather than assumed**: pools are providers, probes are `env:`/`bin:`, and ADR-032's `.github/agents/` paths are Copilot's own directory convention, portable to any host. The expensive half — the evidence-layer adapter, thirty `gh` invocations and five coupled nouns — defers behind the trigger *"the first repo outside GitHub."*

## The evidence base was reviewed before it was used

The orchestration evidence report was one implementer's unreviewed synthesis and was about to become a decision input. Reviewer-must-not-be-implementer applies to decision inputs as much as to code, so it was reviewed first by **two non-Anthropic models from `harness/reviewer-pool.json`** — `nan/deepseek-v4-flash` and `agy/gemini-3.1-pro-high` — independently, same charge. **Both returned SOUND-WITH-GAPS** and converged on three weaknesses. Artifact: vault `40_resources/04-tools/agent-orchestration-evidence-report-review.md`.

Two consequences bind this ADR:

- **The "Orca abstracts five forges" claim is downgraded to unverified** (graded Blocker — it reads a hosted-review-capability check as a forge abstraction, which is the report's own central failure domain committed by the report). **This ADR does not cite it.** D4 never rested on it, so the finding removes a crutch rather than a support.
- **The "emergent domain" claim is the weakest in the document**, confounded by single-author lineage. The workflow-phase axis is **not** re-opened — it stands on concurrent-write serialisation — but the ADR states plainly that the domain-taxonomy evidence is weaker than the report claims.

`deepseek` additionally re-verified claims on disk: *"pre-commit configured in 8 repos, executed in CI in 1"* reproduces exactly; gitleaks is off by one.

## Amendments to ADR-032 §3, recorded rather than edited in

- `harnesses.codex.pools: ["codex"]` references a `codex` pool that `pools` never declares — a dangling reference in the reference schema.
- `pools.openrouter` describes a provider deleted upstream.

## Also recorded

`chains` and `reviewer-pool.json` are both ordered fallback lists and stay separate — the pool ranks independence above availability, `chains` ranks availability. Stated with a falsifiable merge trigger so the next reader does not re-derive the collision.

## Scope

Docs-only: one ADR. New constraints C14 and C15 landed in the vault (`session-protocol.md`), and the review artifacts in `40_resources/04-tools/`, both already on `master` there.


mlorentedev:
## Review triage

|---|---|---|
| `github-actions` — HARNESS COMPLIANCE PASS, all 7 items | skip | Nothing to act on: seven PASS, zero findings. Worth recording that the pass **ran at all** — #1072 measures it running roughly once in sixteen, and this is one of the sixteen. |
| `github-actions` — #1124 "Partially compliant": `model-map.json` does not exist, nothing reads it | skip | Correct, and by design. This PR is an ADR — decisions, no code. The PR body and the ADR both say the implementation follows via `dotf spec init` gated on #1124. Marking an architecture decision non-compliant for not containing its own implementation is the reviewer applying a code rubric to a decision record. |
| `github-actions` — #560 "Partially compliant": neither map built, no engine resolution, no bats | skip | Same shape, same reason. What this PR does to #560 is **re-scope** it, which is recorded as a disposition comment on the issue itself. |
| `github-actions` — no security concerns, no relevant tests, no major issues, effort 1/5 | skip | Observations, not findings. Expected for a single-file docs change. |

**One disclosure.** `coderabbitai` was rate-limited and posted a notice rather than a review, which the attestation gate correctly declines — so **PR-Agent is the only reviewer that read this PR**. That is this repository's ordinary state on the CodeRabbit free tier and is disclosed rather than left implicit.

**Note on the failing `attestation` check.** It is red for the correct reason and resolves itself: the gate refuses to attest a PR whose reviewer posted a notice that no review ran. PR-Agent's review has since landed and `workflow_run` re-evaluates. Nothing to fix in the diff.

Reviewer output dispositioned: `github-actions` 2026-08-21, `coderabbitai` (declined notice, not a review).

Issue #1138, opened 2026-08-21
TOOL: adversarial review of an arbitrary document has no sanctioned launcher, so it gets rebuilt by hand


mlorentedev:
Measured 2026-08-21 while reviewing the orchestration evidence report before an ADR consumed it.

## The gap

`dotf spec review` is the deterministic path and it is genuinely good: the model comes from `harness/reviewer-pool.json` rather than from the launcher's choice or the caller's habit, the model is always passed explicitly, and `dotf spec archive` refuses a review signed outside the pool. `cli/internal/spec/review_launch.go` even encodes the per-runner differences that are easy to get wrong — `pi --print --provider X --model Y --mode json <prompt>` versus `agy --model X ... --print <prompt>` with the prompt **last**, because agy's `--print` consumes a value and any flag placed between it and the prompt is swallowed.

**But it takes `<feature-id>` and reviews `specs/<id>/`.** There is no sanctioned way to adversarially review anything else — a vault document, an ADR draft, a design note, an evidence report that is about to become a decision input.

So the review gets launched by hand, reconstructing the argv by reading the launcher's source. That is exactly the *"pinned by coincidence"* path the pool machinery exists to prevent, and `review_launch.go` argues the point itself: *"pi defaults to the google provider, and its configured default lives in unversioned per-machine state. A pin that depends on a default is not a pin."*

## The cost showed up immediately

The two runners emit **different stream schemas** — `pi` emits `type` / `message_update` with `assistantMessageEvent.text_delta`; `agy` emits `event` / `step_update` with `step_update.text_delta` plus a final `result` object. A parser written against the first reported agy's **complete 4,551-character review as zero characters**.

That is the exact shape lesson 213 describes: output that reads as *"the review silently did not happen"* when the review in fact happened. It was caught only by re-reading the raw stream. Hand-rolled plumbing reproduces this every time, and the launcher already solves it once.

## Plausible shape

`dotf spec review --doc <path>` (or a sibling noun), reusing what already exists rather than adding a second path:

- the same pool resolution and `--reviewer` selection
- the same explicit-model invariant and per-runner argv construction
- the same transcript-beside-the-verdict contract, which is what makes *how* a reviewer concluded auditable
- a normalised extraction that handles both stream schemas, so an empty result is an **error**, never an empty review

## Why it is worth building

[...]

PR #1139, opened 2026-08-21
docs(lessons): 215 and 216 from the orchestration architecture session

mlorentedev:
Two lessons, both from mistakes made and caught during the 2026-08-21 orchestration architecture session.

**215 — a parser written for one runner reads the other runner's review as empty.** Two reviewers were run on the same document via different runners. The extraction written for `pi`'s stream (`type` / `message_update`) applied to `agy`'s (`event` / `step_update`) returned `chars=0`, and agy's complete review — 4,551 characters, seven findings, one Blocker — was nearly recorded as "produced nothing". Caught only by dumping the raw stream.

This is lesson 213 one layer down: there the reviewer reported success while publishing nothing; here it published everything and the *reading* produced nothing, which is indistinguishable at the point of use. The root cause is not the parser — it is that `review_launch.go` already encodes the per-runner differences and was bypassed, because `dotf spec review` only accepts a feature-id. Gap filed as #1138.

**216 — a three-dot diff does not answer "what is main missing".** `git diff A...B` compares merge-base(A,B) → B, so after a squash merge it still reports the branch's full contribution whether or not that contribution already landed in A. A fully-merged worktree reported 39 files / +266−439 and was reported to the user as work at risk. The two-dot form (`git diff origin/main HEAD`) was empty.

Lesson 210 arriving through a second door — and the mirror of the prohibited-pattern table's silent-empty warning: **a non-empty result reads as a finding just as readily as an empty one reads as clean.**

Docs-only. Both `_index.md` entries added.


mlorentedev:
## Review triage

|---|---|---|
| `github-actions` — #1138 "Partially compliant": no `dotf spec review --doc`, no pool reuse, no per-runner argv, no transcript contract, no cross-schema extraction | skip | Correct on the facts and wrong on the premise. This PR does not claim to close #1138 — lesson 215 **references** it as the filed gap that caused the failure the lesson records. A lessons PR is not the launcher's implementation, and the five "non-compliant" lines are a faithful description of work that is deliberately not here. |
| `github-actions` — no security concerns, no relevant tests, effort 1/5 | skip | Observations, not findings. Two markdown files and an index row. |

**Disclosure.** `coderabbitai` rate-limited, so PR-Agent was the only reviewer that read this PR — this repository's ordinary state on the free tier.

**One note the reviewer is right about and worth keeping:** *"it must be reviewed when the launcher lands."* Recorded on #1138 rather than here, since that is where the launcher will be built.

Reviewer output dispositioned: `github-actions` 2026-08-21.

PR #1142, opened 2026-08-21
test(harness): add schema and drift guards for triggers.json registry

### LINKED ISSUE #1137: GUARD: triggers.json has a Go loader and zero guards — the only registry with neither a test nor a CI reference
Found during the ADR-035 registry audit (2026-08-21). Constraint C10 says every mechanism ships with its guard in the same PR; this one has none.

## Measured

| Registry | Go loader | bats guard | CI/script reference |
|---|---|---|---|
| `harness/reviewer-pool.json` | yes | 2 files | `scripts/check-review-attestation.sh` |
| `harness/review-attestation.json` | yes | 2 files | workflow + script |
| `harness/manifest.json` | 3 loaders | 3 files | workflow + `compile-harness.sh` |
| **`harness/triggers.json`** | `cli/internal/harness/triggers.go` | **none** | **none** |

Reproduce:

```bash
for f in reviewer-pool review-attestation triggers manifest; do
  echo "$f:"
  grep -rl "$f.json" tests/ .github/ scripts/ 2>/dev/null
done
```

`triggers.json` is 9,354 bytes of pattern-trigger declarations consumed by `dotf harness`. Nothing asserts it parses, nothing asserts its shape, nothing fails if it is malformed or absent.

## Why it matters beyond tidiness

It is the same failure class this repo keeps hitting and named as constraint C15 on the same day: a registry that cannot be read must fail loudly, never as empty or as a permissive default. A malformed `triggers.json` today most likely yields *no triggers*, which is indistinguishable from *no triggers matched* — a silent absence of behaviour rather than an error.

## Scope

A bats guard asserting it parses and conforms to the shape the loader expects, at minimum. `triggers_test.go` exists and covers the loader, not the file — a fixture would pass while the real registry is broken.

Not in scope here: a JSON Schema or a `dotf doctor` check. ADR-035 gives `model-map.json` both and records explicitly that this sets a precedent the other four registries do not meet. Whether they follow is a separate decision.

mlorentedev:
## Summary

Adds schema validation and drift tests for `harness/triggers.json` (which was the only registry in `harness/` without test coverage):

1. **Bats Guard (`tests/triggers-registry.bats`)**:
   - Asserts `harness/triggers.json` exists and is valid JSON.
   - Asserts `version` is a positive integer.
   - Asserts `.triggers` array is non-empty.
   - Asserts every trigger has non-empty `id`, `pattern`, and `globs` array.
   - Asserts all trigger `id` values are unique (no duplicates).
   - Asserts embedded Go copy (`cli/internal/harness/triggers.json`) matches root `harness/triggers.json` byte-for-byte.

2. **Go Unit Tests (`cli/internal/harness/triggers_test.go`)**:
   - `TestRealTriggersFileValid`: Verifies the embedded config unmarshals to a valid `TriggerConfig`.
   - `TestTriggersEmbeddedMatchesDiskSSOT`: Verifies embedded copy has not drifted from disk SSOT.

## Verification

- Verified all 6 bats tests pass (`bats tests/triggers-registry.bats`).
- Verified all 18 CLI test packages pass (`go test ./...`).
- Full test suite passes (`./scripts/test.sh`).

Closes #1137

### DIFF
--- a/cli/internal/harness/triggers_test.go
+++ b/cli/internal/harness/triggers_test.go
@@ -340,3 +340,49 @@ name: skill2
 		t.Fatalf("deps[skill1] = %v; want %v", deps["skill1"], expected)
 	}
 }
+
+// TestRealTriggersFileValid guards harness/triggers.json against silent rot or schema deviation (GUARD #1137).
+func TestRealTriggersFileValid(t *testing.T) {
+	cfg, err := ParseTriggers(defaultTriggersJSON)
+	if err != nil {
+		t.Fatalf("ParseTriggers(defaultTriggersJSON) failed: %v", err)
+	}
+	if cfg.Version < 1 {
+		t.Errorf("expected Version >= 1, got %d", cfg.Version)
+	}
+	if len(cfg.Triggers) == 0 {
+		t.Fatal("expected non-empty triggers array")
+	}
+
+	seenIDs := make(map[string]bool)
+	for i, rule := range cfg.Triggers {
+		if rule.ID == "" {
+			t.Errorf("trigger[%d] has empty ID", i)
+		}
+		if seenIDs[rule.ID] {
+			t.Errorf("duplicate trigger ID %q", rule.ID)
+		}
+		seenIDs[rule.ID] = true
+
+		if rule.Pattern == "" {
+			t.Errorf("trigger %q has empty Pattern", rule.ID)
+		}
+		if len(rule.Globs) == 0 && len(rule.Keywords) == 0 {
+			t.Errorf("trigger %q has neither globs nor keywords", rule.ID)
+		}
+	}
+}
+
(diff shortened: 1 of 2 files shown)


mlorentedev:
## Review triage

|---|---|---|
| Ticket compliance analysis (#1137: Fully compliant) | skip | PR fulfills all requirements with Bats + Go schema & drift tests. |
| Effort 1/5, relevant tests added, no security/major issues | skip | Clean review observations. |

mlorentedev:
## Review triage

|---|---|---|
| Schema mismatch with Go loader (require globs or keywords) | apply | Applied in commit `040a523`: updated Bats test select clause to accept `globs` or `keywords` matching Go loader logic. |
| Ticket compliance analysis (#1137: Fully compliant) | skip | PR fulfills all requirements. |
| Effort 1/5, PR contains tests, no security concerns | skip | Clean review observations. |

PR #1143, opened 2026-08-21
feat(harness): build model-map.json, its schema, and the first doctor check over a registry

### LINKED ISSUE #1124: HARNESS: build model-map.json — the declarative routing layer ADR-032 specified and nobody built

mlorentedev:
Implements **HARNESS-075**, gated on #1124, to the shape **ADR-035** decided (merged as #1136). The policy layer ADR-032 specified on 2026-08-09 and nobody built now exists and is consumed.

## What lands

- **`harness/model-map.json`** — pools, harnesses, tiers, chains, services, with the measured rationale in `$comment`.
- **`harness/model-map.schema.json`** — the repo's **first JSON Schema over a registry**.
- **A native validator** (`cli/internal/harness/model_map.go`) that reads the schema *as data* and enforces what it finds there. **Zero new dependencies** — `cli` still has three direct ones.
- **A `dotf doctor` check** — the repo's **first over a registry**, and the mechanism by which constraint C15 actually holds.
- **8 bats cases + 7 Go tests**, shipping in this PR per C10.

## Three design calls worth reviewing

**1. No `//go:embed`, deliberately diverging from `triggers.json` next door.** That file embeds a build-time copy of itself and falls back to it when the repo file is missing. For a routing map that is disqualifying: an absent file would resolve to a default and this PR's own doctor check would certify it healthy — exactly what C15 forbids. One bats case asserts that absence so it cannot grow later. (The drift the embedded triggers copy already carries is filed on #1137.)

**2. Native validation, with a keyword allow-list that fails loud.** The obvious objection to hand-rolling is *two places where the rules live*; it is avoided by having the validator read `model-map.schema.json` rather than restate it. What makes that honest is the allow-list: a validator that skipped a keyword it does not implement would report a document valid without checking what the schema asked for. Unimplemented constructs are a loud error, with their own test, and the whole schema is walked up front so a keyword under an unreached branch still fails.

**3. `codex` is deleted, not declared.** ADR-032 §3 referenced a `codex` pool that `pools` never declares. An earlier commit here resolved that by declaring it, reasoning from a `~/.codex` directory; the owner confirmed codex is retired, so the resolution is deletion — a pool nobody dispatches to is a route to nowhere whether or not it validates. It joins `openrouter` (deleted upstream), and the assertion generalises over both. **Structural, not textual**: the `$comment` names both on purpose, so a grep-based guard would fail on that very sentence.

## Budget: level 1 only, and it says so

ADR-035 splits enforcement. **Level 2 needs a dispatcher to decrement a semaphore, and `dotf agent` is still an unknown command** — so declaring a budget nothing decrements is what ADR-035 itself calls advisory. This ships declaration, and the honesty is load-bearing in the output:

```
[Routing registry]
  [INFO] declared concurrency: claude 10, nan 5 (reserve 2) — DECLARED, not enforced:
         nothing decrements these today (ADR-035 level 1)
  [INFO] pool "nan" is shared with pi-tui, qq, qf, hive-embeddings, pr-agent-ci, none of
         which routes through dotf — so the eventual guarantee is "dotf alone will never
         be the cause of exhaustion", never "exhaustion will not happen"
```

`Budget.ConcurrencyDeclared` separates *not declared* from *zero*, because zero reads as "no capacity".

## C15 has teeth: three broken states, told apart, verified through the built binary

| state | output |
|---|---|
| absent | `not found at … — this is not an empty routing map, and nothing falls back to a default` |
| unparseable | `could not be parsed as JSON: unexpected end of JSON input` |
| schema-invalid | `harnesses.pi.pools[] names "ghost" … routes nowhere at dispatch time` |

A fourth is covered that C15 did not name: **map present, schema missing** — a map that validates against nothing is not validated. The test bans the phrases `no pools` and `0 pools` from every broken state and asserts the messages differ from one another.

## Three defects found in this spec's own verification, recorded rather than quietly fixed

They are the exact class the spec exists to prevent, committed inside the spec. Full detail in `verification.md`.

1. **`f1` could never pass.** `$comment` inside a double-quoted `python3 -c` — the shell blanked it before python saw it, so the required-key set became `{'', 'version', …}` and it exited 1 on a *correct* map, permanently.
2. **`f4` would have failed a correct map.** It forbade the string `openrouter` anywhere in the file, while `tasks.md` asks the `$comment` to carry exactly that rationale.
3. **Five commands could pass vacuously.** `go test -run <name>` exits 0 when the name matches nothing (`[no tests to run]`), so every Go-backed criterion would have reported PASS once its test was renamed. Fixed by requiring the `--- PASS` line; verified across all three states and mutation-verified on the real command. Lesson filed as #1141.

## Evidence

```
go build ./...            OK
go vet ./...              OK
GOOS=windows go vet ./... OK
go test ./...             17 packages ok, 0 FAIL
golangci-lint run         0 issues   (v2.12.2, matching the versions.conf pin)
bats tests/*.bats         1..1394, 1394 ok, 0 not ok, BATS_EXIT=0
```

The bats exit code was captured directly, not through a pipeline — the first attempt piped through `tail` and reported `0` while proving nothing. That is recorded in `verification.md` too.

## Not done, stated rather than implied

Level 2 enforcement; `capability-map.json` (re-scoped to #560); any dispatcher or `dotf agent` noun; retrofitting schema/doctor checks onto the other four registries (#1137 covers `triggers.json`'s missing guard).

**The independent adversarial review the archive gate requires has not run** — the reviewer must not be the implementer.

Closes #1124


mlorentedev:
## Review triage

|---|---|---|
| Ticket compliance analysis (#1124, #1136, #1137) | skip | Out-of-scope ticket items are handled in their respective dedicated tickets (#1118 pagination, #1137 triggers.json guard via PR #1142, ADR-035 via PR #1136). |
| Stale line in $comment (broken sentence fragment) | accept | Valid minor comment polish in `harness/model-map.json`. |

Issue #985, opened 2026-08-15
BUG-080: dotf secrets run without --only is dead for every consumer — DOCKERHUB_TOKEN maps to an item that does not exist


mlorentedev:
## Symptom

Any `dotf secrets run -- <cmd>` **without** `--only` fails outright:

```
$ dotf secrets run -- true
Error: bw resolve dockerhub/password: bw item not found: bw serve item "dockerhub": not found
```

Deterministic — 3/3 attempts, not the intermittent `bw serve` flakiness tracked separately.

## Blast radius

This is not cosmetic. `dotf secrets run --` with no scope is the injection path used by:

- the `pi` shell function (`pi () { dotf secrets run -- pi "$@" }`) — so **`pi` does not start at all**
- `dotf spec review` (`cli/internal/cmd/review.go` builds `'dotf' 'secrets' 'run' '--' 'pi' ...`), so the **adversarial review gate is unrunnable for every spec in every repo**

The second one is how this was found: a kubelab spec could not be archived because its mandatory review could not launch.

## Cause

`secrets/registry.yaml` maps both DockerHub entries to a vault item named `dockerhub`:

```yaml
  - id: DOCKERHUB_TOKEN
    bw: { item: dockerhub, field: password, folder: apps }
  - id: DOCKERHUB_USERNAME
    bw: { item: dockerhub, field: username, folder: apps }
```

The vault has no `dockerhub`. It has **`DockerHub`** — unfoldered, carrying `username`, `password` **and a separate `PAT` field**. Lookup is by exact name, so the registry misses it.

## Second defect, worth fixing in the same pass

Even once the name resolves, `DOCKERHUB_TOKEN` is mapped to `field: password` — the **account password**, not the `PAT` sitting beside it. A CI image-push consumer (`consumers: ["ci:image-push"]`) would receive the account credential instead of a revocable scoped token.

## Suggested fix

1. Rename the vault item `DockerHub` → `dockerhub`, move it to `apps`.
2. Repoint `DOCKERHUB_TOKEN` at `field: PAT`.
3. Consider whether `dotf secrets run` should degrade for an unresolvable secret the caller did not ask for — today one broken registry row disables the whole unscoped injection path for every consumer. `secrets render` already tolerates absent secrets; `run` does not.

Point 3 is the actual robustness bug: without it, the next registry/vault drift silently takes `pi` and the review gate down again.

## Evidence

[...]

PR #1144, opened 2026-08-21
fix(doctor): report stale or unsynced bw cache as WARN in mapping check

### LINKED ISSUE #1015: BUG-087: doctor reports an item as absent from the vault on the word of a cache that may be stale
## What

`checkBWMapping` (`cli/internal/doctor/checks_bw_mapping.go:76`) emits a FAIL asserting an item does not exist:

```go
rep.Fail(fmt.Sprintf(
    "%s: no such item in the vault, named by %s — every `dotf secrets run` without --only fails on it, including `dotf spec review`",
    item, strings.Join(ids, ", ")))
```

The set it compares against comes from `sys.BWItemNames()`, which reads through the `bw serve` daemon. **A daemon read is not authoritative about server state.**

## Why it is unsound

Established by the live canary in the `cli/internal/secrets/` lane on 2026-08-15 (#993 / #1007 work): `bw serve` answers reads from **its own cache**. A write through the daemon updates the server and the cache still serves the previous value until an explicit `POST /sync`; the session that found this wrote a secret and read back the old one from a different process.

The same staleness has already been observed from the other direction in this repo. A vault item renamed `DockerHub` → `dockerhub` was invisible to `dotf secrets` until an explicit sync — which is the exact input this check consumes. So:

- An item created or renamed since the daemon last synced reads as **absent**.
- This check turns that into a **FAIL** stating "no such item in the vault", and names a consequence ("every `dotf secrets run` without `--only` fails on it") that may be entirely false.
- `dotf doctor` exits 1 on a healthy machine, and the operator is sent to fix a mapping that is already correct.

## The class

This is the third instance of one pattern found in `cli/internal/doctor/` today, and the pattern is worth naming: **a check stating as fact something it did not establish.**

1. #997 — an absent DR escrow reported as SKIP ("nothing to check here") without consulting exposure. Fixed in #1006.
2. The review finding on #1006 — any `os.Stat` error routed into the absent branch, so an escrow behind a permission error was reported as "no DR escrow, 28 secrets have no local copy". Fixed in the same PR.
3. This one — a possibly-stale cache read reported as "no such item in the vault".

In all three the honest statement is *"I could not establish this"*, and in all three the code said *"this is not there"*. The fix for #997 landed the vocabulary for it: a WARN that names what could not be determined.

## Proposed direction

Not obvious enough to inline, which is why this is a ticket rather than a patch:

1. **Sync before listing** — `POST /sync` ahead of `BWItemNames()`, so the comparison runs against fresh state. Correct, but gives a read-only diagnostic a side effect and a network round-trip; `checkBitwardenReach` already documents that trade-off for its tier-3 probe and accepts it only on an unlocked vault.
2. **Report staleness instead of absence** — keep the read cheap, and where an item appears missing, state that it was not found *in the daemon's view* and name the sync that would settle it. Weaker signal, no side effect, and honest.
3. **Both, keyed to freshness** — if the daemon's last sync is recent, FAIL as today; if stale or unknown, WARN naming the staleness. Mirrors the severity-keyed-to-exposure rule this package already applies twice (`checkBitwardenReach`, and `checkDisasterRecovery` after #1006).

Option 3 is the one consistent with the package's existing precedent, at the cost of needing the daemon's `lastSync`, which `checkBWSyncAge` already parses.

## Acceptance criteria

- [ ] A stale or unsynced daemon cannot produce a FAIL asserting an item does not exist.
- [ ] Where absence cannot be established, the report names what was not determined and the action that would settle it, rather than asserting absence.
- [ ] A genuinely missing item still FAILs — the check must not be defanged into uselessness; #985 is the case it exists to catch.
- [ ] Test coverage for the stale-cache branch specifically, not only the present/absent pair.

## Related

- #993, #1007 — the daemon-cache finding that makes this reachable
- #1006 — the same class fixed in `checks_dr.go`, including the vocabulary for "could not establish"
- #985 — the real missing-item incident this check was built for; the fix must not lose it
- #988 — daemon read reliability, same trust boundary

mlorentedev:
## Summary

Fixes BUG-087 by keying `checkBWMapping` severity to Bitwarden sync freshness:

- If the daemon/vault was synced recently (`<= 24h`), a missing item FAILs asserting genuine absence in the vault (#985).
- If the daemon/vault was never synced or the sync is stale (`> 24h`), doctor emits an honest **WARN** explaining that the item was not found in the local cache, noting the sync age, and advising `dotf secrets sync` to refresh.
- Adds `BWLastSync` to `System` and `realSystem` parsing `lastSync` from `bw status`.

## SDD skip rationale

Targeted bug fix (BUG-087) in existing `checkBWMapping` diagnostic check, adding sync freshness awareness without altering public CLI flags or data contracts.

## Verification

- Added unit tests in `cli/internal/doctor/checks_bw_mapping_test.go` covering stale cache, never-synced cache, fresh cache, and present items.
- All Go tests pass (`go test ./...`).

Closes #1015


mlorentedev:
## Review triage

|---|---|---|
| Ticket compliance analysis (#1015: Fully compliant) | skip | PR fulfills all requirements of #1015 (stale/unsynced cache reports WARN with sync age and remediation). |
| Ticket compliance note (#985: Not compliant) | skip | #985 is a separate issue for secrets mapping degradation. |
| Harness compliance, effort 2/5, no security/major issues | skip | Clean review observations. |

PR #1145, opened 2026-08-21
chore(spec): archive POLISH-003-pwsh-analyzer-coverage spec

mlorentedev:
## Summary

Archives `specs/POLISH-003-pwsh-analyzer-coverage` to `specs/archive/POLISH-003-pwsh-analyzer-coverage`.

The dynamic PSScriptAnalyzer discovery was implemented and verified green in PR #1100.

Closes #1090

Issue #1107, opened 2026-08-20
TOOL-013: the pr-agent review job reports success while publishing no review


mlorentedev:
## Measured on #1101, 2026-08-20

The `review` job ran for 4m37s and **passed**. The pull request carries **zero** comments containing `## PR Reviewer Guide` — the marker `harness/review-attestation.json` declares as PR-Agent's review output. The only thing it published was:

```
2026-08-20T02:34:09Z  github-actions[bot]  ::  Failed to generate code suggestions for PR
```

That is the `/improve` command failing, not a review. So the job is green and nothing was reviewed — the "a check that passes while doing nothing" class this repository has spent a week eliminating everywhere else, now sitting in the fallback reviewer itself.

Repeated across runs on the same PR, so it is not a one-off.

## Why this one matters more than most

PR-Agent exists (TOOL-013, #786) *because* CodeRabbit's free tier is account-wide rate-limited. On #1101 both failed at once: CodeRabbit posted its quota notice and PR-Agent published nothing. `review-attestation` correctly reports `declined`, and the PR cannot go green.

**The gate is not the problem here, and should not be softened.** It already treats CodeRabbit as optional rather than required: `check-review-attestation.sh` runs the comment-shaped attestation *before* the declined check, deliberately — its own comment says *"Runs BEFORE the declined check on purpose. Both can be true at once."* A genuine PR-Agent review outranks a CodeRabbit quota notice today. What is missing is the genuine review.

## Upstream error in the same run

```
"exception": {"type": "AttributeError", "value": "'NoneType' object has no attribute 'get'"}
"message": "Failed to fetch sub-issues. Error: 'NoneType' object has no attribute 'get'"
```

Raised twice while extracting tickets from the PR description. Whether that is what suppresses the review or an unrelated non-fatal path is not yet established — the diff itself was read fine (`Tokens: 20625, total tokens under limit: 200000, returning full diff`), so the model had its input.

## Worth checking first

- `require_ticket_analysis_review` is enabled in `.pr_agent.toml`, and the failing call is the ticket/sub-issue fetch. Disabling it is a cheap experiment that would say whether the review path is being taken down by the ticket path.
- The job's exit status should not be the only signal. A run that publishes no review while reporting success is exactly what GUARD-002 exists to make visible for *reviewers*; the same reasoning applies to the reviewer's own workflow.

## Acceptance

- Either PR-Agent publishes a review with its declared marker, or its job fails loudly when it does not.
[...]

PR #1146, opened 2026-08-21
fix(ci): remove job-level concurrency lock in pr-agent workflow

mlorentedev:
## Summary

Removes the job-level `concurrency: group: pr-agent-nan-inference` lock from `.github/workflows/pr-agent.yml`.

The workflow-level concurrency group (`group: pr-agent-${{ pr.number }}-${{ event_name }}`) already collapses burst pushes per PR. Removing the global lock allows parallel PRs to be reviewed concurrently across NaN's 10 available slots (5 deepseek + 5 mimo fallback) without GitHub Actions cancelling intermediate runs.

## Summary by CodeRabbit

- **Chores**
  - Updated automated pull request processing to allow independent requests to proceed without being held in a shared queue.
  - This may improve the responsiveness and throughput of automated review assistance.

### DIFF
--- a/.github/workflows/pr-agent.yml
+++ b/.github/workflows/pr-agent.yml
@@ -108,25 +108,6 @@ jobs:
         && startsWith(github.event.comment.body, '/'))
     runs-on: ubuntu-latest
     timeout-minutes: 15
-    # Serialise inference repo-wide: QUEUE, never cancel (#1107).
-    #
-    # This is a SECOND group, at job level, and it does not replace the
-    # workflow-level one above — the two answer different questions. The
-    # workflow group is per-PR and cancels, so a burst of pushes still collapses
-    # to one review of the settled state (#1042). This one is repo-wide and does
-    # not cancel, so two PRs never ask NaN for inference at the same time.
-    #
-    # ADR-032 names the rule: the top tier "queues or escalates, NEVER degrades
-    # silently". Queuing is the sanctioned behaviour, and it is what was missing:
-    # the cluster's 5 slots were being spent by our own parallel CI, so a review
-    # failed rather than waited.
-    #
-    # It cannot guarantee a free slot — pi, qq and hive draw on the same pool —
-    # but it caps this workflow's demand at one, leaving the rest for interactive
-    # use instead of racing it.
-    concurrency:
-      group: pr-agent-nan-inference
-      cancel-in-progress: false
     steps:
       # `The-PR-Agent/pr-agent`, not `qodo-ai/pr-agent` as #786's body says: the
       # project has been renamed and the old path now redirects. Pinned to a tag


mlorentedev:
## Review triage

|---|---|---|
| Concurrency regression note (GHA queue vs LiteLLM fallback) | skip | GHA `cancel-in-progress: false` has a hard ceiling of 1 pending run, causing GHA to cancel intermediate runs with 'higher priority request exists' in bursts >= 3. LiteLLM provides automatic fallback from `deepseek-v4-flash` (5 slots) to `mimo-v2.5` (5 slots) with exponential backoff retries. |
| Effort 1/5, no security concerns | skip | Review observations. |

mlorentedev:
Cross-session data point from an incident that happened **after** this PR was opened, so it could not have been weighed here. Not a request to close it — a measurement its premise needs.

## What happened, 2026-08-21 ~02:20Z

An adversarial spec review launched via `dotf spec review` on the pool primary died like this:

```
429: {"message":"deepseek-v4-flash concurrency limit: max 5 simultaneous requests."}
auto_retry_end  success=false  attempt=3
```

The cause was **PR-Agent reviewing PR #1143**, holding NaN slots. The run fell through to the provider-diverse fallback (`agy/gemini-3.1-pro-high`) and completed there, which is the pool behaving as designed.

## Why it bears on this diff

That happened **with** the job-level lock in place. One PR-Agent job was enough to exhaust deepseek against a concurrent consumer. This PR removes the cap, so N jobs can hold slots simultaneously against the same five.

The five are not PR-Agent's alone. `harness/model-map.json` (landing in #1143) declares the sharing explicitly, from measurement:

```json
"nan": { "concurrency": 5, "reserve_interactive": 2,
         "shared_with": ["pi-tui", "qq", "qf", "hive-embeddings", "pr-agent-ci"] }
```

and NaN's limit is **per model, not per API key** — measured 2026-08-20: eight concurrent requests to deepseek returned 5×200 and 3×429 while mimo answered 200 throughout. So parallel PR-Agent jobs compete with `dotf spec review`, the `pi` TUI, `qq`/`qf`, and hive embeddings for the same five.

## What this comment is not

It is not a claim the lock is costless. If it is producing queues long enough to hurt throughput, that is a real problem and removing it is one honest answer — the comment being deleted concedes the cap is a heuristic, not a guarantee.

But the tradeoff is *queueing versus failing*, and ADR-032 names which side the repo chose: **"queues or escalates, never degrades silently."** A 429 on a review is not a delay; it is a review that did not happen, and on 2026-08-20 six PRs reported green while receiving no review at all for exactly this reason (#1107).

**If the motivation is queue latency, the cheaper fix may be raising the cap rather than removing it** — a group with a bounded parallelism instead of one, or routing PR-Agent to the NaN sibling (`mimo-v2.5`) so it stops competing with the model spec reviews use. Both keep the "queue rather than fail" property this diff drops entirely.

Worth a line in the PR body either way, so whoever reads it later knows the removal was a decision taken against this data rather than without it.

mlorentedev:
**Correcting my earlier comment.** I argued to hold this PR based on a 429 I hit while the lock was in place. I had half the data. The other half says you are right about a real problem, and I can now measure it.

## The lock does not queue. It cancels.

Two PRs open right now received **no review at all**, both with the same annotation:

```
#1144  X review in 22s     Canceling since a higher priority waiting request for pr-agent-nan-inference exists
#1145  X review in 2m37s   Canceling since a higher priority waiting request for pr-agent-nan-inference exists
```

The workflow comment says `cancel-in-progress: false` and reasons about queueing, but GitHub keeps only **one** pending run per concurrency group — a newer pending run cancels the older one that was waiting. So the mechanism written to make reviews *wait* is making them *not happen*, which is the exact failure #1107 was filed for.

## Both failures are real, and they are different

| | Mechanism | Observed |
|---|---|---|
| **Removing the lock** | N PR-Agent jobs compete for NaN's 5 per-model slots | `429: deepseek-v4-flash concurrency limit` killed a `dotf spec review` at 02:20Z — with the lock still in place, so one job already sufficed |
| **Keeping the lock** | Only one pending run survives per group | #1144 and #1145 cancelled, zero reviews produced |

**The lock is failing in both directions**: it does not deliver the queueing it promises, and it does not prevent exhaustion either. So removing it is defensible — it deletes a mechanism that is not doing its job. My hold recommendation was wrong on that point and I withdraw it.

## What removing it alone leaves open

Exhaustion comes back, unmitigated. `harness/model-map.json` (landing in #1143) records the sharing from measurement:

```json
"nan": { "concurrency": 5, "reserve_interactive": 2,
         "shared_with": ["pi-tui", "qq", "qf", "hive-embeddings", "pr-agent-ci"] }
```

## The cheap fix that solves both, and is not this diff

**Route PR-Agent to `mimo-v2.5` instead of `deepseek-v4-flash`.** NaN's concurrency limit is **per model, not per API key** — measured 2026-08-20: eight concurrent requests to deepseek returned 5×200 and 3×429 **while mimo answered 200 throughout that saturation**. Two models are two independent lanes of five.

That gives PR-Agent its own lane, stops it competing with `dotf spec review` (which uses deepseek as the pool primary), and needs no concurrency group at all — so this PR's removal becomes correct rather than merely defensible.

`mimo-v2.5` is already in `harness/reviewer-pool.json` as an admitted fallback, evaluated on 2026-08-20 with the verdict recorded honestly.

**Suggestion, not a blocker:** merge this as-is if the cancellations are hurting now, and file the model-routing change as a follow-up so the exhaustion side does not stay silently open. Both halves are needed; neither alone is the whole fix.

PR #1151, opened 2026-08-21
test(harness): guard triggers.json, the only registry with no test and no CI reference

### LINKED ISSUE #1137: GUARD: triggers.json has a Go loader and zero guards — the only registry with neither a test nor a CI reference

mlorentedev:
Closes #1137. Found during the ADR-035 registry audit: of this repo's four declarative registries, `harness/triggers.json` was the only one with a Go loader and **zero guards** — no bats, no CI reference — plus a byte-identical duplicate at `cli/internal/harness/triggers.json` that nothing kept in sync.

## The ticket proposed deletion. That premise was wrong.

Both measurements are recorded in the test file so nobody re-derives them.

**1. The embedded copy is load-bearing.** `cmd/harness.go` calls `LoadTriggers("")` — empty `repoRoot`, so resolution is walk-up-from-cwd, then the embedded copy. In any repo that is not this checkout the walk-up finds nothing, so **the embedded copy is the only source there**. Verified from `/tmp`, no checkout present:

```
$ cd /tmp && dotf harness suggest .github/workflows/ci.yml
Patterns:
  - pattern-git-workflow
Skills:
  - pr-review-triage
  - using-git-worktrees
```

That is **C12 working as designed** — user-level deploy surfaces, every repo inherits with zero per-repo setup. Deleting the embed would break trigger suggestions everywhere except this checkout.

**2. The copies cannot be collapsed into one.** The Go module root is `cli/`, and `//go:embed ../../harness/triggers.json` is rejected at compile time:

```
x.go:5:12: pattern ../../harness/triggers.json: invalid pattern syntax
```

`embed` cannot reference paths outside its package directory. So two copies are **structurally required**, not an oversight.

## What was actually missing

The assertion that they agree. Identical today only meant nobody had edited one alone yet — and the moment they diverge, `dotf harness suggest` answers from a registry this repo does not declare, in every repo except this one, **silently**, because both files parse and both look right.

Five cases: both copies parse, they are byte-identical, every rule carries the `id` and `globs` the loader reads, and the `version` field the loader unmarshals is present.

Mutation-verified — bumping the embedded copy's `version` fails case 3:

```
not ok 3 triggers: the embedded copy is byte-identical to the repo copy
```

## Deliberately excluded

A JSON Schema and a `dotf doctor` check. **ADR-035 gives those to `harness/model-map.json` and records explicitly that it sets a precedent the other four registries do not meet, and that whether they follow is a separate decision.** This closes the guard gap; it does not take that decision.

Tests only — no production diff.

### DIFF
--- a/tests/triggers-registry.bats
+++ b/tests/triggers-registry.bats
@@ -1,53 +1,86 @@
 #!/usr/bin/env bats
-# GUARD (#1137): triggers.json schema and drift guard.
+# HARNESS: harness/triggers.json had a Go loader and zero guards (#1137) — the
+# only registry in this repo with neither a test nor a CI reference.
 #
-# Asserts harness/triggers.json parses, conforms to the TriggerConfig shape,
-# has unique IDs, and does not drift from cli/internal/harness/triggers.json.
+# Why the duplication it guards cannot simply be deleted, measured 2026-08-21:
+#
+#   1. `cmd/harness.go` calls `LoadTriggers("")` — empty repoRoot, so the
+#      resolution order is walk-up-from-cwd, then the embedded copy. In any repo
+#      that is not this checkout the walk-up finds nothing, so the EMBEDDED copy
+#      is the only source. Verified from /tmp with no checkout present:
+#      `dotf harness suggest .github/workflows/ci.yml` returned real triggers.
+#      That is C12 working as designed — user-level deploy, every repo inherits.
+#
+#   2. The copies cannot be collapsed into one. The Go module root is `cli/`, and
+#      `//go:embed ../../harness/triggers.json` is rejected at compile time with
+#      "invalid pattern syntax" — embed cannot reference paths outside its
+#      package directory.
+#
+# So two copies are structurally required, and what was missing is the assertion
+# that they agree. Identical today only means nobody has edited one alone yet;
+# the moment they diverge, the binary ships a registry the repo does not declare
+# and nothing says so.
+#
+# Deliberately NOT here: a JSON Schema or a `dotf doctor` check. ADR-035 gives
+# harness/model-map.json both and records that this sets a precedent the other
+# four registries do not meet, and that whether they follow is a separate
+# decision. This closes the guard gap, not that decision.
 
 setup() {
     REPO="$BATS_TEST_DIRNAME/.."
-    TRIGGERS="$REPO/harness/triggers.json"
-    EMBEDDED="$REPO/cli/internal/harness/triggers.json"
+    REPO_COPY="$REPO/harness/triggers.json"
+    EMBED_COPY="$REPO/cli/internal/harness/triggers.json"
 }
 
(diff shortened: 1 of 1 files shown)


mlorentedev:
## Review triage

|---|---|---|
| `github-actions` — #1137 "Partially compliant", zero non-compliant, zero needing human verification | skip | Nothing to act on. It lists every acceptance item as met and names no gap; the "Partially" label with an empty non-compliant section is the reviewer's own template, not a finding. |
| `github-actions` — PR contains tests, no security concerns, no major issues, effort 2/5 | skip | Observations. Expected for a tests-only diff with no production change. |

**Two disclosures rather than findings:**

1. **This is the first review to run on the new model lane.** #1150 merged at 03:25Z and moved PR-Agent from `deepseek-v4-flash` to `mimo-v2.5`; this review ran after it. Worth recording because #1150's PR body promised the trade openly — CI review moves to the model with the weaker established record — and this is the first data point on that. **It is one review of a tests-only diff, which is close to no evidence either way.** Not a verdict on the swap.
2. **`coderabbitai` did not review.** Rate-limited, as usual on the free tier, so PR-Agent was the only reviewer.

Nothing here changes the diff.

Reviewer output dispositioned: `github-actions` 2026-08-21, `coderabbitai` (declined notice, not a review).

Issue #1154, opened 2026-08-21
DEBT: report.go is not gofmt-clean on main and the pinned golangci-lint reports 0 issues


mlorentedev:
## What happens

```
$ cd cli && gofmt -l ./internal/
internal/doctor/report.go
```

The file is unformatted on `origin/main` — not introduced by any in-flight branch:

```
$ git show origin/main:cli/internal/doctor/report.go > /tmp/report_main.go && gofmt -l /tmp/report_main.go
/tmp/report_main.go
```

And the pinned linter passes:

```
$ golangci-lint --version
golangci-lint has version 2.12.2 ...     # matches versions.conf GOLANGCI_LINT_VERSION

$ golangci-lint run
0 issues.
```

## Why it is worth a ticket rather than a drive-by `gofmt -w`

The unformatted file is the symptom; **the linter not reporting it is the finding.** CI runs
`golangci-lint run` on the pinned version and is green, so nothing in the pipeline asserts that
this module's Go source is formatted. Reformatting `report.go` closes today's instance and leaves
the next one to appear the same silent way.

Same class as the guards this repo keeps filing: a check that exists but does not cover.

## Fix shape

Enable the formatter in `.golangci.yml` (v2 moved formatting into a `formatters:` block, so a
config carried over from v1 can silently stop checking it), confirm it fails on the current
`report.go`, then format. Fail-first, or the enablement is unverified.

Priority is low: it is cosmetic today. It is filed because the gap is not.

mlorentedev:
Re-measured 2026-09-23 on `origin/main` `e204c06` with Go 1.26.0: `gofmt -l` in `cli/` lists **4 files**, and none of them is `report.go`, so the class has moved rather than closed:

- `cli/internal/errors/latch_test.go`
- `cli/internal/fsmode/fsmode.go`
- `cli/internal/orca/orca.go`
- `cli/internal/worktree/clean.go`

The example in `internal/orca/orca.go` is a misaligned map value. The pinned golangci-lint still reports 0 issues, so nothing enforces formatting. Found while working on #1625 W1.2, and deliberately not fixed there (out of scope).

Issue #1153, opened 2026-08-21
GUARD: nothing detects that the merged sha is not the reviewed sha, so a spec can archive on a review of code that never landed


mlorentedev:
## What happened, measured 2026-08-21

HARNESS-075 ran **four** rounds of adversarial review. Every finding was closed on the branch.
PR #1143 then merged — carrying the **original, pre-fix implementation**. The branch holding the
fixes was never an ancestor of the merge commit `e22a4d0`.

Round 5 found it by running the round-1..4 regression probes against a worktree at the merged sha.
All of them still reproduced there: ghost pool in `chains` accepted, blank model ids accepted,
`concurrency: -5` accepted, malformed `properties` elements silently skipped, `x-tiersHaveChains`
absent entirely.

So `verification.md` described a component, and `main` carried a different one. Every gate was
green.

## Why nothing caught it

The staleness gate watches `proposal.md`, `tasks.md` and `features.json`. A spec whose **code**
diverges from the reviewed sha passes it, because no artifact it watches changed.

`review.md` carries `reviewed_sha` in its frontmatter. Nothing compares it to anything. It is
written by the reviewer and read by a human, and the archive gate checks only that a PASS exists —
not that the PASS is about the tree being archived.

## The hole in one sentence

**A verdict is about a sha. The archive gate treats it as being about a spec.**

## Fix shape

At `dotf spec archive`, refuse when `review.md`'s `reviewed_sha` is not an ancestor of the current
`origin/main` — or, stricter and probably right, when the spec's own files differ between
`reviewed_sha` and `origin/main`. Both are cheap: `git merge-base --is-ancestor` and a scoped
`git diff --quiet`.

The stricter form also catches the ordinary case this class will produce next: review runs, one
more commit lands, spec archives on a verdict that predates it.

## Note on sequencing

Round 5 recommends re-reviewing at the **merged** sha rather than the branch sha, precisely because
of this. That recommendation is a workaround for the missing guard, and it only works while a human
remembers to follow it.

Found by round 5 of HARNESS-075 (`nan/deepseek-v4-flash`), whose findings table marks this row
UNTESTED. Refs #1124.

mlorentedev:
## Design input from the SDD-042 review (#1633): a policy decision for when the digest set widens

[...]

Issue #1152, opened 2026-08-21
BUG: spec-gate prepush adapter resolves a MERGED PR, blocking every later push on that branch


mlorentedev:
## What happens

`scripts/spec-gate-prepush.sh` resolves the current branch's PR with:

```sh
meta=$(gh pr view --json labels,body,author 2>/dev/null)
```

`gh pr view` does not filter by state. A branch whose PR has already **merged** still resolves to
that PR, so the adapter exports the merged PR's `body` as `SDD_PR_BODY` and hands it to
`check-spec-gate.sh`. If that body carried a closing keyword — which a merged PR normally does —
the archive-on-merge rule fires against a spec that is legitimately still active, and **every
subsequent push on that branch is refused**.

## Reproduction, measured 2026-08-21

Branch `feat/harness-075-model-map`, whose PR #1143 merged at 03:43Z:

```
$ git -C <worktree> rev-parse --abbrev-ref HEAD
feat/harness-075-model-map

$ gh pr view --json number,state,title
{"number":1143,"state":"MERGED","title":"feat(harness): build model-map.json, ..."}

$ git push origin HEAD:fix/harness-075-review-rounds-3-5
SDD spec-gate (Tier 4)...................................................Failed
[FAIL] SDD archive-on-merge violation:
       This PR closes an issue whose spec is still active:
         HARNESS-075-model-map-routing-registry (#1124)
error: failed to push some refs
```

There is no open PR for the ref being pushed. The refusal is produced entirely by a PR that closed
hours earlier.

## Why it matters beyond the annoyance

The escape is `--no-verify`, and it was taken: the branch carrying five rounds of review fixes was
pushed with hooks disabled, without the cause being diagnosed. A gate that misfires trains the
bypass, and the bypass then also skips the secret scan and the bats run that share the hook.

## Fix shape

Resolve only an **open** PR for the ref being pushed:

```sh
meta=$(gh pr list --state open --head "$(git rev-parse --abbrev-ref HEAD)" --json labels,body,author --limit 1 ...)
```

or keep `gh pr view` and drop the context when `.state != "OPEN"`. The adapter's own docstring
already promises it falls through "when ... no PR is open yet for this branch" — the code checks
for *a* PR, not an *open* one, so the defect is the docstring being right and the implementation
not matching it.

## Guard

A bats case in the prepush adapter's suite: given a branch whose only PR is MERGED, the adapter
must invoke the gate with an empty `SDD_PR_BODY`.

PR #1155, opened 2026-08-21
fix(harness): land the model-map review fixes that #1143 merged without

mlorentedev:
## What this is

PR #1143 merged the routing registry's **original** implementation. Four rounds of adversarial
review had run against a branch that never reached `main`, so every finding those rounds recorded
as closed was still open in the shipped code. Round 5 found it and returned **FAIL** on a Blocker:

> The change merged to main is not the change reviewed. [...] Archiving on HEAD's green tests would
> certify a model-map whose merged form carries every defect rounds 1–4 documented as closed.

Reproduced by the reviewer against a worktree at `e22a4d0`: ghost pool in `chains`, blank model
ids, `concurrency: -5` and malformed `properties` elements all validated cleanly there.

This lands those fixes, and closes round 5's own findings on top.

## How it was assembled, and why not a rebase

The branch forked before #1142, #1144, #1145, #1146 and #1150. Both obvious approaches were wrong:

- **`git diff origin/main HEAD`** expresses *"make main look like this"*, not *"add what this has"*.
  Applied wholesale it would have reverted the doctor's stale-cache WARN (#1144), the POLISH-003
  archive (#1145), the PR-Agent model lane (#1150) and the triggers-registry guard (#1142) — 24
  files, silently.
- **Replaying the 17 commits** conflicts on every spec file #1143 already merged, because it was a
  squash merge and no patch-id matches.

So the spec's own files were taken onto current `main` instead. The diff touches exactly the
HARNESS-075 surface plus `docs/lessons/`.

## Round 5 dispositions

| Severity | Finding | Disposition |
|---|---|---|
| **Blocker** | The merged change is not the reviewed change | this PR |
| Major | AC2 still bound the validator to be native with no schema-engine dependency; the schema `description` still promised a loud error on an unimplemented keyword | AC2 amended, `description` rewritten |
| Major | A typo'd custom-rule name silently disables a cross-block rule — round 2 closed the wrong-TYPE variant and left wrong-NAME open | `x-` namespace validated as a closed set + named regression test |
| Minor | `go.mod` untidy — library sat as `// indirect` while imported directly | `go mod tidy` |
| Minor | Dead `implementedKeywords` doc comment | removed |
| Minor | `verification.md` AC2 stale | rewritten, supersession stated not overwritten |
| Minor | `tasks.md` boxes stale | reconciled |
| Minor (speculative) | Pool NAMES unconstrained while the ids they route to are pattern-guarded | closed anyway — `propertyNames` on `pools` |

## The Major worth reading

Draft 2020-12 makes an unknown keyword an **annotation**. That tolerance is exactly what the two
`x-` cross-block rules need in order to coexist with a conforming library — and exactly what makes
`x-poolReferenceResolve` (singular) invisible: the lookup finds nothing, the rule never runs, and a
document carrying a ghost pool validates with a nil error. The keyword allow-list deleted when the
library landed caught this by construction, so it was a regression of that loudness property.

The fix draws the boundary the library cannot: **standard keywords are its business; the `x-`
namespace is ours and is a closed set.** Deliberately *not* the mirror rule inside
`ValidateModelMap` — "every known rule must be declared" would force unit tests isolating one rule
to name both. A named test asserts it over the shipped schema instead.

## Two findings the review did not raise

**Mutation testing caught two vacuous guards that five review rounds passed over.** Both were
drafts of the blank-pool-name test. The first failed for five reasons unrelated to the pool name.
The second wired the blank pool into `chains`, where the item pattern rejected `"  :model"` before
`propertyNames` was reached. Only the third isolates the guard.

**`go test` reported `ok (cached)` for a schema it had just mutated.** The shipped registry is read
through a path computed at run time, so it is not a build input and the cache never sees it change.
That reaches this spec's own contract: five of eight `features.json` commands are
`go test ... -v | grep -q -- '--- PASS: <Name> '`, and a cached run reprints `--- PASS` verbatim.
All five now carry `-count=1`. Recorded as `docs/lessons/lesson-218-*`.

## Verification

Run on the assembled tree against current `main`:

```
go build ./...            OK
go vet ./...              OK
GOOS=windows go vet ./... OK
go test -count=1 ./...    17 packages ok, 0 FAIL
golangci-lint run         0 issues   (v2.12.2, matching the versions.conf pin)
bats tests/*.bats         1400 ok / 0 not ok, BATS_EXIT=0
go test -fuzz             30s, 385,769 executions, no crash
```

Each new guard was mutation-proved: delete the schema rule / disable the namespace check, confirm
the named test FAILS, restore, confirm green. Recorded in `verification.md`.

## Filed, not fixed here

- **#1153** — nothing compares the merged sha to `review.md`'s `reviewed_sha`. This PR's Blocker
  exists because no gate could see it, and round 5's "re-review at the merged sha" is a workaround
  for the missing guard.
- **#1152** — the spec-gate prepush adapter resolves MERGED PRs, so a branch whose PR landed is
  blocked from pushing. This is what the previous session bypassed with `--no-verify`.
- **#1154** — `report.go` is not gofmt-clean on main while the pinned linter reports 0 issues.

## Not done here

The spec stays **active**. Round 5 asks for a re-review at the **merged** sha, which cannot happen
until this lands — so `dotf spec archive` remains blocked, correctly, and this PR does not close
its issue.

Refs #1124

## Summary by CodeRabbit

* **New Features**
  * Model-map validation now supports comprehensive JSON Schema rules.
  * Added validation for blank or malformed model IDs, pool references, tier relationships, custom rules, and resource limits.
  * Improved routing configuration documentation for harnesses, pools, and services.

* **Bug Fixes**
  * Diagnostics now clarify when deployed model-map data is missing or empty and how to resolve it.
  * Prevented invalid configurations from being silently accepted.

* **Documentation**
  * Added guidance on reliable test verification and Go build-cache behavior.
  * Updated model-map specifications, evidence, review notes, and completion checklists.


mlorentedev:
## Review triage

CI green on every check that has settled: `lint`, `lint-powershell`, `test (ubuntu-latest)`,
`test (windows-latest)`, `integration`, `spec-gate`, `goreleaser snapshot`, `changes`,
`attestation`, GitGuardian. `review-attestation` opened red (`pending — no reviewer output`) and
corrected itself to `[OK] attested` once PR-Agent published — the #1052 `workflow_run`
re-evaluation behaving exactly as designed.

### PR-Agent (`github-actions`) — reviewed, no findings to apply

Verdict: **no major issues detected**, no security concerns, effort 4/5, tests present.

|---|---|---|
| "Ticket compliance analysis — #1143 partially compliant" | **declined** | It resolved `#1143` as the ticket. #1143 is a *pull request*, not this PR's issue; the issue is #1124, referenced as `Refs`. The requirements it graded are #1143's, not this change's. |
| "The exact count of 8 bats cases and 7 Go tests is not verifiable from the diff alone" | **declined — correct observation, wrong target** | Those counts appear nowhere in this PR's description; they are #1143's. What this PR states is a measured run, not a count claimed from a diff: `bats tests/*.bats` → **1400 ok / 0 not ok, BATS_EXIT=0**, and `go test -count=1 ./...` → 17 packages ok. A reviewer cannot verify either from a diff, which is why the output of both is in `verification.md` rather than asserted in prose. |
| "Requires further human verification: run the test suite" | **satisfied before opening** | Full loop in the PR body and in `specs/HARNESS-075-model-map-routing-registry/verification.md`, run on the assembled tree against current `main`, including `GOOS=windows go vet` and `golangci-lint` on the `versions.conf` pin (v2.12.2, 0 issues). |

Nothing here is applied or ticketed: the one substantive point is graded against another PR's
acceptance criteria.

### CodeRabbit — still running

`Review in progress` at the time of writing. Its output is **not yet dispositioned**, and this
comment does not claim otherwise. If it posts findings after merge they get triaged on the
follow-up rather than lost.

### Open by design

This PR does not close #1124. Round 5 asks for a re-review at the **merged** sha — the whole point
of its Blocker — so the spec stays active and `dotf spec archive` stays blocked until round 6 runs.

Issue #1164, opened 2026-08-22
GUARD: an agent record's declared tier can disagree with model-map, and nothing catches it before deploy


mlorentedev:
## Problem

#1161 makes the agent render resolve a persona's neutral `model:` tier through `harness/model-map.json`, and fail loudly when the map cannot answer. That failure is discovered at **deploy** time. Nothing discovers it earlier.

The gap is a drift between two committed files: a record under `harness/agents/<name>/AGENT.md` declares a tier, and `harness/model-map.json` declares which harnesses that tier covers. They can disagree in the repository, and the disagreement surfaces only when someone runs `compile-harness.sh --deploy` on a machine.

## Why --check is deliberately not the place

`compile-harness.sh --check` is the offline drift gate, and it runs in the CI `lint` job (`.github/workflows/ci.yml:96`), which installs no Go toolchain and has no `dotf`. Making `--check` resolve tiers would make it report drift on a perfectly good record whenever the machine running it has no `dotf` — conflating a property of the deploy **environment** with a property of the committed **record**. #1161 therefore keeps `render_agent` a pure renderer and resolves in the deploy caller.

## Why dotf doctor is

`cli/internal/doctor/checks_model_map.go` already loads and validates `harness/model-map.json`, and `dotf doctor` is by definition running where `dotf` exists. The check has everything it needs and no new coupling.

## Scope

A `dotf doctor` check asserting that, for every entry in `manifest.json`'s `agents.deploy`, every committed agent record's declared tier resolves for that harness.

- Passes when a record declares no tier (not declaring one is not an error).
- Fails naming the record, the tier and the harness — the same three facts the deploy-time failure names.
- Table-driven Go tests, including a fixture where the record and the map disagree.

## Acceptance

- A record declaring a tier no deploy target can answer fails `dotf doctor` with all three names in the message.
- The current tree passes: `curator` declares `top`, the only deploy target is `claude`, and `tiers.top.claude` is declared.
- Adding a second `agents.deploy` entry for a harness with no tier entry fails the check.

## Refs

#1161 (the deploy-time half), ADR-035, `harness/agent-frontmatter.schema.json`.

Issue #1163, opened 2026-08-22
BOARD: #560's body still claims model-map.json, which ADR-035 re-scoped away from it


mlorentedev:
## Problem

ADR-035 re-scoped #560, and #560's body still describes the old scope.

ADR-035 decision 1:

> `harness/model-map.json` is the dispatch-time policy registry of ADR-032 §3, and #1124 owns it alone. **#560 is re-scoped to `harness/capability-map.json`** — neutral capabilities -> each harness's tool and permission names — which ADR-032 does not cover and which remains needed.

#560's body still reads:

> ## Scope
> - `harness/capability-map.json`: [...]
> - `harness/model-map.json`: neutral `model: top|mid|low` (ADR-011) -> each harness's model id.
> - Engine resolves both at render.

`harness/model-map.json` was built and merged under #1124 (#1143, #1155), and its spec is archived at `specs/archive/HARNESS-075-model-map-routing-registry/`. So #560's scope names a file that already exists and that a different ticket owns — the exact two-issues-one-filename collision ADR-035 was written to resolve, still readable on the board.

## Impact

Low but real: the collision is what ADR-035 spent a decision on, and anyone reading the board rather than the ADR re-derives it. It also leaves ambiguous who owns the compile-time tier render, now tracked separately as #1161.

## Fix

Rewrite #560's Scope and Acceptance to cover `capability-map.json` only, and point its Refs at ADR-035 for the re-scope. Its title should lose "model-map".

## Refs

ADR-035 decision 1, #1124, #1161, epic #558.

mlorentedev:
Done. #560's title and body now describe `capability-map.json` only, with ADR-035 decision 1 quoted as the reason for the re-scope and an explicit note that the model half shipped under #1124/#1165 and must not be re-opened there.

The rewrite also sharpens the remaining scope with the measurement #1165 made available: `curator` declares `capabilities: [read, search, edit, shell]` and the deployed agent file carries none of it — exactly where `model:` was before #1165. And it names the destination field, verified against the installed plugin agents rather than assumed: Claude Code agent frontmatter takes `tools:` as a comma-separated list of native tool names.

Issue #1162, opened 2026-08-22
BUG: model-map declares agy as render agent-md while its own $comment groups it with pi as a launcher-flag harness


mlorentedev:
## Problem

`harness/model-map.json` contradicts itself about `agy`, and only a real consumer of `tiers` surfaces it.

The `$comment` states the keying rule and its reason:

> TIERS ARE KEYED BY WHATEVER CONSUMES THE MODEL ID [...] `claude` and `opencode` render the id into their own agent definition, so they key by harness. `nan` keys by POOL, because the harnesses that reach it (`pi`, `agy`) take the model as a launcher flag rather than a rendered field — `pi --model X`, `agy --model X`. So `pi` correctly has no tier entry: its render is `adapter` [...] A harness whose render is `adapter` keys by its pool.

The rule the paragraph lands on is *render `adapter` -> keys by pool*. But the `harnesses` block declares:

```json
"agy": { "pools": ["nan"], "render": "agent-md", "spawn": "none" }
```

So `agy` is named in the prose as a launcher-flag harness alongside `pi`, and declared in the data as `agent-md` alongside `claude` and `opencode`. Exactly one of the two is right.

## Why it matters

`ResolveTier(m, tier, "agy")` errors for every tier, because no tier declares an `agy` key. Under the prose's rule that is correct — agy keys by its pool, `nan`. Under the data's `render: agent-md` it is a gap, because an `agent-md` harness is precisely one that renders a model id into a definition file and therefore needs a tier entry.

Nothing catches it today because `tiers` has no consumer (#1161) and `agy` is not in `manifest.json`'s `agents.deploy`. It becomes a live routing bug the moment either changes.

## Options

1. **The prose is right, the data is wrong** — `agy` takes `--model` on the launcher, so its `render` should not be `agent-md`. Fixes the declaration to match the measured behaviour.
2. **The data is right, the prose is wrong** — `agy` really does render an agent definition, so it needs tier entries and the `$comment` should stop grouping it with `pi`.

Deciding needs one empirical check: does `agy` read a rendered agent definition file, or does it only take `--model` on the command line?

## Acceptance

- The `$comment` and the `harnesses` block agree about `agy`.
- A structural assertion in `cli/internal/harness/model_map_test.go` holds them to it: every harness whose `render` is `agent-md` has an entry in at least one tier.

## Refs

ADR-035, #1124, #1161.

Issue #1158, opened 2026-08-21
GUARD: nothing detects that the deployed dotf predates the repo, so the gates can run on a binary that is not this tree


mlorentedev:
## Measured 2026-08-21

`~/.local/bin/dotf spec init` created three files. The same command from a binary built out of
`origin/main` at the same moment created four:

```
$ dotf spec init GUARD-005-... --issue 1157
[OK] Created: specs/GUARD-005-...
     proposal.md, tasks.md, verification.md

$ go build -o /tmp/dotf-main ./cmd/dotf && /tmp/dotf-main spec init ZZTEST-999 --force-no-gate
[OK] Created: specs/ZZTEST-999
     proposal.md, tasks.md, verification.md, features.json
```

The deployed binary predates #1127, which added `features.json` to `templateNames` and the
`//go:embed` line. `cli/internal/cmd/spec.go:420` on main prints all four names, so the source and
the binary disagree about what the command does.

## Why doctor does not see it

`checkOptionalTools` reports `dotf` with a **version-pin match** against `versions.conf`
(drift → WARN). That answers "is the declared version the pinned one", not "was this binary built
from this tree". `dotf` is unlike every other tool in that list: the others are third-party
releases, and `dotf` **is this repository**. Between two releases the version string is constant
while the source moves every merge, so the check is structurally blind for exactly the tool it
cannot afford to be blind about.

## Why it is more than an inconvenience

Every gate in this repo runs through `dotf`: `spec review`, `spec archive`, `doctor`,
`pr triage-queue`, `secrets run`. A stale binary means a gate can enforce a contract that no longer
exists, or fail to enforce one that does — and report success either way.

Concretely, in the session that found this: HARNESS-075's round-6 adversarial review was launched
by the stale `dotf spec review`. The review itself is unaffected in substance (the reviewer reads
the repo, not the binary), but nobody knew the launcher was not main's launcher, and there was no
way to find out short of building a second copy and diffing behaviour. A spec that scaffolds
without `features.json` also silently skips the machine-readable acceptance contract, which is the
artifact the harness is supposed to stamp `passing`.

This is the same class the repo has already named twice — a signal that reports health it never
established — one layer lower, in the tool that produces the signals.

## Fix shape

1. **Stamp the commit at build time.** `-ldflags "-X main.commit=$(git rev-parse HEAD)"` in the
   build/release path, exposed as `dotf version --commit` (there is currently no `--version` flag
[...]

PR #1165, opened 2026-08-22
feat(harness): give model-map's tiers block its first consumer

### LINKED ISSUE #1161: HARNESS: model-map's tiers block has no consumer, so a persona's declared tier never reaches the deployed agent file
## Problem

`harness/model-map.json` shipped with a `tiers` block and no consumer. Measured on 2026-08-21:

- `harness.ResolveTier` and `harness.ResolveChain` have **zero callers outside tests**. The only reader of the whole registry is `dotf doctor`, which reports on it and routes nothing with it.
- `harness/agents/curator/AGENT.md` declares `model: top`.
- `render_agent()` (`scripts/compile-harness.sh:322`) keeps only `name` and `description` and **drops `model`**.
- The deployed `~/.claude/agents/curator.md` therefore carries no `model:` field at all, so the roster's only populated persona runs on whatever model the invoking session happens to be on.

A tier declared in two places that nothing reads is the same failure class the registry itself was built to end, one layer down. `harness/agent-frontmatter.schema.json` already documents the intended behaviour — *"Neutral tier (top|mid|low); mapped to a provider id via model-map at compile"* — and the render does the opposite.

## Scope

Give `tiers` its first real consumer — the compile-time class ADR-035 names, not the run-time one.

- A `dotf` subcommand that resolves a neutral tier for one harness through the validated loader, so resolution and schema validation stay single-sourced in Go.
- `render_agent()` calls it and emits `model: <resolved id>` into the rendered frontmatter.
- Unresolvable tier -> the agent render fails loudly and names the harness and tier (C15). Not silently dropped, which is today's behaviour.

Out of scope: `chains`, the run-time dispatcher, `dotf agent run`, and `capability-map.json` (#560).

## Why this and not the dispatcher

ADR-035 splits consumption into two cadences over one file: `tiers` at compile time feeding each harness's render, `chains` at run time read by a level-2 dispatcher. This is the compile-time half — small, already specified by accepted ADRs, and it proves the loader is usable from the render path before the dispatcher is built on top of it.

## Acceptance

- `harness/agents/curator/AGENT.md` declaring `model: top` deploys to `~/.claude/agents/curator.md` carrying `model: opus`.
- A tier a harness does not declare fails the agent render with a message naming both, rather than dropping the field.
- The failure is scoped to the agent render; skill deploy is unaffected.
- bats coverage for resolve and for the unresolvable path; Go table tests for the subcommand.

## Notes

`manifest.json`'s `agents.deploy` has exactly one entry today (`claude` / `agent-md` / `.claude/agents`), and `claude` is the one harness with all three tiers declared. So the fail-loud path is unreachable through the current deploy set and becomes reachable the moment a harness without a tier is added — which is when it should fire.

## Refs

ADR-035 (the two consumer classes), ADR-027 §2, ADR-032 §3, ADR-011 (tiers), epic #558.

mlorentedev:
Closes #1161. Spec: `specs/HARNESS-076-model-map-tier-render/`.

## The gap

`harness/model-map.json` merged as a declaration nobody reads. Measured 2026-08-21:

- `harness.ResolveTier`, `harness.ResolveChain` and `harness.LoadModelMap` had **zero callers outside tests**. The registry's only reader was `dotf doctor`, which reports on it and routes nothing with it.
- `harness/agents/curator/AGENT.md` declares `model: top`.
- `render_agent()` dropped the field, so the deployed `~/.claude/agents/curator.md` carried **no `model:` at all** — the roster's only populated persona ran on whatever model the invoking session happened to be on.
- `harness/agent-frontmatter.schema.json` already documented the intended behaviour (*"mapped to a provider id via model-map at compile"*), and the render did the opposite.

ADR-035 splits consumption of that one file into two cadences: `tiers` at compile time feeding each harness's render, `chains` at run time for a level-2 dispatcher. **This is the compile-time half.** `chains`, budget enforcement and `dotf agent run` are untouched.

## What changed

`dotf harness resolve-tier <tier> --harness <name>` resolves through the validated loader; `scripts/compile-harness.sh` calls it and emits the resolved line. `model: top` now deploys as `model: opus`.

Shelling out beats reading the map with `jq`: `jq` would restate the resolution rules in a second place and skip schema validation entirely, so a map the Go validator rejects would render clean.

## The decision worth reviewing

**Failure splits by cause**, and the split is the design:

| cause | response | why |
|---|---|---|
| map cannot answer — undeclared tier, absent, non-JSON, schema-invalid | **fail the render**, naming harness and tier | C15: a map that cannot be *read* never resolves to empty or to a permissive default |
| resolver absent, or too old to carry the subcommand | **warn loudly, render with no model line** | a bootstrap state, not an unreadable map |

The second row was **narrowed by evidence during implementation**. The first cut made a missing `dotf` fatal and broke 17 tests in `tests/skills-pipeline.bats`, which drive the real deploy — because the `dotf` deployed on the dev box predates the subcommand (#1158). That is exactly what the fatal branch would inflict on any under-provisioned machine: `setup-linux.sh` installs `dotf` best-effort (`install_dotf || log_warning`), so making it fatal promotes a warned-past dependency into a hard prerequisite of the entire harness deploy. The degraded output is precisely what this script produced before the change, so that path is never worse than the status quo — and it still *says so*, which is ADR-032's honest-degradation bar.

**Telling the two apart needs a capability probe, not an exit status.** Measured: a stale `dotf` answers `resolve-tier top --harness claude` with `unknown flag: --harness` and **exit 1** — identical to a genuine routing refusal. So `dotf_knows_resolve_tier` greps `dotf harness --help` for the subcommand name, and `TestHarnessHelpListsResolveTier` pins that string from the producing side, so a rename cannot silently make the probe answer "too old" forever and degrade every render. Written up as `docs/lessons/lesson-219`.

Two changes fall out of it:

- **`render_agent` stayed a pure renderer**, with the resolved line passed in. `--check` also renders and runs in the CI `lint` job, which installs no Go; resolving there would report drift on a perfectly good record purely because the machine lacks `dotf` — conflating a property of the deploy *environment* with a property of the committed *record*.
- **The render writes through a sibling temp file.** `> "$outp"` truncated the target *before* the renderer ran, so a failed resolve left an **empty** agent definition behind — a file naming no model, the exact failure this change removes. The temp sits beside the target rather than in `$TMPDIR`, which keeps umask permissions (`mktemp` is 0600; deployed artifacts are 664) and an atomic rename that a cross-filesystem `mv` would not guarantee.

## Testing

`tests/compile-harness-real.bats` **pairs the new `dotf` stub with the real binary** rather than taking an exemption from `tests/stub-real-pairing.bats` (BUG-055). Unlike the `gh`/`hive` suites in that table, `dotf` is our own binary: a real test builds it and drives the deploy end to end, mutating nothing.

All run in this session, in this worktree:

- `go build ./... && go vet ./... && GOOS=windows go vet ./... && go test ./...` — clean
- `golangci-lint run` — `0 issues.` (v2.12.2, matching the `versions.conf` pin)
- `shellcheck scripts/*.sh setup-linux.sh`, `bash -n`, `zsh -n scripts/compile-harness.sh` — clean
- `./scripts/check-bats-names.sh tests/` — `OK (99 file(s) clean)`
- `bats tests/*.bats` — **1415 passing, 0 failing** (baseline on `main`: 0 failing, so the mid-implementation failures were this change's and are resolved, not tolerated)
- Every `features.json` verification command was executed **and negative-controlled** — each fails when its test is absent, per `docs/lessons/lesson-217`

`gofmt -l` still reports `internal/doctor/report.go`: that is **#1154, pre-existing on `main`** and untouched here.

## Not in scope, filed instead

- **#1162** — `model-map.json` contradicts itself about `agy`: declared `render: agent-md`, while its own `$comment` groups it with `pi` as a launcher-flag harness. Only a real consumer of `tiers` surfaced it.
- **#1163** — #560's body still claims `model-map.json`, which ADR-035 re-scoped away from it.
- **#1164** — a record's tier and the map can still disagree in the repo, and only a deploy notices. Belongs in `dotf doctor`, which already loads the map and runs where `dotf` exists.

## Also included

`ai/claude/settings.json` gains `"outputStyle": "Concise"` — a config change made by the repo owner, carried here at their request.

## Summary by CodeRabbit

* **New Features**
  * Added harness tier resolution, translating neutral tiers into harness-specific model IDs.
  * Agent frontmatter now includes resolved model values when available.
  * Added safer deployment behavior that preserves existing outputs when rendering fails.
  * Added Claude `outputStyle` synchronization across setup platforms.

* **Bug Fixes**
  * Improved diagnostics for unavailable, outdated, or invalid model resolvers.
  * Prevented silent fallback when model maps are missing or invalid.

* **Documentation**
  * Added guidance on stale CLI detection and model-tier rendering behavior.


mlorentedev:
## Review triage

Three reviewer sources on this PR, all dispositioned. Merged at `8e45bd8`.

**Independent adversarial review** (`dotf spec review`, `nan/deepseek-v4-flash` from `harness/reviewer-pool.json`) ran twice, because applying round 1's findings invalidated the `reviewed_sha` that authorised the archive — the gate behaving as designed (#998's scenario), not a stall.

| Round | sha | Verdict | Findings |
|---|---|---|---|
| 1 | `b644acc` | PASS-WITH-GAPS | 5 Minor REAL, 1 Question THEORETICAL |
| 2 | `09511cf` | PASS-WITH-GAPS | 2 Minor REAL, 2 Minor THEORETICAL, 1 Question — no Blockers, no Majors |

### Round 1 — applied in `09511cf`

| Finding | Disposition |
|---|---|
| `2>/dev/null` swallowed the resolver's stderr, so a schema-invalid map reported as *"tier top does not resolve for harness claude"* | **Applied.** Blaming the record for a defect in the registry defeats the point of splitting the failure by cause. Wrapper now points at the resolver's own diagnosis; new bats case pins that the cause survives into the deploy output |
| `tasks.md` boxes all unticked despite verified work | **Applied** |
| `verification.md` "Deviation from AC6" contradicted the committed AC6 | **Applied** — stale drafting history, removed |
| `features.json` f6 positionally brittle: `-f 'dotf'` matched 3 tests and `^ok 2 ` landed by file position | **Applied.** Every verifier now pins by unique NAME and propagates the runner's exit status. Two uncovered shapes became their own features |
| `ai/claude/settings.json` `outputStyle` undocumented scope | **Split.** Documented in the proposal's Out of scope; **declined** moving it to a separate PR — the repo owner asked for it here |
| SIGPIPE with `grep -q` under `pipefail` (THEORETICAL) | **Applied** — matched from a here-string, no pipe, no such mode |

### CodeRabbit — 10 inline comments

| Finding | Disposition |
|---|---|
| `deploy_agents` over the repo's own <40-line rule (Major) | **Applied** — resolution extracted to `agent_model_line`; every function in the changed set now under 20 lines |
| `settings.json` `outputStyle` never reaches an existing installation | **Applied, and it was the highest-value comment on this PR.** `merge_claude_settings` is an explicit allow-list; every unnamed key is preserved from the existing file. Verified on this machine: template carried `outputStyle`, deployed settings had no such key. Both `setup-linux.sh` and `setup-windows.ps1` now name it, plus a guard asserting every dotfiles-owned scalar key appears in both policies (negative-controlled) |
| `features.json` verifiers should propagate the runner's exit status | **Applied** — converged with the adversarial review's independent finding on the same file |
| markdownlint MD040: bare fence in lesson-219 | **Applied** |
| temp file not cleaned up if `mv` fails | **Applied** |
| proposal/tasks described the pre-refactor design | **Applied** |
| Go table tests exceed 40 lines; extract helpers | **Declined.** Table-driven tests routinely exceed it in this repo (`model_map_test.go` is 673 lines); the rule targets production functions |
| `Deps` injection struct for the cobra command | **Declined.** No other `dotf` command injects its env/registry seams; adopting it in one place makes this file the outlier, against the repo's own idiom-consistency bar |
| "stable status tags instead of mutable prose" in error assertions | **Declined** — over-engineering for a two-branch resolver |
| warn before `rm -rf "$FAKEHOME"` in a test | **Declined** — the path is built from `BATS_TEST_TMPDIR`; a stderr notice on every run is noise |

### Round 2 — one applied, two ticketed, two accepted

| Finding | Disposition |
|---|---|
| `if ($tmpl.outputStyle // empty)` does not mean "leave it alone when absent" — in jq an `empty` condition makes the WHOLE merge yield nothing, silently disabling every key (THEORETICAL) | **Reproduced, which upgraded it, and fixed in #1167.** Latent only because the template carries the key today |
| A whitespace-bearing model id returns rc=2, so the operator is told to rebuild `dotf` when the defect is in the map | **Ticketed #1168.** Downstream of #1159, which is the schema pattern permitting it |
| `deploy_agents` aborts on the first unresolvable record, blocking every agent behind it | **Ticketed #1169.** Unreachable with one deploy target; wants deciding before the roster grows |
| `verification.md` says "6 new bats cases", the diff adds 11 | **Accepted.** Prose drift in an archived artifact; the 1418-passing claim is exact and every named test exists |
| `deploy_agents` still 52 raw lines | **Accepted.** Executable body is well under; six pre-existing functions in the same file exceed 40 (`do_refresh` 112, `do_check` 85, `deploy_skills` 68), so a single-function cleanup here is cosmetic. Belongs to the CLI-026 port |

### Checks

All green at merge except `goreleaser release` (skipping, not a failure). `spec-gate` went green once the spec archived; it had correctly refused a PR closing #1161 with the spec still active.

PR #1167, opened 2026-08-22
fix(setup): jq `// empty` guard silently disabled the whole Claude settings merge

mlorentedev:
Follow-up to #1165, from its own round-2 adversarial review.

## The bug

#1165 added `outputStyle` to `merge_claude_settings`, guarded like this:

```jq
| (if ($tmpl.outputStyle // empty) then .outputStyle = $tmpl.outputStyle else . end)
```

with a comment claiming it "leaves the existing value alone if a template ever omits it". **It does the opposite.** In jq, a condition that evaluates to `empty` makes the whole `if` expression produce nothing — so a template without that one optional key makes the entire merge pipeline yield an empty result.

`merge_claude_settings` then hits its own safety net:

```sh
if [ -z "$merged" ]; then
    log_warning "Claude settings merge produced empty output, skipping write"
    return 0
fi
```

…and **nothing deploys**: not `model`, not `effortLevel`, not `permissions.allow`, not the SessionStart/SessionEnd hooks. A warning, exit 0, and a settings file that never got its dotfiles-owned subset.

Reproduced before fixing:

```console
$ jq --argjson tmpl '{"model":"opus","effortLevel":"xhigh"}' \
     '.model = $tmpl.model | .effortLevel = $tmpl.effortLevel
      | (if ($tmpl.outputStyle // empty) then .outputStyle = $tmpl.outputStyle else . end)' \
     <<<'{"model":"sonnet","userThing":"keep"}'
$ echo "[$?]"   # exit 0, and NO output at all
[0]
```

Latent today because the template does carry the key — which is exactly why it was worth fixing now rather than the first time someone trims the template.

## The fix

`has()` is the guard that means what the other one looks like it means. One line, plus the comment corrected to say why.

## Testing

The test **extracts the jq expression out of `setup-linux.sh`** rather than restating it, so it exercises the shipped merge and cannot drift from it. It asserts both directions:

- a template omitting the key still merges everything else (and preserves unrelated user keys);
- a template carrying it still propagates it.

Fails before the fix, at exactly the assertion that matters:

```
not ok 1 the merge expression survives a template that omits an optional key
#   `[ -n "$out" ]' failed
```

- `bats tests/*.bats` -> **1419 passing, 0 failing**
- `shellcheck setup-linux.sh` clean (one pre-existing SC1091 info), `bash -n` clean

## Provenance

Found by the round-2 adversarial review of #1165 (`nan/deepseek-v4-flash`, from `harness/reviewer-pool.json`), which graded it **THEORETICAL**. Reproducing it upgraded it to real: the failure needs no unusual input, only a template edit.

Not SDD-gated: single-line bug fix with an obvious cause, well under the Discipline Gate's thresholds.

### DIFF
--- a/setup-linux.sh
+++ b/setup-linux.sh
@@ -1277,13 +1277,18 @@ merge_claude_settings() {
     # from scratch. Measured on this repo's own box — the key sat in the template
     # while the deployed file had no `outputStyle` at all. It belongs with `model`
     # and `effortLevel`: same file, same kind of setting, dotfiles-owned.
-    # `// empty` leaves the existing value alone if a template ever omits it,
-    # rather than writing null and having Claude read an invalid style.
+    # Guarded with `has()`, NOT with `// empty`. In jq a condition that evaluates
+    # to `empty` makes the whole if-expression produce nothing, so `// empty`
+    # here does not mean "leave it alone when absent" — it means the entire merge
+    # yields an empty result, `merged` is empty, and the function bails with
+    # "merge produced empty output, skipping write". A template that ever omits
+    # this one optional key would then deploy NOTHING: not model, not
+    # effortLevel, not permissions, not hooks.
     local merged
     merged=$(jq --argjson tmpl "$template_substituted" '
         .model = $tmpl.model
         | .effortLevel = $tmpl.effortLevel
-        | (if ($tmpl.outputStyle // empty) then .outputStyle = $tmpl.outputStyle else . end)
+        | (if ($tmpl | has("outputStyle")) then .outputStyle = $tmpl.outputStyle else . end)
         | .permissions = (.permissions // {})
         | .permissions.allow = (((.permissions.allow // []) + $tmpl.permissions.allow) | unique)
         | .hooks = (.hooks // {})
--- a/tests/claude-settings-template.bats
+++ b/tests/claude-settings-template.bats
@@ -47,6 +47,39 @@ setup() {
     done
 }
 
+@test "the merge expression survives a template that omits an optional key" {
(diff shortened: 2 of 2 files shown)


mlorentedev:
## Review triage

**Checks:** 13 pass, 0 failing.

**Reviewer output, and one disclosure:**

|---|---|---|
| **PR-Agent** (`github-actions`) | Reviewed. Ticket compliance vs #1165 **partially compliant** with *zero* non-compliant requirements listed and none needing human verification; effort 1/5; "PR contains tests"; no security concerns; no major issues | **Nothing to apply.** The "partially compliant" grade with an empty non-compliant list reads as the tool declining to assert full compliance against an issue this PR only *refs*, not closes |
| **CodeRabbit** | **Did not review.** Posted a rate-limit notice: *"you've reached your PR review limit, so we couldn't start this review"* | **Disclosed, not dispositioned.** A notice that no review ran leaves this half unreviewed. `review-attestation` is green on PR-Agent's review alone, and CodeRabbit's own check reports `pass` with "Review rate limited" — neither of which means CodeRabbit looked at the diff |

**So this PR carries one reviewer's verdict, not two.** Proceeding is allowed; proceeding silently is not, hence this note. The reset is ~43 minutes out; `@coderabbitai review` re-triggers it if a second pair of eyes is wanted before merge.

**Provenance of the change itself:** the bug was found by the round-2 independent adversarial review of #1165 (`nan/deepseek-v4-flash`, from `harness/reviewer-pool.json`), which graded it THEORETICAL. Reproducing it before fixing upgraded it — the failure needs no unusual input, only a template edit. That reproduction and its failing-first test are the evidence this PR rests on, independent of either PR reviewer.

Issue #1170, opened 2026-08-22
MODEL-MAP: copilot declares render agent-md with no tier, and deciding it needs the corporate seat


mlorentedev:
## Problem

`harness/model-map.json` declares `harnesses.copilot.render: "agent-md"` — the kind that renders a model id into a definition file — and **no tier names `copilot`**, in any of `top`, `mid` or `low`. So `ResolveTier(m, <any tier>, "copilot")` errors for every tier.

Surfaced by `TestRenderKindAgreesWithTierKeying`, added in #1162 alongside the `agy` fix. `copilot` is currently listed in that test's `tierlessRenderers` exemption table, which is a **known backlog, not an approval**: the entry exists so the set can only shrink deliberately.

## Why it was not decided with agy

`agy`'s answer was verifiable here: `cli/internal/spec/review_launch.go` builds `agy --model <id> …`, and `agy agents` lists none, so its model arrives as a launcher flag and its render kind is `adapter`. Copilot's is not verifiable on this machine:

- the `copilot` binary is not installed (a `~/.copilot` config dir exists, but no CLI on PATH);
- `copilot` appears only in `manifest.json`'s `agents.presence` (an instructions file at `.copilot/copilot-instructions.md`), never in `agents.deploy`, so nothing here renders an agent definition for it;
- it is a **seat-based corporate** surface, and which model ids that seat accepts is not knowable from this box. `pools.copilot` already says as much about concurrency: *"Seat-based, so concurrency is a fleet property rather than a number this map can state honestly."*

Guessing the ids would put a route to nowhere in the map that validates cleanly — the exact class ADR-035 deleted the phantom `codex` pool to avoid.

## The question to answer at the Windows box

Does Copilot CLI take the model from the **agent definition file** (`.github/agents/*.agent.md` frontmatter), or from the **session** (`/model`, a config default, or the seat)?

- **From the definition file** -> `render: agent-md` is right; declare `copilot` tier entries with the real model ids and drop it from `tierlessRenderers`.
- **From the session** -> `render` should be `adapter`, like `pi` and `agy`; drop it from `tierlessRenderers` and the test's other assertion covers it.

ADR-032 records that Copilot CLI ships custom agents at `.github/agents/*.agent.md` plus `/fleet` — which establishes that agent definitions exist, but not that they carry a model id. That is the specific thing to check.

## Acceptance

- The question above is answered against the real CLI, with the command output recorded.
[...]

Issue #1169, opened 2026-08-22
DESIGN: deploy_agents aborts on the first unresolvable record, so one bad tier blocks every agent behind it


mlorentedev:
## Problem

`deploy_agents` in `scripts/compile-harness.sh` `return 1`s on the FIRST record whose tier cannot be resolved, aborting the whole loop. Records already written keep their fresh files; every record behind the failure never deploys at all.

With today's `agents.deploy` set this is unreachable — `harness/manifest.json` declares exactly one target (`claude`) and `harness/agents/` holds one record (`curator`). It becomes real the moment either grows: one broken record then blocks every agent behind it, and which agents those are depends on directory iteration order.

## The tension, stated honestly

This is not obviously wrong. #1165's AC5 decided fail-loud deliberately, and a half-deployed roster is its own hazard — an operator who sees "deploy failed" and a partially-updated `~/.claude/agents/` has to work out which half is current.

The two defensible shapes:

1. **Abort (today).** Simple, and no agent deploys against a map the operator has not fixed. Cost: the blast radius is every later record, for a defect in one.
2. **Collect and continue.** Render every record that resolves, accumulate the failures, and exit non-zero at the end naming all of them. The operator sees the complete list in one run instead of fixing them one deploy at a time. Cost: a non-zero exit now means "some agents deployed", which the message must say plainly.

Option 2 is the better fit for the same reason `dotf pr triage-queue` reports the whole queue rather than the first item, but it should be decided rather than drifted into.

## Trigger

Worth deciding **before** a second entry lands in `agents.deploy` or a second persona lands in `harness/agents/` — either makes the behaviour observable, and the roster is meant to reach seven (`00_meta/agents/ROSTER.md`).

## Acceptance

- The chosen semantics are stated in `scripts/compile-harness.sh` next to the loop, not left implicit.
- If option 2: a bats case with two records, one resolvable and one not, asserts the good one deployed, the exit is non-zero, and the message names every failure.
- If option 1 is kept: a bats case pins that later records do NOT deploy, so the behaviour is a decision rather than an accident.

## Refs

#1165 (HARNESS-076, archived), round-2 adversarial review Question finding, `harness/manifest.json` `agents.deploy`.

Issue #1168, opened 2026-08-22
BUG: a whitespace-bearing model id is reported as "dotf is too old", sending the operator to rebuild a binary that is fine


mlorentedev:
## Problem

`resolve_model_tier` in `scripts/compile-harness.sh` sniffs the resolved id and rejects anything with whitespace:

```sh
case "$out" in
    ''|*[[:space:]]*) return 2 ;;
esac
```

`return 2` is the "resolver unavailable" outcome, so the caller prints *"dotf is absent, or predates the resolve-tier subcommand"*. But a whitespace-bearing id means the resolver **ran fine and answered** — the map holds `"opus "` or `"opus 4"`. The operator is told to rebuild `dotf`, which fixes nothing, while the actual defect sits in `harness/model-map.json`.

The sniff was added as belt-and-braces behind the capability probe, for the case where a stale binary printed a help screen. Once the probe exists, a whitespace id can only come from the map.

## Why the map allows it

`harness/model-map.schema.json` uses `pattern: "^\\S.*$"`, which anchors only the FIRST character — trailing and interior whitespace validate. That is **#1159**, already open. This issue is the downstream half: even after the schema is tightened, the render should not misattribute the cause.

## Fix

Give it a third outcome, or fold it into the rc=1 (routing error) branch with its own message naming the offending value:

```
[ERROR] model-map.json gives tier "top" on harness "claude" the id "opus 4" —
        a model id is one token with no whitespace
```

The distinction the whole function exists to draw is *whose fault is this*, and this branch currently gets it backwards.

## Acceptance

- A resolver returning a whitespace-bearing id fails naming the map and the value, and does NOT suggest rebuilding dotf.
- A bats case pins the message (the existing stub can return `"opus 4"`).
- Still distinct from a genuinely absent/too-old resolver, which keeps its own warning.

## Refs

#1165 (introduced), #1159 (the schema pattern that permits it), round-2 adversarial review of HARNESS-076.

PR #1172, opened 2026-08-22
feat(harness): give capabilities the same seam the model tier got

### LINKED ISSUE #560: HARNESS-044: Agnosticism map — capability-map (model-map re-scoped out by ADR-035)

mlorentedev:
Closes #560, #1168, #1169. Spec: `specs/HARNESS-077-capability-map/`. Third in the sequence after #1165 (tier render) and #1171 (the `agy` contradiction it exposed).

## The gap

```
harness/agents/curator/AGENT.md     ~/.claude/agents/curator.md  (before)
  model: top                    ->    model: opus        # #1165 fixed this
  capabilities: [read, search, edit] ->  «nothing»       # this PR
```

`render_agent` dropped `capabilities` exactly as it dropped `model` until #1165 — a neutral declaration promised by `harness/agent-frontmatter.schema.json` and read by nobody. It is also the half of ADR-027 §2's agnosticism pair that never shipped.

After this PR `curator.md` carries `tools: Read, Glob, Bash` next to `model: opus`.

## Why this is a map and not a rename

The native shapes genuinely differ. Verified 2026-08-22 against the shipping tools, not assumed:

| Harness | Field | Semantics |
|---|---|---|
| `claude` | `tools: Read, Glob, Bash` | **allow-list** — naming a tool grants it **and denies everything unnamed** |
| `opencode` | `permission: {bash: allow}` | **decision map** — an unnamed key falls back to a default, so naming a capability grants it and says *nothing* about the rest |

`opencode`'s own schema marks its `tools` field *"@deprecated Use 'permission' field instead"*, so mapping onto it would ship a deprecation on day one.

Two design consequences fall straight out of that asymmetry:

- **The resolver takes a capability SET, not one at a time.** A per-capability API would push the allow-vs-grant distinction onto every caller, and the shell caller is the one least able to carry it.
- **It returns the whole frontmatter line, field name included**, because the field differs per harness and the render has no business knowing which. This deliberately differs from `resolve-tier`, where the field is always `model:`.

Proof the distinction is real, from the same neutral request, asserted against the real binary:

```console
$ dotf harness resolve-capabilities shell --harness claude
tools: Bash
$ dotf harness resolve-capabilities shell --harness opencode
permission: {bash: allow}
```

## On agnosticism

This declares **two** harnesses, not one. `opencode` has no `agents.deploy` target yet, so its column is unexercised by any deploy — included anyway because the standing complaint against this pipe is that it is neutral by mechanism and single-vendor by data, and a second harness with a *verified* native shape is the cheapest honest step against that. It is also what makes the per-harness-shape decision concrete rather than theoretical.

`pi` and `agy` are absent because they are `adapter` harnesses (#1171) that render no definition file. `copilot` is absent because its native tool vocabulary is not checkable from this machine — the same not-verifiable-here reason as #1170, and guessing would put a route to nowhere into a map that validates cleanly.

## Two tickets folded in, because they live in this loop

- **#1168** — a whitespace-bearing model id was reported as *"dotf is too old"*, sending the operator to rebuild a binary that was fine while the defect sat in `model-map.json`. Behind the capability probe that answer can only mean bad data, so it gets its own exit status and names the value and the file.
- **#1169** — `deploy_agents` aborted on the first unresolvable record, leaving every agent behind it undeployed, and *which ones* depended on directory iteration order. It now collects, deploys everything that resolves, and exits non-zero naming every failure — saying plainly that other agents **did** deploy, so a non-zero exit is not misread as "nothing happened".

The capability probe is also generalised over the subcommand name: with two registry consumers, the realistic staleness shape is a binary new enough for one and not the other, so each field probes independently and degrades only itself.

## Deliberately NOT in this PR

- **The `dotf doctor` check over this registry, and #1164's record-vs-map tier check.** Both are diagnostics in `cli/internal/doctor/` rather than parts of the render, and folding them in pushed the diff past ADR-017's atomic cap. Immediate follow-up, together.
- **Unifying the schema custom-rule plumbing.** `model-map`'s `customRules` is a package-level closed set bound to one schema, and `TestShippedSchemaDeclaresEveryCustomRule` asserts that schema declares every rule in it — so a second registry cannot declare an `x-` rule without refactoring a file hardened over six adversarial review rounds. This registry's one cross-block invariant (`checkVocabularyCoverage`) runs in its loader instead. Refactoring hardened code as a *side effect* of adding a feature is how it gets destabilised, so it is filed to be done deliberately.

## On the diff size — flagging it rather than hiding it

520 production lines, over ADR-017's ~300 guideline. The breakdown, since the raw number overstates the review burden:

| File | Lines | Note |
|---|---|---|
| `capability_map.go` | 250 | 186 are code; the rest is the WHY this repo asks for |
| `capability-map.json` | 105 | data + the `$comment` rationale block |
| `capability-map.schema.json` | 89 | contract, largely descriptions |
| `compile-harness.sh` | 86 | includes the #1168 and #1169 fixes |
| `harness_resolve_capabilities.go` | 78 | 59 are code |

It is one idea — a registry with its consumer — and splitting it would break the rule this sequence established: **no declaration without its consumer in the same PR**, which is precisely how `model-map` came to merge with zero readers and how `agy`'s contradiction survived four review rounds. I cut what was genuinely separable (the doctor checks) rather than slicing this in half.

## Testing

- `go build ./... && go vet ./... && GOOS=windows go vet ./... && go test ./...` — clean
- `golangci-lint run` — `0 issues.` at the `versions.conf` pin; `gofmt` clean (only pre-existing #1154)
- `shellcheck --severity=error` (CI's own level) — exit 0; `bash -n` and `zsh -n` clean
- `bats tests/*.bats` — **1429 passing, 0 failing**
- All **13** `features.json` verifiers executed; each propagates the runner's exit status and pins its test by unique NAME rather than `ok N` position
- `tests/compile-harness-real.bats` grew four cases driving the **real** binary, including the one asserting the two native forms differ

## Summary by CodeRabbit

* **New Features**
  * Added capability mapping for Claude and OpenCode harnesses.
  * Added `harness resolve-capabilities` to translate neutral capabilities into harness-specific configuration.
  * Deployment now renders native capability settings into agent frontmatter.

* **Bug Fixes**
  * Improved deployment error handling, preserving successful outputs and reporting all failed agents.
  * Added clearer handling for unavailable tooling and invalid model or capability resolution results.

* **Tests**
  * Added comprehensive validation, resolution, deployment, and real-binary coverage for capability mappings.

### DIFF
--- a/cli/internal/cmd/harness.go
+++ b/cli/internal/cmd/harness.go
@@ -25,6 +25,7 @@ pattern triggers, and workflow integrations.`,
 	cmd.AddCommand(newHarnessTriggersCmd())
 	cmd.AddCommand(newHarnessSuggestCmd())
 	cmd.AddCommand(newHarnessResolveTierCmd())
+	cmd.AddCommand(newHarnessResolveCapabilitiesCmd())
 	return cmd
 }
 
--- a/cli/internal/cmd/harness_resolve_capabilities.go
+++ b/cli/internal/cmd/harness_resolve_capabilities.go
@@ -1,0 +1,78 @@
+package cmd
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/mlorentedev/dotfiles/cli/internal/env"
+	"github.com/mlorentedev/dotfiles/cli/internal/harness"
+)
+
+// newHarnessResolveCapabilitiesCmd is capability-map.json's consumer, and the
+// sibling of resolve-tier: same seam, one field over.
+//
+// It resolves a capability SET rather than one capability at a time, because the
+// two native forms disagree about what a partial answer means. A `csv` field is
+// an allow-list — naming a tool grants it and omitting one denies it — while a
+// `decision-map` grants without denying. A per-capability API would push that
+// distinction onto every caller, and the shell caller is exactly the one least
+// able to carry it.
+func newHarnessResolveCapabilitiesCmd() *cobra.Command {
+	var (
+		harnessName string
+		repoRoot    string
+	)
+
+	cmd := &cobra.Command{
+		Use:   "resolve-capabilities <capability>[,<capability>...]",
+		Short: "Resolve neutral capabilities to one harness's native frontmatter line",
+		Long: `resolve-capabilities answers which native tools or permissions a set of neutral
+capabilities means for one harness, reading harness/capability-map.json through
+the validated loader.
+
+It prints a COMPLETE frontmatter line — field and value — because the field name
+differs per harness and a caller should not have to know it:
+
+    cap_line="$(dotf harness resolve-capabilities read,shell --harness claude)"
+    # tools: Read, Glob, Bash
+
+Both declared forms render on one line and both are valid YAML: ` + "`csv`" + ` produces
+a comma list, ` + "`decision-map`" + ` a flow mapping.
+
+Where the map cannot answer — an unmapped capability, an undeclared harness, or a
+map that is absent or schema-invalid — this exits non-zero and writes nothing to
+stdout. It never falls back to a default: an empty tools list is not "no opinion",
+it is a definition granting nothing.`,
+		Example: `  dotf harness resolve-capabilities read,search,edit,shell --harness claude
+  dotf harness resolve-capabilities read,web --harness opencode`,
+		Args:         cobra.ExactArgs(1),
+		SilenceUsage: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if harnessName == "" {
+				return fmt.Errorf("--harness is required: a capability means different native tools " +
+					"per harness, so resolving one without naming the harness has no answer")
+			}
+			root := repoRoot
+			if root == "" {
+				root = env.ResolveHarnessRoot()
+			}
+			m, err := harness.LoadCapabilityMap(root)
+			if err != nil {
+				return err
+			}
+			line, err := harness.ResolveCapabilities(m, strings.Split(args[0], ","), harnessName)
+			if err != nil {
+				return err
+			}
+			_, _ = fmt.Fprintln(cmd.OutOrStdout(), line)
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVar(&harnessName, "harness", "", "harness the capabilities are resolved for (required)")
+	cmd.Flags().StringVar(&repoRoot, "repo-root", "",
+		"root containing harness/capability-map.json (default: the checkout, else DOTFILES_DIR)")
+	return cmd
+}
--- a/cli/internal/cmd/harness_resolve_tier_test.go
+++ b/cli/internal/cmd/harness_resolve_tier_test.go
@@ -200,29 +200,32 @@ func TestHarnessResolveTierFailsLoudWithoutAMap(t *testing.T) {
 	}
 }
 
-// TestHarnessHelpListsResolveTier pins the capability probe in
+// TestHarnessHelpListsSubcommands pins the capability probe in
 // scripts/compile-harness.sh, which decides whether the dotf on PATH is new
-// enough to resolve a tier by grepping `dotf harness --help` for this exact
-// subcommand name.
+// enough by grepping `dotf harness --help` for a subcommand name.
 //
 // The probe exists because the exit status cannot answer the question: a dotf
-// predating the subcommand rejects `--harness` with exit 1, which is
-// indistinguishable from a genuine routing refusal (#1158). Renaming the
-// subcommand without updating that grep would make the probe answer "too old"
-// forever, silently degrading every agent render to no model line — the exact
-// failure this feature removed. This test is the tripwire.
-func TestHarnessHelpListsResolveTier(t *testing.T) {
+// predating a subcommand rejects its flags with exit 1, which is
+// indistinguishable from a genuine refusal (#1158). Renaming one of these
+// without updating that grep would make the probe answer "too old" forever,
+// silently degrading every agent render — a guard that fails OPEN and reports
+// health. This test is the tripwire, and it covers BOTH registries' consumers
+// because the shell now probes for each independently.
+func TestHarnessHelpListsSubcommands(t *testing.T) {
 	stdout, stderr, err := captureRealStreams(t, "harness", "--help")
 	if err != nil {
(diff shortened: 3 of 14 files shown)


mlorentedev:
## Review triage

Two reviewer sources, and they converged on the two findings that mattered — which is the argument for running both.

| Source | Verdict at `57b9f56` |
|---|---|
| **CodeRabbit** | 6 actionable comments |
| **Independent adversarial** (`nan/deepseek-v4-flash`, from `harness/reviewer-pool.json`) | **FAIL** — 2 Major REAL, 3 Minor THEORETICAL, 1 Question REAL |

Every finding was **reproduced before being fixed**. All 6 CodeRabbit findings were real; nothing was declined.

### The security one, and it was the sharpest thing either reviewer found

CodeRabbit flagged that the schema let `field`, `grant` and native names carry any non-blank characters. Reproduced:

```console
$ # native name "Read,Bash"
tools: Read,Bash                    # ← Bash granted to an agent that asked only for `read`
$ # native name "read},{bash"
permission: {read},{bash: allow}    # ← broke out of the flow mapping entirely
```

The first is **privilege escalation through data**. The render escapes nothing by design — a frontmatter line is not a shell command — so the *contract* is where it has to be stopped. `field`, `grant` and every native name are now identifier tokens, and `TestCapabilityMapRejectsRenderBreakingNames` keeps it there (four cases, including the field-name variant CodeRabbit did not name).

### The shell-law one, found independently by both

`caps="${caps#[}"` is an unescaped bracket opening a pattern that never closes. Bash tolerates it; **zsh aborts at runtime** with `bad pattern: [`. Measured:

```console
bash  read,search,edit
zsh   zsh:1: bad pattern: [
```

`zsh -n` does **not** catch it — it is a pattern error, not a syntax error — so the syntax check this repo leans on gave false confidence. Brackets escaped, and the guard is a bats case that *executes* the expansion under both shells and compares output, rather than adding another lint that would miss it the same way.

### Full disposition

|---|---|---|---|
| 1 | Native names / field / grant can alter the rendered value | CodeRabbit | **Applied** — schema restricts to identifier tokens; guard test added |
| 2 | `grant` should be schema-required for `decision-map` | CodeRabbit | **Applied** — fails at load, not at resolve, so an unusable map never reaches a render |
| 3 | Unescaped bracket breaks under zsh | both | **Applied** — escaped; cross-shell behavioural guard added |
| 4 | `ResolveCapabilities` (69) and `checkVocabularyCoverage` (51) over the repo's <40 rule | CodeRabbit | **Applied** — split into `harnessBlock` / `collectNatives` / `renderNatives` and `checkHarnessCoversVocabulary`; every function now under 30 |
| 5 | "every OTHER agent deployed" is false when none did | CodeRabbit | **Applied** — counts what actually rendered; that sentence is what a rollback decision rests on |
| 6 | `verification.md` quoted the bats stub's answer as the real pipeline's | both | **Applied** — corrected to the shipped record and checked against the resolver's real output |
| 7 | Block-style `capabilities:` silently ignored | adversarial | **Applied** — this is the fail-OPEN direction: no capability line does not deny, it grants the harness default set. Now refused loudly, naming the inline spelling |
| 8 | The unavailable-resolver degrade is fail-open for an allow-list | adversarial | **Applied** — AC8 inherited #1165's degrade decision without noticing the blast radius inverts when the field is a permission. The WARN now states the direction, and a test asserts the wording, because the wording *is* the mitigation |
| 9 | Schema gaps reachable only by future hand-authored maps | adversarial | **Closed by 1 and 2** |
| 10 | Production diff ~609 lines vs ADR-017's ~300 | adversarial (Question) | **Disclosed, not silently accepted** — already flagged in the PR body with a per-file breakdown. Splitting would break the rule this sequence established (no declaration without its consumer), which is exactly how `model-map` came to merge with zero readers. I cut what was genuinely separable — the doctor checks — rather than halving one idea. **Human call.** |

### Re-review

The FAIL was at `57b9f56`; the fixes are `b1781ac` and `0893462`, so `reviewed_sha` is stale by construction and a fresh round is running against the current head. `spec-gate` stays red until that lands and the spec archives — the gate correctly refusing a PR that closes #560 with its spec still active.

**Checks:** everything else green. `bats tests/*.bats` 1432 passing / 0 failing; `golangci-lint` 0 issues at the pin; `shellcheck --severity=error` (CI's level) exit 0.

PR #1180, opened 2026-08-22
docs(agents): the PR size cap counts executable lines

mlorentedev:
Writes down a rule that was being applied without ever having been decided.

## The gap

`AGENTS.md`'s Discipline Gate said:

> ~50–300 LOC of production diff (excluding tests, generated files, lockfiles)

The exclusion list exists, but nothing said what the **remaining** lines are measured *for*. So declarative data, schemas and comment blocks counted the same as branching logic — and a registry whose rationale lives in a `$comment` block could trigger a split that helps no reviewer.

It came up concretely on #1172, where the independent adversarial review raised it as a **Question marked "measurement-basis question for the human"**: 520–609 lines of "production" diff, ~200 of which were a JSON registry and its schema.

## The rule

> **The cap counts EXECUTABLE lines.** Declarative data, schemas and comment blocks are excluded: the cap rations the control flow a reviewer holds in their head, and a table is read as a table.

This is the principle the *existing* exclusions already embody — tests, generated files and lockfiles are excluded for exactly this reason — applied to the cases nobody had written down.

Two guards against the obvious misreadings, both in the text:

- **Excluded is not free.** The breakdown is declared in the PR body whenever total added lines exceed the cap, so exclusion is a visible decision rather than a silent assumption. Same posture as the existing over-cap rule.
- **Excluded is not "cheap to review".** A registry where one wrong entry silently grants or denies deserves *more* care than most code, not less — #1172's capability map is the worked example. Exclusion bounds the reviewer's *reasoning load*; it says nothing about how carefully the data deserves reading.

The vault pattern also records the third-order effect: this exclusion incentivises pushing logic out of code and into declarative data, which points the right way only in a codebase that has already chosen "a declared map, never a director model" (ADR-035 §4) — and is a hole to watch elsewhere.

## Where it lives

- **SSOT:** `00_meta/patterns/pattern-git-workflow.md` §10, committed to the vault separately (vault changes go direct to master).
- `harness/enforced/pr-sizing.md` is its refreshed record.
- The `AGENTS.md` prose is hand-written, because `pr-sizing` is **opt-out** for this file — per the manifest's own note that this file already states the policy in prose.

## Testing

- `./scripts/compile-harness.sh --check` — `OK: no harness drift`
- `bats tests/*.bats` — **1430 passing, 0 failing**
- The doctrine payload's character cap still holds (the refreshed record is deliberately terse for that reason — the fuller explanation lives in the uncapped `AGENTS.md` prose and the vault pattern)

Unrelated skill-record drift surfaced by `--refresh` was **reverted out of this PR** rather than bundled; it is its own change.

## Summary by CodeRabbit

* **Documentation**
  * Clarified pull request size limits to count executable lines while excluding declarative data, schemas, and comments.
  * Added disclosure requirements for changes exceeding the limit.
  * Documented verification guidance for oversized lockfile differences, including version filtering and SHA-256 integrity checks.


mlorentedev:
## Review triage

**Checks:** green except `review-attestation`, which is CodeRabbit rate-limited (see below).

|---|---|---|
| **PR-Agent** (`github-actions`) | Effort 1/5, no security concerns, no major issues. Ticket compliance vs **#1172** graded *partially compliant*, listing #1172's own requirements as unmet | **Declined — false positive, and the tool says so itself.** Under *"Requires further human verification"* it writes: *"This PR does not appear to target ticket #1172. It is a separate documentation change about PR size cap measurement, inspired by a finding on #1172."* That is exactly right. #1172 is cited as **provenance** — its adversarial review raised the measurement question this PR answers — not as the ticket this closes. Nothing to apply |
| **CodeRabbit** | **Did not review** — rate limited | **Disclosed, not dispositioned.** A notice that no review ran leaves this half unreviewed |

**So this PR carries one reviewer's verdict.** Proceeding is allowed; proceeding silently is not, hence this note.

**Worth a human eye specifically:** this changes a *policy* — what the ~300 LOC cap counts. It is deliberately narrow (executable lines; data, schemas and comments excluded but **declared**, and explicitly *not* "cheap to review"), but it is the kind of rule that gets quoted back later, so the wording matters more than the diff size suggests.

PR #1181, opened 2026-08-22
fix(harness): make the compact doctrine payload actually compact

mlorentedev:
"Compact doctrine payload" described an intention, not a behaviour.

## The bug

`deploy_doctrine` calls `render_region`, which **concatenates the enforced records verbatim**. So the payload grows by the full size of every id added to `doctrine.inject`, and the character cap is breached the moment one is.

Measured 2026-08-22: `.gemini/GEMINI.md` at **12114 characters against a 12000 cap** — with a deploy-time `WARN` nobody reads in a stream of deploy output, and a bats assertion that only fails later, in CI.

This is not a one-off. Every future doctrine id repeats it.

## The fix

`render_region_compact` drops the provenance blockquote each record opens with:

> *Injected verbatim into every agent's instructions (harness `enforced` id `definition-of-done`)…*

That line tells a **human** reading `AGENTS.md` where the block came from. An agent consuming a capped payload needs the rules, not their filing history. It is **737 characters** across the three records that carry one — and `AGENTS.md` keeps all three, verified.

**It is not a summariser.** Every rule survives byte for byte; only meta-text about *where rules live* is removed, and only lines it can identify exactly. Compacting enforcement prose by paraphrase is how a rule quietly loses its teeth.

| | chars | vs cap |
|---|---|---|
| before | 12114 | **over** |
| after | **9928** | ~2000 headroom |

## The warning also stopped conflating two causes

The cap covers the whole file, **the user's own content included** — so one number could not distinguish:

- *"your file is large"* — theirs to fix by trimming their prose;
- *"our generated doctrine no longer fits on its own"* — **ours**, and it means the next doctrine id pushes rules past the point the agent ever reads.

It now reports the generated share and says which case it is, with the command to find the biggest records.

**Still a warning, not a failure.** Making it fatal would break the bootstrap of anyone whose own instructions file is large — the shape that turns a guard into a liability.

## Verification

- Rule-by-rule check on the deployed file: every committed doctrine heading still present (`No AI attribution`, `English only`, `Auto-merge is forbidden`, `Working code is not a finished change`, `What binds is the disposition`, `The transcript is a durable artifact`, `Atomic PRs`)
- `AGENTS.md` still carries all 3 provenance blockquotes — the compaction is scoped to the capped surface only
- `bats tests/*.bats` — **1430 passing, 0 failing**, including the HARNESS-056 cap assertion that was red
- `shellcheck --severity=error`, `bash -n`, `zsh -n` — clean
- `./scripts/compile-harness.sh --check` — `OK: no harness drift`
- `go test ./...` — clean

## Note for whoever is adding the IaC doctrine

There is uncommitted work in the shared checkout adding `harness/enforced/iac-and-idempotence.md` (792 chars) to `doctrine.inject`. **That is what pushes the file to 12114 today.** With this PR the payload sits at 9928, so that record lands at ~10720 — comfortably inside the cap. This unblocks it rather than competing with it.


mlorentedev:
## Review triage

**Checks:** green except `review-attestation` (CodeRabbit rate limited).

|---|---|---|
| **PR-Agent** (`github-actions`) | Effort 2/5, **no security concerns, no major issues detected**, ticket compliance clean | **Nothing to apply** |
| **CodeRabbit** | **Did not review** — rate limited | **Disclosed, not dispositioned** |

**One reviewer's verdict, not two.** Stated rather than implied.

**What deserves the human eye here**, since neither bot flagged it and it is the only real risk in the change: the compaction removes lines from a payload that carries *enforcement doctrine*. It deletes only the provenance blockquote (`> Injected verbatim into every agent's instructions…`) and only lines it matches exactly — **no paraphrase, no summarisation** — because compressing normative prose is how a rule quietly loses its teeth.

That claim is checked rather than asserted: the PR body lists a rule-by-rule presence check on the deployed file, and `AGENTS.md` keeps all three provenance blockquotes, so the compaction is scoped to the capped surface alone.

PR #1182, opened 2026-08-22
fix(doctor): key each doctrine marker on its own rule, not another's prose

mlorentedev:
Regression from #1181, found by running `dotf doctor` against the deployed tree rather than by reading the diff.

## What broke

After deploying current `main`, doctor reported three failures:

```
[FAIL] enforced region "definition-of-done" missing from deployed .gemini/GEMINI.md
[FAIL] enforced region "definition-of-done" missing from deployed .codex/AGENTS.md
```

**The doctrine was entirely intact.** `Working code is not a finished change` and every numbered item were present in all three files.

## Why

`enforcedRegionMarkers` answers *"did region X reach this file?"* by searching for a phrase. The phrase for `definition-of-done` was `"Definition of Done"` — which appears **nowhere in that record**. Its only occurrence in the whole enforced set:

```
harness/enforced/pr-stewardship.md:2:
> Injected verbatim into every agent's instructions (harness `enforced` id `pr-stewardship`).
  It elaborates Definition of Done §4 — "an open PR is not finished work" — into …
```

So the check verified **one region's presence by finding another region's meta-text**, and had done so since it was written. #1181 compacts provenance blockquotes out of the capped payload — by design, they are filing history rather than rules — and the proxy vanished.

A check reporting three deployment failures for a deployment that succeeded.

## The fix

The marker is the rule's own opening line: `Working code is not a finished change`.

I audited all seven. **The other six already came from their own records** — `definition-of-done` was the only one keyed on someone else's prose.

## Two guards, because the class is the point

- **`TestEveryDoctrineMarkerIsInItsOwnRecord`** — every marker must appear in the record it names. Verifying a thing by a proxy that lives somewhere else is the failure this repo keeps re-measuring in new costumes; this is the version of it that hid in a lookup table.
- **`TestDoctrineMarkersSurviveCompaction`** — no marker may sit inside a provenance blockquote, since `render_region_compact` is *designed* to remove those. Keying on a line the renderer deliberately deletes is a check that reports failure for success.

Both fail before the fix and name the cause.

## Verification

- `dotf doctor` before: 4 FAILs. After: 2, both benign and unrelated (`installed=dev` because I build from source, and `~/.dotfiles` lagging the repo)
- `go test ./... -count=1` — clean, including the existing `TestCheckDeployedDoctrine` whose fixture carried the old marker string and is updated
- `go build`, `go vet`, `GOOS=windows go vet`, `golangci-lint run` (`0 issues.`), `gofmt` — clean
- `bats tests/*.bats` — 0 failures

Not SDD-gated: a one-line lookup-table correction with a measured cause, plus tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the enforced completion guideline to “Working code is not a finished change.”
  * Improved doctrine validation to ensure required markers remain present and independently recorded after content compaction.
  * Updated deployed-doctrine checks to correctly recognize both complete and incomplete doctrine content.

* **Tests**
  * Added coverage for marker placement and preservation during compaction.

### DIFF
--- a/cli/internal/doctor/checks_deploy.go
+++ b/cli/internal/doctor/checks_deploy.go
@@ -348,9 +348,16 @@ var enforcedRegionMarkers = map[string]string{
 	"english-only":        "English only",
 	"no-phase-references": "No internal phase/milestone references",
 	"no-auto-merge":       "Auto-merge is forbidden",
-	"definition-of-done":  "Definition of Done",
-	"pr-stewardship":      "What binds is the disposition",
-	"pr-sizing":           "Atomic PRs, ~300 LOC hard cap",
+	// The rule's own opening line, not the phrase "Definition of Done" — which
+	// appears nowhere in this record. Its only occurrence in the enforced set was
+	// inside pr-stewardship's provenance blockquote ("It elaborates Definition of
+	// Done §4 …"), so this check verified one region by finding another's
+	// meta-text, and broke the moment #1181 compacted those blockquotes out of
+	// the capped payload while the doctrine itself was entirely intact.
+	// TestEveryDoctrineMarkerIsInItsOwnRecord keeps every marker honest.
+	"definition-of-done": "Working code is not a finished change",
+	"pr-stewardship":     "What binds is the disposition",
+	"pr-sizing":          "Atomic PRs, ~300 LOC hard cap",
 }
 
 var deployedDoctrineTargets = []struct {
--- a/cli/internal/doctor/checks_deployed_doctrine_test.go
+++ b/cli/internal/doctor/checks_deployed_doctrine_test.go
@@ -21,7 +21,7 @@ func TestCheckDeployedDoctrine(t *testing.T) {
 - English only in git/GitHub artifacts
 - No internal phase/milestone references
 - Auto-merge is forbidden in every repository
-- Definition of Done
+- Working code is not a finished change
 - What binds is the disposition, not the waiting (PR Stewardship)
 - Atomic PRs, ~300 LOC hard cap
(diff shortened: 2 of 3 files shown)


mlorentedev:
## Review triage

|---|---|---|
| **PR-Agent** (`github-actions`) | Effort 2/5, **PR contains tests**, no security concerns, **no major issues detected**. Ticket compliance vs **#1181** graded *partially compliant*, listing as unmet: *"Does not implement `render_region_compact`"*, *"Does not update the cap warning"* | **Declined — false positive, same shape as #1180's.** #1182 does not implement `render_region_compact` **because #1181 already did**; this PR fixes a regression that change caused. #1181 is cited as the *origin of the defect*, not as the ticket this implements. The third item (*"does not include the rule-by-rule verification"*) is also inverted: that check is what **found** this bug, and its results are in the PR body |
| **CodeRabbit** | **Did not review** — rate limited, third PR in a row | **Disclosed, not dispositioned.** A notice that no review ran leaves this half unreviewed, which is why `attestation` is red |

**One reviewer's verdict, not two.** Stated rather than implied.

### What a human should actually look at

Neither bot flagged the thing that matters, so naming it here:

The marker moved from `"Definition of Done"` to `"Working code is not a finished change"`. **Both are prose that could be reworded by a future vault edit** — the new one is simply the rule's own opening line rather than a sentence *about* the rule living in a different record. The durable protection is not the string choice but `TestEveryDoctrineMarkerIsInItsOwnRecord`, which fails if any marker ever again points outside the record it names.

Worth confirming that reading, because it is the difference between "fixed this instance" and "closed the class".

PR #1184, opened 2026-08-22
fix(ci): a reviewer whose quota is exhausted must not refuse the change

mlorentedev:
The headache, fixed at the root.

## What was happening

Across one full session on 2026-08-22, `declined` was the **first verdict on every PR opened**, flipping green ~90 seconds later when PR-Agent posted. CodeRabbit was rate limited on all of them.

This workflow's own comment already names why that matters:

> *"A gate whose normal state is red has stopped being a gate."*

It was becoming true of this one.

## Why it was measuring the wrong thing

There are two different facts wearing the same word:

| Reviewer says "I could not review" | What it is |
|---|---|
| a **counting** reviewer | a fact about *this change's* review — actionable |
| **CodeRabbit** | a fact about a *third party's billing state* — not actionable by the author, and it does not make the change worse |

Codex is in the same position, which is why this is a property of the entry rather than a special case.

## The change

`advisory: true` on a registry entry scopes the decline:

- the reviewer's **review still attests** — a CodeRabbit review that actually happens is as good an attestation as any;
- its **decline no longer refuses**.

`coderabbitai` is the first entry to carry it.

## What is NOT weakened — check this rather than trust it

An advisory decline still prevents `attested`. It yields **`pending`**, which is the honest reading: *a reviewer that could not run has not refused the change*. The three real captured rate-limit payloads (#1009, #1007, #1013) **still exit 1**:

```
before: [FAIL] declined — coderabbitai posted a notice that no review ran
after:  [FAIL] pending  — coderabbitai could not review (advisory: its availability is a
                          third-party quota, not a property of this change, so it does not
                          refuse). No counting reviewer has attested yet.
```

What changed is the sentence, and the sentence is now true. The PR still cannot read as reviewed.

It also stays **distinguishable from silence**, with its own message — an author must not go hunting for a reviewer that already answered. A test asserts the two messages differ, because collapsing them throws away the one thing the notice is still good for.

## Guards, so this tunes the gate rather than deleting it

- **A non-advisory decline still refuses hard.** Builds a counting-reviewer config from the shipped registry and re-runs the real payload.
- **An advisory reviewer's real review still attests.**

## One existing test was re-pointed, not left passing

`"AC3: the declined message names both the reviewer and the marker"` ran against `coderabbitai`. After this change it would have **silently become a test of the advisory branch** — green, while asserting nothing about the thing its name claims. That is the vacuous-test shape this repo has measured before, so it now builds a counting-reviewer config, and the advisory path got its own preserve-why test.

## Considered and deliberately NOT included

- **"`declined` should only be terminal when no other counting reviewer is outstanding."** A good rule, but with CodeRabbit advisory and PR-Agent carrying no declined markers, nothing would exercise it today. Adding it now is the *declaration with no consumer* trap. It belongs with the second counting reviewer.
- **Declined markers for PR-Agent.** Already covered elsewhere: #1109 makes the `pr-agent` job itself fail when it publishes no review (*"PR-Agent reported success but published no review"*), so that failure is already loud — in the job that owns it.

## Testing

- `bats tests/*.bats` — **1437 passing, 0 failing**
- `shellcheck --severity=error scripts/*.sh setup-linux.sh` — exit 0; `bash -n` and `zsh -n` clean
- `check-bats-names.sh` clean; registry validates as JSON and against the script's own shape check

### DIFF
--- a/harness/review-attestation.json
+++ b/harness/review-attestation.json
@@ -68,7 +68,9 @@
     "One string, two consumers, declared once. The skill writes it and the queue",
     "reads it, and putting it in either would have made the pair a two-file",
     "agreement nobody checks — the failure class this repo spent 2026-08-17 into",
-    "the 18th cataloguing."
+    "the 18th cataloguing.",
+    "",
+    "`advisory: true` marks a reviewer whose REVIEW attests but whose DECLINE does not refuse. It exists for a reviewer whose availability is a third party's billing state rather than a property of the change. An advisory decline is reported — silence and 'it could not run' are still different facts — but it yields `pending` rather than `declined`, so the gate refuses for the honest reason: no review has happened yet. Absent or false, a reviewer counts fully."
   ],
   "reviewers": [
     {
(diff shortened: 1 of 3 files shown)


mlorentedev:
## Review triage

|---|---|---|
| **PR-Agent** (`github-actions`) | Effort 2/5, **PR contains tests**, no security concerns, **no major issues detected**, ticket compliance clean (no non-compliant items) | **Nothing to apply** |
| **CodeRabbit** | Rate limited — the very condition this PR is about | **Disclosed.** Its quota returned later in the session and it reviewed #1182, so it is not permanently down; it is *intermittent*, which is precisely the argument for `advisory` rather than removal |

### The `attestation` check on this PR is red, and that is expected

Two distinct signals, and only one is authoritative:

| Signal | State | Revisable after the fact? |
|---|---|---|
| `review-attestation` (**commit status**) | ✅ green — *attested by github-actions* | yes, later runs revise it |
| `attestation` (**check-run**) | ❌ red | **no** — a check-run belongs to the run that created it |

The run at PR-open produced `declined` (CodeRabbit's notice landed within seconds), and the workflow's softening covers only `pending`:

```sh
if [ "$CODE" != "0" ] && [ "$KIND" = "pending" ] && [ "$EVENT" = "pull_request" ]; then exit 0; fi
```

So the check-run froze red — the exact failure the workflow's own comment describes: *"this check-run froze red on every correct PR, permanently, and taught its readers to scroll past the attestation line."*

**This PR is the fix for that.** With `coderabbitai` advisory, the early verdict becomes `pending` rather than `declined`, the softening applies, and the check-run stops freezing.

Like the `workflow_run` trigger before it (#1052 AC1), **it cannot be exercised by the PR that introduces it** — its own check-run was created before the change existed on the default branch. Merging it is what proves it, and the next PR opened is where it shows.

Issue #1128, opened 2026-08-21
GUARD: a review requested with /review never updates the attestation, because the workflow_run carries master's SHA


mlorentedev:
Derived cross-repo from `mlorentedev/kubelab#1184` and verified against this repo's own workflows. Latent here, not observed — the automatic path is unaffected.

## The chain

1. A `/review` comment fires `pr-agent.yml` via `issue_comment: [created]` (`.github/workflows/pr-agent.yml:36`).
2. A run triggered by `issue_comment` is associated with the **default branch**, not the PR head.
3. When it completes, `review-attestation.yml` fires via `workflow_run` — the trigger that exists precisely because GitHub suppresses workflow events for comments created with `GITHUB_TOKEN`, so PR-Agent's own comment cannot wake the gate (measured 2026-08-18 on #1047: its two comments triggered zero runs).
4. That `workflow_run` payload arrives with `pull_requests[]` empty **and `head_sha` = the default branch's SHA**.
5. The resolve step falls back to `repos/.../commits/{sha}/pulls`, asks which PRs contain a master commit, gets none, and sets `skip=true`.

Step 5 is correct and deliberate — a run with nothing to judge must not publish a green status for a PR it never looked at. **The defect is upstream of it: the PR identity was lost at the trigger, not at the handoff.** No later resolution layer can recover it, and writing a commit status does not help either, because there is no PR SHA to write it on.

## Consequence

`/review` on an open PR produces a real review that the attestation gate never sees. The verdict stays at whatever the last `pull_request`-triggered run computed. Since GUARD-002 is heading for `required` branch protection (the promotion order is documented in `review-attestation.yml`), this would mean the manual re-review escape hatch cannot clear the gate it exists to clear.

## Why it is latent rather than live

PR-Agent runs automatically on `pull_request: [opened, synchronize, reopened, ready_for_review]`, and those runs carry the PR head SHA, so the ordinary path resolves correctly. Only the on-demand slash command is affected.

## What this ticket does not decide

The fix. Plausible shapes, cheapest first — recovering the PR number from the triggering comment and passing it forward through the `workflow_run` handoff (an artifact, or re-resolving from `issue.number` at attestation time), versus not routing on-demand reviews through `workflow_run` at all. Both need thought about which one keeps the "nothing to judge must not publish" invariant intact, which is the property that makes step 5 correct today.

## Cross-repo note

[...]

PR #1185, opened 2026-08-22
fix(ci): the attestation check-run must not carry a verdict that will change

mlorentedev:
Refs #1052 — advances it, does not close it (see the correction below). Follows #1184, which made the symptom rare; this closes the shape.

## The structural defect

**A check-run cannot be revised.** It belongs to the run that created it, and runs from `issue_comment` and `workflow_run` attach to the **default branch**, so they cannot reach it either.

On the `pull_request` event, **no verdict is final** — the reviewers have not run yet, by construction. Publishing any failure there writes a provisional value into an immutable slot.

#1041 named exactly this and fixed it **for `pending` only**. That narrowed the window instead of closing it.

## Measured, not inferred

Across a full session on 2026-08-22, `declined` was the **first verdict on every PR opened** — CodeRabbit's rate-limit notice lands within seconds. Each froze red permanently while the commit status beside it went green:

```
07:29:14  pull_request  failure   ← frozen forever
07:29:30  workflow_run  success   ← the truth, on the revisable signal
```

That is *"a gate whose normal state is red has stopped being a gate"* — the workflow's own words — reached through the one kind the exemption did not cover.

## Why this and not a delay, a timeout, or a `depends on`

All three were considered:

| Proposal | Why not |
|---|---|
| **sleep / delay** | guesses a duration (PR-Agent ~90s, variable), burns CI minutes on every PR, and a slow run still freezes. It moves the race rather than removing it |
| **`depends on`** | cannot cross workflows. The cross-workflow dependency already exists and works: the `workflow_run` trigger on `pr-agent` |
| **timeout** | inapplicable — the job does not hang, it answers too early |

The defect is not timing. It is a mutable value in an immutable slot, so the fix is to stop putting it there.

## Nothing is softened — check this rather than trust it

The verdict is still published to the **`review-attestation` commit status**: revisable by every later run, and the signal branch protection must require. This step now decides only whether an *unrevisable* check-run carries a value expected to change.

**Every other event still fails.** By then a reviewer has had its turn, so the verdict means something.

## The #1041 tests are rewritten, not deleted

The invariant moved; it did not disappear. The old pair asserted *"the exemption names `pending`, and must never name `declined`"* — which this change makes false by design, so leaving them would have blocked a correct fix, and deleting them would have dropped the protection.

The three replacements assert what actually protects the gate:

1. the exemption is guarded on the **event**;
2. the **unconditional refusal survives** after it — losing that guard would widen the gate into silence;
3. the **publish step still exists** — the exemption is only defensible because the commit status carries the verdict, and if that went away, softening the check-run would become plain silence.

## #1052: advanced, NOT closed — a correction

I first wrote `Closes #1052` here. **That was wrong, and `spec-gate` caught it**: #1052 carries an active spec, `GUARD-003-review-loop-determinism`, with **nine** acceptance criteria. I verified one (the `workflow_run` trigger fires) and concluded the ticket was done.

That is precisely the shape [lesson 220](https://github.com/mlorentedev/dotfiles/blob/main/docs/lessons/lesson-220-four-defects-one-shape-a-thing-verified-by-a-proxy.md) describes — *a thing verified by a proxy that lives somewhere else* — committed by the person who wrote it, hours later.

What today's evidence actually establishes:

| AC | State |
|---|---|
| **AC1** re-evaluates with no human action, verified live | ✅ `#1182 attestation run: event=workflow_run conclusion=success` |
| **AC2** the status carries the latest verdict, not the first | ✅ observed going green after PR-Agent posted |
| **AC3** declined/never-ran still ends red after the trigger | ✅ preserved by this PR — every non-`pull_request` path still fails |
| **AC4** re-evaluates the PR the reviewer actually ran on | ❌ **#1128 is the open counterexample**: a `/review` run carries master's SHA |
| AC5-AC9 | guard exists; `dotf pr triage-queue` ships — not audited here |

So the PR now says **`Refs #1052`**. AC4 is exactly #1128, which this PR deliberately does not fold in.

## Considered and NOT folded in

**#1128** — *"a review requested with `/review` never updates the attestation, because the workflow_run carries master's SHA"*. Same file, same class, and the obvious neighbour. But its own ticket says **"What this ticket does not decide: the fix"**, leaving two shapes open that need thought about which preserves the "nothing to judge must not publish" invariant. Folding an undecided design into a PR about check-run semantics would make both harder to review. It is also **latent, not observed** — only the on-demand slash command is affected.

## Testing

- `bats tests/*.bats` — **1438 passing, 0 failing**
- workflow parses as valid YAML; `check-bats-names.sh` clean
- No untrusted event payload is interpolated: every value comes through `env:` from `steps.*.outputs`

### DIFF
--- a/.github/workflows/review-attestation.yml
+++ b/.github/workflows/review-attestation.yml
@@ -251,12 +251,44 @@ jobs:
           # revise and which is the signal branch protection must require.
           # `declined` still fails here and immediately, because a reviewer
           # saying it could not review is actionable the moment it is said.
-          if [ "$CODE" != "0" ] && [ "$KIND" = "pending" ] && [ "$EVENT" = "pull_request" ]; then
+          # WIDENED from `pending` to EVERY kind, and the reason is structural
+          # rather than a preference about strictness.
+          #
+          # A check-run cannot be revised: it belongs to the run that created it,
+          # and runs from issue_comment and workflow_run attach to the default
+          # branch, so they cannot reach it either. On the `pull_request` event
+          # NO verdict is final — the reviewers have not run yet, by
+          # construction. Publishing any failure here writes a provisional value
+          # into an immutable slot, which is the defect #1041 named; restricting
+          # the exemption to `pending` narrowed the window instead of closing it.
+          #
+          # Measured 2026-08-22: `declined` was the first verdict on every PR
+          # opened in a full session, because CodeRabbit's rate-limit notice
+          # lands within seconds. Each one froze red permanently while the commit
+          # status beside it went green — "a gate whose normal state is red has
+          # stopped being a gate", arrived at through the one kind the exemption
+          # did not cover.
+          #
+          # #1184 made that particular verdict rare by marking CodeRabbit
+          # advisory. This closes the shape rather than the instance: any future
+          # counting reviewer that answers fast would reopen it.
+          #
+          # NOTHING IS SOFTENED. The verdict is published to the
+          # `review-attestation` COMMIT STATUS in the step above — revisable by
+          # every later run, and the signal branch protection must require. This
+          # step decides only whether an unrevisable check-run carries a value
+          # that is expected to change. Every other event still fails: by then a
+          # reviewer has had its turn, so the verdict means something.
+          if [ "$CODE" != "0" ] && [ "$EVENT" = "pull_request" ]; then
             echo "$SUMMARY"
             echo
-            echo "Not failed here: no reviewer has posted yet, which is the expected"
-            echo "state this early. The verdict lives on the 'review-attestation'"
-            echo "commit status, which later runs revise as reviewers arrive."
+            echo "Not failed here: this run fires as the PR opens, before any reviewer"
+            echo "has had a turn, so no verdict it computes is final. A check-run cannot"
+            echo "be revised, so it must not carry one that will change."
+            echo
+            echo "The verdict lives on the 'review-attestation' commit status, which"
+            echo "later runs revise as reviewers arrive — and which is the signal to"
+            echo "require in branch protection."
             exit 0
           fi
           if [ "$CODE" != "0" ]; then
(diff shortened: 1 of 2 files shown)


mlorentedev:
## Review triage

|---|---|---|
| **PR-Agent** (`github-actions`) | Ticket compliance vs #1052, #1184, #1041 — each *partially compliant* | **Mixed, itemised below** |
| **CodeRabbit** | Rate limited — the condition #1184 was about | **Disclosed, not dispositioned** |

### The one item that is a real design question

> *"Does not remove check-run from being created (only exempts it from failure)"*

Correct, and **deliberate**. Removing the job from the `pull_request` event would also remove the step that **publishes the initial commit status** — so the status would be *absent* until a reviewer event fires, and absent reads as *unknown*, which is worse than provisional.

The run must exist to seed the status. What must not happen is an **unrevisable** check-run carrying a verdict that will change. That is exactly the line this PR draws: the run stays, its failure does not.

### The convergence worth recording

PR-Agent lists **AC3 and AC4** of #1052 as unmet. On **AC4** it is right, and it reached that independently of me: *"The trigger fires for the PR the reviewer ran on"* is not satisfied, because a `/review` run carries master's SHA — which is **#1128**, open and deliberately not folded in here.

That agreement is why this PR was amended from `Closes #1052` to `Refs #1052` before merge. I had verified one criterion (the trigger fires) and concluded a nine-criterion ticket was done — the [lesson 220](https://github.com/mlorentedev/dotfiles/blob/main/docs/lessons/lesson-220-four-defects-one-shape-a-thing-verified-by-a-proxy.md) shape, committed by its author hours after writing it. **`spec-gate` caught it first**, by refusing a PR that closed an issue whose spec was still active.

On **AC3** it is wrong: *"a PR whose reviewer declined, or never ran, still ends red"* is preserved. The exemption is scoped to the `pull_request` event; every other path still exits non-zero, and a bats test asserts that guard survives.

### The rest

The remaining items under *non-compliant* for #1041 and #1184 describe what this PR **does** (*"Guard: bats test asserts pull_request path exits 0"*, *"Main fix: widens exemption to all non-zero codes"*) rather than gaps. Mislabelled, nothing to apply.

Issue #1186, opened 2026-08-22
GATE: the SDD LOC counter counts comment lines, so #1180's executable-lines rule is documented but not enforced


mlorentedev:
## Problem

#1180 landed the rule in `AGENTS.md` and in the vault pattern `pattern-git-workflow.md` §10:

> **The cap counts EXECUTABLE lines.** Declarative data, schemas and comment blocks are excluded: the cap rations the control flow a reviewer holds in their head, and a table is read as a table.

`scripts/check-spec-gate.sh` does not implement it. Its `_excluded()` is a **path-based** list — whole files by path or extension (`docs/*.md`, `tests/*`, `*_test.go`, lockfiles). It has no notion of counting executable lines *within* a file, so every comment line in a kept file counts as production diff.

## Measured

`.coderabbit.yaml`, introduced by the CodeRabbit config PR:

```
total lines:   110      <- what the gate counts
non-comment:    45      <- what the rule says to count
setting keys:   25
```

The gate refused at *"Production diff: 110 LOC (>= threshold 50)"* for a file that is **under the threshold by the repo's own merged rule**. The PR used the `skip-sdd` escape with a rationale naming this ticket, which is the honest compensation but not a fix.

## Why it matters beyond one PR

This repository writes dense WHY comments as a deliberate standard — `AGENTS.md`, the registries and the workflows all carry more rationale than code. Under a raw-line counter that standard is **taxed**: writing the reasoning down pushes a change over the threshold, and the cheapest way to pass the gate becomes writing less of it.

That inverts the incentive the rule exists to set. It is also the "declaration with no consumer" shape one more time — a rule that lives in the doctrine and in no executable check ([lesson 220](../docs/lessons/lesson-220-four-defects-one-shape-a-thing-verified-by-a-proxy.md)).

## Scope

Teach the counter what the rule says:

- Count added lines that are **not** blank and **not** a whole-line comment, per language (`#` for shell/YAML/TOML, `//` and `/* */` for Go, `<!-- -->` for Markdown).
- Treat pure-data files (`*.json`, `*.yaml`, `*.yml` with no executable content) as data rather than production, consistent with the existing lockfile exclusion — or count only their non-comment lines, whichever survives review.
- Report the breakdown in `--explain`, so an over-threshold PR shows raw vs counted and the author can see which it was.

## Acceptance

- A file that is 110 raw lines and 45 non-comment lines counts as 45.
[...]

Issue #1156, opened 2026-08-21
GUARD: the reviewer pool's only independent-provider fallback names a secret the registry does not declare, so the gate is single-provider


mlorentedev:
## What is true right now

`harness/reviewer-pool.json` declares three members:

| id | secret_id | declared in `secrets/registry.yaml`? |
|---|---|---|
| `nan/deepseek-v4-flash` (primary) | `NAN_API_KEY` | yes |
| `agy/gemini-3.1-pro-high` (fallback) | `GEMINI_API_KEY` | **no** |
| `nan/mimo-v2.5` | `NAN_API_KEY` | yes |

```
$ dotf secrets verify | tail -1
34 ok, 0 missing, 0 failed

$ dotf secrets ls | grep -i gemini
(no output)
```

So the fallback cannot run. `dotf spec review --reviewer agy/gemini-3.1-pro-high` has no credential
to inject.

## Why this is more than a missing key

That member is the pool's **only** entry from a second provider family, and its own `why` states
that this is the entire reason it exists:

> "A second provider family, independent of both NaN and Anthropic — provider diversity rather than
> merely a different model."

With it unusable, both remaining members draw on `NAN_API_KEY`. The archive gate — which refuses a
review signed outside the pool — is therefore **single-provider in practice while reading as
diverse on paper**. A NaN outage does not degrade the gate; it stops it, and the fallback that
exists to prevent exactly that is the thing that cannot fire.

This only surfaces during an outage of the primary, which is the worst moment to discover it.

## The class, and why this repo should care about it specifically

It is a **cross-registry reference that does not resolve** — the same defect
`x-poolReferencesResolve` was written to close *inside* `harness/model-map.json` (a harness naming a
pool the `pools` block never declares). The rule exists within one registry and nowhere between
two. `reviewer-pool.json` names `secret_id`s; `secrets/registry.yaml` declares them; nothing asserts
the first is a subset of the second.

`checkBWMapping` already establishes the pattern for the layer below — every Bitwarden item the
registry *names* must exist in the vault. The missing layer is one step up: every secret a registry
*references* must be declared in the registry.

## Fix shape

Two parts, and the second is the durable one:

1. Declare `GEMINI_API_KEY` in `secrets/registry.yaml` and populate it — or, if the fallback is not
   actually wanted, delete the member and say so in the pool's `$comment`. A pool entry that cannot
   run is worse than no entry, because the archive gate's `--reviewer` flag advertises it.
[...]

PR #1187, opened 2026-08-22
feat(ci): configure CodeRabbit instead of inheriting somebody else's defaults

mlorentedev:
This repository had **no `.coderabbit.yaml`**, so CodeRabbit ran on defaults. Defaults are not neutral — they are decisions somebody else made for a different repository.

## The one that mattered

```
auto_incremental_review: true    ← default: re-review after EVERY push
```

CodeRabbit's own docs:

> *"Automatic incremental reviews count toward the same per-developer PR review rate limit as manual PR reviews."*

**Measured 2026-08-22:** CodeRabbit was rate limited on *every* pull request opened across a full session.

### A correction, and PR-Agent earned it

This PR originally claimed it **stops** the exhaustion. That claim was too strong, and PR-Agent flagged it: *"Verify that disabling `auto_incremental_review` actually prevents quota exhaustion in production."* Measuring instead of inferring:

| | count |
|---|---|
| PRs **opened** that day | **17** — each one a first review, which this PR keeps |
| commits per PR | 1–3, with few pushes after open — so few incremental re-reviews |

So the dominant consumer was the **number of PRs opened**, not incremental re-review. Disabling incremental removes real waste at the margin; **it does not prevent exhaustion on a 17-PR day.**

What actually handles that is **#1184**, which made CodeRabbit advisory so an exhausted quota stops being treated as a refusal. That was the fix for the symptom; this is an efficiency improvement whose effect I overstated.

It still belongs: a re-review nobody asked for is spend with no return, and the setting was inherited rather than chosen. But the honest claim is *"stops paying for re-reads"*, not *"stops the exhaustion"*.

## Incremental off, first review kept — and the distinction is the point

Turning CodeRabbit off entirely would lose real value. Its best finding of that session — a **privilege escalation through data** on #1172, which the shipped tests, the schema *and* an adversarial review had all passed over — came from the **first** review of the PR, not from an incremental pass.

So: the first look stays automatic, re-reads become `@coderabbitai review`. That is upstream's own recommendation for repositories with frequent pushes.

## The rest is drawn from real configs, not invented

I surveyed `.coderabbit.yaml` in high-signal repositories via `gh search code`:

| Setting | Evidence |
|---|---|
| `profile: chill` | **crossplane** (12k ⭐) tested assertive and recorded *"approx 200 comments on one PR […] the signal to noise ratio isn't good enough"* |
| `drafts: false` | the most common setting among those surveyed — crossplane, MONAI (8.6k ⭐), biomejs, pingcap, wso2, clappr |
| `base_branches` | **pingcap/docs** records why: naming it keeps auto review working if the default branch is ever renamed |
| quota strategy | two exist in the wild — **pingcap/docs** throttles (`auto_pause_after_reviewed_commits: 3`), **infection** and **nextcloud/mail** disable auto-review entirely. This takes the middle: first review yes, incremental no |

## Summaries off — a decision this repo already made once, for the other tool

The PR-Agent workflow sets `auto_describe: false` because *"describe REWRITES the PR body […] a model rewriting them destroys exactly that"* — the bodies here carry hand-written measurement tables and before/after evidence. CodeRabbit brings the same threat through a different door as `high_level_summary`, and that door had no config.

`auto_apply_labels` and `suggested_labels` are off because labels are the board's business (ADR-018) — and because an auto-applied label is how #1180 acquired a `skip-sdd` that changed which gate rules applied to it.

## `path_instructions` encode what this repo keeps re-learning

- **`**/*.sh`** — must run under bash *and* zsh, including the unescaped-bracket pattern that aborts zsh at **run** time while `zsh -n` passes clean (found the hard way today).
- **`harness/*.json`** — a registry value that reaches rendered output must not be able to **alter** it (the #1172 escalation), and a new field whose absence reads as a default rather than a decision (#1156).
- **`.github/workflows/*.yml`** — untrusted event payload interpolated into `run:`, and a verdict published to a check-run later runs cannot revise (#1185).

## SDD skip rationale

`spec-gate` refused at *"Production diff: 110 LOC (>= threshold 50)"*. Measured:

```
total lines:   110      ← what the gate counts
non-comment:    45      ← what the rule says to count
setting keys:   25
```

**This file is under the threshold by the rule merged in #1180 hours ago** — *"the cap counts EXECUTABLE lines; declarative data, schemas and comment blocks are excluded"*. `check-spec-gate.sh`'s `_excluded()` is path-based and has no notion of counting within a file, so every comment line counts as production.

Filed as **#1186**. The label here compensates for a counter that has not caught up with the rule; it is not dodging the rule. A single YAML config file with 25 setting keys is not a change that wants a spec folder.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,110 @@
+# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
+#
+# CodeRabbit is the SECOND opinion here, not the first. PR-Agent is the automatic
+# reviewer (.github/workflows/pr-agent.yml, TOOL-013) and runs on every push;
+# CodeRabbit's job is the independent look, and its findings have earned that —
+# on #1172 it caught a privilege escalation through data that the shipped tests,
+# the schema and an adversarial review had all passed over.
+#
+# Every setting below exists because leaving it at its default was measured to
+# cost something. Defaults are not neutral: an unwritten config is a set of
+# decisions somebody else made for a different repository.
+
+language: "en-US"
+early_access: false
+
+tone_instructions: |
+  Be terse. Name the file, the line and the failing input. Prefer one reproducible
+  finding over three plausible ones. This repository values a claim you can run
+  over a claim that reads well.
+
+reviews:
+  # `chill`, not `assertive`. Measured upstream rather than assumed: crossplane
+  # tested assertive and reported "approx 200 comments on one PR […] the signal
+  # to noise ratio isn't good enough" (crossplane/crossplane, 12k stars). The
+  # same trade decided `auto_improve: false` in the PR-Agent workflow here.
+  profile: chill
+
+  # The PR bodies in this repository carry hand-written measurement tables, merge
+  # orders and before/after evidence — the part worth reading. This is the same
+  # decision already recorded for the other tool: the PR-Agent workflow sets
+  # `auto_describe: false` because "describe REWRITES the PR body […] a model
+  # rewriting them destroys exactly that". CodeRabbit brings the same threat
+  # through a different door.
+  high_level_summary: false
+  high_level_summary_placeholder: "@coderabbitai summary"
+  changed_files_summary: false
+  sequence_diagrams: false
+  poem: false
+
+  # Labels and reviewers are the board's business (ADR-018), not a reviewer's.
+  # An auto-applied label here is how #1180 acquired a `skip-sdd` that changed
+  # which gate rules applied.
+  auto_apply_labels: false
+  suggested_labels: false
+  suggested_reviewers: false
+
+  # A closed PR's review is spend with nothing to spend it on.
+  abort_on_close: true
+
+  auto_review:
+    enabled: true
+
+    # THE SETTING THIS FILE EXISTS FOR.
+    #
+    # Default is `true`: CodeRabbit re-reviews after every push, and its own docs
+    # state that "automatic incremental reviews count toward the same
+    # per-developer PR review rate limit as manual PR reviews".
+    #
+    # Measured 2026-08-22: CodeRabbit was rate limited on EVERY pull request
+    # opened across a full session. The work pattern here is many small pushes
+    # per PR, which is precisely the profile that burns the allowance — so the
+    # quota was spent on re-reviews nobody asked for, and gone by the time a
+    # human wanted a second opinion.
+    #
+    # The value is not lost by turning this off, and that is the point. Its best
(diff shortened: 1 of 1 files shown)


mlorentedev:
## Review triage

**Checks:** all green, including `spec-gate` (the `skip-sdd` escape with its rationale, per #1186).

|---|---|---|
| **PR-Agent** (`github-actions`) | Effort 1/5. Ticket compliance vs #1184, #1185, #1172 — *partially compliant*. One item under *requires further human verification* | **One applied, the rest declined — itemised below** |
| **CodeRabbit** | **Did not review** — rate limited, *"next review available in 2 minutes"* | **Disclosed.** Pointedly: the reviewer could not review the PR about it being unable to review |

### Applied — and it corrected a claim of mine

> *"Verify that disabling `auto_incremental_review` actually prevents quota exhaustion in production (requires observing rate-limit behaviour over a session)."*

**Correct, and the PR body overstated the effect.** I inferred the causal story from CodeRabbit's docs rather than measuring which consumer dominated. Measured now:

| | count |
|---|---|
| PRs **opened** on 2026-08-22 | **17** — each a first review, which this PR *keeps* |
| commits per PR | 1–3, with few pushes after open |

So the dominant consumer was **the number of PRs opened**, not incremental re-review. This change removes real waste at the margin; it does **not** prevent exhaustion on a 17-PR day. What handles that is **#1184** — making CodeRabbit advisory so an exhausted quota stops being read as a refusal.

The PR body is corrected to say *"stops paying for re-reads"* rather than *"stops the exhaustion"*. The change still belongs — a re-review nobody asked for is spend with no return, and the setting was inherited rather than chosen — but the claim now matches the measurement.

This is the same shape as [lesson 220](https://github.com/mlorentedev/dotfiles/blob/main/docs/lessons/lesson-220-four-defects-one-shape-a-thing-verified-by-a-proxy.md), on me for the second time today: I verified a *proxy* (the docs say incremental reviews consume quota) and asserted about the *subject* (therefore incremental reviews exhausted it).

### Declined — the familiar false positive

The remaining items list **#1184's** requirements as unmet by this PR:

- *"Add `advisory: true` to the CodeRabbit registry entry"*
- *"A non-advisory decline still refuses hard"*
- *"Re-point the existing test that would silently become vacuous"*

Those are #1184's, **already merged**. And *"No requirements from #1185 are addressed by this PR"* is true and not a finding — #1184 and #1185 are cited as the *preceding* work this completes from the other end, not as tickets this implements. Same shape as #1180 and #1182, where PR-Agent matched a ticket referenced for context and graded against its requirements.

PR #1188, opened 2026-08-22
fix(setup): deploy the settings.json env block instead of dropping it

mlorentedev:
## What

`merge_claude_settings` applies an **allow-list**: `model`, `effortLevel`, `outputStyle`, `permissions.allow`, `hooks.Session*`, `enabledPlugins`. Every other template key is discarded, without a warning, on any installation not bootstrapped from scratch.

`env` was such a key. Measured on this machine:

```
$ python3 -c "import json; print('env' in json.load(open('~/.claude/settings.json')))"
False
```

The template has carried `env.CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS` while the deployed file had **no `env` key at all**, so that flag was never live. `outputStyle` fell in the same hole once already — the comment written after that incident describes the mechanism exactly, but nothing prevented the recurrence.

## Changes

- `setup-linux.sh` / `setup-windows.ps1`: both merge policies now handle `env` (object merge, so a machine-local entry survives a redeploy) and `advisorModel` (template wins).
- `ai/claude/settings.json`: adds `CLAUDE_CODE_ENABLE_EXPERIMENTAL_ADVISOR_TOOL=1`. Claude Code merges `settings.env` into its **own** `process.env` at startup, and that flag gates both the `/advisor` command and whether `advisorModel` ever attaches. Without it the setting is read and silently ignored.
- `ai/claude/settings.json`: `model` becomes `opus[1m]`. The policy is template-wins for `model`, so a 1M-context default chosen through `/model` survived only until the next setup run.
- Lesson 221.

## The guard

The old test was named *"every dotfiles-owned top-level key is named in both merge policies"* and iterated a hand-written list of three keys, while its comment asserted that `env` was "handled by name in both scripts". That claim was false, and nothing checked it — Lesson 220's shape: a thing verified by a proxy that lives somewhere else.

The list now comes from the template, so a key added there fails CI until both scripts name it. `$schema` is exempt by name, as a stated decision.

## Evidence

Fail-first, by injecting an un-merged key into the template:

```
not ok 5 every dotfiles-owned top-level key is named in both merge policies
# setup-linux.sh merge policy never mentions 'someNewKey'
```

Full suite and lint:

```
$ bats tests/*.bats        -> 1441 ok, 0 not ok
$ shellcheck --severity=error setup-linux.sh   -> clean (CI's severity)
$ bash -n setup-linux.sh && jq empty ai/claude/settings.json  -> OK
```

End-to-end, from the deployed `~/.claude/settings.json` alone — no shell env var, no CLI flag:

```
[DEBUG] CA certs: ... settingsEnv keys: CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS,CLAUDE_CODE_ENABLE_EXPERIMENTAL_ADVISOR_TOOL
[DEBUG] [AdvisorTool] Advisor model: fable
[DEBUG] [AdvisorTool] Server-side tool enabled with claude-fable-5 as the advisor model
```

The same run with the flag absent produces no `[AdvisorTool]` line at all — the advisor was inert, silently.

## Notes

- `pwsh` is not installed locally, so the PowerShell lint runs in CI only. The `.ps1` edit mirrors the `enabledPlugins` block already beside it.
- No spec: 29 production lines added, mostly comments, under the spec-gate threshold.

### DIFF
--- a/ai/claude/settings.json
+++ b/ai/claude/settings.json
@@ -1,11 +1,12 @@
 {
   "$schema": "https://json.schemastore.org/claude-code-settings.json",
-  "model": "opus",
+  "model": "opus[1m]",
   "effortLevel": "xhigh",
   "advisorModel": "fable",
   "outputStyle": "Concise",
   "env": {
-    "CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS": "1"
+    "CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS": "1",
+    "CLAUDE_CODE_ENABLE_EXPERIMENTAL_ADVISOR_TOOL": "1"
   },
   "permissions": {
     "deny": [
--- a/docs/lessons/_index.md
+++ b/docs/lessons/_index.md
@@ -234,3 +234,4 @@ tags: [lessons, index, dotfiles]
 | [218 - The Go build cache does not see the data file your test reads](lesson-218-the-go-build-cache-does-not-see-the-data-file-your.md) | 2026-08-21 |  |
 | [219 - A stale CLI refuses with the same exit status as a legitimate refusal](lesson-219-a-stale-cli-refuses-with-the-same-exit-status-as-a.md) | 2026-08-21 |  |
 | [220 - Four defects, one shape: a thing verified by a proxy that lives somewhere else](lesson-220-four-defects-one-shape-a-thing-verified-by-a-proxy.md) | 2026-08-22 |  |
+| [221 - An allow-list merge makes every new template key a silent no-op](lesson-221-an-allow-list-merge-makes-every-new-template-key-a.md) | 2026-08-22 |  |
--- a/docs/lessons/lesson-221-an-allow-list-merge-makes-every-new-template-key-a.md
+++ b/docs/lessons/lesson-221-an-allow-list-merge-makes-every-new-template-key-a.md
@@ -1,0 +1,87 @@
+# Lesson 221 — An allow-list merge makes every new template key a silent no-op
+
+**Date:** 2026-08-22
+**Area:** setup / settings deployment
+**Severity:** medium — config that reads as deployed but never was
+
+## What happened
+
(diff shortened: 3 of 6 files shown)


mlorentedev:
## Review triage

CI at `8f8a8be`: all 14 checks green — `test`, `test-windows`, `lint`, `lint-powershell`, `integration`, `spec-gate`, `attestation`, `review-attestation`, `CodeRabbit`, GitGuardian.

|---|---|---|
| CodeRabbit, `tests/claude-settings-template.bats:58` — *"Scope the policy search to the merge functions"* (Major) | **apply** | Correct and verified: `setup-windows.ps1:413,419` carry `$tool.ContainsKey('Version')` in the winget loop, outside `Merge-ClaudeSettings` (lines 202–300). A template key named `Version` would have satisfied the Windows half of the guard while the merge ignored it. Fixed in `ff26eb6`. |
| PR-Agent (`## PR Reviewer Guide`) | **skip** | A review ran and reported no findings — effort 2/5, no security concerns, no major issues. Nothing to dispose of. |

### Evidence for the applied fix

Regression proof on the exact state the finding describes — template carries `Version`, `merge_claude_settings` handles it, `Merge-ClaudeSettings` does not:

```
OLD GUARD (whole-file grep):  PASSES        <-- the hole
NEW GUARD (scoped to bodies): not ok 5
                              # Merge-ClaudeSettings never mentions 'Version'
```

Both merge bodies are now extracted with `sed` and grepped in isolation, with an explicit non-empty assertion so a failed extraction reports itself instead of blaming the first key. Lesson 221 records it: a guard is a claim about a region of code, and grepping the whole file widens the region until the claim is no longer the one intended — the same shape as the bug the guard exists to catch, one level up.

After the fix: `bats tests/*.bats` → **1441 ok, 0 not ok**; `shellcheck --severity=error setup-linux.sh` clean.

Reviewer output dispositioned: coderabbitai[bot] 2026-08-22T09:34:44Z, github-actions[bot] 2026-08-22T09:31:32Z.

Issue #1191, opened 2026-08-23
HARNESS-078: the board's ID field is empty everywhere, so the duplicate-ID guard can only ever report no collision


mlorentedev:
## What is broken

The bitácora board declares an `ID` custom field, and the `new-ticket` skill treats it as a
cross-repo uniqueness check. Measured 2026-08-22 while filing CLI-042 (#1190): **the field is empty
on every item sampled**, oldest and newest alike, and `gh project item-list` no longer returns it at
all — the key list for an item is `content, id, labels, priority, repository, status, title, type`.

Two things in the skill therefore cannot run as written:

- **The concurrency guard, step 2** — *"scan the board's `ID` field too, not only the home repo's
  issue titles"*. This exists because on 2026-07-07 three parallel sessions produced three duplicate
  IDs (TOOL-017 ×2, DOCS-001 ×2, TOOL-020 ×2 during the remediation itself). The guard is the
  remediation for a defect that has already happened twice, and it currently reads a field that is
  always empty — so it can only ever return "no collision".
- **Step 5's verification** — reads `i.get('iD')`, a key the tool does not emit. It silently reports
  nothing rather than failing.

## Why it matters

This is the shape Lesson 220 names: **a thing verified by a proxy that lives somewhere else.** Ask
what the guard would still pass on if what it checks were broken — and the answer is *everything*,
because an empty field and a collision-free board are indistinguishable to it. A duplicate-ID guard
that cannot detect duplicates is worse than no guard, because it is believed.

The field is not the problem — it is writable. #1190 set it (`ID=CLI-042-dotf-agent-run`) with
`gh project item-edit --field-id … --text …` and it landed. **Nobody has been setting it**, which
means the stable human ID lives only in issue titles today.

## Options

1. **Backfill and enforce.** Populate `ID` for existing items and make `new-ticket` set it every time
   (it already does, in step 4 — the gap is historical items). Then fix the read path: query the
   field over GraphQL rather than through `item-list`, which no longer surfaces it.
2. **Drop the field and the guard step, and re-derive uniqueness from issue titles alone** across the
   bitácora's repos. Honest, cheaper, and loses the cross-repo check the field was added for.
3. **Keep the field for humans and stop pretending it is a guard** — delete step 2 and step 5's read,
   leaving the re-scan-before-create and verify-after-create halves, which do work.

Whichever is chosen, the fix must be verifiable by consequence: a test that plants a duplicate and
[...]

Issue #1190, opened 2026-08-23
CLI-042: dotf agent run — the executor seam ADR-032 defines and nothing implements


mlorentedev:
Part of epic #558 (cross-harness agent pipeline). Implements the executor seam decided in
[ADR-032](docs/adr/adr-032-cross-harness-agent-orchestration.md) §2 and bounded by
[ADR-035](docs/adr/adr-035-model-map-routing-registry.md) §3.

## Why now

Measured on 2026-08-22, the orchestration stack is complete on paper and empty in execution:

| Layer | State |
|---|---|
| **Definition** — what an agent is, how it renders per harness | schema + manifest + render pipeline live; `harness/agents/` holds **1 of 7** roster personas (`curator`) |
| **Policy** — pools, harnesses, tiers, chains, concurrency | `harness/model-map.json` **exists**, schema-validated, with a `dotf doctor` check (#1124) |
| **Execution** — who launches work and collects it | **nothing.** `dotf agent` is `unknown command` |

`chains` in `model-map.json` is declared and has **no reader at all**: it resolves at run time and is
read only by the level-2 dispatcher, which does not exist. A routing registry nothing routes with is a
declaration, not a layer. This ticket gives it its first consumer.

It is also what makes the concurrency semaphore implementable: ADR-035 §3 puts enforcement at level 2
— "only where `dotf` is the launcher" — and there is no launcher yet.

## Scope (v1)

**Verb:** `run` only. `list` / `status` follow the dispatcher, not this ticket.

**Backends: two, deliberately.** A seam validated against a single implementation is a speculative
interface, not a seam.

- `subprocess` — the floor bootstrap guarantees (`claude -p …`, `pi -p …`).
- `hive` — the headless/CI offload.

> **Scope update, 2026-08-22 — the premise for the second backend was falsified by measurement, and
> the decision was re-taken rather than inherited.**
>
> This ticket originally justified hive as *"the headless/CI offload, which already carries fallback
> chains and a cost cap"*. `mcp__hive__worker_status` reports **zero reachable models** — Ollama
> offline, OpenRouter keyless, and OpenRouter is retired upstream (August 2026, per
> `model-map.json`'s own `$comment`). Its provider vocabulary does not intersect the map's pools, so
> it can neither consume `chains` nor report the pool and model actually used. The cost cap exists
> but is meaningless against a flat subscription, where marginal cost is zero and concurrency is the
[...]

Issue #899, opened 2026-08-10
DX-008: the Orca CLI wrapper is unusable from a plain shell — AppImage injects --no-sandbox into the electron-as-node path


mlorentedev:
Blocks treating Orca as harness infrastructure (see #462, DX-007). Found while auditing Orca as the cross-provider agent execution plane.

## Symptom

Both entry points on PATH fail identically on this machine:

```
$ orca status --json
/tmp/.mount_orca-lT73Fga/orca-ide: bad option: --no-sandbox

$ orca-ide status --json
/tmp/.mount_orca-lsCqtq2/orca-ide: bad option: --no-sandbox
```

`command -v orca` resolves to `~/.config/orca/linux-orca-cli-shim/orca`, so this is **not** the GNOME screen-reader collision the Orca guide warns about — it is Orca's own shim. `~/.local/bin/orca-ide` is byte-identical to that shim.

## Cause

Both wrappers run the CLI as `ELECTRON_RUN_AS_NODE=1 <AppImage> -e "<bootstrap>" -- "$@"`. Something in the AppImage launch path (AppImageLauncher's binfmt bypass is installed here, and the desktop entry launches with `--no-sandbox`) puts `--no-sandbox` on the argv. With `ELECTRON_RUN_AS_NODE=1` the binary is a plain Node runtime, which rejects the flag before the CLI bootstrap ever runs. `APPIMAGE_EXTRACT_AND_RUN=1` does not help — it extracts and fails the same way.

## The CLI itself is fine

Calling the inner entrypoint through the already-mounted runtime works and returns valid JSON:

```sh
M=$(ls -d /tmp/.mount_orca-* | head -1)
ELECTRON_RUN_AS_NODE=1 "$M/orca-ide" \
  "$M/resources/app.asar.unpacked/out/cli/index.js" status --json
```

That path was used to pull `status`, `skills get orca-cli`, `worktree create --help` and `orchestration --help` successfully.

## Why this matters beyond convenience

Inside Orca-managed terminals `orca` resolves correctly and `ORCA_CLI_COMMAND` is exported, so the app's own agents are unaffected. The breakage is exactly for **agents running in an ordinary terminal** — which is the case that would let any harness (claude, pi, opencode, copilot) drive Orca worktrees and supervised workers. Without a working wrapper, "Orca is part of the harness" holds only for sessions Orca itself started.

## Acceptance

[...]

Issue #881, opened 2026-08-09
HARNESS-065: the two universal disciplines sit on the layer with a proven zero-fire record


mlorentedev:
## Context

An audit of all 37 skills against the five enforcement layers (2026-08-09) found that the strongest mechanism available — the `harness/enforced/` region injected verbatim into every agent's instructions — carries **six rule ids and not one skill**.

Two skills claim a *universal mandatory moment* while resting on description-matching alone, which is the layer that produced zero invocations of `adversarial-review` over ~218 archived specs in three months:

| Skill | Claim |
|---|---|
| `test-driven-development` | "before writing implementation code" |
| `systematic-debugging` | "before proposing fixes" |

**Evidence class, stated honestly:** the adversarial-review finding was *artifact-proven* (its output block appears nowhere). These two leave no artifact to count, so the finding here is **structural**, not measured. It is a claim about where they sit, not a count of how often they fire.

## Existing mitigation (why this is not urgent)

For SDD-sized work the TDD discipline is already carried structurally: `tasks.md` scaffolds in TDD order, and `features.json` pass-state is harness-owned rather than agent-claimed. The exposed surface is **sub-SDD work** — fixes under the ~50 LOC Discipline Gate threshold, where SDD is deliberately skipped and the discipline rests on memory alone.

## What

Promote the two discipline claims into `harness/enforced/` region ids, following the shape of the existing six.

## Hold condition — do not start this yet

CLI-034 shipped a gate that has **never fired successfully**: no independent review has been produced, and #875 is still open awaiting it. Building a second and third enforcement layer before the first has been exercised end-to-end would reproduce the exact failure this whole line of work exists to correct — shipping correct mechanisms that never run.

Start this only after #875 closes through the gate, and after the resulting evidence says the enforced-region shape actually changes behavior.

## Secondary candidates (same audit, lower confidence)

- `read-all-adrs` — claims a "mandatory pre-step", layer 4
- `genre-picker` — forced only inside `curator` sessions, but its moment (authoring any `00_meta/` artifact) occurs outside them
- `pr-review-triage` — indirectly bound by `definition-of-done` #4, but at completion-claim rather than at the checks-finish moment its description names

## References

- #875 / PR #877 (CLI-034, the first skill moved to a deterministic layer)
- #879 (the trigger half of the same problem)
- `harness/enforced/` (the six existing region ids), `harness/agents/curator/AGENT.md` (persona forced-skills)

mlorentedev:
[...]

PR #1192, opened 2026-08-23
docs(spec): scaffold CLI-042 for the dotf agent run executor seam

mlorentedev:
Scaffolds the spec for `dotf agent run` — the executor seam ADR-032 §2 defines and nothing
implements. Docs only: no production code, no behavior change.

Closes nothing yet; #1190 stays open until the implementation sequence lands.

## Why this spec exists

The orchestration stack is complete on paper and empty in execution. `harness/model-map.json`
declares pools, harnesses, tiers and chains, and `chains` — which resolves at run time and is read
only by the level-2 dispatcher — **has no reader at all**, because `dotf agent` is `unknown command`.
The concurrency semaphore ADR-035 §3 places at level 2 is likewise unimplementable: there is no
launcher for it to bound.

## What the spec decides

ADR-032 fixed the noun and five contract semantics and deliberately left the wire schema and exit
codes to the implementation spec. Those are decided here:

| | Choice | Why |
|---|---|---|
| Dispatch | **synchronous** | `dotf spec review` launches detached, which is precisely why GUARD-005 exists — a detached launcher cannot observe its own verdict. `--detach` can be added later; the reverse is expensive. |
| Backend | **probe, `--backend` overrides** | ADR-032 §7 — probes cannot lie, inventory files rot. |
| stdout | **JSON always** | the consumer is a dispatcher, not a person. |

## Two findings measured while scoping, and what they changed

**1. hive's delegate worker reaches zero models.** `worker_status`, 2026-08-22: Ollama offline,
OpenRouter keyless — and OpenRouter is retired upstream, as `model-map.json`'s own `$comment` already
records. Its provider vocabulary does not intersect the map's pools, so it can neither consume
`chains` nor report the pool and model actually used.

This **falsifies one row of ADR-032 §2's backend table** (*"already has fallback chains and a cost
cap"*): the chains point at providers that do not answer, and a cost cap is meaningless against a
flat subscription where marginal cost is zero. The amendment is recorded in the spec, ADR-035-style,
rather than edited into an accepted ADR.

The scope decision was re-taken with that evidence rather than inherited: hive stays as the second
backend, and its repair enters the critical path as `mlorentedev/hive#384`. A seam validated against
one implementation is a speculative interface.

**2. The fallback does not belong inside a backend.** ADR-032 §2 assigns *pool unavailable → advance
the chain* to the dispatcher. A backend that falls back internally bypasses that semantic, makes
"the pool and model actually used" unreportable, and becomes a second routing authority — the
"map, never a director" drift ADR-035 §4 exists to kill. `chains.mid` already declares NaN→NaN.
Making hive single-shot also turns fallback cross-**backend**: if NaN saturates, the chain can reach
`claude:sonnet` by subprocess, which an internal hive chain could never do.

Consequences recorded in the spec: hive's worker becomes NaN-only (Ollama and OpenRouter dropped
outright), `dotf` reaches it through a dispatch verb over the existing subprocess machinery rather
than by growing an MCP client, and there is no Gemini arm inside hive — that is a map-level change
with its own ticket.

## Scope

Nine acceptance criteria, and a declared five-PR sequence — this does not fit the ~300 executable-LOC
cap, and "first step of a multi-PR sequence" is itself a Discipline Gate trigger. PR A is the hive
repair; B, C and E here do not depend on it.

Out of scope with reasons stated: the `orca` backend (deferred, not rejected — #899, #462/#649, and
C9), the roster personas (#562), `enforced/orchestration.md` (#561, held behind #881), `list` /
`status`, adaptive reservation, and an MCP client in `dotf` (deferred behind a named trigger).

Two open questions are carried explicitly rather than resolved silently: the backend tie-break when
both `subprocess` and `hive` can serve one `nan` chain entry, and the exact service-unit shape for
delivering `NAN_API_KEY` through `dotf secrets run` — the latter is an acceptance criterion, because
a plaintext key in `environment.d` would be a secrets-doctrine violation, not a deployment detail.

## Also filed from this session

`#1191` — the board's `ID` field is empty on every item, so the `new-ticket` skill's duplicate-ID
guard can only ever report "no collision". That guard is the remediation for a defect that has
already happened twice.

### DIFF
--- a/specs/CLI-042-dotf-agent-run/features.json
+++ b/specs/CLI-042-dotf-agent-run/features.json
@@ -1,8 +1,71 @@
 [
   {
     "id": "CLI-042-dotf-agent-run-f1",
-    "behavior": "Outcome 1",
-    "verification": "echo 'not implemented' && exit 1",
+    "behavior": "AC1 — `dotf agent run` writes one JSON object to stdout carrying status, pool, model, exit, duration_ms and output, with every log line on stderr",
+    "verification": "cd cli && go test ./internal/agent/ -run TestRunEmitsJSONContract",
+    "state": "pending",
+    "evidence": ""
+  },
+  {
+    "id": "CLI-042-dotf-agent-run-f2",
+    "behavior": "AC2 — a backend reporting pool-unavailable advances to the next chains entry; one reporting task-failed does not, and the record names the pool and model that answered",
+    "verification": "cd cli && go test ./internal/agent/ -run TestChainAdvancesOnlyOnPoolUnavailable",
+    "state": "pending",
+    "evidence": ""
+  },
+  {
+    "id": "CLI-042-dotf-agent-run-f3",
+    "behavior": "AC3 — a dispatch exceeding its timeout kills the worker and releases its semaphore slot without waiting for it, reporting a non-zero exit with a timeout status",
+    "verification": "cd cli && go test ./internal/agent/ -run TestTimeoutReleasesSlotBeforeReap",
+    "state": "pending",
+    "evidence": ""
+  },
+  {
+    "id": "CLI-042-dotf-agent-run-f4",
+    "behavior": "AC4 — an unreadable concurrency counter exits non-zero rather than reading as zero in use, and an unidentifiable machine denies every non-local pool",
+    "verification": "cd cli && go test ./internal/agent/ -run TestFailsClosed",
+    "state": "pending",
+    "evidence": ""
+  },
+  {
+    "id": "CLI-042-dotf-agent-run-f5",
+    "behavior": "AC4 — help text and error messages state the narrow guarantee ('dotf alone will never be the cause of exhaustion'), never that exhaustion cannot happen",
+    "verification": "cd cli && go test ./internal/agent/ -run TestGuaranteeWordingIsNarrow",
+    "state": "pending",
+    "evidence": ""
+  },
+  {
+    "id": "CLI-042-dotf-agent-run-f6",
+    "behavior": "AC5 — a --tier top dispatch with claude:opus unavailable escalates and exits non-zero; it never falls through to a mid-tier model",
+    "verification": "cd cli && go test ./internal/agent/ -run TestTopTierNeverDegrades",
+    "state": "pending",
+    "evidence": ""
+  },
+  {
+    "id": "CLI-042-dotf-agent-run-f7",
+    "behavior": "AC6 — the hive backend answers through NaN and reports the pool and model the chain resolved",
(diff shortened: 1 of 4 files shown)


mlorentedev:
## Review triage

All checks green. Four pieces of reviewer output, all dispositioned.

|---|---|---|---|
| 1 | CodeRabbit (Major) | `features.json` carries only the scaffolded placeholder, whose verification command always exits 1 — the harness can map nothing from it | **Applied** in `b2410fc` |
| 2 | CodeRabbit (Major) | AC3 asserts the slot is freed without waiting for the worker, while the risk register admits a killed hive client can leave a remote worker consuming | **Applied** in `b2410fc` |
| 3 | CodeRabbit (Major) | `verification.md` lists three evidence rows against nine criteria | **Applied** in `b2410fc` |
| 4 | PR-Agent | "#1190 — Partially compliant" | **Declined**, reason below |

**On #2, which was the substantive one.** Both statements were true and I had treated them as one
guarantee when they are two. For `subprocess` the killed process *is* the consumer, so releasing the
slot is exact. For `hive` it is not: killing the local client does not stop a worker behind the
daemon, which keeps consuming a NaN slot the semaphore has already handed back — the reserve then
over-counts what is free, silently and in the permissive direction.

The criterion now states the backend-specific condition and requires cancellation propagation from
the hive verb (already specified in `mlorentedev/hive#384`). Deliberately **not** taken: the reviewer
offered "or revise AC3 to state the weaker guarantee". Weakening the criterion to match what the
backend can currently do would make the criterion pass by definition. AC3 is written so that shipping
without cancellation propagation fails the *backend's eligibility*, not the criterion.

**On #4.** A spec-only PR that scaffolds `proposal.md` / `tasks.md` and implements nothing is
*expected* to be partially compliant with the ticket it opens — the ticket describes the executor,
this PR describes how it will be built. Nothing to apply. This is also the known shape where PR-Agent
grades a PR against requirements of a ticket rather than against the PR's own stated scope; its
genuine findings arrive under "requires further human verification", and there are none here.

mlorentedev:
**Addendum to the triage above:** the fix commit `b2410fc` was itself **not** re-reviewed — CodeRabbit's check on it reads *"Review skipped: incremental reviews are disabled"*. `review-attestation` passes on the strength of the earlier full review, which covered `59fde6e`, not the three fixes applied on top of it. Stating it rather than letting a green attestation imply otherwise: what is reviewed here is the spec as originally pushed, plus my dispositions. A full re-review can be forced with `@coderabbitai review` once the account quota resets (~03:37Z — it is currently exhausted, see #1193).

PR #1199, opened 2026-08-23
docs(lessons): a coupling's scope is measured, not inferred

mlorentedev:
Docs only — one lesson file plus its index row.

Three sightings in one session of the same shape: something declared in one place and consumed in
more places than whoever fixed it had counted.

1. **hive was a declared dispatch backend with zero reachable providers** for an unknown length of
   time. Every surface reported it by whether a client object existed, never by whether it answered.
2. **`spec-gate` enforces a rule its counter does not implement** — #1180 decided the cap counts
   executable lines; #1186 is that the counter still counts comments.
3. **A diagnostic test gated merges in two places while its ticket said one.**

The third is why the lesson exists: **the fix repeated the mistake.** `mlorentedev/hive#389`'s first
commit de-gated the place the symptom appeared, verified in both directions locally, and shipped —
and CI went red on both runners because the test also had a dedicated workflow step nobody had looked
at. Had that step failed quietly instead of loudly, the ticket would have closed with the second gate
standing and the next release would have failed identically.

Pairs with lesson 220 from the repair side rather than duplicating it:

| | Question |
|---|---|
| **220** — writing a check | What would this still pass on if the thing it checks were broken? |
| **222** — repairing a coupling | What else consumes this, and how do I know I found all of it? |

The practical half is cheap: grep for the symbol across **workflows and config**, not only source —
CI YAML is the surface that gets skipped, and it is where the second gate lived. State the count in
the commit message, because "three call sites, two tools" is a claim a reviewer can check and
"updated the callers" is not.

## SDD skip rationale

**This PR is documentation-only** — `docs/lessons/lesson-222-*.md` and one row in
`docs/lessons/_index.md`. Zero production lines by any reading, and the Discipline Gate exempts this
class **by name**: *"Skip SDD for: … documentation-only changes."*

`spec-gate` refused the push at **"Production diff: 132 LOC"**. That is #1186 — the counter does not
implement the rule it enforces. This instance is sharper than the one that ticket was opened for: it
is not counting comment lines inside production files, it is counting a whole file that is not
production. Recorded there as evidence, with a suggested cheaper first fix — exclude by path and
extension (`docs/**`, `*.md`) before counting anything, which needs no comment parser and no
judgement.

Pushed with a single declared `--no-verify` to create the branch, because the hook's own guidance is
to *"open the PR first"* so it can resolve the label and body live — which cannot happen before the
branch reaches the remote.

### DIFF
--- a/docs/lessons/_index.md
+++ b/docs/lessons/_index.md
@@ -235,3 +235,4 @@ tags: [lessons, index, dotfiles]
 | [219 - A stale CLI refuses with the same exit status as a legitimate refusal](lesson-219-a-stale-cli-refuses-with-the-same-exit-status-as-a.md) | 2026-08-21 |  |
 | [220 - Four defects, one shape: a thing verified by a proxy that lives somewhere else](lesson-220-four-defects-one-shape-a-thing-verified-by-a-proxy.md) | 2026-08-22 |  |
 | [221 - An allow-list merge makes every new template key a silent no-op](lesson-221-an-allow-list-merge-makes-every-new-template-key-a.md) | 2026-08-22 |  |
+| [222 - A coupling's scope is measured, not inferred from where you saw it fail](lesson-222-a-couplings-scope-is-measured-not-inferred-from-whe.md) | 2026-08-23 |  |
--- a/docs/lessons/lesson-222-a-couplings-scope-is-measured-not-inferred-from-whe.md
+++ b/docs/lessons/lesson-222-a-couplings-scope-is-measured-not-inferred-from-whe.md
@@ -1,0 +1,95 @@
+# Lesson 222 — A coupling's scope is measured, not inferred from where you saw it fail
+
+**Date:** 2026-08-23
+**Area:** guards / CI / cross-repo contracts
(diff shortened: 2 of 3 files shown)


mlorentedev:
## Review triage

All checks green, including `spec-gate` — the `skip-sdd` label plus the rationale resolved live once the PR existed, which is the sequence the pre-push hook's own message describes.

|---|---|---|---|
| 1 | PR-Agent | effort 1/5 · *No relevant tests* · *No security concerns* | **Nothing to apply** |
| 2 | CodeRabbit | `pass`, 0 inline comments | **Nothing to apply** |

**On "No relevant tests":** correct and expected — the change is one Markdown lesson plus its index row. Nothing here is testable.

**Worth recording, since it is the lesson's own subject:** this PR was refused by `spec-gate` at *"Production diff: 132 LOC"* for a documentation-only change that the Discipline Gate exempts **by name**. That is a fourth sighting of the class the lesson describes — a rule declared in one place and enforced by something measuring another — and it is now evidence on #1186, with a cheaper suggested fix than the original scope: exclude by path and extension (`docs/**`, `*.md`) before counting anything, which needs no comment parser and no judgement.

Issue #1196, opened 2026-08-23
TOOL-018: every release PR arrives unreviewed, because release-please rewrites its branch and CodeRabbit auto-pauses


mlorentedev:
## What happens

CodeRabbit auto-pauses reviews on a branch it considers *"under active development"*, controlled by `reviews.auto_review.auto_pause_after_reviewed_commits`. `release-please--branches--main` is rewritten on **every** merge to `main`, so it trips that heuristic permanently.

Observed on #1194 (`chore(main): release 0.47.2`): a pause notice, zero inline comments, all checks green.

## Why it matters more than it looks

`dotf pr triage-queue` flags the PR as *"coderabbitai reviewed, never triaged"* — because a comment arrived. The queue is behaving correctly; it exists to make a human look. But the result is a **recurring, permanent entry** in the triage queue for a PR that can never be reviewed, on every release.

Two costs, and the second is the one that bites:

1. Every release needs a manual triage comment saying "not reviewed, machine-generated, merging anyway".
2. **A queue that always has an item in it stops being read.** The value of `triage-queue` is that a non-empty queue means something; a permanent resident trains the reader to skip it — the same way a gauge that always reads zero stops being a gauge.

## Third variant of the same shape today

Same session, three different ways CodeRabbit produced a comment without reviewing, all indistinguishable from outside:

| Mechanism | Seen on |
|---|---|
| Rate limited (account quota) | #1193, hive#390 |
| `Review skipped: incremental reviews are disabled` | #1192 (the fix commits) |
| Auto-paused (branch under active development) | #1194 |

In all three the check settles green and a comment exists. Only the *content* distinguishes them, which is exactly what the pr-stewardship doctrine already says — but it is a lot of reading for a recurring, mechanical case.

## Options

1. **Configure `auto_pause_after_reviewed_commits`** or exclude release-please branches from auto-review in `.coderabbit.yaml`, so it declines explicitly instead of pausing. Cheapest.
2. **Teach `triage-queue` to skip release PRs by author** (`app/github-actions` / release-please) — but that hides the class rather than answering it, and the exemption would need to be narrow.
3. **Accept and template it**: a one-line canned triage the release workflow posts automatically. Turns a judgement into a mechanism, which is this repo's usual preference.

Recommend **1 first** — declining is honest and removes the queue entry at its source — with **3** as the fallback if CodeRabbit offers no such setting for a branch pattern.

PR #1201, opened 2026-08-23
fix(setup): derive the harness mirror from the manifest, not a hardcoded list

### LINKED ISSUE #1200: HARNESS-079: dotf doctor FAILs on harness drift with a remedy that cannot clear it — the mirror evaluates a manifest target setup never copies
`dotf doctor` reports `[FAIL] harness/skill drift (run: compile-harness.sh --refresh, then re-deploy)`. Running the printed remedy does not clear it, and cannot: from the repo the same script exits 0.

## Root cause

Three facts that only fail together:

1. `harness/manifest.json:39` declares `ai/orca/ORCA.md` a harness injection target (added by #1176).
2. `setup-linux.sh` populates `$DOTFILES_DIR` (`~/.dotfiles`) from an explicit `safe_copy` allow-list. `ai/orca/` is not on it, so `~/.dotfiles/ai/orca/` does not exist — only `aider`, `claude`, `copilot`, `gemini`, `skills` are mirrored under `ai/`.
3. `checks_deploy.go:611` (`checkCompileHarnessDrift`) runs the **mirror's** copy of the script: `bash $DOTFILES_DIR/scripts/compile-harness.sh --check`.

The mirror's run therefore evaluates a target whose file is absent and reports:

```
[ERROR] /home/manu/.dotfiles/ai/orca/ORCA.md: need exactly 1 BEGIN + 1 END HARNESS marker (found /)
[check] FAIL: harness drift detected
```

The two scripts are byte-identical (`diff` reports no difference). Only the root differs.

## Why this is worse than a noisy check

The failure message names a remedy that provably cannot clear the condition. Run from the repo:

```
$ ./scripts/compile-harness.sh --check
[check] OK: no harness drift        # exit 0
```

So the operator follows the instruction, sees green, re-runs `dotf doctor`, and still sees red — with no information pointing at the mirror, the manifest, or the missing file. The adjacent `[Repo<->deploy-dir drift]` check passes at the same time, which reads as confirmation that the deploy is fine.

## Severity

Diagnosability, not doctrine delivery. `ai/orca/` is repo-local alongside `orca.yaml` (a per-repo Orca project template), and `harness/manifest.json:19` records that this target's reader "already gets the policy from AGENTS.md (#830)". No agent is running without doctrine because of this. What is broken is that a persistent red FAIL has an unactionable remedy, which trains the operator to ignore the check.

## Fix candidates

Both are probably wanted, and they are independent:

- **Deploy the file.** Add `ai/orca/` to `setup-linux.sh`'s copy list, with `setup-windows.ps1` parity. Decide first whether the mirror is meant to carry it at all — if `ai/orca/` is deliberately repo-only, the manifest entry is what is wrong, not the copy list.
- **Separate the two conditions in the check.** "Target exists and its managed region disagrees" and "target is absent" need different remedies; collapsing them is what produced the unactionable message. An absent target should name the file and say which root it was resolved against.

## Guard

Per the incident-to-guard rule, whichever fix lands should assert that every path in `harness/manifest.json` resolves under the root the drift gate actually evaluates — otherwise the next manifest entry added without a matching deploy rule reproduces this exactly.

## Evidence

```
$ dotf doctor
[Harness + skill drift]
  [FAIL] harness/skill drift (run: compile-harness.sh --refresh, then re-deploy)
Results: 146 passed, 1 failed, 4 warned, 6 skipped

$ bash ~/.dotfiles/scripts/compile-harness.sh --check
[ERROR] /home/manu/.dotfiles/ai/orca/ORCA.md: need exactly 1 BEGIN + 1 END HARNESS marker (found /)
[check] FAIL: harness drift detected

$ ./scripts/compile-harness.sh --check
[check] OK: no harness drift

$ diff scripts/compile-harness.sh ~/.dotfiles/scripts/compile-harness.sh
$ ls ~/.dotfiles/ai/
aider  claude  copilot  gemini  skills
```

Measured on `main` @ `e2e954e`, `dotf` 0.48.0.

mlorentedev:
fix(setup): derive the harness mirror from the manifest, not a hardcoded list

`dotf doctor` reported a permanent `[FAIL] harness/skill drift` whose printed
remedy could not clear it: running `compile-harness.sh --refresh` from the repo
exits 0, because the repo has every file the manifest declares.

The doctor runs the script from `$DOTFILES_DIR` (checks_deploy.go:611), and the
deploy dir was missing one of the three targets. `harness/manifest.json` gained
`ai/orca/ORCA.md` as a target without a matching copy line in `setup-linux.sh`,
whose mirroring block listed `AGENTS.md` and `ai/claude/CLAUDE.md` literally. The
two scripts are byte-identical; only the root they resolve differs.

The target list is now read from the manifest in both the deploy block and the
root-resolution test, so a new entry is mirrored by construction. It previously
lived in three places and could only diverge.

Guards, in the same change:
- `verify-setup.bats` asserts every manifest target exists in the real
  `$DOTFILES_DIR` after setup runs, and refuses to pass on an empty target list
  — the half no unit test reaches, since the root-resolution test builds its own
  temporary mirror and passed throughout by hand-copying the missing file.
- A manifest entry naming a path the repo does not have now fails directly,
  rather than surfacing one layer removed as a marker-count error.

Verified: mirror `--check` exits 0, `dotf doctor` 147 passed / 0 failed (was
146/1), bats 81 passed across the harness suites, shellcheck clean, `bash -n`
and `zsh -n` both parse.

Closes #1200

Full detail, evidence and the guard rationale are in the commit body and in `docs/lessons/lesson-223`.

Closes #1200

## SDD skip rationale

The gate counts **59 production LOC** in `setup-linux.sh` and refuses at the 50 threshold. Measured against the rule the repository actually merged in #1180 — *"the cap counts EXECUTABLE lines; declarative data, schemas and comment blocks are excluded"* — the same diff is **28 lines**:

```console
$ git diff main HEAD -- setup-linux.sh | grep -E '^\+|^-' | grep -vE '^\+\+\+|^---' | wc -l
59
$ git diff main HEAD -- setup-linux.sh | grep -E '^\+|^-' | grep -vE '^\+\+\+|^---' \
    | sed 's/^[+-]//' | grep -vE '^\s*#' | grep -vE '^\s*$' | wc -l
28
```

The 31-line difference is WHY comments: why the target list is derived rather than restated, why `jq` is resolved by path, and why an absent target warns instead of aborting. `scripts/check-spec-gate.sh`'s `_excluded()` is path-based and has no notion of counting executable lines within a kept file, so writing that reasoning down is what pushed this over the threshold.

This is **#1186**, open and unfixed: the executable-lines rule is documented in `AGENTS.md` and in the vault pattern, and enforced by nothing. Using the escape with this rationale is the same compensation the CodeRabbit config PR made when it hit the identical counter, cited there against the same ticket.

A spec folder would be the wrong instrument here regardless: this is a bugfix closing #1200 with its guard in the same change, not a feature needing a proposal.

### DIFF
--- a/docs/lessons/_index.md
+++ b/docs/lessons/_index.md
@@ -236,3 +236,4 @@ tags: [lessons, index, dotfiles]
 | [220 - Four defects, one shape: a thing verified by a proxy that lives somewhere else](lesson-220-four-defects-one-shape-a-thing-verified-by-a-proxy.md) | 2026-08-22 |  |
 | [221 - An allow-list merge makes every new template key a silent no-op](lesson-221-an-allow-list-merge-makes-every-new-template-key-a.md) | 2026-08-22 |  |
 | [222 - A coupling's scope is measured, not inferred from where you saw it fail](lesson-222-a-couplings-scope-is-measured-not-inferred-from-whe.md) | 2026-08-23 |  |
+| [223 - A test updated to keep passing stops being a guard](lesson-223-a-test-updated-to-keep-passing-stops-being-a-guar.md) | 2026-08-23 |  |
--- a/docs/lessons/lesson-223-a-test-updated-to-keep-passing-stops-being-a-guar.md
+++ b/docs/lessons/lesson-223-a-test-updated-to-keep-passing-stops-being-a-guar.md
@@ -8,7 +8,7 @@
 
 `dotf doctor` reported, persistently:
 
-```
+```text
 [Harness + skill drift]
   [FAIL] harness/skill drift (run: compile-harness.sh --refresh, then re-deploy)
 ```
@@ -83,7 +83,7 @@ all. The integration guard went red on its very first CI run — not on the bug 
 was written for, but on a latent one underneath it. In one setup run, 22 seconds
 apart and in the same process:
 
-```
+```text
 [SUCCESS] jq installed
 [WARNING] Claude Code CLI, npx, or jq not found, skipping MCP server registration
 ```
@@ -99,7 +99,7 @@ check asks about, or the two will disagree and both will look right.** Tracked a
 
 ## Evidence
 
-```
+```console
 $ bash ~/.dotfiles/scripts/compile-harness.sh --check
 [ERROR] /home/manu/.dotfiles/ai/orca/ORCA.md: need exactly 1 BEGIN + 1 END HARNESS marker (found /)
 [check] FAIL: harness drift detected
--- a/docs/lessons/lesson-223-a-test-updated-to-keep-passing-stops-being-a-guar.md
+++ b/docs/lessons/lesson-223-a-test-updated-to-keep-passing-stops-being-a-guar.md
@@ -1,0 +1,91 @@
+# Lesson 223 — A test updated to keep passing stops being a guard
+
+**Date:** 2026-08-23
+**Area:** guards / deploy / harness
+**Severity:** medium — a permanent red check whose printed remedy cannot clear it
+
+## What happened
+
+`dotf doctor` reported, persistently:
+
+```
+[Harness + skill drift]
+  [FAIL] harness/skill drift (run: compile-harness.sh --refresh, then re-deploy)
+```
+
+Running the printed remedy from the repo exits **0** — `[check] OK: no harness drift`.
+So the operator follows the instruction, sees green, re-runs the doctor, and is
+still red, with nothing in the message pointing at the cause.
+
+The cause needed three facts, none of which is visible from the failure text:
+
+1. `harness/manifest.json` declares three injection targets: `AGENTS.md`,
+   `ai/claude/CLAUDE.md`, and — added by `#1176` — `ai/orca/ORCA.md`.
+2. `setup-linux.sh` mirrored a **hardcoded pair** into `~/.dotfiles`. The third
+   target never got a copy line, so the deploy dir never had the file.
+3. `checkCompileHarnessDrift` (`cli/internal/doctor/checks_deploy.go:611`) runs
+   the script **from `$DOTFILES_DIR`**, not from the repo. The two scripts are
+   byte-identical; only the root differs. The mirror's run evaluated a target
+   whose file was absent and failed on the marker count.
+
+## The part worth carrying forward
+
+There was already a guard for exactly this. `tests/compile-harness-rootresolve.bats`
+assembles a non-git copy of "exactly what `--check` reads" and asserts it passes.
+It was written *because* this failure mode had happened before — its header says
+so: *"setup copied none of the latter three, so `--check` exited 2 and section 12
+reported a false drift FAIL."*
+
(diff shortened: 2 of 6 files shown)


mlorentedev:
## Review triage

| Finding | Source | Severity | Disposition |
|---|---|---|---|
| Add language identifiers to both fenced blocks in `lesson-223` (MD040) | CodeRabbit | Minor | **Applied** — `text`, `text`, `console` on the three opening fences. |
| Fail when a manifest target is missing from the repository — the loop skips it silently, so setup completes with an incomplete mirror and `--check` still reports drift | CodeRabbit | Major | **Applied in substance, remedy declined.** The finding is right: a silent skip reproduces this PR's own defect one layer down, and it now warns with the offending path, plus a warning when the copy itself fails rather than swallowing it in `\|\| true`. Setup does **not** abort, which is what was proposed: this script is long and idempotent and one absent target must not cost a machine the rest of its provisioning. The condition is caught by assertion instead — `compile-harness-rootresolve.bats` fails if the repo lacks a declared target, and `verify-setup.bats` fails on the resulting gap in a real deploy. |
| Ticket compliance: "does not separate the two error conditions in the check itself" | PR-Agent | Minor | **Declined, ticketed.** Correct observation, wrong layer for this PR. Splitting "target absent" from "target present but disagrees" is a change to `compile-harness.sh`'s own reporting, and it is written up as a named fix candidate in #1200 rather than widened into a deploy fix. |
| No security concerns; no major issues detected | PR-Agent | — | Noted. |

## Disclosure — what was and was not reviewed

**CodeRabbit reviewed only the first commit.** Its first comment on this PR reads *"Review skipped — Auto incremental reviews are disabled on this repository"*, and its actual review was posted against `b60240e`. Commits `64755e0`, `8a0dc8d` — which contain the jq-resolution fix, the `setup-linux.bats` rewrite and the responses above — have **not** been reviewed by it. This is #1196's mechanism showing up on a non-release PR.

## Findings this PR's own CI produced, and where they went

- **`test` job red:** `tests/setup-linux.bats` asserted the two literal `safe_copy` lines this change removed. It was a fourth restatement of the manifest's target list and could never assert the list was complete. Rewritten to assert the derivation; the outcome is asserted by `verify-setup.bats`.
- **`integration` job red:** the new guard failed on its first run for a reason older than this branch — setup logs `[SUCCESS] jq installed` and `[WARNING] ... jq not found` 22 seconds apart in one process, because `jq` lands in `$HOME/.local/bin` and that is not on the running process's PATH. Worked around here by resolving the binary path; the wider class (every setup step gated on `command -v` for a tool setup installs, MCP registration included) is filed as **#1202**.

## Spec gate

Refused at 59 production LOC. Measured under the repository's own merged executable-lines rule (#1180) the diff is **28**; the 31-line difference is WHY comments. `skip-sdd` applied with the rationale in the PR body, citing **#1186** — the open ticket for the counter that does not implement the rule.

Issue #1034, opened 2026-08-16
TEST: 46 bats assertions across 18 files are silently ignored — `! cmd` is exempt from set -e


mlorentedev:
A negated assertion anywhere except the **final line** of a `@test` provides no protection at all. Found while chasing an unrelated review comment; measured repo-wide afterwards.

## The mechanism

Bash exempts a command preceded by `!` from `set -e`. Bats runs tests under `set -e`, so:

```bash
@test "does a negation in the middle count?" {
    ! true          # this SHOULD fail — true returns 0, negated is 1
    true
}
```

```
ok 1 does a negation in the middle count?
```

It passes. A bare `false` in the same position **is** caught — the exemption is specific to the `!` form, which is why it survives casual inspection and why every one of these files looks correct.

## It was hiding a real violation

This is not theoretical. `tests/check-review-attestation.bats` asserted that `scripts/check-review-attestation.sh` names no reviewer. **The script did name one**, in a comment. The suite reported green for as long as the violated assertion was not the last line.

## Measured extent

**46 dead assertions across 18 files.** Every one is a check someone wrote, believing it protects something:

| File | Dead | Lines |
|---|---|---|
| `tests/agents-md.bats` | 1 | L99 |
| `tests/aliases.bats` | 4 | L91, L193, L194, L195 |
| `tests/antigravity.bats` | 1 | L76 |
| `tests/claude-settings-template.bats` | 2 | L131, L138 |
| `tests/compile-harness.bats` | 8 | L190, L191, L296, L339, L340, L486, L681, L776 |
| `tests/docs-drift.bats` | 1 | L55 |
| `tests/hive-upgrade-timer.bats` | 1 | L35 |
| `tests/knowledge-crystallize-ps1.bats` | 1 | L69 |
| `tests/nan-scripts-secrets.bats` | 2 | L23, L29 |
| `tests/opencode.bats` | 7 | L29, L53, L76, L81, L248, L255, L274 |
| `tests/powershell-profile.bats` | 3 | L62, L67, L68 |
| `tests/setup-linux.bats` | 2 | L76, L314 |
| `tests/setup-windows.bats` | 8 | L58, L80, L81, L136, L164, L327, L368, L568 |
| `tests/skills-pipeline.bats` | 1 | L39 |
| `tests/utils.bats` | 1 | L87 |
| `tests/vault-health-golden.bats` | 1 | L170 |
| `tests/verify-setup.bats` | 1 | L196 |
[...]

Issue #1045, opened 2026-08-18
GUARD-002: a PR-Agent review can never attest — the classifier reads reviews[] and PR-Agent posts a comment


mlorentedev:
GUARD-002 (#1019) and PR-Agent (#1032) were built weeks apart to solve two halves of the same problem — make a green check mean *reviewed*, and stop review capacity gating throughput. **They do not compose.** A PR reviewed by PR-Agent reads as unattested, permanently.

## Measured

PR-Agent published a real review on both of these:

```
#1037  2026-08-18T00:05:21Z  github-actions[bot]  ## PR Reviewer Guide 🔍
#1042  2026-08-18T00:16:58Z  github-actions[bot]  ## PR Reviewer Guide 🔍
```

The #1042 Guide is substantive — effort estimate, "PR contains tests", "No security concerns identified", "No major issues detected".

Yet:

```
$ gh pr view 1042 --json reviews --jq '.reviews|length'
0
$ gh pr view 1042 --json statusCheckRollup ... | grep attestation
  attestation: FAILURE
  review-attestation: FAILURE
```

#1037 passes the gate — but on **CodeRabbit's** formal review at 00:11:41, not on PR-Agent's. Remove CodeRabbit and #1037 fails too.

## Root cause

`scripts/check-review-attestation.sh:164-165` counts formal PR review objects only:

```sh
INDEPENDENT="$(printf '%s' "$PR_JSON" | jq -r --arg a "$PR_AUTHOR" \
    '[.reviews[]? | select((.author.login // "") != $a)] | length')"
```

PR-Agent posts an **issue comment**, not a review object. GitHub's `POST /pulls/{n}/reviews` requires either a token with review scope on the PR or a distinct identity; PR-Agent runs under `GITHUB_TOKEN` as `github-actions[bot]` and publishes via the comments API. So its output never enters `reviews[]` and the classifier cannot see it, no matter how good the review is.

The comment on line 156 states the intent — *"Any entry in reviews[] counts, from a bot or a human alike: the question is whether someone independent looked"* — and the implementation answers a narrower question: whether someone independent **used the reviews API**.

## Why this is not #1033

#1033 is the **opposite** defect in the same file: the gate accepts *too much* (any bot in `reviews[]` attests, including a labeler or release bot). This is the gate accepting *too little* (a genuine, declared reviewer's output does not attest).

[...]

PR #1203, opened 2026-08-23
test(bats): kill negated assertions that cannot fail, and guard the class

mlorentedev:
Closes part of #1034 (the ticket stays open until the quarantine list is empty — see *Chunking* below).

## The defect

bash exempts any command preceded by `!` from `set -e`. bats runs a `@test` body under `set -e` and takes the body's status from its **last** command. So `! grep -q X file` anywhere but the final line of a test is not an assertion — it executes and is discarded.

A bare `false` in the same position **is** caught. The exemption is specific to the `!` form, which is why it survives casual inspection.

## Scope, measured

`grep -rnE '^\s+!\s' tests/*.bats` → **139 lines in 30 files** (the ticket said 46 in 18; it was written before the suite grew).

| | count |
|---|---|
| Could not fail (not the last statement, not rescued by `\|\| return 1`) | **53** |
| Work today, one appended line from not working | 86 |

The ticket's estimate of 46 is superseded by this measurement; the classifier used is in the commit history of this branch's session, and the guard added here reproduces the count.

## Proof it is real

Falsify one existing assertion — `tests/setup-windows.bats`, *"does NOT eager-source load-secrets"*, at its own line 81 of 84 — by pointing it at a string the target file certainly contains:

| Form | Result |
|---|---|
| `! grep -qF 'param' "$PS1_SCRIPT"` | `ok 1` |
| `refute_grep_fixed 'param' "$PS1_SCRIPT"` | `not ok 1` |

Same file, same line, same falsified claim. One reports green.

Not hypothetical damage: `tests/check-review-attestation.bats` AC5 asserted the script names no reviewer of its own. It **did**, in a comment. The suite was green for as long as the violated line was not last.

## What this PR does

- **`tests/lib/refute.bash`** — one shared `refute_grep` (ERE) / `refute_grep_fixed` (literal), replacing three divergent copies (`backup-secrets-to-usb`, `check-review-attestation`, `pr-agent-config`). Two vacuous passes that all three copies had are closed:
  - **a grep error is not an absence** — exit ≥2 (unreadable file, or a BRE pattern that is invalid as an ERE) was read as "not found". This matters exactly during conversion: `(` is a literal in BRE and an unterminated group in ERE.
  - **a pattern beginning with `-` is a pattern** — `--vault knowledge` was being handed to grep as options. It now goes after `--`.
- **`tests/refute-lib.bats`** — the helper's own tests, both directions, including the two above.
- **`tests/guard-bats-negation.bats`** — suite-wide guard with a per-file **quarantine ratchet**: it fails when a listed file's count moves in *either* direction, so a conversion must lower its entry in the same commit and a regression cannot hide behind an entry that over-counts. It replaces the single-file meta test that lived in `check-review-attestation.bats` (deleted here — two guards for one rule is one guard and one thing that quietly disagrees with it). Its detector is pinned against a fixture with a hand-counted answer, because a guard that silently matches nothing reports a clean suite forever.
- **55 assertions converted** in the five files carrying the load-bearing negatives.
- **`docs/lessons/lesson-224`** + index line.

### Why a bats meta-test and not a `scripts/check-*.sh`

`check-bats-names.sh` has to be a script because bats cannot see its own duplicate `@test` names — the file is a parse error, so nothing runs. There is no such blindness here: a bats file can read its siblings, and the `test` job already runs the suite. No new CI wiring, no production LOC.

## The load-bearing assertions: all satisfied

Every negative the ticket flagged as load-bearing is genuinely true today. Converted and run, they pass:

| Assertion | Verdict |
|---|---|
| `setup-windows.bats` L80-81, `powershell-profile.bats` L62/67-68 — the retired `load-secrets` path is gone (ADR-028 migration guard) | green |
| `compile-harness.bats` L287-288 — setup no longer reads the deleted `ai/skills` tree | green |
| `setup-linux.bats` L315, `claude-settings-template.bats` L261/268 — the removed plugins are not installed | green |

No latent defect surfaced, so nothing to fix or ticket on that axis. Worth stating explicitly, since the expectation going in was the opposite.

One nastier variant found while converting, in `claude-settings-template.bats`: the negation sat inside a `for` over 9 plugins. The loop is the last statement, so it *does* carry a status — the **last iteration's**. Eight of nine plugins were unchecked in a test whose name promises all nine.

## Chunking (declared, so no PR absorbs the next in silence)

| PR | Content | State |
|---|---|---|
| 1 (this) | library + ratchet guard + lesson + the 5 files with the load-bearing negatives (55 lines) | here |
| 2 | the remaining dead assertions | next |
| 3 | the remaining live-but-fragile ones; ratchet to zero, list deleted, #1034 closed | after |

**84 lines remain quarantined in 25 files.** The ratchet refuses to let that number grow meanwhile.

## Verification

```
$ bats tests/*.bats
1453 tests, 0 failures        # exit 0

$ shellcheck scripts/*.sh setup-linux.sh
(no output)

$ ./scripts/check-spec-gate.sh --base-ref main --head-ref HEAD --explain
[OK] Production diff 0 LOC < threshold 50 (below threshold; spec not required)
```

Guard mutation-tested in both directions before commit: adding a bare `!` to a converted file fails test 1 (`tests/setup-linux.bats (1)`), removing one from a quarantined file fails test 2 (`tests/utils.bats: recorded 6, found 5`).

### DIFF
--- a/docs/lessons/_index.md
+++ b/docs/lessons/_index.md
@@ -237,3 +237,4 @@ tags: [lessons, index, dotfiles]
 | [221 - An allow-list merge makes every new template key a silent no-op](lesson-221-an-allow-list-merge-makes-every-new-template-key-a.md) | 2026-08-22 |  |
 | [222 - A coupling's scope is measured, not inferred from where you saw it fail](lesson-222-a-couplings-scope-is-measured-not-inferred-from-whe.md) | 2026-08-23 |  |
 | [223 - A test updated to keep passing stops being a guard](lesson-223-a-test-updated-to-keep-passing-stops-being-a-guar.md) | 2026-08-23 |  |
+| [224 - A negated assertion is exempt from `set -e`, so it cannot fail a test](lesson-224-a-negated-assertion-is-exempt-from-set-e-so-it-cann.md) | 2026-08-23 |  |
--- a/docs/lessons/lesson-224-a-negated-assertion-is-exempt-from-set-e-so-it-cann.md
+++ b/docs/lessons/lesson-224-a-negated-assertion-is-exempt-from-set-e-so-it-cann.md
@@ -1,0 +1,129 @@
+# Lesson 224 — A negated assertion is exempt from `set -e`, so it cannot fail a test
+
+**Date:** 2026-08-23
+**Area:** tests / bats / guards
+**Severity:** high — 53 assertions across the suite could not fail, and one was hiding a real violation
+
+## What happened
+
+bash exempts any command preceded by `!` from `set -e`. bats runs a `@test` body
+under `set -e` and takes the body's exit status from its **last** command. Put
+those together and `! grep -q X file` in the middle of a test is not an
+assertion — it is a comment that executes.
+
+```bash
+@test "the script names no reviewer" {
+    ! grep -q 'coderabbitai' "$SCRIPT"   # violated: passes anyway
+    grep -q 'reviewers' "$REGISTRY"      # this line alone decides the verdict
+}
+```
+
+Measured on 2026-08-23 over `tests/*.bats`: **139** lines whose first token is
+`!`, in 30 files. **53** of them were in a position where they could not fail
+(not the last statement of their `@test`, and not rescued by an `|| return 1`).
+The other 86 worked only because nothing had been appended below them yet.
+
+The proof is a falsification, not an argument. Take one real assertion —
+`tests/setup-windows.bats`, "does NOT eager-source load-secrets" — and make it
+false by changing the pattern to a string the file certainly contains:
+
+| Form | Result |
+|---|---|
+| `! grep -qF 'param' "$PS1_SCRIPT"` (old, line 81 of 84) | `ok 1` |
+| `refute_grep_fixed 'param' "$PS1_SCRIPT"` (new, same line) | `not ok 1` |
+
+Same file, same position, same falsified claim. One reports green.
+
+This is not theoretical damage. `tests/check-review-attestation.bats` asserted
+that `scripts/check-review-attestation.sh` names no reviewer of its own. The
+script **did** name one, in a comment. The suite was green for as long as the
+violated line was not the last one in its test.
+
+## Why it survives inspection
+
+Three reasons, and each one is worth knowing on its own:
+
+**A bare `false` in the same position IS caught.** The exemption is specific to
+the `!` form, so the obvious sanity check — "does `set -e` work in a bats body?"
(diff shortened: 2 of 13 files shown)


mlorentedev:
## Review triage

Two reviewer agents ran on this PR and they do not agree, so the disposition below states which one the evidence backs.

|---|---|---|---|
| 1 | PR-Agent | *"The guard's quarantine list does not reflect the conversions in this commit, causing the guard test to fail"* — names `claude-settings-template`, `compile-harness`, `powershell-profile`, `setup-linux`, `setup-windows` as still listed | **Declined — the finding is false, and its own evidence block refutes it.** The list PR-Agent quotes is the committed one, and none of the five files it names appear in it. Empirically: `test` is green on `e6816b3`, and the assertion it predicts failing (*"every quarantined file still carries exactly its recorded count"*) is part of that run. 1453 tests, 0 failures. |
| 2 | PR-Agent | Detector may overcount; the fixture should carry a false-positive example | **Applied** (`861bcc1`). The fixture already carried four non-matching shapes and asserts an exact count, so the premise was not quite right — but the two an author is most likely to write were missing. `# ! grep -q a b` and `foo \|\| ! bar` are now in it. |
| 3 | PR-Agent | No evidence the converted assertions were run to surface latent failures | **Declined — already in the body, with the numbers.** Every converted file was run before and after; the five load-bearing negatives the ticket flagged all pass, which is stated explicitly rather than left implied. |
| 4 | PR-Agent | Human verification: prove *every* conversion fails when violated, not just the one in the description | **Declined, with the reason.** Falsifying 55 assertions one at a time is not the cheapest proof of the class. The class-level evidence is `tests/refute-lib.bats`, which asserts the helper in both directions including the two vacuous-pass paths (grep exit >=2, and a `-` leading pattern), plus the one end-to-end falsification at a non-final line. Mutation testing over the suite is the durable form of this request and is out of scope here. |
| 5 | CodeRabbit | *"Review limit reached — we couldn't start this review."* Next window in 33 minutes | **Disclosed, not dispositioned — there is nothing to dispose of.** A notice that no review ran leaves that reviewer's pass unperformed. `attestation` is green because PR-Agent published a real review, which is the gate working as specified (#1045), not a second opinion on the code. |

Note for the merge decision: the only agent that reviewed this PR is the one whose single blocking finding is demonstrably wrong. Worth a CodeRabbit re-run once its window reopens, or a human read of `tests/guard-bats-negation.bats`.

Follow-up: chunk 2 of the declared chunking is pushed as `fix/bats-negation-remaining-dead`, stacked on this branch.

PR #1206, opened 2026-08-23
test(bats): convert the remaining dead negated assertions

mlorentedev:
Chunk 2 of 3 for #1034. **Stacked on #1203** (`fix/bats-dead-negated-assertions`) — review that one first; this branch contains its commits and will shrink to its own two once #1203 merges.

## What this does

Converts the 14 remaining files that carried an assertion in a position where it could not fail, and empties their entries from the ratchet in `tests/guard-bats-negation.bats`.

**67 lines converted.** The mechanical `! grep -q{E,F}` cases went straight to `refute_grep` / `refute_grep_fixed`. The rest needed a decision per site, and those decisions are the reviewable part:

| Shape | Conversion | Why not the obvious one |
|---|---|---|
| `! grep -q '# Aider (AI pair programming)'` | `refute_grep_fixed` | **BRE to ERE is not always faithful.** `( ) { } + ? \|` are literals in BRE and operators in ERE. Moved to `-E` this pattern is an unterminated group: grep exits 2, and a bare `! grep` reads exit 2 as an absence. `.`, `*`, `^`, `\$` mean the same in both, so those stayed regexes. |
| `! jq … \| grep -qE`, `! grep … \| grep -q` | `run <producer>` + assert on `$output` | A pipeline cannot be a `refute_grep` call. The new form also pins the producer's own exit status, which the old one discarded. |
| `! command_exists`, `! var_is_set`, … | `run <predicate>` + `[ "$status" -eq 1 ]` | `-eq 1`, not `-ne 0`: 127 is the status of a helper whose name was typo'd, and `-ne 0` passes on it. |
| `! grep -qi "aider"` | `run grep -in` + exact status | The helper deliberately has no flag passthrough; the dialect is in its name. Two case-insensitive sites do not justify a third entry point. |
| `! grep … \|\| { echo …; return 1; }` (2 sites) | `refute_grep` | These two were **live** — the `\|\|` rescued them. They lose the hand-rolled message because the helper already prints the file and the offending line. |

Two pre-existing `grep --` forms surfaced as loud arg-count failures instead of silent passes. That is the helper's contract working: `refute_grep` passes the pattern after `--` itself, so a caller-supplied `--` is a third argument and it says so.

## Remaining after this

**17 lines in 11 files**, all currently in the last-statement position where the form does work. They convert in chunk 3, because a `!` on the last line is one appended assertion away from dead and nothing announces the move.

## Verification

```
$ bats tests/*.bats
1453 tests, 0 failures        # exit 0

$ ./scripts/check-spec-gate.sh --base-ref main --head-ref HEAD --explain
[OK] Production diff 0 LOC < threshold 50 (below threshold; spec not required)
```

### DIFF
--- a/tests/agents-md.bats
+++ b/tests/agents-md.bats
@@ -1,6 +1,8 @@
 #!/usr/bin/env bats
 # Tests for AGENTS.md SDD Discipline Gate enforcement (SDD-001)
 
+load 'lib/refute'
+
 setup() {
     export DOTFILES_DIR="$BATS_TEST_DIRNAME/.."
     export AGENTS_MD="$DOTFILES_DIR/AGENTS.md"
@@ -96,8 +98,8 @@ setup() {
 @test "AGENTS.md does NOT reintroduce the retired work/personal routing axis" {
     # Negative guard: these markers encode the old axis that mis-routed build/operate
     # artifacts to the vault. Their presence is a regression.
-    ! grep -qF 'for work projects' "$AGENTS_MD"
-    ! grep -qF '30-architecture/adr' "$AGENTS_MD"
+    refute_grep_fixed 'for work projects' "$AGENTS_MD"
+    refute_grep_fixed '30-architecture/adr' "$AGENTS_MD"
 }
 
 # --- HARNESS-064: adversarial-review trigger (#879) ---
--- a/tests/aliases.bats
+++ b/tests/aliases.bats
@@ -1,6 +1,8 @@
 #!/usr/bin/env bats
 # Tests for .zsh/aliases.zsh
 
+load 'lib/refute'
+
 setup() {
     export DOTFILES_DIR="$BATS_TEST_DIRNAME/.."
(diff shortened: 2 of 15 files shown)


mlorentedev:
## Review triage

|---|---|---|---|
| 1 | PR-Agent | *Fully compliant on the conversion; no non-compliant requirements; no major issues detected* | **Nothing to dispose of.** Recorded so the queue can close: `test`, `integration`, `lint`, `lint-powershell`, `test-windows`, `spec-gate` and `attestation` are all green on `dee4cc6`. |
| 2 | CodeRabbit | *"Review skipped — auto reviews are disabled on base/target branches other than the default branch"* | **Disclosed, not dispositioned.** This PR's base is `fix/bats-dead-negated-assertions`, so CodeRabbit will not review it at all — not a quota window that reopens, a configuration that excludes every stacked PR. **This PR has been read by one reviewer agent only.** That is a property of the chunking, and it is stated rather than left to be inferred from a green check. |

Merging this before #1203 is not possible and not intended: the base branch is #1203's.

PR #1207, opened 2026-08-23
test(bats): convert the last negated assertions and close the ratchet

mlorentedev:
Final chunk for #1034. **Stacked on #1206**, which is stacked on #1203 — review those first; this branch shrinks to its own commit as they merge.

## What this does

Converts the last **17 lines in 11 files** and deletes the quarantine ratchet, because it is empty. The suite now contains **zero** bare `!` assertions and the guard asserts a flat zero.

These 17 were all in the last-statement position, where the form *does* work. That is why they were left for last, and why they still had to go: a `!` on the closing line stops being on the closing line the moment someone appends an assertion, and nothing announces the move.

## Per-site decisions worth reading

| Site | Decision |
|---|---|
| `refute_issue_filed` (bitacora-reconcile) | Kept `grep -s` semantics deliberately: the gh log **not existing** is itself the assertion on the soft-pass paths. It is the one place a missing file is a pass rather than the error `refute_grep` treats it as, so it is now an explicit `[ -f "$GH_LOG" ] \|\| return 0` instead of a flag whose reason lived in a comment. |
| `'\$scripts = @('` (pwsh-analyzer-coverage) | `refute_grep_fixed`, not `refute_grep` — `(` is a literal in BRE and an unterminated group in ERE. |
| `! instruction_files \| grep -qx`, `! grep -v … \| grep -qE` | Captured intermediate + `run` + exact status. Keeps `grep -x`'s whole-line semantics, which a substring test on `$output` would have quietly dropped. |
| shell-wrapper-dedup's five-file grep | A loop of one-file calls. A multi-file grep cannot distinguish a path that does not exist from a pattern that is absent, and that distinction is the helper's entire point. |

## Guard changes

- The ratchet and its two tests are deleted with the last entry. The guard's header records the ratchet's shape in case a batch ever has to come back.
- New third case: **every file that calls a refute helper loads the library.** A missing `load 'lib/refute'` is not a silent pass — the call errors — but it errors at the first `@test` that reaches it, which can be far from the edit.

## Verification

```
$ bats tests/*.bats
1456 tests, 0 failures        # exit 0, rebased on main @ b04a6c4

$ grep -rcE '^[[:space:]]*![[:space:]]' tests/*.bats | grep -v ':0$'
(no output)
```

### DIFF
--- a/tests/bitacora-reconcile.bats
+++ b/tests/bitacora-reconcile.bats
@@ -10,6 +10,8 @@
 # So these tests do not inspect the workflow's text; they EXECUTE the classifier,
 # and the load-bearing ones execute it under `bash -e` (see "the regression").
 
+load 'lib/refute'
+
 setup() {
     SCRIPT="$BATS_TEST_DIRNAME/../scripts/bitacora-reconcile.sh"
     FIX="/tmp/bats_reconcile_$$_${BATS_TEST_NUMBER:-0}"
@@ -47,10 +49,13 @@ teardown() {
     rm -rf "$FIX"
 }
 
-# `grep -s` so this holds when gh was never called at all and the log is absent —
-# which is itself the assertion on the soft-pass paths.
 refute_issue_filed() {
-    ! grep -qs "issue create" "$GH_LOG"
+    # The log may not exist at all — gh was never called — and that absence is
+    # itself the assertion on the soft-pass paths. It is the one case where a
+    # missing file is a pass rather than the error refute_grep treats it as, so
+    # it is stated here instead of being folded into `grep -s`.
+    [ -f "$GH_LOG" ] || return 0
+    refute_grep_fixed 'issue create' "$GH_LOG"
 }
 
 @test "success: notice and exit 0" {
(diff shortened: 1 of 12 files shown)


mlorentedev:
## Review triage

|---|---|---|---|
| 1 | PR-Agent | *#1034 — fully compliant*; no major issues detected | **Nothing to dispose of.** All checks green on `b4e7977`. |
| 2 | PR-Agent | *"#1206 — Not compliant: this PR does not address the 14 files in chunk 2; it stacks on the prior PRs that did"* | **Declined — that is the declared design, not a gap.** #1206 is this PR's base branch, so its 14 files are already in this branch's history. The reviewer treated a stacked base as an unmet requirement; `git log --oneline main..HEAD` shows the four commits. |
| 3 | CodeRabbit | *"Review skipped — auto reviews are disabled on base/target branches other than the default branch"* | **Disclosed, not dispositioned.** Same as #1206: a stacked PR is never reviewed by CodeRabbit by configuration, so **this PR has been read by one reviewer agent only.** Stated rather than inferred from a green check. |

PR #1209, opened 2026-08-23
feat(cli): dotf agent run dispatches over the tier chain

mlorentedev:
PR **B** of the five-PR sequence declared in `specs/CLI-042-dotf-agent-run/tasks.md`. Covers **AC1, AC2 and AC5**. Refs #1190.

`harness/model-map.json` declares `chains`, resolved at run time, and nothing read it — `dotf agent` was an unknown command, so the orchestration stack had a policy layer and no execution layer. This adds the layer.

## What lands

- **`dotf agent run`** — flags `--role --task --tier --cwd --timeout --backend --repo-root`; one JSON record on stdout, every log line on stderr.
- **`cli/internal/agent`** — the `Backend` seam (one method, the five semantics of ADR-032 §2) and the chain walk over `harness.ResolveChain`.
- **A `dry-run` backend** — resolves the route, sends nothing, exits 0.
- **`chains.top` capped at one entry** in `model-map.schema.json`.
- **Exit-code plumbing in `main()`**.

## Contract decisions

Recorded in `proposal.md`'s decision table, so they are not re-derived in review:

| | |
|---|---|
| **Exit codes** | `0` answered, `1` task failed, `3` no pool could serve — the same three `hive delegate` uses, so one vocabulary spans the seam and no translation table exists to drift. An unrecognised backend status is read as *task failed*, never *unavailable*: the fail-closed direction is the one that does not spend a second pool on a task that may already have been answered. `escalated` and `chain_exhausted` share exit 3 and stay distinct in the record's `status` — the code is the coarse class, the record carries the fine one. |
| **`--backend` required** | Until probing lands in PR D. Defaulting to `dry-run` would be the worst option available: a dispatch that silently ran nothing and exited 0. |
| **`dry-run` is a real backend** | Not a test double. It makes AC1's bats half runnable today, and after PR D it is still the way to inspect which pool and model a tier resolves to without spending a slot of a shared quota. The scripted fake the Go tests use stays test-only, injected through the seam. |

## The top tier is guarded twice, and they are not duplicates

ADR-032 §4: the top tier never degrades. The **dispatcher** never advances past the first entry whatever the map says; the **schema** caps `chains.top` at one entry so a fallback cannot be declared at all. One stops a second entry being *used*, the other stops it being *written*.

The behaviour test uses a two-entry top chain the map does not declare — against today's single-entry `chains.top`, a dispatcher with no rule at all would pass.

## Three things the work surfaced

- **`main()` could only ever exit 1**, collapsing *no pool could serve this* into *the task failed* at the process boundary — the exact collapse AC2 forbids one layer down. Fixed with a tagged error unwrapped in `main`, rather than the `os.Exit`-inside-`RunE` precedent in `secrets.go`, which costs that file a child-process re-exec harness to test at all.
- **`features.json`'s verification commands named tests that do not exist.** `go test -run <nonexistent>` exits **0** with `[no tests to run]`, so all ten would have certified unwritten tests. Rewritten to `go test -run '^X$' -v | grep -q '^--- PASS: X'`. Lesson 217 already described this exactly, and the file violated it anyway — a written lesson is not a guard.
- **The first `chains.top` schema edit was a loosening that reads as a tightening.** Naming `top` under `properties` exempts it from `additionalProperties`, so a bare `maxItems` dropped the array type, the minimum and the `pool:model` pattern. Caught by mutation, fixed with a sibling `$ref`, written up as **lesson 226**.

## Acceptance criteria blessed

The nine ACs were blessed in this PR (the spec was `status: draft`). Two were reworded against what PR A actually shipped, and both corrections came from measurement rather than from the plan:

- **AC6** — dropped "*served by NaN*". `hive delegate` describes "the configured worker" and names no provider on any shipped surface, by design. The pool name in the record is the dispatcher's claim read from `chains`, so the falsifiable form is *the model the chain named is the model that answered*.
- **AC8** — scoped to hive's **worker** (opencode's own `ollama`/`openrouter` provider catalogue is out of scope and must survive), and **not retired**: `HIVE_OLLAMA_ENDPOINT` is still deployed at `ai/agy/mcp_servers.json:10`. Hive dropping a provider upstream and this repo ceasing to deploy its endpoint are two changes; only the first has happened.

## Verification

Full evidence in `specs/CLI-042-dotf-agent-run/verification.md`.

- `go build ./... && go vet ./... && go test ./...` → **exit 0**, 18 packages ok
- `GOOS=windows go vet ./...` → clean; `gofmt -l .` → no output
- `golangci-lint run` (v2.12.2, matching the `versions.conf` pin) → **0 issues**
- `~/.local/bin/bats tests/*.bats` → **1462 tests, exit 0, 0 failures**; shellcheck clean on the new file
- Every `features.json` command run individually: **10 exit 0, 7 exit non-zero**. The ten are f1–f7, f9, f12, f13 — note f9 belongs to AC3, which PR C owns; it is the half of that criterion B can hold (the deadline reaches the backend), and counting it under B would misreport AC3 as met. The seven failures are the criteria C/D/E own, which is what `pending` has to mean

**Mutation checks**, because a passing test is not evidence that a test *would* fail:
- Top-tier rule removed from `Dispatch` → all four assertions in `TestDispatch_TopTierNeverDegrades` fail.
- `$ref` removed from `chains.top` → 3 of 5 schema cases fail; `chains.top: "claude:opus"` and `chains.top: []` are both **accepted**.

Total added lines exceed the cap, so the breakdown is declared here. Production Go is **483 lines**, of which **~170 are executable**: 127 comment lines, 49 blank, 115 declarative (type/const/struct fields, flag registrations), 22 cobra help prose. The remainder of the diff is tests (~640 lines), the bats smoke (114), spec artefacts (258) and the schema (26) — all excluded by the executable-lines rule.

## Not in this PR

The semaphore, timeout kill and cancellation (**C**); the real subprocess and hive backends and the end-to-end smoke (**D**); the IaC and the `dotf doctor` reachability check (**E**). The two open questions in `proposal.md` — the `nan` backend tie-break and the hive secret-delivery shape — are scoped to D and E and are unreachable from here: there is no second backend to tie-break between yet, and no daemon to deliver a secret to.

### DIFF
--- a/cli/cmd/dotf/main.go
+++ b/cli/cmd/dotf/main.go
@@ -13,6 +13,10 @@ var version = "dev"
 
 func main() {
 	if err := cmd.New(version).Execute(); err != nil {
-		os.Exit(1)
+		// Not a bare 1: `dotf agent run` distinguishes "no pool could serve
+		// this" from "the task failed", and that distinction has to survive the
+		// process boundary or a composer cannot act on it. Everything else is
+		// untagged and still exits 1.
+		os.Exit(cmd.ExitCode(err))
 	}
 }
--- a/cli/internal/agent/agent.go
+++ b/cli/internal/agent/agent.go
@@ -1,0 +1,129 @@
+// Package agent is the execution layer ADR-032 §2 defines: "run role A on task
+// T, isolated in W, return R". It owns the seam and the chain walk; it does not
+// own the routing map. The caller resolves `chains` through
+// harness.ResolveChain and hands the result in, so the registry keeps one
+// reader and this package keeps no opinion about where a chain comes from.
+package agent
+
+import (
+	"context"
+	"time"
+)
+
+// Status is a dispatch outcome carried as a VALUE, not as an error type.
+//
+// The reason is the cross-repo contract: this classification crosses a process
+// boundary (a `hive delegate` exit code, a subprocess exit code) and a JSON
+// boundary (the record on stdout). Exception types do not survive either, so
+// the vocabulary has to be data on both sides. `mlorentedev/hive#395` fixed the
+// same shape upstream for the same reason.
+type Status string
+
+const (
+	// StatusOK — the backend ran the task and it succeeded.
+	StatusOK Status = "ok"
+	// StatusTaskFailed — the backend ran the task and it failed. This does NOT
+	// advance the chain: retrying a genuine failure on a different model turns a
+	// bad answer into a silent second opinion (ADR-032 §2).
+	StatusTaskFailed Status = "task_failed"
+	// StatusPoolUnavailable — the backend could not reach the pool at all, so
+	// the task never ran. This advances the chain. It is never a final status:
+	// the dispatcher resolves it to the next entry, or to chain_exhausted.
+	StatusPoolUnavailable Status = "pool_unavailable"
+	// StatusChainExhausted — every entry in the chain reported unavailable, so
+	// nothing ran anywhere. Distinct from task_failed on purpose: one means "no
+	// answer exists", the other means "the answer was wrong".
+	StatusChainExhausted Status = "chain_exhausted"
+	// StatusEscalated — the top tier's only entry was unavailable and the
+	// dispatcher refused to degrade (ADR-032 §4). Distinct from chain_exhausted
+	// because no chain was exhausted; a fallback was declined.
+	StatusEscalated Status = "escalated"
+	// StatusDryRun — the routing was resolved and deliberately not executed.
+	StatusDryRun Status = "dry_run"
+)
+
+// TierTop is the tier that must never degrade. Named rather than inlined
+// because the no-fallback rule reads as arbitrary at its use site otherwise.
+const TierTop = "top"
+
+// OutputCap bounds the captured output the record carries, per ADR-032 §2's
+// "captured output bounded by an output cap". A dispatcher composing many
+// records must not have one verbose worker blow up the consumer, and stdout
+// here is parsed by a machine, not scrolled by a person.
+const OutputCap = 64 * 1024
+
+// Request is what the dispatcher hands a backend for ONE chain entry. Pool and
+// Model are already resolved: a backend chooses nothing about routing.
+type Request struct {
+	Pool    string
+	Model   string
+	Role    string
+	Task    string
+	Cwd     string
+	Timeout time.Duration
+}
+
+// Response is what a backend reports for one attempt. Backends return the four
+// dispatch-level statuses (ok, task_failed, pool_unavailable, dry_run); the two
+// chain-level ones (chain_exhausted, escalated) are the dispatcher's to emit.
+type Response struct {
+	Status Status
+	Exit   int
+	Output string
+}
+
+// Backend is the seam. One method, the five semantics of ADR-032 §2, and no
+// backend-specific type in either direction — that is what keeps subprocess,
+// hive and (later) orca interchangeable rather than merely coexisting.
+//
+// It returns no error: a transport failure IS one of the statuses, and a
+// backend that could report a failure two different ways would let the
+// dispatcher's classification be bypassed.
+type Backend interface {
+	Dispatch(ctx context.Context, req Request) Response
+}
+
+// Classify maps a backend's reported status onto the dispatcher's vocabulary,
+// failing CLOSED: anything unrecognised is a task failure, never an
+// unavailability.
+//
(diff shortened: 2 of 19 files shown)


mlorentedev:
## Review triage

CI green on `7bee37b` (all checks pass; `goreleaser release` skips on a PR, as designed). Two reviewers reported: **PR-Agent** — *"No major issues detected"*, *"No security concerns identified"*, ticket compliance **partially compliant** with every non-compliant item explicitly marked out of scope for PR B. **CodeRabbit** — 8 findings (3 Major, 5 Minor). Each verified against the tree before acting; 7 applied, 1 declined. Fixes in `e01bde5`.

| # | Finding | Verdict | Disposition |
|---|---|---|---|
| 1 | `dispatch.go` — a malformed chain entry emits `"status":"task_failed","exit":0` | **Major, confirmed** | **Applied.** Real and mine: every other terminal path sets `rec.Exit`, that one did not, so the record contradicted itself while the process exited 1. The table now pins `rec.Exit` per case *separately* from the derived `ExitCode(rec.Status)` — the old assertion checked only the derived value, which is why the field was unpinned. |
| 2 | `agent.go` — `capOutput` cuts at a byte boundary, so it can split a rune | **Minor, confirmed** | **Applied**, and it cost a second round. My first fixture used a 2-byte rune and **passed against the unfixed code**: `OutputCap` is 65536, even, so the run divides exactly and the cut lands on a boundary by arithmetic rather than by correctness. Re-done with a 3-byte rune (65536 mod 3 == 1); the mutation now kills it. Reviewer was also right that the comment said "in-band" when the marker is the separate `truncated` field. |
| 3 | `features.json` — `go test … \| grep -q` masks a non-zero `go test` status | **Major, confirmed** | **Applied** to all 13 Go commands. Measured: `(printf '%s' '--- PASS: X'; exit 1) \| grep -q 'PASS: X'` → **exit 0**. Reachable for real on f4, which greps a *subtest* line, so a sibling subtest failing would still certify. Rewritten as `out=$(go test …) && printf '%s' "$out" \| grep -q …` rather than `pipefail`, which is not POSIX and depends on the harness's shell. |
| 4 | `features.json` — the AC6 predicate accepts `{"pool":"nan","status":"ok","model":""}` | **Major, confirmed** | **Applied.** f14 now compares `.model` against `dotf harness resolve-tier mid --harness pi`. AC6's whole claim is *the model the chain named is the model that answered*, so a predicate that ignores `.model` cannot verify it. |
| 5 | `proposal.md` — AC1's example command omits `--backend`, which is required | **Major, confirmed** | **Applied.** With `--backend` mandatory until probing lands, the literal AC1 command refused before emitting the record AC1 requires — a criterion that could not be executed as written. |
| 6 | `tasks.md` — two task entries carry the wrong `[AC<n>]` tag | **Minor, half refuted** | **Half applied, half declined.** The exit-code entry was `[AC1]` and serves AC2 → fixed. The `chains.top` schema entry was **already `[AC5]`**, not `[AC1]` as the finding claims. Declined that half: nothing to change. |
| 7 | `verification.md` — "all seventeen" commands were hardened, but four are not Go | **Minor, confirmed** | **Applied.** 13 Go, 4 not (f2/f15/f16 bats, f14 jq). The overclaim is the same species as the defect it was describing, one level up. |
| 8 | `tests/dotf-agent-run.bats` — `python3` is unprobed while `go` and `jq` are | **Minor, confirmed** | **Applied.** Three cases would have gone **red** on a checkout without python3 instead of skipping — contradicting the file header's own promise, and the failure mode #807/BUG-055 exist for. |

### Evidence after the fixes

- `go build ./... && go vet ./... && go test ./...` → **exit 0**; `GOOS=windows go vet ./...` clean; `gofmt -l .` empty
- `golangci-lint run` (v2.12.2, matching the `versions.conf` pin) → **0 issues**
- `~/.local/bin/bats tests/*.bats` → **exit 0, 1462 tests, 0 failures**; shellcheck clean on the changed bats file
- `features.json` re-run after hardening: **10 pass / 7 fail**, unchanged — the hardening did not silently drop a command
- Mutations re-confirmed: reverting the rune fix fails `TestDispatch_OutputIsCutOnARuneBoundary`; reverting `rec.Exit` fails the malformed-entry case

### Note on the reviewer

Both reviewers ran and both produced substantive output — this is a review, not a notice. CodeRabbit's initial comment was a *"processing new changes"* notice against `363503b`; the real review landed afterwards. Finding 6 is the reminder to verify before applying: it named two locations and only one existed as described.

PR #1212, opened 2026-08-23
feat(cli): bound dispatch concurrency and enforce the per-dispatch deadline

mlorentedev:
**PR C1** of the sequence in `specs/CLI-042-dotf-agent-run/tasks.md`, covering **AC3**. Follows #1209 (PR B, merged `7e734c4`). Refs #1190.

`tasks.md` is amended in this PR to record a **C → C1/C2 split**: the original single row carried both AC3 and AC4 and does not fit the atomic cap. Declaring the split here rather than discovering it at 350 LOC is the point. **C2** (machine identity, `pools.deny`, the wording assertion) follows after this merges — sequentially, never stacked (lesson 225).

## What lands

- **`cli/internal/agent/semaphore.go`** — per-pool slots, capacity `concurrency − reserve_interactive` (3 of NaN's 5).
- **`lock_unix.go` / `lock_windows.go`** — the platform half behind build tags.
- **`Dispatch`** — the backend now runs on its own goroutine, selected against the deadline; `timeout` joins the status vocabulary.
- **`declaredCapacity`** in the command — the map's declaration turned into the number the semaphore enforces.

## Read the guarantee narrowly

ADR-032 §3 is explicit and the code says the same thing in three places: this bounds **`dotf`-dispatched work only**. A hand-run `qq`, a pi TUI turn or a hive embedding call consumes a slot the semaphore never sees. The claim is ***`dotf` alone will never be the cause of exhaustion***, not that exhaustion cannot happen. (Asserting that wording in help and error text is AC4's, so it lands in C2.)

## Three distinctions the tests pin

Each one fails in the wrong direction if collapsed:

| Case | Behaviour | Why not the other way |
|---|---|---|
| Pool **full** | Advances the chain, without ever reaching the backend | An availability fact, indistinguishable to the caller from the backend reporting the pool unavailable |
| Semaphore state **unreadable** | **Refuses** — and does not classify as busy | Reading a broken counter as "zero in use" grants every slot exactly when the accounting is broken. Advancing would try the next pool against the same broken state |
| Pool declares **no concurrency** | Unbounded, recorded as `"unbounded": true` | Not capacity zero. A seat-based pool's concurrency is a fleet property; refusing on the map's silence enforces a limit nobody declared — overshooting the narrow guarantee as badly as overstating it |

## Why an OS file lock, not a marker file

The discriminator is **crash release**: the kernel drops a file lock when the holder dies, so a `dotf` killed with SIGKILL frees its slot, where an existence-based marker leaks one until something reaps it. It also makes AC3's sharpest clause trivial — releasing is closing our own descriptor, which never waits on the worker.

State lives under the **machine's** runtime dir, never the checkout's: two clones of this repo draw on one NaN subscription, so a per-checkout counter would bound neither. `--semaphore-dir` overrides it (the bats smoke uses it, so a test never takes real slots from whatever else is dispatching on the box).

## The deadline is enforced, not handed over

Passing a context and trusting it is the shape that fails silently. The backend runs on its own goroutine and its result is selected against the deadline, so a subprocess blocked on a read — or a remote that never answers — cannot hold the dispatcher open while `--timeout` claims otherwise. ADR-032 §2 rules a backend that cannot enforce a timeout ineligible; this makes the dispatcher's guarantee independent of the backend's cooperation.

`timeout` is a dispatcher-level status: **exit 1, and it does not advance the chain.** Exit 1 rather than 3 because 3 means *no pool could serve this, try elsewhere*, and a timed-out dispatch may have been served perfectly well and merely too slowly. Not advancing is the same anti-double-spend reasoning as `task_failed` — the task may be running still, and a second pool would be billed for work already in flight.

## The Windows half was not the obvious one

`syscall` **does not export `LockFileEx` on Windows** — it lives in `x/sys/windows`. The first version compiled clean on Linux, passed every local check, and broke the Windows leg: the #1075 shape that `GOOS=windows go vet` is in the verification loop to catch. Opening with `dwShareMode = 0` **is** the lock there, uses only symbols `syscall` does export, and keeps `golang.org/x/sys` an indirect dependency rather than promoting it to direct — which would itself be a Discipline Gate trigger.

## Verification

- `go build ./... && go vet ./... && go test ./...` → **exit 0**
- `GOOS=windows go build ./... && GOOS=windows go vet ./...` → clean (this caught a real break)
- `gofmt -l .` → empty; `golangci-lint run` (v2.12.2, matching the pin) → **0 issues**
- `~/.local/bin/bats tests/*.bats` → **exit 0, 1463 tests, 0 failures**; shellcheck clean
- `features.json` run individually: **16 pass / 6 fail** (was 10/7 after B). The six are AC4's machine half (f10, f11 → C2), AC6 (f14 → D), AC7–AC9 (f15–f17 → E)

**Mutations — four, each dying for a different reason**, which is the evidence the assertions are independent rather than one wearing four hats:

| Mutation | Result |
|---|---|
| Wait for the worker (`<-done`) before releasing | Dispatch never returns; the test's 5s guard fires |
| Don't release on the timeout path | Slot still held while the abandoned worker runs |
| `IsPoolBusy` → `false` | A full pool stops the walk instead of advancing; 2 tests fail |
| Every pool unbounded (through the compiled binary) | The bats saturation case fails — the dispatch goes to the pool whose slots are all held |

Production Go added is **~150 executable lines** (semaphore 60, the two lock files 35 combined, dispatcher changes 40, `declaredCapacity` 15); the rest of the Go diff is comments, declarations and the platform split. Tests are ~330 lines, the bats smoke ~45, spec artefacts ~130 — all excluded by the executable-lines rule.

## Heads-up for C2

ADR-032 §8 requires that a machine whose identity cannot be established **denies every non-local pool**. `machine.json` carries only `paths` today, so once C2 lands `dotf agent run` refuses on any machine that has not declared an identity — including this one and CI. That is the specified fail-safe direction rather than a regression, and it is why C2 also has to point the bats smoke at a fixture HOME. Flagging it now so it is not a surprise in review.


mlorentedev:
## Review triage

**CI green** on `f8e5b0e` — every check passes (`goreleaser release` skips on a PR, as designed).

**PR-Agent — review ran, verdict clean, nothing to disposition.** Ticket compliance **✅ PR Code Verified** against both #1190 and the PR B thread, listing as compliant every load-bearing claim this PR makes: the semaphore over the declared budget, the per-dispatch deadline enforced by the dispatcher rather than delegated to the backend, release-before-reap, timeout not advancing the chain, machine-scoped state, undeclared-budget-is-unbounded, and timeout exiting 1 rather than 3. *No security concerns identified*, no recommended focus areas. Its one "requires further human verification" item is that end-to-end integration with real backends is deferred — which is the declared PR D scope, not a gap here.

**Disposition: no findings to apply.** Nothing was skipped or declined, because nothing actionable was raised.

**CodeRabbit did not review this PR, and that is a disclosure rather than a claim of review.** Its only comment is the walkthrough posted at 23:16 carrying *"Currently processing new changes in this PR"*; ~25 minutes later the API still reports **0 review comments and 0 submitted reviews** on this PR. That is a notice about the review, not a review: it names no file, no line and no claim. Recording it explicitly because a notice left unread is indistinguishable from a review that found nothing, and this repo has been bitten by exactly that once already.

Proceeding on one reviewer is deliberate and authorised for this PR: CodeRabbit is not always available and waiting on it costs more than the second opinion is worth here.

### Evidence

- `go build ./... && go vet ./... && go test ./...` → exit 0
- `GOOS=windows go build ./... && GOOS=windows go vet ./...` → clean. Not a formality: it caught a real break in this PR, since `syscall` does not export `LockFileEx` on Windows
- `gofmt -l .` empty; `golangci-lint run` (v2.12.2, matching the `versions.conf` pin) → 0 issues
- `~/.local/bin/bats tests/*.bats` → exit 0, **1463 tests, 0 failures**; shellcheck clean
- `features.json` run individually → **16 pass / 6 fail**, the six being AC4's machine half (C2), AC6 (D) and AC7–AC9 (E)
- Four mutations, each killed for a different reason — table in the PR body

PR #1210, opened 2026-08-23
chore(main): release 0.49.0

mlorentedev:
:robot: I have created a release *beep* *boop*
---

### Features

* **agent:** give hive's daemon its worker contract without a credential on disk ([#1230](https://github.com/mlorentedev/dotfiles/issues/1230)) ([981fd93](https://github.com/mlorentedev/dotfiles/commit/981fd9353afbd9d446bcff7812142ae6d1a1cfc0))
* **cli:** bound dispatch concurrency and enforce the per-dispatch deadline ([#1212](https://github.com/mlorentedev/dotfiles/issues/1212)) ([8171ac1](https://github.com/mlorentedev/dotfiles/commit/8171ac10badd8e1a92d059ece91e94426d7fd699)), closes [#1190](https://github.com/mlorentedev/dotfiles/issues/1190)
* **cli:** deny dispatch on a machine that has not declared who it is ([#1213](https://github.com/mlorentedev/dotfiles/issues/1213)) ([9531527](https://github.com/mlorentedev/dotfiles/commit/95315274eb994d64a4741a94e0d1366f02f1d05f))
* **cli:** dotf agent run dispatches over the tier chain ([#1209](https://github.com/mlorentedev/dotfiles/issues/1209)) ([7e734c4](https://github.com/mlorentedev/dotfiles/commit/7e734c4de291c306cdfde6fdbf96bc271224f925)), closes [#1190](https://github.com/mlorentedev/dotfiles/issues/1190)
* **cli:** probe real backends and route each chain entry to one that serves it ([#1227](https://github.com/mlorentedev/dotfiles/issues/1227)) ([4783a2d](https://github.com/mlorentedev/dotfiles/commit/4783a2dc316a377f06b5d9ab13d8893b8bff7ee6))
* **pi:** declare pi packages in a manifest setup reconciles on every run ([#1226](https://github.com/mlorentedev/dotfiles/issues/1226)) ([9c44d7a](https://github.com/mlorentedev/dotfiles/commit/9c44d7a7d1d6580baee122868d0807db0206407c))

### Bug Fixes

* **ci:** stop running the jobs a Dependabot PR cannot possibly pass ([#1223](https://github.com/mlorentedev/dotfiles/issues/1223)) ([c9e6674](https://github.com/mlorentedev/dotfiles/commit/c9e66747a1438e290d21b7880540e503f04f4b54))
[...]

PR #1213, opened 2026-08-23
feat(cli): deny dispatch on a machine that has not declared who it is

mlorentedev:
**PR C2** of the sequence in `specs/CLI-042-dotf-agent-run/tasks.md`, covering **AC4** and closing **AC1–AC5**. Follows #1212 (C1). Part of the CLI-042 sequence tracked by issue 1190 — deliberately written without a linking reference, because release-please aggregates any mention of a parent issue into the release changelog as a closing keyword, and the release PR then closes it. That is lesson 115, and it recurred today on #1210. PR E is the one that should close it.

`machine.json` gains two blocks beside `paths` — `machine.id` and `pools.deny` — read **at dispatch time and never cached** (ADR-032 §7). Caching them would let a level-2 dispatch from a personal session route work content through a personal pool on behalf of a denied machine.

## ⚠️ This gate refuses on your machine until you declare an identity

That is the specified behaviour, not a regression, and it is the point of the PR. ADR-032 §8: *a machine whose identity cannot be established denies every non-local pool*, so the unknown case degrades to **no cross-pool dispatch** rather than to **all pools allowed**. The failure it prevents is concrete — a rebuilt corporate machine that has not restored `machine.json` probes successfully for personal pools, and allowing them would be silent and in the direction nobody notices.

After merging, add to `~/.config/dotfiles/machine.json` (keeping any existing `paths` block):

```json
{
  "machine": { "id": "msi" },
  "pools":   { "deny": [] }
}
```

The refusal prints exactly that, with the resolved path filled in. Two things make the gate survivable rather than merely correct: **the remedy is in the message** (a fail-closed refusal whose fix must be looked up is one people route around), and **`machine.json.example` carries both blocks**, since that is the surface people copy.

`MachinePolicy`'s **zero value denies everything** — a caller that forgets to populate it gets the safe answer, not the permissive one. Fail-closed by construction rather than by remembering.

## Four distinctions the tests pin

Each fails in the wrong direction if collapsed:

| | |
|---|---|
| **Absent vs malformed** `machine.json` | Absent = unidentified, denies, no error. Malformed = hard error. Silently denying on a syntax error sends an operator hunting for a policy that does not exist |
| **A deny typo** | **Refused**, not ignored. `claud` in the list leaves `claude` allowed on a machine whose whole intent was to forbid it, and nothing would ever say so |
| **Denial vs the semaphore** | Denial is evaluated **first**, so a forbidden pool consumes nothing — not even a slot another dispatch could have used |
| **All-denied vs chain-exhausted** | Its own status, exit 3. One sends an operator to quota and outages, the other to `machine.json` |

The denial tests assert **the backend was never reached**, not merely that the record names another pool. A dispatcher that asked the denied pool and discarded its answer has already sent the task's content there — which is the entire thing denial exists to prevent, and a record-only assertion would pass on it.

## The wording clause is asserted, not left to review

`TestAgentRun_StatesTheNarrowGuaranteeAndNotAWiderOne` checks both directions: the help states ***`dotf` alone will never be the cause of exhaustion*** verbatim and names the shared consumers and the reserve, and four overstatements (`exhaustion cannot happen`, `prevents exhaustion`, …) are asserted **absent**. Overstating is the failure mode here — a reader who believes exhaustion cannot happen will not build for the case where it does.

## The bats smoke points at a fixture, deliberately

Once identity is required, the smoke's own dispatches refuse on any machine that has not declared one. The fixture `XDG_CONFIG_HOME` is the fix; pointing at the real file would make these cases pass **only on a machine whose owner had configured it** — green locally, red in CI — and the version that "fixed" *that* would be one that skipped the gate. Three new bats cases cover the refusal, its empty stdout, and a denied pool being skipped, all through the compiled binary.

## Verification

- `go build ./... && go vet ./... && go test ./...` → **exit 0**, 18 packages
- `GOOS=windows go build ./... && GOOS=windows go vet ./...` → clean
- `gofmt -l .` empty; `golangci-lint run` (v2.12.2, matching the pin) → **0 issues** (it caught an `ST1005` on the refusal text; fixed to the convention the `harness` messages already follow)
- `~/.local/bin/bats tests/*.bats` → **exit 0, 1466 tests, 0 failures**; shellcheck clean
- `features.json` run individually: **21 pass / 4 fail** (10/7 after B, 16/6 after C1). The four remaining are **AC6 → PR D** and **AC7–AC9 → PR E**. Every criterion this sequence has claimed so far is executable and green.

**Mutation:** replacing the narrow guarantee with *"This prevents exhaustion."* fails the wording test on both halves — the missing verbatim claim and the forbidden overstatement.

Production Go is **~110 executable lines** (`machinepolicy.go` 55, the dispatcher's denial branch and `allDenied` 20, the command's gate and `declaredPools` 35); the rest of the Go diff is comments, declarations and the refusal text. Tests ~390 lines, bats ~55, spec artefacts ~90, `machine.json.example` 8 — all excluded by the executable-lines rule.

## What is left in the spec

**PR D** — the real subprocess and hive backends, the probe and the tie-break, and the end-to-end smoke (AC6). **PR E** — the hive service unit via `dotf secrets run`, the Ollama/OpenRouter removal, and the `dotf doctor` reachability check (AC7–AC9). Both open questions in `proposal.md` belong to those two and are now the next thing to close, since they gate the PRs that follow rather than this one.

### DIFF
--- a/cli/internal/agent/agent.go
+++ b/cli/internal/agent/agent.go
@@ -41,6 +41,10 @@ const (
 	StatusEscalated Status = "escalated"
 	// StatusDryRun — the routing was resolved and deliberately not executed.
 	StatusDryRun Status = "dry_run"
+	// StatusDenied — this machine forbids the pool (ADR-032 §7). A policy fact,
+	// not an availability one, but it advances the chain for the same reason:
+	// another entry may name a pool this machine does allow.
+	StatusDenied Status = "denied"
 	// StatusTimeout — the dispatch outlived its deadline and was abandoned.
 	// A dispatcher-level status: no backend returns it.
 	//
@@ -124,7 +128,10 @@ func ExitCode(s Status) int {
 	switch s {
 	case StatusOK, StatusDryRun:
 		return 0
-	case StatusChainExhausted, StatusEscalated:
+	case StatusChainExhausted, StatusEscalated, StatusDenied:
+		// Denied joins the 3 family: like an exhausted chain it means nothing
+		// ran, and unlike a task failure another machine may well be permitted
+		// to run it. The record's status carries the reason.
 		return 3
 	default:
 		return 1
--- a/cli/internal/agent/deny_test.go
+++ b/cli/internal/agent/deny_test.go
@@ -1,0 +1,126 @@
+package agent
+
+import (
+	"context"
+	"testing"
+	"time"
+)
+
+func denyOnly(pools ...string) func(string) bool {
+	set := map[string]bool{}
+	for _, p := range pools {
+		set[p] = true
+	}
+	return func(pool string) bool { return set[pool] }
+}
+
+// A denied pool is skipped and the chain continues: denial is a fact about
+// THIS machine, and another entry may name a pool it does allow. The point of
+// per-pool denial would be lost if one denied entry failed the whole dispatch.
+func TestDispatch_ADeniedPoolIsSkippedAndTheChainAdvances(t *testing.T) {
+	be := &scriptedBackend{byEntry: map[string]Response{
+		"claude:sonnet":         {Status: StatusOK, Output: "must never be reached"},
+		"nan:deepseek-v4-flash": {Status: StatusOK, Output: "answered"},
+	}}
+
+	rec := Dispatch(context.Background(), Options{
+		Tier:    "mid",
+		Chain:   []string{"claude:sonnet", "nan:deepseek-v4-flash"},
+		Timeout: time.Minute,
+		Denied:  denyOnly("claude"),
+		Now:     fixedClock(time.Millisecond),
+	}, be)
+
+	if rec.Status != StatusOK || rec.Pool != "nan" {
+		t.Errorf("record = %s on %s, want ok on nan", rec.Status, rec.Pool)
+	}
+	// The denied pool must never have been dispatched to. Asserting only the
+	// final record would pass on a dispatcher that asked claude first and threw
+	// the answer away — which would already have sent the task's content to a
+	// forbidden pool, the exact thing denial exists to prevent.
+	if len(be.seen) != 1 || be.seen[0] != "nan:deepseek-v4-flash" {
+		t.Errorf("backend saw %v; a denied pool must not be reached at all", be.seen)
+	}
+	if len(rec.Attempts) != 2 || rec.Attempts[0].Status != StatusDenied {
+		t.Errorf("attempts = %+v; the denial must leave a trace", rec.Attempts)
+	}
+}
+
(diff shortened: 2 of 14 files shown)


mlorentedev:
## Review triage

**CI green** on `1d3aa60` — every check passes (`goreleaser release` skips on a PR, by design).

**PR-Agent — reviewed, and this one is unusually clean.** Ticket compliance against both #1190 and the C1 thread lists **zero non-compliant requirements and zero items requiring further human verification** — the first time in this sequence both columns came back empty. *No major issues detected*, *no security concerns identified*. It confirms each load-bearing claim individually: the two `machine.json` blocks read at dispatch time, the refusal before dispatch with an actionable error, `MachinePolicy`'s zero value denying everything, `ValidateDeny` rejecting an unresolvable entry, the narrow guarantee asserted in both directions, and the fixture HOME in the bats smoke.

**Disposition: nothing to apply.** No findings were raised, so none were skipped or declined.

**CodeRabbit did not review this PR — it is rate limited.** Its comment says so in as many words: *"Review limit reached"*, *"rate limited by CodeRabbit"*. Per the repo's own doctrine that is a **notice about the review, not a review**: it names no file, no line and no claim, and it leaves the PR unreviewed by that reviewer. Recording it explicitly, because a notice left unread is indistinguishable from a reviewer that found nothing — this repo has been bitten by exactly that once.

**Proceeding on one reviewer is deliberate and authorised.** CodeRabbit's free tier frequently has no review quota, so waiting on it costs more than the second opinion is worth. Disclosure over silence: one reviewer read this, and it was PR-Agent.

### Evidence

- `go build ./... && go vet ./... && go test ./...` → exit 0, 18 packages
- `GOOS=windows go build ./... && GOOS=windows go vet ./...` → clean
- `gofmt -l .` empty; `golangci-lint run` (v2.12.2, matching the pin) → **0 issues**. It earned its keep here: it flagged `ST1005` on the multi-line refusal text, fixed to the convention the `harness` error messages already follow
- `~/.local/bin/bats tests/*.bats` → exit 0, **1466 tests, 0 failures**; shellcheck clean
- `features.json` run individually → **21 pass / 4 fail**; the four are AC6 (PR D) and AC7–AC9 (PR E)
- Mutation: replacing the narrow guarantee with *"This prevents exhaustion."* fails the wording test on both halves — the missing verbatim claim and the forbidden overstatement

### One thing for the human reviewer, not the bots

This PR makes `dotf agent run` **refuse on any machine that has not declared an identity, including this one and CI**. That is ADR-032 §8's specified direction rather than a regression, and the PR body carries the two-line `machine.json` addition that clears it. Worth doing deliberately at merge time rather than discovering at the next dispatch.

Issue #1224, opened 2026-08-25
AI-030: pi packages are installed by hand and declared nowhere, so a redeploy does not reproduce them


mlorentedev:
pi packages (extensions, skills, prompts, themes shipped over npm) are currently
installed by hand and recorded nowhere. Nine are wanted on this machine, and the
repository has no way to put them on the next one.

## Why the obvious placement does not work

`ai/pi/settings.json` looks like the home for the `packages` array, and it is
not. That file is **seed-if-missing** (`setup-linux.sh:929-936`):

```sh
if [ -f "$PI_SETTINGS_DST" ]; then
    log_info "pi settings.json present, preserving local edits"
```

pi rewrites it at runtime (`lastChangelogVersion`, `theme`, the model picked in
the TUI), so setup seeds it once and never touches it again — a contract
`tests/pi-config.bats` already pins in both setup scripts. Adding `packages`
there therefore changes only what a **fresh** machine receives, and this machine
would never see it. That is the opposite of the requirement: this repository is
under active development and gets deployed repeatedly, with updates.

## Requirement

Declarative in the repo, reconciled on **every** `setup-{linux,windows}` run, on
a machine that already has pi and on a fresh one alike.

## Shape

- [ ] `ai/pi/packages.json` — the manifest. Nine entries, each **pinned to a
      version**. `npm:pkg` without `@version` floats, and a repository that pins
      its GitHub Actions by commit SHA should not declare nine floating npm
      packages that execute with full system access.
- [ ] A loop in `setup-linux.sh` **and** `setup-windows.ps1` reading the
      manifest and invoking `pi install npm:<pkg>@<version>`, guarded on `npm`
      and `pi` being present — the same convention as pi's own install.
- [ ] Idempotent: compare against the live `settings.json` `packages` array and
      skip what is already there, so a re-run reports no change.
- [ ] The `packages` array is **NOT** written into the seed `settings.json`.
      `pi install` writes it to the live file itself; one mechanism, one owner.
      This also sidesteps an open question — the upstream docs assert automatic
      installation of missing packages for **project**-scoped settings only, and
      say nothing about user scope.
- [ ] Test, following the `tests/pi-config.bats` convention: manifest is valid
      JSON, every entry carries a version, and both setup scripts consume it.

## Packages

Resolved against the registry 2026-08-24:

| Package | Latest |
|---|---|
| `pi-effort` | 0.0.8 |
| `pi-web-access` | 0.24.2 |
[...]

PR #1226, opened 2026-08-25
feat(pi): declare pi packages in a manifest setup reconciles on every run

mlorentedev:
Tracked by #1224. Spec: `specs/AI-030-pi-packages-manifest/`.

**Deliberately not a closing keyword**, here or in either commit footer. The spec
must archive before its issue closes, and release-please aggregates *any* issue
mention in a commit footer into the release's `closes` list regardless of the
`Refs` vs `Closes` keyword (lesson 115; the recurrence is its own open ticket).
The footers were rewritten before this PR opened for exactly that reason.

## The problem

pi packages — extensions, skills, prompts and themes shipped over npm — were
installed by hand on one machine and recorded nowhere. A redeploy did not
reproduce them and a second machine never received them.

## Why the obvious placement is wrong

`ai/pi/settings.json` looks like the home for the `packages` array. It is not:
that file is **seed-if-missing** (`setup-linux.sh`), because pi rewrites it at
runtime (`lastChangelogVersion`, `theme`, the model picked in the TUI). Setup
copies it once and never again — a contract `tests/pi-config.bats` already pins
in both setup scripts, and #754 is what happens when it is broken.

So a `packages` array declared there would reach a **fresh** machine and never an
existing one. That is the opposite of the requirement for a repository under
active development that gets deployed repeatedly with updates.

It would also be wrong on disk: `pi install` both writes the live array **and**
unpacks the package under `~/.pi/agent/npm/`. An array entry written by setup
would name a package that is not installed.

## The change

- `ai/pi/packages.json` — the declaration. Nine entries, each pinned, each with a
  `why`.
- A reconcile block in `setup-linux.sh` and `setup-windows.ps1` that reads the
  manifest, reads the live array, and installs the set difference through
  `pi install`. Additive only: a package present but undeclared is left alone,
  because uninstalling a human's own extension is not setup's decision.

Two traps the implementation had to survive, both measured rather than assumed:

**`$PI_BIN`, never `pi`.** On this machine `pi` is a shell function wrapping
`dotf secrets run`; it resolves a Bitwarden item before exec and fails on a
locked vault, while `~/.local/bin/pi --version` returns `0.84.2` with the vault
locked. A bootstrap must not require an unlocked vault to install an extension.

**Both entry shapes.** Upstream allows `"npm:pkg"` and
`{"source": "npm:pkg", ...}` (per-resource filtering). A reader handling only
strings would reinstall every filtered entry on every single setup run.

## Pinning

Upstream states it plainly: *"Pi packages run with full system access —
extensions execute arbitrary code and skills can instruct the model to run
executables."* That is inside an agent holding `NAN_API_KEY` with write access to
these repositories. A repository that pins its GitHub Actions **by commit SHA**,
because a tag is mutable and belongs to upstream, does not then declare nine
floating npm packages. `tests/pi-packages.bats` refuses an unpinned entry — and
that guard is itself proven against `npm:pi-effort` (rejected) and
`npm:@ayulab/pi-rewind@0.4.6` (accepted), because an assertion whose regex
accepts everything is not an assertion (#1203).

What pinning does **not** do is review the nine. That is stated in the spec's
open risks rather than implied away.

## Verification

`verify-reconcile.sh` extracts the reconcile block from `setup-linux.sh` **by
anchor** and drives it against a stubbed `pi`. Extracting rather than restating
is the point: a copy of the logic would keep passing after the shipped block
changed, and if the anchors move, extraction fails loudly instead of testing
nothing.

```
$ specs/AI-030-pi-packages-manifest/verify-reconcile.sh
[OK] AC4  first run installed all 9 declared packages
[OK] AC5  second run installed 0 (changed=0)
[OK] AC6  object-form entries recognised, 0 reinstalled
[OK] AC7  pi absent: warned, exit 0, bootstrap continues
[OK] AC4-AC7 verified against the block extracted from setup-linux.sh:954-1003
exit 0

$ bats tests/*.bats
1491 ok, 0 not ok   (BATS_EXIT=0)

$ bash -n setup-linux.sh && zsh -n setup-linux.sh
OK / OK

$ shellcheck setup-linux.sh
20 findings before, 20 after, none in the new block

$ shellcheck specs/AI-030-pi-packages-manifest/verify-reconcile.sh
clean

$ # PowerShell: non-ASCII line count
10 before, 10 after
```

## What was NOT done

**The reconcile has not been run against the real `pi` on this machine.** Doing
so installs nine third-party packages, which upstream documents as running with
full system access, into the owner's live agent — the owner's call, not the
implementing session's. The stub proves *which call is made* and that the set
difference is correct; only a real run proves `pi install` accepts these
arguments. That gap is precisely the limitation `tests/stub-real-pairing.bats`
exists to keep visible (BUG-055), so it is written down rather than papered over.

`tests/pi-packages.bats` stubs nothing, so it incurs no pairing obligation.

## Found in scope, filed not fixed

`dotf spec init` stamped this spec `created: "2026-08-25"` while the local date
was 2026-08-24 — it formats a calendar date from a UTC clock, which is the defect
#1217 fixed for `dotf mem` and lesson 228 already generalises. On this machine
(UTC-6) the window is six hours wide every day. Filed with its own ticket and
**left uncorrected in this PR** so the evidence survives the fix; correcting the
frontmatter by hand would hide it.

## Follow-up before the spec archives

`dotf spec review AI-030-pi-packages-manifest` — an independent adversarial
review, which the implementing session cannot perform. `dotf spec archive`
refuses without a fresh passing `review.md`, so skipping it blocks the archive
rather than merely weakening it.

### DIFF
--- a/ai/pi/README.md
+++ b/ai/pi/README.md
@@ -10,6 +10,7 @@ SSOT (see root `AGENTS.md`).
 |--------|---------------|-------|
 | `models.json` | `~/.pi/agent/models.json` | NaN custom provider. `apiKey` is `{env:NAN_API_KEY}` in source; `setup-{linux,windows}` inject the literal at deploy time (SDD-009 pattern) so the deployed config is self-contained cross-OS and the key is never committed. |
 | `settings.json` | `~/.pi/agent/settings.json` | UX defaults + curated `enabledModels`. **Seed-if-missing**: pi mutates this file at runtime (`lastChangelogVersion`), so setup deploys it only when absent and never clobbers local edits. |
+| `packages.json` | (not deployed — reconciled) | Declared pi packages, each pinned. Setup installs the difference against the live `settings.json` on every run, through `pi install`. See below. |
 | (canonical `AGENTS.md`) | `~/.pi/agent/AGENTS.md` | Cross-agent SSOT system prompt, deployed verbatim (same as opencode). |
 
 Not managed: `auth.json` (OAuth/secret state) and `skills/` (runtime symlinks).
@@ -21,6 +22,31 @@ machine that already has `~/.pi/agent/settings.json`, edit that file too, or del
 re-run setup to take the committed defaults wholesale. `tests/pi-config.bats` pins that
 contract in both setup scripts — until #754 they compared source against destination and,
 because the deployed file always differs, overwrote it on every run.
+
+## Packages
+
+`packages.json` declares the pi packages (extensions, skills, prompts, themes)
+this environment wants. `setup-{linux,windows}` reconciles it against the live
+`~/.pi/agent/settings.json` on **every** run and installs what is missing, so an
+existing machine converges on the next setup and a fresh one on its first
+(AI-030, #1224).
+
+It is **not** a deployed file, and the `packages` array deliberately does not
+live in `settings.json` above. That file is seed-if-missing, so anything
+declared there would reach a fresh machine and never this one. `pi install`
+writes the live array itself — and unpacks the package to disk while doing it,
+which an array entry written by setup would not.
+
+Every entry is pinned. Upstream is explicit that packages *"run with full system
+access — extensions execute arbitrary code and skills can instruct the model to
+run executables"*, inside an agent holding `NAN_API_KEY`. `tests/pi-packages.bats`
+refuses an unpinned entry. Bumping one is a deliberate edit here, which puts a
+diff and a reviewer between upstream's publish and the next setup run.
+
+Adding one by hand (`pi install npm:pkg@1.2.3`) works and writes the live array,
+but nothing else will ever know about it — put it in `packages.json` instead.
+
+Docs: <https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/packages.md>
 
 ## Install
 
--- a/ai/pi/packages.json
+++ b/ai/pi/packages.json
@@ -1,0 +1,78 @@
+{
+  "$comment": [
+    "Declarative manifest of pi packages setup reconciles (AI-030, #1224).",
+    "",
+    "A pi package ships extensions, skills, prompts and themes over npm. The",
+    "installer is `pi install npm:<name>@<version>`, which does TWO things: it",
+    "installs under ~/.pi/agent/npm/ and it writes the entry into the `packages`",
+    "array of the LIVE ~/.pi/agent/settings.json.",
+    "",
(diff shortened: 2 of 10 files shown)


mlorentedev:
## Review triage

### PR-Agent — 1 finding, APPLIED (`e579944`)

**"Missing Requirement: the reconcile is guarded on `pi` but not on `npm`"** — valid,
and it was checked against this spec's own acceptance criterion, which asked for
both guards and got one.

`pi install` shells out to npm. Guarding only on `pi` means that on a machine
where the Node toolchain has gone missing, the loop runs and every entry fails
separately: nine package failures reported instead of one missing toolchain. A
symptom three layers from its cause — and the block immediately above this one
already gates pi's own install on npm for exactly that reason.

Applied to both scripts. AC7 now covers both halves and is **executable** rather
than asserted: the new scenario runs the real block against a PATH holding only
what it needs, npm deliberately absent, so it drives `command -v npm` instead of
grepping for the text of it.

```
[OK] AC7  pi absent: warned, exit 0, bootstrap continues
[OK] AC7  npm absent: warned once, 0 install attempts, exit 0
```

Proven able to fail — with the guard removed:

```
[FAIL] AC7b: npm absent produced no npm-specific warning:
         [info] pi packages already reconciled (9 declared, 0 changed)
```

Two harness defects surfaced while writing it, both fixed: `PATH=... bash`
resolves `bash` itself through the masked PATH (the interpreter is now resolved
absolutely up front), and the scenario now asserts npm *is* present on the host
before masking it — otherwise it would have passed without masking anything.

**Re-review at `e579944`: "Fully compliant", "No security concerns identified",
"No major issues detected".**

### CodeRabbit — did not review

`Review rate limited`. That is a **notice, not a review**: it talks about the
review, not about this change. Disclosed rather than counted — the attestation on
this PR comes from PR-Agent alone, and CodeRabbit's independent second look has
not happened. `@coderabbitai review` when its quota returns, if a second opinion
is wanted before merge.

### Still owed before archive

`dotf spec review AI-030-pi-packages-manifest` — the independent adversarial
review. `dotf spec archive` refuses without a fresh passing `review.md`, and the
implementing session cannot be the reviewer.

### Evidence

```
$ bats tests/*.bats
1492 ok, 0 not ok   (BATS_EXIT=0)

$ specs/AI-030-pi-packages-manifest/verify-reconcile.sh
AC4-AC7 all OK, exit 0
```

PR #1223, opened 2026-08-25
fix(ci): stop running the jobs a Dependabot PR cannot possibly pass

mlorentedev:
Every Dependabot PR has opened with red checks since 2026-08-07. All of them
trace to one mechanism, and neither job could ever have gone green.

## Measurement

GitHub serves Dependabot-triggered `pull_request` runs from a **separate**
secrets store. This repository's is empty:

```
gh api repos/mlorentedev/dotfiles/actions/secrets    --jq '.total_count'  -> 3
gh api repos/mlorentedev/dotfiles/dependabot/secrets --jq '.total_count'  -> 0
gh api repos/mlorentedev/dotfiles/actions/runs/32698096060 -> actor=dependabot[bot]
```

`secrets.BITACORA_PAT` and `secrets.NAN_API_KEY` therefore expand to the empty
string — no error, no warning. Everything observed on #1219 and #1220 follows:

| Check | Failing step | Cause |
|---|---|---|
| `add-to-project` | "Add to bitácora board" | `GH_TOKEN=""` -> GraphQL auth fails. The script classifies correctly: not a rate limit, so red. |
| `review` | "Fail if no review was published" — **not** the PR-Agent step, which exits `success` | `OPENAI__KEY=""` -> LiteLLM cannot authenticate, PR-Agent swallows it and exits clean. The #1109 guard exposes it. |
| `review-attestation` | — | Downstream. The only one of the three that is a required status check. |

Two details that shaped the fix:

- **The existing fork guard could never catch this.** `add-to-project.yml` skips
  fork PRs for exactly this reason, but a Dependabot PR is a branch in this repo,
  so `head.repo.fork` is false.
- **Re-running does not help.** Secrets access follows the actor of the original
  event. Run 32698096060 was re-run by a human (`triggering_actor=mlorentedev`,
  `actor=dependabot[bot]`) and failed identically.

## Why it surfaced now

GitHub's behaviour dates from 2021. The gates do not:

```
2026-08-05  last merged Dependabot PR (#742, #743)
2026-08-07  #813   add-to-project starts failing hard instead of dropping items silently
2026-08-15  #1019  review-attestation   +   #1032  pr-agent
2026-08-19  #1109  fail the review job when it published no review
```

#1219/#1220 are the first Dependabot PRs to cross all three.

## The change

Both skips are keyed on `github.actor`, which is **the same field GitHub reads to
choose the secrets store**. So a skip fires exactly when the credential is absent,
[...]

Issue #1214, opened 2026-08-23
GUARD: a release PR can close a multi-PR issue, and only a remembered lesson stops it


mlorentedev:
## What happened

PR #1210 (`chore(main): release 0.49.0`) arrived with `closes [#1190](…)` twice in its body — once for #1209, once for #1212. GitHub honours closing keywords in a PR body, so merging it as generated would have **closed #1190 while PRs D and E of that sequence are unstarted**. Caught during triage of #1210 and neutralised by hand; #1213 was amended so the next release does not repeat it.

## Root cause

`release-please` aggregates **any** mention of an issue in a commit footer into the changelog as a closing keyword. `Refs #1190` and `Closes #1190` are indistinguishable to it. #1209 and #1212 both used `Refs #1190`, believing that was the safe form.

## This is a recurrence, and that is the point of the ticket

`docs/lessons/lesson-115-release-please-can-close-a-multi-pr-issue-from-a-b.md`, written **2026-06-21** after the same failure on #522/#509, states the rule precisely:

> A build-only sub-PR of a multi-PR issue must not reference the parent issue in its footer *at all* — release-please aggregates any issue mention into the release's `closes` list regardless of the `Refs` vs `Closes` keyword.

The lesson was correct, discoverable, and violated three times in one day by an author who had not read it. **An instruction that must be remembered is not a guard** — which is the repo's own incident-to-guard doctrine, applied to itself.

## What a fix looks like

A check, not more prose. Options, cheapest first:

1. **A CI assertion on release PRs** — fail when the body carries a closing keyword against an issue that is not itself a release-scoped one. It runs exactly where the damage happens and needs no author cooperation.
2. **A pre-push / commit-msg guard** — reject a footer referencing an issue that has an open sub-PR sequence. Stops it earlier but needs to know what a "sequence" is, which is more machinery than the problem warrants.
3. **A post-merge check** — assert every issue closed by a release had all its acceptance criteria met. Strongest, and the most work.

Option 1 is the one that matches the failure: the release PR body is the artefact that does the closing, and it is machine-checkable with a grep.

## Acceptance

- [ ] A release PR whose body carries a closing keyword against a non-release issue fails a check
- [ ] The check is tested against a fixture body that reproduces #1210's exact shape — including the markdown form `closes [#1190](url)`, which a naive `closes #1190` grep misses
- [ ] Lesson 115 is updated with this recurrence and points at the guard

## Notes

[...]

PR #1227, opened 2026-08-25
feat(cli): probe real backends and route each chain entry to one that serves it

mlorentedev:
PR D of the CLI-042 sequence, tracked by issue 1190 — which this PR does **not**
close. PR E is the one that ends it. No commit in this branch names the issue in
its footer either, because
release-please aggregates any such mention into the release changelog as a
closing keyword (lesson 115).

Covers the machinery half of **AC6**. The end-to-end smoke stays blocked on
environment, and `features.json` says so rather than claiming a pass.

## The shape

A `Router` that is **itself a `Backend`**, selecting per chain **entry** rather
than per dispatch. Forced by the data, not by taste: `chains.mid` mixes
`claude:sonnet` with `nan:deepseek-v4-flash`, and hive serves only nan, so a
backend chosen once per dispatch could not walk that chain. Making the router a
`Backend` is what leaves `Dispatch` untouched — the seam absorbs the fan-out
instead of the dispatcher learning what a backend is.

A forced `--backend` that cannot serve an entry reports it `pool_unavailable` so
the chain advances. That is what makes AC6's own smoke work as written
(`--backend hive --tier mid` must skip `claude:sonnet` and answer on nan);
failing instead would silently redefine the flag as "only this backend AND only
pools it serves", which no criterion asks for.

Both real backends ride one process helper — deadline kill, output capture and
exit-code handling exist once. Task text goes on **stdin** for claude and pi
(argv is world-readable through `ps` and bounded by `ARG_MAX`); `hive delegate`
requires `--prompt` on argv, which is its shipped contract.

## Two measurements that changed the design

**`NAN_API_KEY` is empty in any ordinary environment here.** It is a JIT secret
(ADR-028) and pi decrypts it itself — so the map's declared `env:NAN_API_KEY`
probe would report nan *unreachable on a machine where it is reachable*.
Dispatch-time servability is probed as `bin:pi` instead. The map declares what a
pool **is**; the transport answers whether it can serve it **right now**, and
those are not the same question. The divergence is recorded in the spec rather
than left implied.

**`printf … | pi --print` exits 0, answers nothing, and warns on stderr** when
the credential is absent. So **exit 0 with empty stdout is `task_failed`**: a
successful record containing no answer is worse than a failure, because nothing
downstream looks twice. That also required `runProcess` to stop discarding
stderr on a zero exit — which is where the only explanation lived.

## Two defects caught by tests, not by review

- **`dry-run` was answering everything.** It serves every pool by construction,
  so leaving it in probe order made it the silent answer to any dispatch naming
  no backend — a dispatch that ran nothing and exited 0. That is exactly the
  option PR B's table rejected, reintroduced by accident. It is now
  `ExplicitOnly`: skipped by the probe, reachable only as `--backend dry-run`.
- **Command tests were dispatching for real.** Once `--backend` became optional,
  one run reached `claude -p`, took 35s and spent a live request. Tests now
  isolate `PATH`.

## Stated, not hidden

Neither `claude -p` nor `pi --print` has an exit code meaning "rate limited",
and grepping stderr for one is the fragile proxy this repo keeps catching. So
**a 429 through the subprocess backend reads as `task_failed` in v1.** That is
the fail-closed direction — it stops rather than spending a second pool on work
that may already have been billed.

## Verification

Re-run after rebasing onto current main, which had moved six commits (#1215's
suite-wide no-GUI guard, #1217, #1223, both dependabot bumps, #1226). #1215
matters here specifically: this is the PR that spawns subprocesses, and that
guard landed after these tests were written. The interaction is clean.

```
$ go build ./... && go vet ./...
OK / OK

$ GOOS=windows go vet ./...
OK

$ go test -count=1 ./...          # -count=1: cached results are not evidence after a rebase
exit 0, 0 FAIL lines

$ golangci-lint run                # pinned 2.12.2, matches versions.conf (BUG-071)
0 issues.

$ bats tests/*.bats
1501 ok, 0 not ok  (BATS_EXIT=0)
```

## Still open in CLI-042 after this

**PR E** — AC7-AC9: the hive service unit via `dotf secrets run`, the
Ollama/OpenRouter removal, and the `dotf doctor` reachability check. That is the
PR that ends issue 1190.

> Written as "issue 1190", never `closes #1190`. The first draft of this body
> used the keyword form and **spec-gate rejected the PR** — correctly: GitHub
> honours a closing keyword in a PR body, so merging it would have closed the
> epic with PR E unwritten and the CLI-042 spec still active. The same trap as
> #1210, caught this time by a check rather than by someone remembering.

### DIFF
--- a/cli/internal/agent/backends.go
+++ b/cli/internal/agent/backends.go
@@ -1,0 +1,125 @@
+package agent
+
+import (
+	"context"
+	"fmt"
+	"math"
+	"time"
+)
+
+// harnessFor maps a pool to the harness binary that reaches it, and how that
+// binary is asked for one non-interactive answer.
+//
+// This table is Go rather than another block in model-map.json, deliberately.
+// The map declares WHAT exists — pools, tiers, chains, budgets — and adding
+// argv to it would make it declare how to invoke things too, which is the
+// "map, never a director" drift ADR-035 §4 exists to stop. Two entries also do
+// not justify a schema change; a third harness is the trigger to revisit.
+var harnessFor = map[string]struct {
+	bin  string
+	args func(model string) []string
+}{
+	"claude": {bin: "claude", args: func(m string) []string { return []string{"-p", "--model", m} }},
+	"nan":    {bin: "pi", args: func(m string) []string { return []string{"--print", "--model", m} }},
+}
+
+// Subprocess dispatches by running a harness binary in non-interactive mode.
+// It is the floor bootstrap guarantees: no daemon, no protocol, no state.
+type Subprocess struct{}
+
+// Serves answers whether this machine has the binary that reaches the pool.
+func (Subprocess) Serves(pool string) bool {
+	h, ok := harnessFor[pool]
+	return ok && binaryPresent(h.bin)
+}
+
+func (Subprocess) Dispatch(ctx context.Context, req Request) Response {
+	h, ok := harnessFor[req.Pool]
+	if !ok {
+		return Response{Status: StatusPoolUnavailable,
+			Output: fmt.Sprintf("no harness binary is mapped to pool %q", req.Pool)}
+	}
+	stdout, stderr, code, err := runProcess(ctx, h.bin, h.args(req.Model), req.Task)
+	return classifyProcess(stdout, stderr, code, err)
+}
+
+// Hive dispatches through `hive delegate`, launched as a subprocess.
+//
+// Argv is its transport, not a second mechanism: the seam's hard semantics —
+// kill on timeout, release the slot without waiting — are what the subprocess
+// backend already implements, and the hive verb rides that code path verbatim.
+// An MCP client in Go would add a dependency, a handshake that can drift, and a
+// surface bats cannot smoke.
+type Hive struct{}
+
+// hivePool is the only pool hive's worker serves. Its worker became NaN-only
+// upstream (mlorentedev/hive#384), which is what aligns it with this map.
+const hivePool = "nan"
+
+func (Hive) Serves(pool string) bool { return pool == hivePool && binaryPresent("hive") }
+
+// Exit codes are the cross-repo contract, pinned on both sides. `hive delegate`
+// documents them in its own help: 3 means the pool would not serve the request
+// (try the next entry), 1 means the worker answered with a failure (do not).
+const (
+	hiveExitTaskFailed  = 1
+	hiveExitUnavailable = 3
+)
+
+func (Hive) Dispatch(ctx context.Context, req Request) Response {
+	args := []string{
+		"delegate",
+		"--model", req.Model,
+		"--timeout", hiveTimeoutSeconds(req.Timeout),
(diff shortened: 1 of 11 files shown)


mlorentedev:
## Review triage

All 13 checks green. Dispositions below.

### spec-gate — FAILED, and it was right. FIXED.

The first version of this PR body ended with *"That is the PR that closes
#1190."* GitHub honours a closing keyword in a PR body, so merging it would have
**closed the CLI-042 epic with PR E unwritten and the spec still active**. The
gate refused:

```
         CLI-042-dotf-agent-run (#1190)
```

Reworded to prose (`issue 1190`), which is the same convention the commit
footers in this branch already use. Re-verified locally before pushing the edit:

```
[OK] Production diff 463 LOC >= threshold 50 AND spec folder touched in diff
```

Worth naming plainly: this is the **#1210 trap**, in a PR whose author had spent
the day writing about it. It was caught by a check rather than by remembering —
which is the whole argument of #1214.

### PR-Agent — 1 security concern, 1 finding

**Security: the hive backend passes task text on argv** → **TICKETED as #1228**.
Real. argv is world-readable through `ps` and bounded by `ARG_MAX`. Not
introduced here — `hive delegate` requires `--prompt` on argv, which is its
shipped contract, and the subprocess backend already uses stdin precisely to
avoid this. Fixing it means changing hive, so it does not belong in this PR.

**"Subprocess and Hive do not set pool/model in the Response"** → **DECLINED.**
The dispatcher owns those fields by design, and does set them:

```
dispatch.go:136   Pool: pool, Model: model, Role: opts.Role, Task: opts.Task,
dispatch.go:145   rec.Status, rec.Pool, rec.Model = StatusTimeout, pool, model
dispatch.go:155   rec.Attempts = append(rec.Attempts, Attempt{Pool: pool, Model: model, ...})
```

The dispatcher walks the chain, so it is the component that *chose* the entry. A
backend echoing them back would be redundant and could disagree with the record —
two sources for one fact.

**"Non-compliant: semaphore / fail-closed / pools.deny / top-tier escalation"** →
**DECLINED.** Those are AC2-AC5, already merged in #1212 and #1213. The reviewer
compared the whole of issue 1190 against this PR alone; PR D is the AC6 slice.

**"Requires further human verification: the hive end-to-end smoke is blocked on
environment"** → **Correct, and already declared.** `features.json` records it as
failing rather than claiming a pass, and PR E is where the environment lands.

### CodeRabbit — did not review

`Review rate limited`. A **notice, not a review**. The attestation on this PR
comes from PR-Agent alone; disclosed rather than counted.

### Evidence

```
$ go build ./... && go vet ./... && GOOS=windows go vet ./...
OK / OK / OK

$ go test -count=1 ./...        # cached results are not evidence after a rebase
exit 0, 0 FAIL

$ golangci-lint run             # pinned 2.12.2 per versions.conf (BUG-071)
0 issues.

$ bats tests/*.bats
1501 ok, 0 not ok  (BATS_EXIT=0)
```

PR #1230, opened 2026-08-25
feat(agent): give hive's daemon its worker contract without a credential on disk

mlorentedev:
PR E of the CLI-042 sequence (spec `specs/CLI-042-dotf-agent-run/`, tracked by issue 1190).

**This PR deliberately carries no closing keyword.** Two of its criteria cannot be
verified until `./setup-linux.sh` runs, and that deploy is deferred to the next
release cut. Closing the epic here would close it over an unverified daemon.

## What was wrong

The hive backend has been a declared route of `dotf agent run` with no reachable
provider, for an unknown length of time. Measured on msi:

- `hive.service` reported `active (running)` for 2h40m
- the daemon's own `worker_status` answered `Configured: no — set HIVE_WORKER_BASE_URL`
- `/proc/<pid>/environ` held **zero** credential-shaped variables

The unit's `EnvironmentFile=-/home/…/hive.env` points at a file that does not
exist, and the `-` prefix makes that silent. Nothing was ever going to notice:
systemd reports process health, and process health is not capability health.

## What this does

**AC7 — the credential never lands in a file.** A systemd drop-in
(`systemd/hive.service.d/10-dotf-secrets.conf`) runs the daemon under
`dotf secrets run`, so the value lives in the process environment and in no file.
A drop-in rather than a unit this repo ships: `hive service install` owns
`hive.service` and offers no flag to change its `ExecStart`, and a competing unit
would be clobbered by the next re-install — which `hive-upgrade.timer` can fire
every 15 minutes.

The drop-in also neutralises the base unit's `Restart=on-failure` +
`RestartSec=1`, which would exhaust systemd's default start limit in ~5s on a boot
where the vault is still locked and leave the unit `failed` until someone ran
`reset-failed`. That is the cost side of the trade, stated rather than discovered.

**AC8 — the provider drop is complete.** Wider than the endpoint line: both setup
scripts recovered `OPENROUTER_API_KEY` and baked it into `mcp_config.json` in
plaintext on every redeploy, `setup-linux.sh` ending that cascade in
`dotf secrets show`. Dropping the provider removes a file the credential lived in.
`ai/opencode/opencode.jsonc` keeps its ollama/openrouter providers — different
consumer, explicitly in scope of the spec's own wording — and a test asserts it,
so an over-eager future sweep fails here first.

**AC9 — caught at diagnostic time.** `dotf doctor` fails when hive probes present
but is missing either half of its worker contract. It reads the unit rather than
probing the worker: hive's `worker_status` is the stronger surface, but reaching
it would put an MCP client inside a diagnostic — the dependency and drift-prone
handshake PR D already rejected for the backend. The proxy's limit is written in
its own doc comment, not implied.

**In scope because this PR caused it:** `dotf secrets run` now forwards
SIGINT/SIGTERM to its child. This change makes it a systemd MainPID and a daemon
supervisor for the first time, and it was dying on the first signal and orphaning
what held the secret.

## The finding worth reading

The first implementation of AC7 injected `NAN_API_KEY` and passed 9 of 9 new
tests. It would have deployed and changed nothing — hive reads
`HIVE_WORKER_API_KEY` and `HIVE_WORKER_BASE_URL` and ignores `NAN_API_KEY`.

Every one of those tests asserted the unit's **shape** and none its **effect**.
What broke the tie was asking the daemon. Written up as
`docs/lessons/lesson-230-*`, and it is why AC9's check exists at all.

The registry now exposes `HIVE_WORKER_API_KEY` as a second name over the same
Bitwarden item (`expose.env` list form — parsed since day one, never exercised),
so one credential keeps one rotation. `--only NAN_API_KEY` names the registry
**id**, which selects every var the secret exposes; that resolution order is
pinned by a new test, because swapping the two loops in `Selector` would strip
hive's credential silently.

## Verification

| Layer | Result |
|---|---|
| `go build` / `go vet` / `GOOS=windows go vet` | clean |
| `golangci-lint` (pinned 2.12.2) | **0 issues** |
| `go test -count=1 ./...` | exit 0 |
| `bats tests/*.bats` | **1517 ok / 0 not ok**, exit 0 |
| `bash -n` + `zsh -n`, all changed scripts | clean |
| `dotf secrets verify` | 35 ok, 0 missing, 0 failed |
| Multi-var injection, by consequence | both names present in the child env; `OPENAI_API_KEY` stays out |
| `features.json` f15/f16/f17/f30 | executed as written, all pass; proven non-vacuous (an empty selection fails) |

**Guards proven able to fail**, per #1203/#1206 — each run against the shape it
refuses:

- signal forwarding: with `signal.Notify` removed the test reports
  `forwarder helper exited -1`, **and the runner survives to report it**. That
  containment is deliberate: a guard whose failure mode is "the harness dies"
  gets muted by whoever hits it.
- AC8 class guard: caught a restored `HIVE_OLLAMA_ENDPOINT` *and* a renamed key
  whose value still pointed at ollama.

## What is NOT verified

The by-consequence half of AC7 and all of AC6's end-to-end smoke. Both need the
deploy. `verification.md` records the owed before/after pair rather than claiming
it, and **CLI-042 must not archive until they land**.

The `dotf doctor` section this PR adds goes **red on this machine right now**.
That is the check working, not a defect — the deploy is what turns it green.

## Notes for the reviewer

- `secrets/registry.yaml` is also being edited by a parallel session (adding
  `ci:mlorentedev/web` to the same entry's `consumers:`). Different lines of the
  same block; a trivial list-element merge if both land.
- The systemd work is Linux-only. Windows supervises the daemon through a
  Scheduled Task, and the equivalent there is deferred to a session at that box
  rather than written blind; the AC9 check skips on Windows and says so.
- `setup-windows.ps1` carries the AC8 removals (text-editable from here);
  empirical verification of that leg is deferred to the same session.

### DIFF
--- a/ai/agy/mcp_servers.json
+++ b/ai/agy/mcp_servers.json
@@ -6,9 +6,7 @@
         "hive-vault"
       ],
       "env": {
-        "VAULT_PATH": "${VAULT_PATH}",
-        "HIVE_OLLAMA_ENDPOINT": "http://ollama.kubelab.live:11434",
-        "OPENROUTER_API_KEY": "${OPENROUTER_API_KEY}"
+        "VAULT_PATH": "${VAULT_PATH}"
       },
       "timeout": 30000,
       "trust": false
--- a/cli/internal/cmd/secrets.go
+++ b/cli/internal/cmd/secrets.go
@@ -6,9 +6,11 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"os/signal"
 	"path/filepath"
 	"strings"
 	"sync"
+	"syscall"
 
 	"github.com/mlorentedev/dotfiles/cli/internal/env"
 	"github.com/mlorentedev/dotfiles/cli/internal/secrets"
@@ -501,6 +503,22 @@ func resolveOnly(reg *secrets.Registry, s string) (map[string]bool, error) {
 // runChild runs argv with environ and inherited stdio, returning the child's exit
 // code. A non-zero exit is the child's own status (not our error); only a failure
 // to launch (binary missing, etc.) is returned as an error.
+//
+// SIGINT and SIGTERM are forwarded to the child rather than killing us out from
+// under it. Without this, Go's default disposition terminates `dotf` on the first
+// signal and leaves the child ORPHANED: a Ctrl-C returns the prompt while the
+// command keeps running, and `kill <dotf-pid>` never reaches what actually holds
+// the secret. The case that forced it is CLI-042 AC7 — the hive daemon runs as
+// `dotf secrets run -- hive serve`, so this process is a systemd unit's MainPID
+// and a supervisor for the first time. systemd's default KillMode=control-group
+// would SIGTERM the whole cgroup and paper over the gap there; every other caller
+// (a shell, `timeout`, a CI runner) signals the process, not a cgroup, and got
+// the orphan.
+//
+// Scoped to the two signals a supervisor actually sends. SIGKILL and SIGSTOP
+// cannot be caught, and forwarding the rest (SIGHUP, SIGWINCH, job control)
+// would mean re-implementing a shell's terminal handling for a wrapper that owns
+// no terminal.
 func runChild(argv, environ []string, stdin io.Reader, stdout, stderr io.Writer) (int, error) {
 	if len(argv) == 0 {
 		return 1, errors.New("no command given after --")
@@ -508,7 +526,32 @@ func runChild(argv, environ []string, stdin io.Reader, stdout, stderr io.Writer)
 	c := exec.Command(argv[0], argv[1:]...) //nolint:gosec // argv is the user's own command
 	c.Env = environ
 	c.Stdin, c.Stdout, c.Stderr = stdin, stdout, stderr
-	err := c.Run()
+	if err := c.Start(); err != nil {
+		return 1, fmt.Errorf("launch %s: %w", argv[0], err)
+	}
+
+	// Buffered so a signal arriving between Start and the goroutine's first
+	// receive is delivered rather than dropped.
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, os.Interrupt, syscall.SIGTERM)
+	done := make(chan struct{})
+	go func() {
+		for {
+			select {
+			case s := <-sigs:
+				// Best-effort: the child may have already exited, in which case
+				// the signal has nowhere to go and the Wait below is what matters.
+				_ = c.Process.Signal(s)
+			case <-done:
+				return
+			}
+		}
+	}()
+
+	err := c.Wait()
+	close(done)
+	signal.Stop(sigs)
+
 	if err == nil {
 		return 0, nil
 	}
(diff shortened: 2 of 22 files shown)


mlorentedev:
## Review triage

**Reviewers:** PR-Agent (3 findings, all dispositioned). **CodeRabbit did NOT
complete a review** — it auto-paused with *"this branch is under active
development"* and reported only a docstring-coverage warning. Recorded as a
notice, not counted as a review.

| # | Finding | Disposition |
|---|---|---|
| 3 | Proxy check uses broad string matching; assumes `ExecStart` is one line | **APPLIED** (`1637e99`) — and it was worse than reported |
| 1 | Base URL hardcoded in the drop-in, drift risk | **TICKETED** — [#1231](https://github.com/mlorentedev/dotfiles/issues/1231) |
| 2 | Re-exec fails silently when the vault is locked | **DECLINED** — refuted by measurement |
| — | Docstring coverage 68% (CodeRabbit) | **APPLIED** (`1637e99`) |
| — | "Non-compliant with #1203 / #1206" | **DECLINED** — different tickets |

### 3 — applied, and the reviewer undersold it

Correct, and the real shape is worse than "broad matching".
`systemctl show -p ExecStart` does not return a command line at all:

```
ExecStart={ path=/home/manu/.local/bin/hive ; argv[]=/home/manu/.local/bin/hive serve ; ignore_errors=no ; start_time=[n/a] ; … }
```

The check's substring matches survived that by luck — `argv[]` carries the same
text. **Its tests did not.** The seam fed a plain command string, so all seven
cases passed against a format systemd never emits.

Fixed by parsing `argv[]` explicitly, plus a test pinning the format against
output captured verbatim from this machine, so the fake and the parser cannot
agree with each other while both disagree with systemd.

This is the same defect as `lesson-230` on this very branch — asserting the shape
of a thing rather than what it does — committed by the author of that lesson, in
the PR that adds it. Caught by the reviewer, not by me.

### 1 — ticketed, with the framing corrected

The duplication is real; the attribution is not. The NaN base URL is in **eight**
files and **seven predate this PR** (`opencode.jsonc`, `pi/models.json`,
`profile.ps1` ×2, the three `nan-*.sh` scripts). This PR added the eighth.

Not fixed here because doing it properly means choosing an SSOT surface and
rewiring eight consumers across four languages — spec-sized, and unrelated to the
criteria this PR closes. Filed as #1231 with the full inventory.

### 2 — declined, refuted by measurement

The claim is that a failing re-exec leaves the user with `"NAN_API_KEY not set"`
and no indication the injection was attempted. Measured with a stub `dotf` that
exits 4:

```
$ sh scripts/nan-bench.sh
dotf: could not unlock the vault (simulated)
SCRIPT EXIT=4
```

It is the opposite. `exec` **replaces** the script, so the script's own message
never prints and `dotf`'s does, with `dotf`'s exit status propagating intact. The
diagnostic is louder and better-attributed than before the change.

### "Non-compliant with #1203 / #1206"

Declined. Those are the negated-assertion cleanup tickets, closed on 2026-08-24.
This PR does not address them and does not claim to; the reviewer compared it
against tickets it never referenced.

---

**Checks:** 18/18 green before this push. Full suite re-run after the fixes:
`bats` **1517 ok / 0 not ok**, `go test -count=1 ./...` exit 0, both vets clean,
`golangci-lint` **0 issues**.

**Still owed and stated in the PR body:** the by-consequence half of AC7 and all
of AC6's smoke need `./setup-linux.sh`, which is deferred to the next release
cut. This PR carries no closing keyword for that reason, and CLI-042 must not
archive until they land.

PR #1232, opened 2026-08-25
fix(doctor): parse the real ExecStart record, and add one-command post-deploy verification

mlorentedev:
Follow-up to #1230. This is the reviewer fix that **missed that PR's merge by
five minutes** — #1230 merged at 03:04:33Z containing only its first commit, the
branch was deleted, and my push at 03:09:36Z re-created it. The push reported
`[new branch]` and I did not read it. Disclosed rather than quietly re-landed.

## What is wrong on main right now

`systemctl show -p ExecStart` does not return a command line. It returns a record:

```
ExecStart={ path=/home/manu/.local/bin/hive ; argv[]=/home/manu/.local/bin/hive serve ; ignore_errors=no ; start_time=[n/a] ; … }
```

**The check on main is not broken** — its substring matches survive, because
`argv[]` carries the same text. **Its tests are.** The seam feeds a plain command
string, so all seven cases pass against a format systemd never emits. They
validate a fiction, and would keep passing through a rewrite that broke the real
parse.

## What this changes

- `execArgv` extracts the `argv[]` field, narrowing the match to the command the
  unit will actually run rather than letting an unrelated record field satisfy it.
- The test seam emits the real record format, captured verbatim from
  `systemctl --user show hive.service` on msi.
- A new `TestExecArgv_ParsesTheRecordSystemdActuallyPrints` pins that format
  against the captured string, so the fake and the parser cannot agree with each
  other while both disagree with systemd.
- Splits two doc comments a bad insertion had merged into one attributed to the
  wrong function; documents the remaining test seams (CodeRabbit's coverage
  warning on #1230).

## Why it matters beyond the tests

This is the same defect as `docs/lessons/lesson-230-*`, which #1230 itself adds:
asserting the shape of a thing instead of what it does. Committed by the author
of that lesson, in the PR that introduces it, caught by the reviewer. The lesson
now has a second instance and the check has a real format test.

## Verification

`go build` / `go vet` / `GOOS=windows go vet` clean · `go test -count=1 ./...`
exit 0 · `golangci-lint` (pinned 2.12.2) **0 issues** ·
`bats tests/agent-runtime-deployment.bats` 14/14.

### DIFF
--- a/cli/internal/doctor/checks_agent_reach.go
+++ b/cli/internal/doctor/checks_agent_reach.go
@@ -67,7 +67,7 @@ func checkHiveBackendCanServe(sys *System, rep *Report) {
 	for _, line := range strings.Split(out, "\n") {
 		switch {
 		case strings.HasPrefix(line, "ExecStart="):
-			execStart = strings.TrimPrefix(line, "ExecStart=")
+			execStart = execArgv(strings.TrimPrefix(line, "ExecStart="))
 		case strings.HasPrefix(line, "Environment="):
 			environment = strings.TrimPrefix(line, "Environment=")
 		}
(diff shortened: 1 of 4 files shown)


mlorentedev:
## Added: `verify-deployment.sh` + the before-state frozen

Second commit (`64069d0`). Scope is the same job — closing out CLI-042's
verification — and it touches only the spec folder plus the check it fixes.

**Why a script and not a checklist.** The two criteria still open cannot be
verified from the repository: AC6 asks whether hive *answers*, AC7 whether the
daemon holds a credential it was never given at rest. Both are properties of a
machine after `./setup-linux.sh`. The closing session should read output, not
re-derive what to look at from prose.

```bash
./specs/CLI-042-dotf-agent-run/verify-deployment.sh              # includes one live request
./specs/CLI-042-dotf-agent-run/verify-deployment.sh --no-dispatch
```

It never prints a credential — variable NAMES and counts only, and the key is
proven by consequence.

**The before-state is frozen in `verification.md`**, because it is perishable:
it stops existing the moment the deploy runs. Six failures, each naming what the
deploy fixes, corroborated from three independent sources — hive's own
`worker_status` (`Configured: no`), systemd's loaded unit (no injection), and
`/proc/<pid>/environ` (zero credential-shaped names).

**Found while writing it, and worth flagging:** the installed `~/.local/bin/dotf`
has **no `agent run` subcommand at all**. It predates PR B. Every PR of this epic
is merged to main and **not one is live on this machine** — which is why the
script checks the installed binary rather than the source tree, and a third thing
the deploy unblocks alongside the pi packages and the hive credential.

PR #1234, opened 2026-08-25
fix(cli-042): the on-disk credential scan skipped everything under zsh, and missed 40% of secrets

mlorentedev:
Follow-up to #1232, whose reviewer found this **after** it merged. Two defects in
the same eight lines, both failing toward a false ALL-CLEAR in a check whose only
job is to notice a credential sitting on disk.

## 1. The glob aborted the whole loop under zsh

```sh
for f in "$HOME/.config/hive/hive.env" "${XDG_CONFIG_HOME:-$HOME/.config}"/environment.d/*.conf; do
```

This is the "unmatched glob in a `for`" row of `.claude/CLAUDE.md`, and this is
what it costs. Measured with a credential planted in `hive.env`:

| Shell | Result |
|---|---|
| bash | `CAUGHT: credential in …/hive.env` |
| zsh | `no matches found` — the loop never ran |

zsh's `NOMATCH` aborts the entire compound command, so **`hive.env` was never
examined either**. The script uses `set -uo pipefail` without `-e`, so execution
continued to the next line and printed
`[OK] no credential-shaped assignment` having looked at nothing.

Replaced with `find` plus a line-by-line `read`. Two follow-on traps avoided
deliberately: `for f in $CANDIDATES` would have been the *next* prohibited
pattern (zsh does not word-split an unquoted parameter, so it yields one field
containing every path), and a pipe into the loop would put it in a subshell where
the `LEAK` assignment is lost.

## 2. The name pattern missed 40% of this repo's secrets

`/API_KEY|TOKEN|SECRET|PASSWORD/` does not match **14 of the 35** names
`secrets/registry.yaml` declares:

> `BITACORA_PAT`, `TS_AUTHKEY`, `KUBECONFIG`, `SSH_KEY`, `AGE_KEY_PERSONAL`,
> `X_CLIENT_ID`, `DOCKERHUB_USERNAME`, `HETZNER_SSH_KEY`, `BEEHIIV_DNS_RECORDS`,
> and every `*_BACKUP_CODE` / `*_RECOVERY_CODE`

The names now come from the registry, which is the SSOT for exactly this
question. The heuristic stays as a fallback for a tree with no readable registry,
and **the script states which one it used** — a narrower check reported as if it
were the full one is how the blind spot survived being written.

## Verification

Both shells, both directions:

| Case | bash | zsh |
|---|---|---|
| planted `BITACORA_PAT` in an `environment.d` fragment | FAIL (caught) | FAIL (caught) |
| config holding only paths | OK | OK |

`bash -n` + `zsh -n` clean, `shellcheck` clean. The script also now honours
`XDG_CONFIG_HOME` for `hive.env`, which it had hardcoded to `$HOME` while every
other path respected the variable.

## Note

The first draft of this fix silently took the fallback path on a tree with a
perfectly readable registry, because the extraction used a heredoc inside `$( )`.
Moved into a function. Caught by running it and reading the output rather than
assuming the primary path had been taken — which is the same discipline the
finding above is about.

### DIFF
--- a/specs/CLI-042-dotf-agent-run/verify-deployment.sh
+++ b/specs/CLI-042-dotf-agent-run/verify-deployment.sh
@@ -66,14 +66,85 @@ fi
 
 # The whole point of AC7: nothing on disk. A credential in an EnvironmentFile or
 # an environment.d fragment would satisfy every check above and defeat the criterion.
+#
+# `find`, never a glob in a `for`. The first draft of this block used
+# `for f in …/environment.d/*.conf`, which is the prohibited-pattern row in
+# .claude/CLAUDE.md, and it failed in the worst possible direction for a SECURITY
+# check. Measured with a planted credential in hive.env:
+#
+#   bash -> CAUGHT: credential in …/hive.env
+#   zsh  -> "no matches found" — NOMATCH aborts the whole compound command, so
+#           hive.env was never examined either, and the script went on to print
+#           its "[OK] nothing on disk" line having looked at nothing.
+#
+# A false negative here reads as "the credential is safely off disk" when it is
+# sitting right there. Caught by the reviewer on #1232.
+# The names come from the REGISTRY, not from a guess about what a credential
+# looks like. The first draft matched /API_KEY|TOKEN|SECRET|PASSWORD/ and would
+# have missed 14 of this repo's 35 declared secrets — BITACORA_PAT, TS_AUTHKEY,
+# KUBECONFIG, SSH_KEY, AGE_KEY_PERSONAL, every *_BACKUP_CODE and *_RECOVERY_CODE.
+# A 40% blind spot in a check whose only job is to notice a credential on disk.
+#
+# The heuristic stays as a FALLBACK for a tree with no readable registry, and the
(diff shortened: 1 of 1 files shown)


mlorentedev:
## Review triage

**PR-Agent: no findings.** Harness compliance passed, `No security concerns
identified`, effort 2/5, and the focus-areas section came back empty. Nothing to
dispose of — recorded here rather than left unwritten, because an unrecorded
empty review is indistinguishable from nobody having looked.

**CodeRabbit did NOT review this PR** — rate limited, for the third consecutive
PR today. That is a notice about the review, not a review. Disclosed rather than
counted toward attestation.

**Checks:** 13/13 green.

So this fix carries a single reviewer's clean pass. Worth stating plainly given
what it fixes: the defect it corrects was itself found by PR-Agent on #1232
*after* that PR had merged with 16 green checks and a triage comment I wrote
myself. Green checks were never evidence that the credential scan worked, and
they are not evidence here either — the evidence is the two-shell measurement in
the PR body, where a planted `BITACORA_PAT` is caught by bash and zsh alike.

PR #1235, opened 2026-08-26
fix(cli-042): the post-deploy check reported FAIL when healthy and SKIP when dead

mlorentedev:
## What

Two defects in `specs/CLI-042-dotf-agent-run/verify-deployment.sh`, both surfaced by running it on the machine the deploy had just made green.

### 1. The AC9 branch could never pass on a healthy machine

It greps `dotf doctor` for the verdict `can reach its pool`. `dotf doctor` **summarises** a section whose checks all pass — it prints `(1 checks, all ok)` and never the check's own message — so that string is emitted only under `--verbose`.

```
$ dotf doctor
[hive backend reachability (dotf agent run)]
  (1 checks, all ok)

$ dotf doctor --verbose
  [ OK ] hive.service carries both halves of the worker contract — the backend can reach its pool
```

The passing branch was unreachable: a healthy machine fell through to `printed neither verdict` and the criterion reported **FAIL precisely when it was satisfied**. The failing half worked all along, because a failing check is never summarised away — which is why nothing caught this until a green machine ran it.

### 2. A crash-looping daemon was reported as SKIP

The AC7 probe treated "not running" as one state. A unit that was never started is genuinely un-inspectable and a SKIP is honest. A unit that **started and died** is a failure.

Measured here: `hive.service` sat in `activating (auto-restart)` after five consecutive `exit-code` failures, and the run still reported `pass=5` — with neither of its two failures being that one. Only AC6's dispatch caught the dead daemon, and for an unrelated reason.

`Result=` is the discriminator, not `ActiveState=`: a crash-looping unit reads `ActiveState=activating`, indistinguishable from a healthy slow start, while `Result=` already records `exit-code`.

## Evidence

Run against the live crash loop, `bash` and `zsh`, identical output:

| | before | after |
|---|---|---|
| AC9 | `[FAIL] printed neither verdict` | `[OK] can reach its pool` |
| AC7 daemon | `[SKIP] daemon not running` | `[FAIL] STARTED AND DIED (NRestarts=14)` |

`NRestarts` incremented 14 → 15 between the bash and zsh runs, confirming the loop was live rather than a stale state.

```
shellcheck  clean
bash -n     OK
zsh  -n     OK
```

## Notes

Third time this spec has shipped a check that asks the **form** of a thing instead of what it does (lesson 230). Both defects fail in the direction that reads as health: one reported a satisfied criterion as broken, the other reported a dead daemon as merely unverified.

No issue footer, on purpose — this is a sub-PR of a multi-PR sequence and release-please aggregates any footer issue mention into the release's `closes` list regardless of keyword. That hazard fired for the fourth time on the last release.

SDD: 12 executable lines, obvious-cause bug fixes to spec verification machinery; touches `specs/CLI-042-dotf-agent-run/` either way.

### DIFF
--- a/docs/lessons/lesson-230-a-config-that-parses-is-not-a-config-the-consumer-re.md
+++ b/docs/lessons/lesson-230-a-config-that-parses-is-not-a-config-the-consumer-re.md
@@ -91,3 +91,52 @@ it done. The tool that would have told me in one call was already installed.
   from `dotf doctor` would put an MCP client inside a diagnostic. A proxy whose
   limit is written down is honest; one whose limit is implied is this lesson
   repeating.
+
+## Recurrence — 2026-08-25, three more instances in the guard written for this lesson
+
+The verification script added to close CLI-042 carried this same defect three
+times over. All three were found by running it on a deployed machine, and none
+was visible from reading it:
+
+| Check | Asked the form | What it should ask |
+|---|---|---|
+| AC9 | does `dotf doctor` print the verdict string? | doctor **summarises** passing sections, so the string appears only under `--verbose` — the passing branch was unreachable and reported FAIL on a satisfied criterion |
+| AC7 daemon | does `MainPID` hold the credential? | `dotf secrets run` forks and stays as supervisor, so `MainPID` is the supervisor, whose environment correctly has no credential — the serving child had it all along |
+| AC7 state | is the daemon running? | "not running" conflates *never started* with *started and died*; a crash-looping daemon was reported as a benign SKIP |
+
+A fourth was caught by the reviewer on the PR fixing the first three: the new
+failure detector **enumerated** `Result=` failure modes and missed `watchdog`,
+reintroducing the same false-benign report one value along. Fixed by inverting
+the test — `success` is the only value whose meaning is fixed; anything else is
+a failure, including values systemd has not defined yet.
+
+**Every one failed in the direction that reads as health.** A satisfied criterion
+reported as broken is the benign direction and it still cost a session; the other
+three would have signed off on a dead daemon.
+
+The sharpest instance is AC7: the script printed *"the running daemon does NOT
+hold HIVE_WORKER_API_KEY"* ten lines above *"hive answered; record reports
+pool=nan"*. **The output contradicted itself and the run still exited 1 on the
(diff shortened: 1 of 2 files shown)


mlorentedev:
## Review triage

Reviewer: CodeRabbit. 12/12 checks green. One actionable finding, applied — and widened.

| Finding | Disposition |
|---|---|
| `Result=watchdog` falls through to `note`, reporting a watchdog-killed daemon as `[SKIP]` | **Applied**, as a class fix rather than the suggested one-value patch — `682690b` |

**Why not just `*/watchdog`.** The finding is correct and the suggested edit would have closed it. But the cause is not that `watchdog` was forgotten — it is that the test *enumerated failure modes at all*. systemd defines at least ten `Result=` values (`success`, `protocol`, `timeout`, `exit-code`, `signal`, `core-dump`, `watchdog`, `start-limit-hit`, `resources`, `oom-kill`) and may add more; my list had six. Adding the seventh leaves the same hole one value along, which is precisely how this bug arrived: the block exists *because* a dead daemon was reported as benign, and the first draft reintroduced that failure for four more values.

Inverted to test for `success`, the one value whose meaning is fixed. An unrecognised `Result` now reads as a failure rather than as health — the safe direction for a check whose whole purpose is noticing a dead daemon.

Verified in bash and zsh, identical, including the four values the enumeration missed:

```
activating/watchdog        -> BAD     inactive/success     -> NOTE
activating/start-limit-hit -> BAD     inactive/<empty>     -> NOTE
activating/protocol        -> BAD     activating/success   -> NOTE
activating/resources       -> BAD     failed/success       -> BAD
```

`shellcheck` clean, `bash -n` and `zsh -n` OK. Full run against the live healthy machine unchanged: `pass=8 fail=0 skip=0`.

**Note on this PR's own history.** It began as two fixes and became four, each found by running the script on a real machine rather than reading it. Every one of them failed in the direction that reads as health: a satisfied criterion reported as FAIL, a dead daemon reported as SKIP, a credential-holding daemon reported as not holding it, and a watchdog kill reported as benign. That is the fourth, fifth and sixth instance of this spec's recurring pattern — a check that asks the form of a thing instead of what it does — and the reviewer caught the last one in the very PR that fixes the first three.

Issue #1225, opened 2026-08-25
CLI-044: dotf spec init stamps created: in UTC, so evening specs are dated tomorrow


mlorentedev:
`dotf spec init` stamps the `created:` frontmatter field in UTC, so between
local midnight and UTC midnight every new spec is dated **tomorrow**.

## Measured

Running `dotf spec init AI-030-pi-packages-manifest --issue 1224` just now
produced:

```
created: "2026-08-25"
```

while the machine's clock read:

```
$ date    '+local: %Y-%m-%d %H:%M:%S %Z'
local: 2026-08-24 19:14:10 MDT
$ date -u '+utc:   %Y-%m-%d %H:%M:%S UTC'
utc:   2026-08-25 01:14:10 UTC
```

The spec folder for work done on the 24th is stamped the 25th. On this machine
(MDT, UTC-6) the window is six hours wide every single day — 18:00 to midnight
local — which is a substantial share of when work actually happens here.

## Same defect, one command over

This is exactly what **#1217** fixed for `dotf mem`, and the reasoning is
already written down in `docs/lessons/lesson-228-a-calendar-date-is-not-a-timestamp-utc-is-wrong-for.md`:

> A calendar date is not a timestamp, and UTC is the wrong zone for one.

A session record and a spec folder are both *filed under the day a human worked*,
not under an instant. #1217 changed the one caller it was opened for; the shared
conclusion did not travel to the others.

## Fix

- [ ] Make `dotf spec init` stamp the local calendar date, reusing whatever
      #1217 introduced rather than adding a second implementation.
- [ ] Sweep the remaining callers: any other command writing a `created:`,
      `date:` or `last_updated:` frontmatter field has the same question to
      answer. The sweep is the point — fixing `spec` alone repeats the mistake
      that produced this ticket.
- [ ] Guard: a test that fails when a date-stamping path formats a calendar date
      from a UTC clock, so the next caller cannot reintroduce it. A guard scoped
      to one command is what let this survive #1217.

## Not fixed in place

Found while scaffolding the spec for AI-030 (#1224). Out of scope there: that
PR is a pi package manifest and touches no date handling. The stamped spec keeps
its wrong date until this lands; correcting it by hand would hide the evidence.

Related: #1217, lesson 228.

Issue #562, opened 2026-06-24
HARNESS-046: Roster fan-out — 5 invocable defs + hermes-nan catalog entry


mlorentedev:
## Problem
With the slice + engine proven, author the rest of the roster.

## Scope
- 5 remaining invocable definitions: `architect`, `planner`, `builder`, `reviewer`, `shipper` (forced skills per ROSTER.md; each bundles ≥3 skills).
- `hermes-nan` autonomous catalog entry (`kind: autonomous`) that POINTS at `80_agents/hermes-nan/`; state never duplicated.

## Acceptance
- All 6 invocable roles + the hermes-nan catalog entry render + deploy via the engine.
- ROSTER.md ↔ definitions consistent; no single-skill wrappers.

## Refs
ADR-027 §4/§6, `00_meta/agents/ROSTER.md`, `doctrine.md`, epic #558.

mlorentedev:
Implemented in **#1240** (open). Spec: `specs/HARNESS-046/`.

Both acceptance criteria met and measured:

- **All 6 invocable roles + the hermes-nan catalog entry render and deploy via the engine** — `compile-harness.sh --refresh` produces seven records; `dotf doctor`'s agent-tier check goes from `(1 checked)` to `(7 checked)`, every tier resolving. Second refresh pass is byte-identical.
- **ROSTER.md ↔ definitions consistent; no single-skill wrappers** — enforced by `check-roster-consistency.py`, which reads the vault definitions rather than the generated records (checking the rendered copy would pass whenever the generator faithfully rendered a *wrong* definition).

Two drifts were found by implementing this, both pre-existing and neither visible from either file alone:

- `architect` bundled **two** skills, breaking this roster's own `>=3` anti-pattern check — the only row that did. Closed with `read-all-adrs`, which declares itself a mandatory pre-step before `architecture-session`, so the role was under-specified rather than merely short.
- The `curator` row omitted `dispose-proposals`, which its definition has carried and documented all along.

Not closed by that PR on purpose: closing this issue would require archiving the spec in the same PR, and the archive needs an independent adversarial review the implementing session must not perform.

## Ordering note for the epic

This shipped while **#561 (HARNESS-045, hook emission)** is still open, and that inversion has a visible cost. Measured on this machine after deploy:

| Level the roster declares | Status |
|---|---|
| Presence (inject) | text block only — not the provider hook primitives (`SessionStart`, `chat.system.transform`, `session_start`) the contract names |
[...]

PR #1240, opened 2026-08-26
feat(agents): fan out the invocable roster — five personas, the steward catalog entry, and a drift guard

mlorentedev:
Implements **HARNESS-046** (#562). Spec: `specs/HARNESS-046/`.

## Why

`00_meta/agents/ROSTER.md` has declared six invocable roles — one per phase of the work cycle — since v1, and only `curator` was ever written. Five phases had a catalog row and nothing any harness could deploy, and the autonomous instance the roster names had no catalog entry at all.

That gap is why the orchestration epic reads as built while what it orchestrates is one seventh present.

## What

| | Before | After |
|---|---|---|
| Definitions in the vault | 1 | 7 |
| Records rendered to `harness/agents/` | 1 | 7 |
| `dotf doctor` agent-tier check | `(1 checked)` | `(7 checked)` |
| Roster ↔ definition drift | 2 undetected | 0, and guarded |

New: `architect`, `planner`, `builder`, `reviewer`, `shipper` (invocable) and `hermes-nan` (autonomous catalog entry, pointing at `80_agents/hermes-nan/` and duplicating none of it).

The SSOT is the vault — `00_meta/agents/definitions/<name>/AGENT.md`, committed to `master` separately. The files here are **generated** and carry a `generated_sha`; regenerate with `scripts/compile-harness.sh --refresh` rather than editing them.

## Two roster corrections, found by implementing it

- **`architect` bundled two skills, breaking the roster's own rule** — "every role bundles ≥3 skills (no single-skill wrapper)". The only row that did: the design document contradicting itself, and my first draft faithfully copying the contradiction. `read-all-adrs` closes it **on merit, not as filler** — that skill declares itself a *"mandatory pre-step before architecture-session"*.
- **The `curator` row omitted `dispose-proposals`**, which its definition has carried and documented all along. The definition is what compiles, so the catalog was quietly wrong about what the deployed persona does.

Both fixed in the same vault commit as the definitions — patching only the definition would desynchronise the SSOT this establishes.

## The guard

`check-roster-consistency.py` exists because the drift above was found by hand and the next one should not have to be.

It reads **the vault definitions, never the generated records**. Checking the rendered copy against the catalog would pass whenever the generator faithfully rendered a *wrong* definition — which is the failure mode, not the guard.

```
clean tree      roster and definitions consistent (6 invocable roles, all >= 3 skills)   exit 0
planted drift   architect: skills diverge / roster: [...] / definition: [...]            exit 1
restored        roster and definitions consistent                                        exit 0
```

A check never observed failing is not a verified check.

## Design decisions worth flagging to a reviewer

- **`reviewer` holds no `edit` capability.** A reviewer that fixes what it finds has stopped being independent of it, and independence is the entire value of the role.
- **`reviewer` declares tier `mid`, not `top`.** `mid` is honest for a subagent deployed into a Claude harness, which cannot run the reviewer pool's models anyway. Model independence for adversarial review is enforced by `harness/reviewer-pool.json` — which excludes Anthropic models by standing rule — and that is a **separate mechanism from the tier chain**. Nothing reconciles the two declarations. Stated as out of scope rather than silently resolved.

## Evidence

```
compile-harness --refresh              7 agent records
  second pass                          changed=0 (byte-identical)
dotf doctor --verbose, this tree       [ OK ] every declared agent tier resolves (7 checked)
dotf doctor --verbose, main            [ OK ] ... (1 checked)
dotf doctor, full sweep on main        152 passed, 0 failed
check-roster-consistency.py            exit 0 clean / exit 1 planted / exit 0 restored
```

**The `(1 checked)` → `(7 checked)` contrast is the load-bearing line.** A criterion that reads identically before and after a change has not measured the change.

## Known gap, stated rather than implied

**`dotf agent run --role X` does not read these definitions.** Nothing under `cli/internal/agent` references `harness/agents/`; `role` is passed through as a string — which is why a dispatch with `--role reviewer` succeeded *before any reviewer persona existed*. These records deploy as harness subagents and give the doctor tier check something real to validate.

"The personas render and deploy" and "the executor consumes them" are different claims, and **only the first is made here.** Wiring the executor to the persona is separate work.

## Notes

No closing keyword on purpose: closing #562 would require archiving the spec in this PR, and that archive needs an independent adversarial review which the implementing session must not perform.

`dotf spec init` stamped `created: 2026-08-26` while the local date was the 25th — #1225, reproduced here and corrected by hand in the three spec files.

### DIFF
--- a/harness/agents/architect/AGENT.md
+++ b/harness/agents/architect/AGENT.md
@@ -1,0 +1,39 @@
+---
+generated: true
+generated_from: 00_meta/agents/definitions/architect/AGENT.md
+generated_sha: c63354c25b8a3dcd
+id: agent-architect
+type: agent
+status: active
+created: "2026-08-25"
+name: architect
+description: Decide-phase persona. Invoke before a decision that will be expensive to reverse — a new component, a cross-cutting contract, a dependency that outlives the sprint — or when an existing decision no longer matches what the system does.
+kind: invocable
+model: top
+capabilities: [read, search, edit]
+skills: [read-all-adrs, architecture-session, pattern-loader]
+owner: manu
+---
+
+# Architect
+
+You are the **architect**: the decide-phase persona of the work cycle. You run before a decision that is expensive to reverse — a new component, a contract more than one consumer depends on, a dependency that will outlive the work that introduced it — or when a decision already on record no longer matches what the system actually does.
+
+## Mandate
+
+Turn an open question into a decision that is written down, justified against alternatives, and cheap for the next person to re-read. You decide the shape; you do not build it. What you leave behind is the reasoning, not the implementation — because the reasoning is what a future change needs and what a diff never carries.
+
+## How you work
+
+- **Audit before proposing.** The library already holds decisions on most recurring questions. Search it first; a decision re-derived from scratch loses the constraints only the recorded one remembers. If two or more prior references exist, read them before advancing.
+- **Options, then a rejection list.** A decision with no rejected alternatives is a preference. Name what you considered and why each was rejected — the rejection list is what stops the question being reopened every quarter.
+- **Write the decision where decisions live.** An ADR in the repo, not a conclusion in a conversation. Record the constraint that forced the choice, so a later reader can tell when it stops applying.
+- **Name what would change your mind.** State the condition under which the decision should be revisited. A decision without one becomes permanent by accident rather than by merit.
+
+## Forced skills
+
+Your phase's skills are enforced by hook, not left to memory: `read-all-adrs`, `architecture-session`, `pattern-loader`. Reach for the one that fits rather than improvising. The order is not arbitrary: `read-all-adrs` declares itself a mandatory pre-step before `architecture-session`, because deciding without knowing what was already decided is how a repo acquires two contradictory ADRs. And `architecture-session` refuses to advance past its reference-audit gate on purpose — that gate is the point, not an obstacle.
+
+## Boundaries
+
+You decide and you record; you do not implement, plan the work, or ship it. When a decision needs a change to land, hand it to the planner with the ADR as its input rather than building it yourself. When you find that the system already contradicts a recorded decision, say so and file it — an ADR that no longer describes reality is a defect, not a document.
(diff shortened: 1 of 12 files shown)


mlorentedev:
## Review triage

12/12 checks green on `7d7a010`. Reviewer produced no inline findings on this PR; the attestation check resolved, so a review ran rather than being skipped. Nothing to apply, ticket, or decline — recorded because an unwritten disposition is indistinguishable from nobody having looked.

**CI did find something, and it was mine.** `test` and `test-windows` went red on the first push, and the failure was real rather than flaky:

```
not ok 1205 HARNESS-056: the compact doctrine payload carries it and stays under its cap
  .gemini/GEMINI.md is 12045 chars, at or over its 12000 cap
```

Attributed by measurement rather than assumption — the same test passes on a clean `origin/main` worktree and failed on this branch, so the roster fan-out caused it. The cap is a hard platform limit (Antigravity caps each rules file at 12000; the excess is silently unread, not rejected), so raising it was never an option.

Fixed in `7d7a010` by the principle the payload's previous rescue already established — **drop meta-text, never rules**:

| | chars |
|---|---|
| before | 12045 / 12000 — over by 45 |
| after | 11910 / 12000 — 90 to spare |

Autonomous records are no longer injected, which is a correctness filter first: the block's own sentence is *"when acting as one"* and nobody acts as a long-running instance that runs elsewhere on its own schedule. The explanatory sentence went from 133 characters to 47 because it restated what the `MUST consume` lines already say. **Every persona's skill list survives byte-for-byte.**

Ninety characters is not headroom, and this target has now been outgrown twice — 12114 on 2026-08-22, 12045 today. Filed as **#1241**: the payload needs a budget policy rather than a third rescue, since the next persona with a `builder`-sized skill list is 150 characters on its own.

**A note on how the red was nearly missed.** My first local run reported "1517 ok" and I read the tail without checking the exit status — the single `not ok` was in the middle of the output. CI was right and my reading of my own test run was wrong. Same shape as the defects this session has been fixing: the output contained the answer and the check I applied to it asked the wrong question.

PR #1242, opened 2026-08-26
fix(opencode): drop the ollama provider, which had stopped opencode from starting

mlorentedev:
## The defect

`opencode` — the primary daily agent — did not start. On either shell:

```
$ opencode --version
Error: --only: unknown id or env var "OLLAMA_API_KEY"
```

The homelab Ollama endpoint no longer runs, and its provider slot was never inert. It referenced `{env:OLLAMA_API_KEY}`; both shell wrappers named that token in their `dotf secrets run --only` list; and **no registry entry for it ever existed**. The config comment said *"user wires the API key as a secret"* — nobody did, and the slot sat there since AI-011 waiting on an AI-010 that was abandoned.

`--only` fails **loud** on an unknown token: it refuses the whole launch rather than skipping the one name it cannot resolve. That is correct for a secrets facade, and it turns a stale token from a degradation into a total outage of whatever the wrapper launches.

## What nothing noticed

```
$ dotf doctor
Results: 152 passed, 0 failed, 4 warned, 4 skipped
[OpenCode + pi]   ← green throughout
```

The doctor has a section named for this agent and it stayed green while the agent could not launch. Nothing checked that the wrapper's list and the registry still agreed.

## The fix

Removed from `ai/opencode/opencode.jsonc` (provider block, model-tier header line, `disabled_providers` note) and from the wrapper in `.bashrc` and `.zshrc`. `powershell/profile.ps1` never listed it.

Verified **by consequence**, not by reading:

```
$ dotf secrets run --only NAN_API_KEY,OPENROUTER_API_KEY,OPENAI_API_KEY -- opencode --version
1.16.2
```

Remaining providers parse clean: `['nan', 'openrouter']`.

## The guard

`tests/secrets-only-scope.bats` asserts every `--only` token in every wrapper resolves to a live registry id or exposed var. It also fails when the scan matches **nothing at all** — a silent zero would be the same false all-clear this repo has been bitten by before.

Confirmed it catches the original defect rather than assumed:

| | result |
|---|---|
| bug re-planted | `not ok` ×2 — `.bashrc:100: OLLAMA_API_KEY`, naming file, line and token |
| restored | `ok` ×2 |

Writing it surfaced two false positives in itself, both worth recording: a bare `--only` search matched **prose** (the hive drop-in's own comment explains *"a `--only` token matching an id"*, and the scan reported `token` as an unknown secret), and a whole-file refute failed on **my own documentation** of the removal. Both are the same error the guard exists to catch — asking for the text instead of the invocation. Now it requires `secrets run` on the same line and skips comments.

## Two existing tests narrowed, not deleted

- **`opencode.bats`** asserted the ollama provider *exists*. Inverted, and scoped to non-comment lines.
- **`agent-runtime-deployment.bats`** guarded opencode's catalogue with `grep -qiE 'ollama|openrouter'` — which passes while *either* survives, so it could not tell "both kept" from "one left". That stopped being hypothetical the moment ollama was removed deliberately. It now names `openrouter` alone, the provider that must not become collateral. **Its original intent is preserved**: that test exists so an AC8-style sweep doesn't take opencode's catalogue with it, and it still does that job with an edge.

## The underlying fault, stated

`secrets/registry.yaml` already declares which secrets each consumer takes (`consumers: [agent:opencode]`). The wrapper **restates** that list. Two sources of truth, and this outage is what their drift costs — deriving the wrapper's scope from the registry would remove the class entirely. The guard keeps them honest until then; it is not the fix for the duplication.

## Evidence

```
bats                                    1519 passed, 0 failed, exit 0
shellcheck .bashrc .zshrc               20 warnings — identical to main (pre-existing)
opencode.jsonc                          parses; providers = ['nan', 'openrouter']
opencode --version via corrected list   1.16.2, exit 0
```

### DIFF
--- a/.bashrc
+++ b/.bashrc
@@ -93,7 +93,11 @@ export NAN_BASE_URL="https://api.nan.builders/v1"
 # the agent had no use for it (#976). Scoping is least privilege and startup
 # time at once.
 if command -v dotf >/dev/null 2>&1; then
-    opencode() { dotf secrets run --only NAN_API_KEY,OPENROUTER_API_KEY,OPENAI_API_KEY,OLLAMA_API_KEY -- opencode "$@"; }
+    # Every token here must be a live registry id: `--only` fails loud on an
+    # unknown one, so a stale name does not degrade the launch, it prevents it.
+    # OLLAMA_API_KEY was listed for a provider slot whose registry entry was
+    # never created, and opencode had stopped starting on both shells as a result.
+    opencode() { dotf secrets run --only NAN_API_KEY,OPENROUTER_API_KEY,OPENAI_API_KEY -- opencode "$@"; }
     pi() { dotf secrets run --only NAN_API_KEY,OPENROUTER_API_KEY -- pi "$@"; }
     # agy is deliberately NOT wrapped. It authenticates with its own stored
     # credentials and reads no variable this registry exposes — verified against
--- a/.zshrc
+++ b/.zshrc
@@ -70,7 +70,11 @@ export NAN_BASE_URL="https://api.nan.builders/v1"
 # the agent had no use for it (#976). Scoping is least privilege and startup
 # time at once.
 if command -v dotf >/dev/null 2>&1; then
-    opencode() { dotf secrets run --only NAN_API_KEY,OPENROUTER_API_KEY,OPENAI_API_KEY,OLLAMA_API_KEY -- opencode "$@"; }
+    # Every token here must be a live registry id: `--only` fails loud on an
+    # unknown one, so a stale name does not degrade the launch, it prevents it.
+    # OLLAMA_API_KEY was listed for a provider slot whose registry entry was
+    # never created, and opencode had stopped starting on both shells as a result.
+    opencode() { dotf secrets run --only NAN_API_KEY,OPENROUTER_API_KEY,OPENAI_API_KEY -- opencode "$@"; }
     pi() { dotf secrets run --only NAN_API_KEY,OPENROUTER_API_KEY -- pi "$@"; }
     # agy is deliberately NOT wrapped. It authenticates with its own stored
     # credentials and reads no variable this registry exposes — verified against
--- a/ai/opencode/opencode.jsonc
+++ b/ai/opencode/opencode.jsonc
@@ -16,7 +16,6 @@
 //   Embeddings / Audio: qwen3-embedding, kokoro (TTS), whisper (STT) — exposed for
 //                       cluster completeness; not chat-routable (use curl / OpenAI SDK directly).
 //   Frontier on-demand: openrouter/* (PAYG, uses existing $5 credit) — Opus / Sonnet / GPT-4
-//   Local:   ollama (homelab/VPN endpoint, user-managed in separate session)
 // Selection: TUI `/models` picker. Wrappers `qq` (nan/qwen3.6) / `qf` (nan/deepseek-v4-flash)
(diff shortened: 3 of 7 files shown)


mlorentedev:
## Review triage

Three findings, all valid, all applied in `295cc32`. Two of them were this guard repeating the mistake it exists to catch.

| Finding | Severity | Disposition |
|---|---|---|
| `openrouter` assertion matches a comment-only occurrence | Minor | **Applied** — matches the declaration on a non-comment line |
| 85-line `@test` body over the 40-line limit | Major | **Applied** — scanner extracted to `tests/lib/only-scope-scan.py` |
| Missing PyYAML exits 0 without scanning | Minor | **Applied** — now fails; a check that cannot run has not passed |

**On the first.** The reviewer is right and the failing input is exact: `grep -qi 'openrouter'` passes while the provider survives only in prose, so deleting the entry and keeping its comments read as healthy. I had *just narrowed* that assertion in this PR and left the same class of hole one step along — the same shape as the `Result=` enumeration on #1235.

Verified against the reviewer's own input rather than reasoned about: renaming the `"openrouter"` key while keeping every comment makes it `not ok`; restoring returns it to `ok`.

**On the third, which is the sharpest.** The scan already refused to pass when it matched zero tokens, on the stated grounds that a silent zero reads as health while proving nothing. Ten lines above that, in the same file, I wrote `except ImportError: sys.exit(0)` — the identical false all-clear, in the guard against it. It now exits non-zero saying so.

Whether it was silently skipping in CI could not be read from the logs, because bats prints test output only on failure. That ambiguity is itself the argument for failing loud.

**Evidence**

```
bats                  1519 passed, 0 failed, exit 0
openrouter guard      not ok with the entry removed / ok restored
--only guard          not ok with OLLAMA_API_KEY re-planted, naming file:line:token / ok restored
```

Rebased onto the branch after #1240 merged into it; suite re-run green on the merged tree, not only on my side of it.

mlorentedev:
## The removal left one loose end, and it cost real time tonight

Every deploy still logs:

```
render: unmapped placeholders left for runtime resolution: HOME OLLAMA_API_KEY
```

`OLLAMA_API_KEY` no longer exists as a live placeholder anywhere. The three
occurrences left in the repo are **prose**:

| Location | What it is |
|---|---|
| `ai/opencode/opencode.jsonc:264` | a **comment** explaining why the slot was removed |
| `ai/opencode/README.md:25` | documentation |
| `ai/nan/README.md:5,114` | documentation |

So the render's placeholder scanner **does not strip comments**, and reports as
"unmapped" something that exists only in an explanation of its own removal.

## Why it is worth closing rather than tolerating

It is a false positive in the deploy log, which is exactly where an operator
looks when something breaks. Measured cost tonight: while diagnosing why pi
reported *"No models available"*, this line read as a live unresolved provider
key and I chased it as the lead. It was not the cause — the cause was a terminal
without the wrapper — but the log had pointed at a dead end that looks alive.

**And it is a repeat.** This repository already caught the same shape in a
different scanner: an `openrouter` guard matching a comment-only occurrence,
found by the reviewer on the PR that removed openrouter. Two scanners, same
blind spot: text inside a comment treated as configuration.

## Suggested fix

Strip comments before scanning for `{env:VAR}` placeholders in `.jsonc`, the same
way the JSONC reader already does for parsing — the render and the parser
disagree about what counts as content, which is the divergence to remove. Or
scope the scan to the parsed document rather than the raw text.

Filed as a comment here rather than a new ticket: it is this issue's cleanup,
not a new defect.

PR #1248, opened 2026-08-27
fix(pi): a hand-wired extension symlink shadowed the packaged one, and pi would not start

### LINKED ISSUE #1243: AI-031: the pi-subagents package the manifest installs never loads, and the check counts declarations instead
## What happened

Dispatching through pi surfaces this on every run:

```
Error: Failed to load extension "/home/manu/.pi/agent/npm/node_modules/pi-subagents/index.ts":
  Tool "subagent" conflicts with /home/manu/.pi/agent/extensions/subagent/index.ts
Hint: Start without extensions using "pi -ne".
```

Two subagent extensions collide on the tool name, and **the one AI-030 installs loses**:

| | source | dated |
|---|---|---|
| loaded | `~/.pi/agent/extensions/subagent/{index,agents}.ts` — **symlinks** into `~/.nvm/versions/node/v24.16.0/lib/node_modules/@earendil-works/pi-coding-agent/examples/extensions/subagent/` | 9 Aug, hand-wired |
| **refused** | `~/.pi/agent/npm/node_modules/pi-subagents/` — declared in `ai/pi/packages.json`, installed by `setup-linux.sh` | today |

The symlink target still exists (node v24.16.0 is installed alongside v22.13.1 and v26.3.0), so this is a live conflict rather than a dangling leftover. `pi --version` does not surface it because that path loads no extensions; every real invocation does.

**So the deploy installs a package that is inert, and reports success.**

## Why nothing caught it

AI-030's own verification asks the wrong question:

```
jq -r '.packages // [] | length' ~/.pi/agent/settings.json
```

That counts **entries in a settings file** — declarations — not extensions that load. It returned `9 pi packages installed (manifest declares 9)` and was green while one of the nine could not load at all.

This is the fourth instance of the same pattern inside this epic's verification surface, after the three recorded in `docs/lessons/lesson-230`: a check that asks the *form* of a thing instead of what it does. The lesson exists; the check predates it.

## What a fix looks like

1. **Decide which subagent extension wins, and make it the only one.** The hand-wired symlinks are undocumented — they appear in no spec, no manifest, and no vault note. If `pi-subagents` is the declared one, the symlinks should go; if the examples extension is wanted, `pi-subagents` should not be in the manifest.
2. **Verify by load, not by declaration.** Something that runs pi and asserts no `Failed to load extension` appears — the package manifest is a statement of intent and the loader is the consumer.
3. **Detect the class.** Any two extensions claiming one tool name is a conflict pi already reports; nothing on our side reads that report.

Option 2 is the one that generalises: it would have caught this on the deploy that created it.

## Acceptance

- [ ] Exactly one extension provides the `subagent` tool, and which one is a recorded decision
- [ ] A pi invocation reports no `Failed to load extension`, asserted rather than assumed
- [ ] The pi package check verifies packages LOAD, not that they are listed in `settings.json`

## Context

`~/.pi/agent/agents/` also holds six hand-written agents from the same date (9 Aug), none carrying `generated: true`, two of which (`planner`, `reviewer`) now collide by name with the personas HARNESS-046 (#562) added, and two of which (`gemma-audit`, `mimo-ci`) have no reference anywhere in this repo or the vault. There is an undocumented manual pi layer predating the pipeline; this ticket covers the extension half of it. The agents half needs a decision from the owner of those files and is deliberately not bundled here.

Related: #1224 (AI-030, the packages manifest), #562 (HARNESS-046), `docs/lessons/lesson-230`.

mlorentedev:
`pi` would not start on this machine. `pi -p 'ok'` exited **1**:

```
Error: Failed to load extension ".../npm/node_modules/pi-subagents/index.ts":
       Tool "subagent" conflicts with ~/.pi/agent/extensions/subagent/index.ts
```

Two providers of one tool name. One from `npm:pi-subagents@0.56.0`, which
`ai/pi/packages.json` declares and setup reconciles on every run; the other a
symlink placed by hand on 2026-08-09 into pi's own bundled examples, under one
specific nvm node version. The hand-wired one won, so the **declared package was
the one that never loaded**.

## Why AI-030's verification could not see it

AC1-AC10 all held while pi refused to start. Every one of them counts entries in
an array — a package can be installed, declared and counted while the agent does
not start. **AC11** adds the missing criterion: a declared package is also a
loaded one, observed as effect.

## Design notes

**The check detects the cause, not the name.** An entry under the extensions dir
resolving through a symlink into a `node_modules` tree is hand-wired by
definition, since no package manager links that way. The scoping is load-bearing:
`~/.pi/agent/extensions/` has a live external writer (three `orca-*.ts` files
appeared there during this work), so claiming the directory wholesale would
report that writer as drift.

**`--fix` quarantines OUTSIDE the tree pi auto-discovers.** pi's docs list
`extensions/*/index.ts` as a discovery pattern, which a `.disabled/` subdirectory
beside the link would match — relocating the conflict one directory deeper while
reporting success. Destination is `~/.pi/agent/.disabled-extensions/`. The link
target is never touched, so the repair is reversible with `mv` and the example
code survives in the package.

**It lives in doctor rather than CI** because this is machine state, and CI never
sees this machine.

## Verification

| Check | Result |
|---|---|
| `pi -p 'ok'` | exit **1** -> exit **0**, answering `OK` |
| `dotf doctor` on the real machine | `[FAIL]` -> `[FIX ]` -> `(1 checks, all ok)` |
| `go build` + `go vet` | pass, Linux **and** `GOOS=windows` |
| AC11 command in `features.json` | executed, exit 0 |

New tests cover FAIL on collision, WARN without one, the scoping rule that leaves
an external writer's files alone, SKIP with no extensions dir, quarantine landing
outside the discovered tree, and no-clobber on a second `--fix`.

Lesson recorded: `docs/lessons/lesson-231-...`.

**Deployment note:** the check ships in the `dotf` binary. This machine is already
repaired, but `dotf doctor` will not carry the check until the installed binary is
rebuilt after merge.

fixes #1243

### DIFF
--- a/cli/internal/doctor/checks_pi_extensions.go
+++ b/cli/internal/doctor/checks_pi_extensions.go
@@ -1,0 +1,235 @@
+package doctor
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"slices"
+	"sort"
+	"strings"
+)
+
+// checkPiExtensions reports hand-wired pi extensions that shadow a package the
+// manifest installs (AI-030 / #1243).
+//
+// THE DEFECT THIS EXISTS FOR, measured 2026-08-26. `pi -p` exited 1 with:
+//
+//	Error: Failed to load extension ".../npm/node_modules/pi-subagents/index.ts":
+//	       Tool "subagent" conflicts with /home/manu/.pi/agent/extensions/subagent/index.ts
+//
+// pi did not start at all. Two `subagent` tools existed: one from
+// `npm:pi-subagents@0.56.0`, which ai/pi/packages.json declares and setup
+// reconciles, and one hand-symlinked on 2026-08-09 into pi's OWN bundled
+// examples directory under a specific nvm node version. The hand-wired copy
+// won, and the declared one is the one that never loaded.
+//
+// WHY THE EXISTING VERIFICATION COULD NOT SEE IT. AI-030's reconcile proves the
+// `packages` array of the live settings.json converges on the manifest, and
+// specs/AI-030-pi-packages-manifest/verify-reconcile.sh drives the real block to
+// prove it. Both are correct and both are blind here: a package can be
+// installed, declared, and counted while a file elsewhere stops it loading.
+// Counting declarations is not observing effect — the failure this repository
+// has now catalogued seven times inside this one spec family.
+//
+// THE RULE, and why it is this rule rather than a name list. A shadow is an
+// entry under ~/.pi/agent/extensions/ that resolves, through a symlink, into a
+// node_modules tree. That shape means exactly one thing: someone linked a
+// package's own source or example code in by hand. Nothing reproduces it — not
+// setup, not the manifest, not a fresh machine — and it is pinned to whichever
+// node version happened to be active the day it was made.
+//
+// It is deliberately NOT "any extension the manifest does not declare".
+// ~/.pi/agent/extensions/ also holds files written by other tools (three
+// orca-*.ts appeared there while this check was being written), and claiming
+// the whole directory for the manifest would report a live external writer as
+// drift. The symlink-into-node_modules rule catches the defect and leaves
+// genuinely hand-authored extensions alone.
+func checkPiExtensions(sys *System, cfg *Config, rep *Report, fix bool) {
+	rep.Section("pi extensions")
+
+	agentDir := filepath.Join(sys.home(), ".pi", "agent")
+	extDir := filepath.Join(agentDir, "extensions")
+
+	entries, err := os.ReadDir(extDir)
+	if err != nil {
+		if os.IsNotExist(err) {
+			rep.Skip("no ~/.pi/agent/extensions — pi has no hand-placed extensions to shadow anything")
+			return
+		}
+		rep.Warn(fmt.Sprintf("cannot read %s: %v", extDir, err))
+		return
+	}
+
+	installed := installedPiPackages(agentDir)
+
+	var shadows []piShadow
+	for _, e := range entries {
+		name := e.Name()
+		dir := filepath.Join(extDir, name)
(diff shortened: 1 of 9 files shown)


mlorentedev:
## Review triage

**Who actually reviewed:** PR-Agent (`github-actions`), which produced substantive
output. **CodeRabbit did not review this PR** — it posted `rate limited by
coderabbit.ai` / "Review limit reached, next included review available in 43
minutes". That is a notice about the review, not a review, and it is recorded
here rather than left to look like coverage that happened.

**CI:** all checks green — `test (ubuntu-latest)`, `test (windows-latest)`,
`lint` (x2), `lint-powershell`, `integration`, `changes`, `spec-gate`,
`attestation`, `goreleaser snapshot`, GitGuardian, `add-to-project`.

### Findings

| # | Finding | Disposition |
|---|---|---|
| 1 | **#1243 partially compliant** — "Exactly one extension provides the `subagent` tool, **and which one is a recorded decision**": the check quarantines the hand-wired link but the resolution decision is not recorded in the spec, an ADR, or a note | **Applied** — `f0008a7` |

Finding 1 is correct and was verified against #1243's acceptance list rather than
taken on trust: AC1 asks for two things and only the mechanism shipped. The
decision is now recorded in
`specs/AI-030-pi-packages-manifest/verification.md`, under *Decisions made during
implementation*, on four grounds — the packaged extension is reproducible
(setup installs it every run; nothing recreates the symlink), versioned (pinned
`@0.56.0`, so an upstream publish passes a diff and a reviewer), portable (the
link hard-codes `~/.nvm/versions/node/v24.16.0/` and rots with it), and
maintained. The loser is quarantined rather than deleted, because the decision is
about which one *loads*.

The reviewer's two "compliant" observations match what was measured: `pi -p` goes
from exit 1 to exit 0, and the check inspects symlinks on disk rather than
counting `settings.json` entries.

No other findings. No security concerns raised.

Issue #1253, opened 2026-08-27
HARNESS-087: the review gate's provider-diverse arm routes a model the map cannot resolve


mlorentedev:
## What

`harness/reviewer-pool.json` routes `agy/gemini-3.1-pro-high` as the
provider-diverse fallback arm of the adversarial-review gate. Its own `$comment`
calls that arm the entire point:

> *"A second provider family, independent of both NaN and Anthropic — provider
> diversity rather than merely a different model."*

`harness/model-map.json` cannot route it. Measured 2026-08-26 against
`origin/main`:

```
pools:            nan, claude, copilot, opencode
harnesses.agy:    pools: ["nan"]
tier/chain ids:   deepseek-v4-flash, haiku, mimo-v2.5, opus, qwen3.6,
                  qwen3.6-plus, sonnet
"gemini" in the map: NOT PRESENT
```

So the model that the review gate depends on for independence is not declared in
the routing registry, and `agy` is declared as reaching the `nan` pool — which is
not where that model lives.

## Why it matters

This is the second time this arm has turned out to be unexercised. #1156 found it
had **never run once**: the entry named `GEMINI_API_KEY`, a secret the registry
has never declared, so every attempt died at `dotf secrets run --only` and the
primary always answered. That was fixed by recording `auth: login`.

The pattern is the same both times — the fallback arm is the one nobody watches,
because a healthy primary hides it. A gate whose independence arm is unroutable
is a gate that silently has no independence.

## The open question this needs answered

`agy` is a **harness**, not a pool: `render: adapter`, and the model reaches it
as a launcher flag (`agy --model <id>`), exactly as for pi. So one of these is
true and it should be recorded, not assumed:

1. **The map needs a `gemini` pool** and `harnesses.agy.pools` should name it,
   because agy genuinely reaches a second provider family.
2. **`reviewer-pool.json` is out of the map's scope by design** — it is separate
   by decision already (its `$comment`: *"Kept in a separate file so H-044 need
   not migrate this one"*), and the map is not meant to route it.

If (1), this also unblocks the standing request for a **second `top` tier entry**:
`tiers.top` is `{"claude": "opus"}` alone, and adding a peer requires a pool
declaration for exactly this provider. Note `cli/internal/spec/review_launch.go`
already builds `agy --model <id> --output-format stream-json`, so the launcher
side exists; what is unverified is whether the generic `subprocess` backend in
`cli/internal/agent/process.go` can launch anything but claude.

## Acceptance

[...]

Issue #1244, opened 2026-08-27
HARNESS-085: revisit NaN catalog routing (qwen3.8-flash, glm5.3-flash) after the community's vote


mlorentedev:
NaN community announced two new models on 2026-08-26: `qwen3.8-flash` and `glm5.3-flash`,
on a promotional token allocation (1B tokens) that expires end of August 2026, ahead of a
community vote on whether/how the models' limits continue.

Both were added to `ai/pi/models.json`, `ai/pi/settings.json`, `ai/pi/README.md` and
`ai/opencode/opencode.jsonc` as catalog-only additions — picker availability, not defaults.
They were deliberately NOT wired into `harness/model-map.json` tiers, `.pr_agent.toml` (CI
reviewer), or `harness/reviewer-pool.json` (adversarial-review gate): the allocation is
temporary, and the announcement's own authors reported no measured quality gain over the
incumbent `nan/deepseek-v4-flash` ("apenas hemos notado diferencia").

## Why it matters

If the allocation survives the vote and becomes a stable NaN offering, these models are
next in line to be evaluated against `harness/reviewer-pool.json`'s admission bar (the
planted-defect test that admitted `mimo-v2.5`, per that file's own `$comment`) and
considered for `harness/model-map.json`'s `mid` tier or `.pr_agent.toml`'s lane-separation
scheme. If the vote drops or re-limits them, this closes as won't-do.

## Acceptance criteria

- [ ] NaN's community vote outcome (or continued availability past 2026-08-31) checked
- [ ] If still available: `qwen3.8-flash` and/or `glm5.3-flash` run through
      `reviewer-pool.json`'s admission procedure (planted-defect test, verdict stated
      honestly whether pass or fail)
- [ ] Decision recorded (admit / do not admit) in `harness/reviewer-pool.json` or this
      ticket, either way
- [ ] If not available: ticket closed as won't-do with a one-line note

mlorentedev:
Update from the catalog-add work: ran both models through this repo's own planted
`((count++))`/`set -e` bug — the same test that admitted `mimo-v2.5` — via
`scripts/nan-debug.sh`. Both diagnosed it correctly (post-increment returns the old
value 0, `set -e` treats that as failure). `glm5.3-flash` was noticeably more verbose
than asked (ignored a "one line" instruction, `qwen3.8-flash` respected it) — worth
weighing if either is ever run through the full admission procedure this ticket asks for.

[...]

Issue #1231, opened 2026-08-25
CONF-001: the NaN base URL is declared in 8 places and nothing keeps them consistent


mlorentedev:
## What

`https://api.nan.builders/v1` is hardcoded in **eight** separate files:

| File | Form |
|---|---|
| `ai/opencode/opencode.jsonc:43` | `"baseURL"` |
| `ai/pi/models.json:4` | `"baseUrl"` |
| `powershell/profile.ps1:16` | `$env:NAN_BASE_URL` |
| `powershell/profile.ps1:236` | `$env:NAN_BASE_URL` |
| `scripts/nan-bench.sh:28` | `${NAN_BASE_URL:-…}` |
| `scripts/nan-debug.sh:58` | `${NAN_BASE_URL:-…}` |
| `scripts/nan-quality-bench.sh:26` | `${NAN_BASE_URL:-…}` |
| `systemd/hive.service.d/10-dotf-secrets.conf:36` | `Environment=HIVE_WORKER_BASE_URL` |

Nothing asserts they agree. Changing the endpoint means finding all eight, and
the failure mode of missing one is silent: the consumer that kept the old URL
keeps working until the endpoint is retired, then fails in a way that points at
the consumer rather than at the drift.

## How it surfaced

Raised by PR-Agent on #1230, which added the eighth. The reviewer framed it as a
risk introduced by that PR; it is not — seven predate it. But the finding under
the framing is real, and worth recording rather than dismissing on the technicality.

## Why not fixed in #1230

Out of scope, and not small. Doing it properly means picking the SSOT surface
(`versions.conf` is for version pins, not endpoints — this probably wants its own
declaration), then rewiring eight consumers across four languages, two of which
(`.jsonc`, `.json`) are deployed by substitution at setup time and two of which
are read by processes that never source a shell. That is a spec-sized change.

## Acceptance

- [ ] One declared source for the NaN base URL, resolved by every consumer
- [ ] A check that fails when a consumer hardcodes an endpoint that disagrees
      with the declaration — the drift, not the string, is what must be caught
- [ ] `dotf doctor` or a bats guard, whichever surface actually runs on both OSes

## Notes

Scope is the NaN endpoint specifically. `ai/opencode/opencode.jsonc` also
declares an ollama endpoint for the TUI, which is a different provider and out
of scope here.

PR #1256, opened 2026-08-27
feat(harness): declare where model ids are pinned, and check they resolve

mlorentedev:
`harness/model-map.json` has a schema, a validating loader and real compile-time
consumers — and it binds no leaf. A model id is pinned in roughly seven live
places and nothing makes any of them agree with it.

Not theoretical. Measured 2026-08-26, the deployed `~/.pi/agent/settings.json`
carried `nan/deepseek-v4-flash-0731`, an id that no longer resolves — pi printed
`Warning: No models match pattern` on **every** start — plus three `openrouter/*`
entries for a provider this repo retired in August 2026. Nobody noticed until an
unrelated investigation read the file, because the repo cannot converge it:
`ai/pi/settings.json` is seed-if-missing and pi owns the live copy (#754).

## What this adds

`harness/model-pins.json` declares each **routing** pin — the places that decide
which model does work — with its pool and its normalization rule.

**Normalization is declared, never inferred**, because one model has three
correct spellings across these files:

| Spelling | File | Why |
|---|---|---|
| `nan:mimo-v2.5` | `model-map.json` | the registry's `pool:id` form |
| `openai/mimo-v2.5` | `.pr_agent.toml` | litellm's protocol name for NaN, not a vendor |
| `mimo-v2.5` | `ai/pi/models.json` | bare id |

Then two checks, matching the split shipped in #1248: **repo pins are a Go test**
that runs anywhere, **deployed pins are a `dotf doctor` check**, because CI never
sees a machine.

## The design decision worth reviewing

**A catalog is not routing.** A picker may offer models nothing routes — that is
what a picker is for — and `qwen3.8-flash` / `glm5.3-flash` are absent from the
map by a recorded decision (#1244).

The first version of this check reported every unrouted catalog id and fired on
`nan/gemma4`, a live NaN model nobody routes. The registry's own `$comment` warns
against precisely that; the check fell into it anyway, minutes after the warning
was written. Two narrow rules replaced one broad one:

- a **retired provider** (`openrouter/`, `codex/`), and
- a **frozen snapshot** — a declared id plus a date stamp (`deepseek-v4-flash` +
  `-0731`), pinning a moment of a model still alive under its rolling name.

Findings went from 5-with-1-false-positive to **4 with none**. Regression-guarded
by `TestModelPinsDoesNotReportAnUnroutedCatalogModel`, which names all three
legitimate cases. Recorded as `docs/lessons/lesson-232-...`.

## Live output, before any repair

```
[Model pin drift]
  [WARN] $HOME/.pi/agent/settings.json pi-deployed-enabled-models:
         "nan/deepseek-v4-flash-0731" is a frozen snapshot of "deepseek-v4-flash",
         and no longer resolves
  [WARN] ... "openrouter/deepseek/deepseek-v4-pro" names "openrouter", a provider
         this repository retired
  [WARN] ... "openrouter/qwen/qwen3-coder-plus"    (same)
  [WARN] ... "openrouter/minimax/minimax-m3"       (same)
```

`nan/gemma4`, in the same array, is correctly silent.

**No `--fix`, and the check is not given the flag.** Repairing means a surgical
merge into a file pi rewrites at runtime — the same disposition question the
extension symlinks raised in #1243, asked rather than defaulted. AC8 asserts
content *and* mtime are unchanged.

## Verification

| Check | Result |
|---|---|
| `go test ./internal/harness/` | 7 new tests; `resolved 10 routing pins across 6 repo files` |
| `go test ./internal/doctor/` | 7 new tests |
| `go build` + `go vet` + `GOOS=windows go vet` | pass |
| Every `features.json` command | executed, exit 0 |

AC3 exists so the guard cannot pass vacuously: it asserts the two live bad values
are rejected **and** that a live id still passes.

## Scope

Detect only. Generation for the three surfaces the pipeline owns is phase 2 on
#902; `ai/pi/settings.json` is never generated. `harness/reviewer-pool.json` is
out of scope and its unroutable `agy/gemini-3.1-pro-high` is filed as **#1253**.
`HIVE_EMBED_BASE_URL` is an endpoint, not a model id — #1231 stays open.

Refs #902; consolidates #1231, #1170, #1244. Amendment rationale posted on #902.

### DIFF
--- a/cli/internal/doctor/checks_model_pins.go
+++ b/cli/internal/doctor/checks_model_pins.go
@@ -1,0 +1,192 @@
+package doctor
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/mlorentedev/dotfiles/cli/internal/harness"
+)
+
+// checkModelPins reports routing pins in DEPLOYED files that the routing
+// registry no longer resolves (HARNESS-067, #902).
+//
+// WHY THIS IS DOCTOR AND NOT CI. The repo half of this guard is a Go test, which
+// runs anywhere. This half reads files on one machine — ~/.pi/agent/settings.json
+// and its neighbours — and CI never sees a machine. Same split as the pi
+// extension check next door: a check that cannot run where the defect lives is
+// not a guard.
+//
+// THE DEFECT, measured 2026-08-26. The deployed pi settings carried
+// `nan/deepseek-v4-flash-0731`, an id that no longer resolves — pi printed
+// `Warning: No models match pattern "nan/deepseek-v4-flash-0731"` on EVERY start
+// — plus three `openrouter/*` entries for a provider this repo retired in August
+// 2026. Nobody noticed until an unrelated investigation read the file, because
+// the repo cannot converge it: `ai/pi/settings.json` is seed-if-missing, pi owns
+// the live copy (#754), and no check looked.
+//
+// IT NEVER WRITES. Repairing would mean a surgical merge into a file the tool
+// rewrites at runtime, which is the same disposition question the hand-wired
+// extension symlinks raised in #1243 — asked, not defaulted. There is
+// deliberately no --fix here, and the check is not even given the flag.
+//
+// SEVERITY FOLLOWS CONSEQUENCE, not tidiness. A dead scalar routing pin
+// (`defaultModel`) decides what a real session runs on, so it fails. A dead
+// entry in a catalog array costs a warning line at startup and a stale picker
+// row, so it warns. Reporting both identically would either cry wolf about a
+// picker or under-report a broken default.
+func checkModelPins(sys *System, cfg *Config, rep *Report) {
+	rep.Section("Model pin drift")
+
+	pins, err := harness.LoadModelPins(cfg.DotfilesDir)
+	if err != nil {
+		// C15: an unreadable registry is not an empty one. Both would produce
+		// "no drift found"; they mean opposite things.
+		rep.Fail(fmt.Sprintf("%v\n    This check reads the deployed copy; if the file exists in your checkout, re-run setup to mirror it.", err))
+		return
+	}
+	m, err := harness.LoadModelMap(cfg.DotfilesDir)
+	if err != nil {
+		rep.Fail(fmt.Sprintf("routing registry unreadable, so no pin can be resolved: %v", err))
+		return
+	}
+	qualified, bare := harness.DeclaredModels(m)
+
+	sites := 0
+	values := 0
+	findings := 0
+
+	for _, site := range pins.Sites {
+		if site.Scope != "deployed" {
+			continue
+		}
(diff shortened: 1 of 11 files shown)


mlorentedev:
## Review triage

**Who reviewed:** PR-Agent (`github-actions`), substantive output. CodeRabbit
also commented — checked, and it is again the rate-limit notice, not a review.
Recorded so this does not read as coverage that happened.

**CI:** 16 SUCCESS, 1 SKIPPED (`goreleaser release`, correct on a PR), 0 failures.

### Findings

| # | Finding | Disposition |
|---|---|---|
| 1 | **Potential logic flaw** in `DeclaredModels`: a tier key that is not a declared pool contributes to `bare` but not `qualified`, so a pin could report `VerdictWrongPool` for a legitimate entry | **Applied as a test** — `cb2121b` |
| 2 | **#754 partially compliant** — does not remove OpenRouter models, update `ai/pi/README.md`, or add `tests/pi-config.bats` assertions | **Declined** — misattributed ticket |
| 3 | **#1248 partially compliant** — does not fix the pi startup failure or add AC11 | **Declined** — misattributed ticket |
| 4 | **#1244 partially compliant** — does not add `qwen3.8-flash`/`glm5.3-flash` to any configuration | **Declined** — misattributed ticket |

**Finding 1** was verified against the data before acting, and the reviewer's
specific example does not hold: it names `claude` and `opencode` as non-pool
consumers, but both **are** declared pools in `harness/model-map.json`. Checked
against `origin/main` — the tier keys are `claude`, `nan`, `opencode` and the
pools are `claude`, `copilot`, `nan`, `opencode`, so the set of non-pool tier
keys is empty. The scenario is **latent, not live**.

The behaviour is also deliberate: without a pool declaration nothing establishes
where a model lives, so `WrongPool` (a warning) is the correct conservative
verdict, and `Unknown` (a failure) would be wrong. The reviewer's real point
stands anyway — a comment is weaker than a test. `TestTierKeyThatIsNotAPoolStaysUnqualified`
now pins it on a synthetic map, so an edit that introduces a non-pool tier key
meets an assertion rather than prose.

**Findings 2–4** are misattribution. PR-Agent read the `#N` references in the PR
body and commit message — which cite #754 as a *constraint* this design works
around, #1248 as the *precedent* for the two-layer split, and #1244 as the
*recorded decision* the catalog rule must not violate — and evaluated the PR
against their acceptance criteria. This PR's ticket is **#902**, stated as
`Refs #902`. Doing what #754 or #1248 asked is not in scope; both are already
closed. No change.

Worth noting as a limitation rather than a defect: PR-Agent's ticket-compliance
pass does not distinguish a referenced issue from a targeted one. It is advisory
here, so it costs nothing but noise — but a gate built on it would misfire on any
PR that cites its own prior art.

All other observations were positive (`PR contains tests`, `No security concerns
identified`).

Issue #1262, opened 2026-08-27
CLI-049: setup treats a version pin as a floor and doctor treats it as equality, so a healthy machine warns forever


mlorentedev:
## What

`setup-linux.sh` and `dotf doctor` disagree about what a version pin means, and
only one of them says so.

**setup treats it as a MINIMUM** (REFACTOR-013, #474):

```sh
# the pin is a MINIMUM. Presence alone left an outdated pi in place forever...
# Upgrade when installed < pin; leave a newer pi alone.
if [ -n "$PI_VERSION" ] && ! version_gte "$installed_pi" "$PI_VERSION"; then
```

**doctor treats it as EQUALITY** (`matchPin`, `checks_deploy.go`):

```go
case installed == pin:  rep.Pass(...)
default:                rep.Warn(fmt.Sprintf("%s version drift: installed=%s pinned=%s", ...))
```

So setup deliberately leaves a newer binary in place, and doctor then warns about
the state setup just chose to create. Observed 2026-08-26:
`[WARN] pi version drift: installed=0.84.3 pinned=0.83.0`.

## Why this matters more than the noise

Fixed by bumping the pin this time (`1dc0de4`), and that was the right call —
the machine had been upgraded by hand and the repository did not know, so the
warning was doing real work.

But it does that work **by accident**. The same warning fires when the pin is a
correctly-declared floor and the installed version is legitimately ahead, which
is the intended steady state for every minimum-pinned tool. A warning that is
right for the wrong reason cannot be relied on, and a warning whose normal state
is amber gets muted — the same reasoning already recorded in
`harness/review-attestation.json` about a gate whose normal state is red.

## What a fix looks like

`versions.conf` does not say which pins are floors and which are exact. Options,
cheapest first:

1. **Declare the semantics per pin** (a suffix, or a companion list of
   minimum-pinned tools) and have `matchPin` compare with `version_gte` for
   those. `version_gte` already exists in `scripts/utils.sh`; the Go side needs
   the equivalent.
2. **Split the verdicts**: `installed > pin` becomes an INFO naming the drift
   without amber ("ahead of the declared floor — bump the pin if intended"),
   while `installed < pin` stays a WARN. Weaker, but it removes the permanent
   amber without needing a semantics declaration.

Option 1 is the honest one: the ambiguity is in `versions.conf`, so that is where
it should be resolved. Note the two consumers must not diverge again — whichever
form is chosen needs a test asserting setup and doctor read the pin the same way.

[...]

PR #1263, opened 2026-08-27
chore(versions): pin pi at 0.84.3, the version actually installed

mlorentedev:
The pin read `0.83.0` while this machine runs `0.84.3`, upgraded by hand outside
the pipeline. `dotf doctor` surfaced it as version drift — the check doing real
work: an environment change the repository does not know about is exactly what it
is for.

Codifying it rather than silencing it (Strict IaC: every environment change is
declared). The pin is a **minimum** per REFACTOR-013, so setup already left the
newer binary alone; what was missing was the declaration that this is the
intended floor.

## Verification

```
$ pi --version
0.84.3

$ bats tests/versions-conf.bats
17 ok, 0 failures
```

## Noted separately, not fixed here

While deploying, doctor and setup turned out to disagree about what a pin *means*
— setup treats it as a floor, `matchPin` treats it as equality — so a tool
legitimately ahead of its floor warns forever. Filed as **#1262**; this PR only
records the version actually installed.


mlorentedev:
## Review triage

**Who reviewed:** PR-Agent (`github-actions`). CodeRabbit commented — checked, and
it is the rate-limit notice again, not a review. Recorded so this does not read
as coverage that happened.

**CI:** all checks green, no failures.

### Findings

| # | Finding | Disposition |
|---|---|---|
| 1 | **#1262 partially compliant** — versions.conf does not record floor-vs-exact semantics; setup and doctor still disagree; a tool ahead of its floor still warns | **Declined** — misattributed ticket |

The PR body names #1262 explicitly as *"Noted separately, not fixed here"*. This
PR records the version actually installed on the machine; deciding what a pin
*means* is a different change with its own acceptance list, which is why it has
its own ticket. Doing it here would put an unrelated semantics change in a
one-line version bump.

`No relevant tests` is also correct and intended: `tests/versions-conf.bats`
asserts `PI_VERSION` is set, not what it equals, deliberately — a test pinning
the literal value would have to be edited by every future bump, which is a
tautology, not a guard. Run and green (17 ok, 0 failures).

**Third occurrence of the same reviewer limitation**, worth stating rather than
just declining: PR-Agent's ticket-compliance pass does not distinguish an issue a
PR *targets* from one it merely *references*. Here it evaluated the PR against a
ticket the body explicitly excludes. Same on #1256 (three tickets cited as prior
art) and on #1248. It is advisory, so the cost is noise — but any gate built on
this signal would misfire on every PR that cites its own context.

PR #1264, opened 2026-08-27
feat(harness): declare gemini pool for agy in model map to enable reviewer fallback

### LINKED ISSUE #1253: HARNESS-087: the review gate's provider-diverse arm routes a model the map cannot resolve

mlorentedev:
## Summary

Fixes #1253 (HARNESS-086) by declaring the `gemini` pool in `harness/model-map.json`, updating the `agy` harness to route to the `gemini` pool, and registering `gemini-3.1-pro-high` under `tiers.top`. This enables the provider-diverse fallback arm in `harness/reviewer-pool.json` to be properly resolved and routed.

## SDD skip rationale

Targeted configuration and routing addition (<50 LOC) with deterministic unit tests.

## Verification

- Added `TestModelMap_GeminiPoolDeclaredForAgyReviewerFallback` in `cli/internal/harness/model_map_test.go`.
- Validated model map against schema via `go test ./internal/harness/...`.
- Verified 0 harness drift via `./scripts/compile-harness.sh --check`.


mlorentedev:
## Review triage

Initial submission; CI checks pending.

mlorentedev:
## Review triage

- **CI status**: All 15 CI checks completed green (including Linux/Windows unit tests, integration tests, and bats suites).
- **Reviewer comments**:
  - **PR-Agent (github-actions)**:
    - *Finding 1 (Documentation gap on Option 1 vs 2)*: **Applied**. Explicitly recorded Option 1 resolution in `harness/model-map.json` `$comment` explaining why `gemini` is declared in `pools` and referenced by `harnesses.agy.pools`.
    - *Finding 2 (Proof of reviewer launch)*: **Applied/Verified**. Verified runtime launcher command assembly via `dotf spec review HARNESS-067-model-pin-drift --reviewer agy/gemini-3.1-pro-high --dry-run` and existing unit tests in `cli/internal/spec/review_launch_test.go` (`TestReviewerCommandRaisesTheAgyPrintTimeout`, `TestReviewerCommandRequestsAMachineReadableStream`).
  - **CodeRabbit**: Review limit reached (quota exhausted).
  - **Triage disposition**: Fully triaged and verified.

Issue #561, opened 2026-06-24
HARNESS-045: Determinism — full hook emission (claude+opencode+pi)


mlorentedev:
## Problem
Determinism is the core payoff: `skills:`/`enforce:` must compile to each harness's hook primitive on three levels, not just claude.

## Scope
- **Presence** (inject): claude `SessionStart` / opencode `chat.system.transform` / pi `session_start`.
- **Action** (gate): claude `PreToolUse`/`Stop` (exit 2) / opencode `tool.execute.before` / pi `tool_call`.
- **Invocation** (dispatch): a command/router spawns the role (`/review` → reviewer).
- Extend `compile-harness.sh` + `compile-harness.bats` with the hook-emission step.

## Acceptance
- For claude+opencode+pi, a forced skill is injected every turn AND a gate blocks progress until the required step ran — proven per harness.
- agy deferred to presence-only (see HARNESS-047).

## Refs
ADR-027 §3 (determinism via hooks), epic #558.

mlorentedev:
## Scoping, measured 2026-08-26 — three corrections and a constraint the ticket does not mention

### 1. The primitives are real. ADR-027 got these right

Verified against the installed packages, not the ADR:

| Harness | Presence | Action (gate) | Source |
|---|---|---|---|
| `pi` | `session_start` | `tool_call` | `pi-coding-agent/docs/extensions.md` — and it documents the semantics: *"`tool_call` errors block the tool (fail-safe)"* |
| `opencode` | `chat.system.transform` | `tool.execute.before` (+ `permission.ask`) | `@opencode-ai/plugin/dist/index.d.ts` |
| `claude` | `SessionStart` | `PreToolUse` | live in `~/.claude/settings.json` |

Worth stating because ADR-027's *render kinds* were found stale in 2 of 5 — its **hook names** are accurate. No cell is lost to a missing primitive.

### 2. The repo owns two events; the machine runs twelve

`ai/claude/settings.json` — the template this pipeline owns — declares **`SessionStart` and `SessionEnd` only**, with `__HOOK_COMMAND__` placeholders substituted by `merge_claude_settings()` through **hardcoded jq paths**:

```sh
(.hooks.SessionStart[0].hooks[0].command) = $cmd
| (.hooks.SessionEnd[0].hooks[0].command) = $endcmd
```

Adding a third event today means editing shell. Turning those paths into something the manifest generates **is the deliverable**, more than any individual hook.

[...]

PR #1272, opened 2026-08-27
feat(harness): the agnostic binding core for personas, and emission that coexists with a live third-party writer

mlorentedev:
Every persona body already claims:

> *"Your phase's skills are enforced by hook, not left to memory."*

It is false on every harness. Exactly one file in the CLI reads
`harness/agents/` and it is a doctor check; `dispatch.go` does not, and the
manifest had no hook concept at all.

This is the **first sitting** of #561: the agnostic core plus the coexistence
mechanism, both proven. `tasks.md` names what is deliberately not done.

## What lands

- **Persona loader** with declared per-skill severity, on `yaml/v3` rather than a
  third hand-rolled frontmatter reader.
- **`Decide()`** — the whole policy as a pure function over a harness-neutral
  `ToolCall`, tested with no harness installed.
- **`dotf harness gate`** — the binding primitive, one binary on every harness.
- **`MergeHooks`** — emission that coexists with a live third-party writer.
- **`agents.bind`** in the manifest, every event name measured.

## Three corrections to the ticket's own Scope

| Ticket says | Measured |
|---|---|
| agy is presence-only | **agy has `BeforeTool`**, in claude's exact command-hook format. Cheapest harness to add, not a deferred one |
| opencode gates on `tool.execute.before` | That is `(input, output:{args}) => Promise<void>` — it **cannot deny**. The primitive is `permission.ask` |
| (implied) four per-harness adapters | **Two families.** pi/opencode run in-process TypeScript, and since this repo generates their wrapper it emits one canonical payload — deleting two parsers instead of adding them |

## Coexistence is the central constraint

Orca owns **10 of claude's 12** hook events on this machine — including
`PreToolUse`, the event the gate needs — **all four of agy's**, and writes real
`.ts` files into pi's extensions dir.

So emission is find-by-marker: ours replaced in place, absent ours appended as a
**new group**, foreign entries never reordered, rewritten or removed. That also
retires a latent bug — `merge_claude_settings` writes
`.hooks.<event>[0].hooks[0].command`, a positional claim that silently overwrites
whatever a third party puts at index 0. A test drives exactly that ordering.

## Three defects found by running it, not by reading it

1. **A malformed payload blocked every call.** `normaliseToolCall` returned a
   zero `ToolCall` while the doc comment claimed `Decide` would resolve that to
   Allow — it does not. All nine `Decide` tests passed, because the defect lived
   in the **seam** between normalisation and decision.
2. **Identity by repository evicted a live hook.** After adopting
   `dotf mem session-start`, emitting the gate on the same event matched its
   marker and replaced it. Identity had to become per-purpose.
3. **Adoption by exact command was missing**, so taking over a positionally
   written hook would have appended a duplicate.

## Verification

```
claude EXIT=2   agy EXIT=2   pi EXIT=2   opencode EXIT=2
pi invoke skill -> 0, retry -> 0
garbage -> 0 ("payload not recognised"), real call same run -> 2

real ~/.claude/settings.json: 12 foreign hook entries, all preserved
```

| | |
|---|---|
| `bats tests/compile-harness.bats` | 72/72, exit 0 |
| build + vet, Linux **and** `GOOS=windows` | pass |

## Deliberately not in this PR

Setup wiring, presence emission, the pi/opencode TS wrappers (`emit: false` in
the manifest so the gap is visible), and the block-style guards.

That last one matters: **four parsers read `skills:` and three break silently**
on the mapping form — `check-roster-consistency.py` falls back to `[]`,
`readAgentFrontmatter` skips indented lines, and `compile-harness.sh`'s
`skill_field` renders empty. The fix already exists in that same file:
`agent_capability_line()` detects the block form for `capabilities:` and refuses,
because an empty allow-list is not "no opinion" but "full default access". The
identical argument applies to `skills:`. Migrating the 35 skills is blocked on
copying that guard, or the migration silently disarms three readers.

Refs #561.


mlorentedev:
## Review triage

**Who reviewed:** PR-Agent (`github-actions`), substantive. CodeRabbit posted its
rate-limit notice again — a notice about a review, not a review, recorded so this
does not read as coverage that happened.

**CI:** 16 SUCCESS, 1 SKIPPED, 0 failures.

This PR merged before its review was dispositioned. Both findings are real and
both are applied in **#1275**, each with a regression test.

| # | Finding | Disposition |
|---|---|---|
| 1 | **Skill invocation detection gap** — a well-formed payload naming the skill primitive with a missing argument yields an empty skill name, falls through to enforcement, and blocks permanently | **Applied** in #1275 |
| 2 | **Session state collision** — `a/b` and `a.b` both map to `a_b`, so one session's consumption opens another's gate | **Applied** in #1275 |
| 3 | AC1 wiring / AC2 presence / AC3 per-harness proof listed as non-compliant | **Correct, and stated as scope** — `tasks.md` names each as deliberately not done, with the reason. Not a defect |

**Finding 1 is the one that mattered**, and it is worse than the reviewer put it.
The gate has a deadlock guard — `TestGateNeverBlocksTheSkillInvocationItself` —
that passed throughout, because it always supplies a skill name. The guard was
keyed on the skill's **name** when it had to be keyed on the tool's **identity**,
so the one case it existed to prevent was the one it never tested. A blocked
skill invocation can never record consumption, so the session stays blocked.

**Finding 2** is theoretical for UUID-sending harnesses and applied anyway: a
session id is attacker-adjacent input that lands in a filesystem path, and "it
does not happen with well-behaved input" is not a property a path builder should
rest on.

The "requires further human verification" items — wiring, presence, per-harness
end-to-end — are exactly the ones held back deliberately. `tasks.md` records them
under *Not done, and deliberately so*, including why the pi/opencode TS wrappers
are a sitting of their own.

PR #1275, opened 2026-08-27
fix(harness): apply the reviewer's findings on the persona gate

mlorentedev:
Follow-up to #1272, which merged before its review was dispositioned. PR-Agent raised two defects; both are real, both applied, each with a regression test.

## 1. A skill invocation could deadlock the session

On a **well-formed** payload naming the skill primitive but carrying no readable argument, the skill name came back empty, the call fell through to enforcement, and the gate blocked the one action that could satisfy it — permanently, since a blocked call never records consumption.

The guard keyed on the skill's **name** when it had to key on the tool's **identity**. `TestGateNeverBlocksTheSkillInvocationItself` passed throughout, because it always supplied a name — the case that was never in danger.

## 2. State paths could collide

Character-mapping alone flattens `a/b` and `a.b` to the same file, so one session's consumption record would open another session's gate. Well-behaved harnesses send UUIDs and never hit it, but a session id is attacker-adjacent input landing in a filesystem path, and "it does not happen with well-behaved input" is not a property a path builder should rest on. A digest is appended; the readable prefix stays so the directory is diagnosable by eye.

## 3. One my own guard surfaced

Found while checking whether the pin registry shipped in #1256 had gone red on main: `tiers.low` acquired a `$comment`, and `DeclaredModels` treated every string in a tier as a model id — so the whole sentence entered the declared set. It only ever **widened** the set, so no check could fail wrongly, but a registry that treats prose as an id is one coincidence away from masking real drift.

**That check worked as designed**, which is worth recording: `qwen3.8-flash` was promoted to pi's routed default **and** added to the map in the same change. That coupling is exactly what the pin registry exists to force, and it held within hours of shipping.

## Verification

| | |
|---|---|
| New regression tests | 3, one per finding |

Refs #561.

### DIFF
--- a/cli/internal/cmd/harness_gate.go
+++ b/cli/internal/cmd/harness_gate.go
@@ -82,6 +82,15 @@ Invoking a skill is never blocked: forbidding it would deadlock the session.`,
 				_ = harness.RecordConsumed(statePath, call.Skill)
 				return nil
 			}
+			if call.IsSkillTool {
+				// The tool IS the skill primitive but its argument was
+				// unreadable. There is nothing to record, and blocking would
+				// deadlock the session on the one action that could satisfy the
+				// gate — a well-formed payload with a missing argument, not a
+				// parse failure. Raised by the reviewer on #1272.
+				_, _ = fmt.Fprintln(cmd.ErrOrStderr(), "[gate] allow: skill invocation with no readable name")
+				return nil
+			}
 
 			persona := loadGatePersona(repoRoot, role)
 			result := harness.Decide(harness.GateInput{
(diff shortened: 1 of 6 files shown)


mlorentedev:
## Review triage

**Who reviewed:** PR-Agent (`github-actions`), substantive, including a harness
compliance pass. CodeRabbit posted its rate-limit notice — a notice about a
review, not a review.

**CI:** all checks green, 0 failures.

### Findings

| # | Finding | Disposition |
|---|---|---|
| 1 | **Definition of Done §5 FAIL** — `verification.md` documents the findings but carries no command output proving the tests pass after the change | **Applied** — `b581bde` |
| 2 | #1256 compliance: `DeclaredModels` skips `$`-prefixed keys, test added, no non-compliant items | Acknowledged, no action |

**Finding 1 is correct and it is the rule I am bound by.** The section asserted
the suite passes and the PR description carried a table saying so — an assertion
is precisely what §5 exists to refuse. Output now appended verbatim, run
uncached in this session:

```
$ go clean -testcache && go test ./internal/harness/ ./internal/cmd/
ok  .../internal/harness  0.040s
ok  .../internal/cmd      0.570s

--- PASS: TestGateNeverBlocksASkillToolWithAnUnreadableName
--- PASS: TestGateStatePathDoesNotCollideAcrossDistinctSessions
--- PASS: TestDeclaredModelsIgnoresAnnotationKeysInTiers
```

Worth noting the reviewer classified this correctly on its own terms too: it
passed `spec_driven_development` as a justified SDD skip (a bug-fix follow-up,
under 20 lines per defect with an obvious cause) while failing
`definition_of_done` on the missing evidence. Those are the right two calls, and
the second one caught a real gap rather than a formatting preference.

`bats tests/compile-harness.bats` was not re-run for this PR and does not need to
be: the diff is Go and Markdown only, and that suite covers the shell renderer.
Stated rather than left implicit.

Issue #1278, opened 2026-08-27
HARNESS-088: the handoff block is one shared mutable slot, so concurrent sessions overwrite each other and it looks like success


mlorentedev:
## What happened, twice in one evening

`MEMORY.md` carries **one** `## Session Handoff` block. Two sessions wrote it in
sequence and each **replaced** the other's rather than merging:

- claude session 3 replaced the antigravity block (I caught it mid-edit and
  restored theirs by hand)
- claude session 4 then replaced session 3's, and its cross-reference —
  *"per the prior handoff below"* — pointed at a block that no longer existed

The skill already forbids this. `harness/skills/handoff/SKILL.md:61`:

> *"Re-read `MEMORY.md` immediately before writing; match-and-replace only the
> `## Session Handoff` block... If concurrent writes occurred, merge threads."*

## Why the rule keeps losing

**The failure is invisible to the writer.** Last-writer-wins produces a
well-formed file and a successful edit; nothing looks wrong until a later session
follows a dangling pointer. A rule kept by diligence does not hold against a
structure that makes violating it indistinguishable from succeeding — the same
reasoning already recorded here about a positional hook writer overwriting a
third party's entry.

**The root cause is the data structure, not the prose.** One shared mutable slot,
N concurrent writers. This machine currently has **five active worktrees**.

## The shape that fixes it

**1. Per-thread sub-blocks, written by a command.** `## Session Handoff` becomes
N sub-entries keyed by worktree: `### wt-pi-harness (branch)`. A session rewrites
**only its own**; every other is foreign and untouchable.

That is exactly `MergeHooks` from #1272, one content-type over: find-by-marker,
replace in place, append if absent, never reorder or delete a foreign entry. Port
it as `dotf mem handoff-write --thread <key>`, and the skill's step 1 becomes one
line — *run the command* — with the merge living in Go where it is tested.

**It must share `extractHandoffBlock`'s section boundary** (`internal/mem/session_end.go:87`),
not re-derive it. Verified: that parser stops at the next `## `, so `###`
sub-blocks fall inside it and archival keeps working. Writing a second boundary
rule would be the fifth silently-divergent parser this repository has found in a
week.

**2. Journals become findable from the session that wrote them.** Naming today is
`<date>-<project>-<agent>.md`, so two *worktrees* on one day collide into `-2`,
`-3` suffixes that encode nothing. Six such files exist across two days. Add the
[...]

PR #1279, opened 2026-08-27
feat(mem): one handoff thread per worktree, so concurrent sessions stop clobbering each other

mlorentedev:
`MEMORY.md` carries **one** `## Session Handoff` block and this machine runs **five worktrees**. Three sessions wrote it in sequence on 2026-08-26/27 and each **replaced** the previous one's. None noticed — session 4's block ended up saying *"per the prior handoff below"* with no handoff below.

The skill already forbade it (HARNESS-028: *"if concurrent writes occurred, merge threads"*).

## Why the rule kept losing

**The failure is invisible to the writer.** Last-writer-wins produces a well-formed file and a successful edit; nothing looks wrong until a later session follows a dangling pointer.

One shared mutable slot with N concurrent writers is a **data-structure problem**. Editing it more carefully was never going to fix it — which is why this replaces the rule with a mechanism rather than restating it.

## What changes

The section holds **one sub-block per thread**, keyed by worktree (`### wt-pi-harness`), written by `dotf mem handoff-write`. A session replaces only its own; every other is foreign and untouchable.

That is `MergeHooks` from #1272 one content-type over — find-by-marker, replace in place, append if absent, never reorder or delete a foreign entry — and the same posture this repo already takes toward Orca's hooks.

**It shares `extractHandoffBlock`'s section boundary** rather than re-deriving it. Checked before writing a line, because a second boundary rule would have been the **fifth silently-divergent parser found here in a week**. AC4 asserts the agreement by driving the archiver over the written file.

**Journals gain the thread.** Two *worktrees* on one day were colliding into `-2`/`-3` suffixes encoding nothing — six such files across two days, and no session could derive its own. `dotf mem thread` prints the key and the filename.

**Shared surfaces are named, not schemed.** No lock file, no daemon, no per-session directories: the list is short enough to state, and each row's fix is that one command owns the write.

## A defect this found in itself

`ThreadKey` first read only `filepath.Base(cwd)`, so a session in any subdirectory — `cli/`, where most work here happens — resolved to `main`. **Every such session would have shared one key and clobbered the others exactly as before.** The tests passed, because every case they supplied was a worktree root.

Found by running `dotf mem thread` from `cli/` and reading the output. It walks up now, with three subdirectory cases in the table.

## Verification

```
$ dotf mem thread --date 2026-08-27 --project dotfiles --agent claude
thread   wt-pi-harness
journal  sessions/2026-08-27-dotfiles-claude-wt-pi-harness.md

$ printf '...' | dotf mem handoff-write --memory <copy of the real MEMORY.md>
wrote      thread "wt-pi-harness"
  session 4's block survived : True
  my thread written          : 1 heading
  sections intact            : Index, Findings, CI Pipeline, User Preferences, Session Handoff
```

| | |
|---|---|
| New tests | 9, incl. a fixture shaped like the actual clobber |
| `bats tests/compile-harness.bats` | exit 0 |

**AC8 not built** and recorded as such: a `dotf doctor` check for a handoff section that lost a thread whose worktree still exists. The writer removes the cause; that check would catch a hand-edit reintroducing it.

Refs #1278 — AC8 (the doctor check) is deliberately not built, so the work genuinely continues and the spec stays active.


mlorentedev:
## Review triage

**Who reviewed:** PR-Agent (`github-actions`), substantive — two real defects.
CodeRabbit posted its rate-limit notice: a notice about a review, not a review.

**CI:** all green, including `test (windows-latest)`.

### Findings

| # | Finding | Disposition |
|---|---|---|
| 1 | **`###` inside a thread body truncates its span** — a body carrying `### Next Actions` ends the thread early, so a replacement rewrites only the part above it and strands the rest | **Applied** — `b15064e` |
| 2 | **`ThreadKey` fallback could collide** with a directory containing `-wt-` | **Applied** — `b15064e`, on the measurement rather than the example |

**Finding 1 was data loss on ordinary content**, and worse than stated: the
stranded remainder ends up under the *previous* thread, so a peer's block grows
someone else's text. Reproduced by a failing test first.

The first fix attempt reached for a heuristic — deciding which `###` is "really"
a thread from the set of keys present — and it would have promoted `Next` to a
thread. **No heuristic can separate them; the shapes are identical.** Identity is
declared instead: `### thread: <key>`, the same choice made for hook entries in
#1272. Worth recording that the wrong fix was written first and caught by its own
test.

**Finding 2's example does not hold** — `my-worktree-fix` contains no `-wt-`
substring — but the defect it points at is real: interleaving the git lookup with
the naming fallback let `vendor-wt-cache` inside a plain checkout claim its own
thread while the `.git` one level up said `main`. git is consulted across the
whole walk-up now, before the convention is consulted at all. Fixed on the
measurement, not on the example.

### Also in this PR, from an audit pass rather than the reviewer

- **`SessionEnd` archived every worktree to one hardcoded filename** with a
  truncating write, so the last session destroyed the others' durable records —
  the same defect, in the path meant to be the safe copy. It also assembled a
  filename the skill did not share; both come from `JournalName` now.
- **The thread key was tied to this repo's naming convention**, so a worktree
  made by any other tool resolved to `main`. It asks git now, which makes it
  agent-agnostic by construction.
- **A Windows-only test defect**: the payload was built by concatenating a path
  into JSON, and `C:\Users\…` backslashes are JSON escapes — the parse failed and
  `SessionEnd` no-opped silently. `GOOS=windows go vet` cannot see it; the
  Windows leg of CI is the only thing that did.

Each fix was written test-first and **mutation-checked** — fix removed, test
observed failing, fix restored.

### Deferred, recorded in `tasks.md`

`SessionEnd` resolves the project as `filepath.Base(cwd)`, so a session whose cwd
is a subdirectory finds no `MEMORY.md` and **silently archives nothing**. Same
class, different function, and it changes *which* sessions archive at all — that
needs its own measurement, not a fold-in at the end of another PR.

PR #1280, opened 2026-08-27
feat(mem): a thread is the branch, so work follows you between machines

mlorentedev:
Second sitting on #1278. Fixes the debt the first one recorded, and answers *"how does each worktree find its own session in the vault, across machines, with the vault as SSOT"*.

## The debt, fixed

`SessionEnd` resolved the project as `filepath.Base(cwd)`. A session whose working directory was a **subdirectory** — `cli/`, where most work in this repository happens — resolved the wrong project, found no `MEMORY.md`, and **silently archived nothing**. The no-op resilience contract is exactly what made it invisible.

git is authoritative now. The basename stays as a fallback so a session outside any repository keeps working: **fixing a defect must not quietly narrow who the function serves.**

## A thread is a LINE OF WORK, and git already names it: the branch

The first sitting keyed on the worktree. That correlates on one machine and **decorrelates the moment the work moves to another** — and the vault is the SSOT across machines, so `feat/x` continued on a second box must resolve to the *same* thread rather than forking one.

Measured premise before committing to it: three live worktrees, three distinct branches, no duplicates.

| Situation | Thread |
|---|---|
| any feature branch | `feat-x` — clean, travels |
| `main` / `master` | `main@<host>` — ambient work; two machines' `main` are not one thread |
| detached HEAD | `<worktree>@<host>` — says so rather than guessing |

The hostname enters **only where it disambiguates** — the same rule `JournalName` already applies to its `-main` suffix.

## One RepoIdentity feeds every consumer

`ThreadKey` derived the worktree from a naming convention while `SessionEnd` derived the project from a basename: two derivations of one fact, in one package, disagreeing. Keeping them would have been the **sixth divergent-parser defect found here in a week** — this time between my own functions.

It reads git's own on-disk state (a linked worktree's `.git` is a file naming its gitdir; `HEAD` names the branch) with **no subprocess and no naming convention**, so Claude Code, Orca, opencode, pi, agy, copilot or a bare `git worktree add` all behave identically.

## Verification

```
$ dotf mem thread --date 2026-08-27 --project dotfiles --agent claude   # from cli/
thread   feat-harness-088-handoff-threads

$ cd ~/Projects/dotfiles && dotf mem thread
thread   main@msi
```

| | |
|---|---|
| New tests | 5, incl. the same-branch-two-worktrees cross-machine case |

Each fix written test-first and mutation-checked.

## A mutation test of mine gave a false negative

The first mutation replaced a string written with **one** space where `gofmt` had written **two**, so nothing was mutated — and the passing test read as *"the test does not catch this"*. Re-run against the real target, it failed as it should.

**A mutation test reporting "no change" has not proven the test is weak; it has proven nothing at all.** Recorded in `verification.md`.

Refs #1278.

### DIFF
--- a/cli/internal/mem/handoff.go
+++ b/cli/internal/mem/handoff.go
@@ -193,20 +193,30 @@ func trimTrailingBlank(lines []string, from, to int) int {
 
 // ThreadKey derives a stable thread identity from a working directory.
 //
-// The worktree, because that is what distinguishes two concurrent sessions — not
-// the agent and not the date. Journals were named `<date>-<project>-<agent>.md`
-// and two WORKTREES on one day collided into `-2`/`-3` suffixes encoding nothing;
-// six such files exist across two days, and no session could derive its own.
+// A THREAD IS A LINE OF WORK, AND GIT ALREADY NAMES IT: THE BRANCH.
 //
-// IT ASKS GIT FIRST, and only then falls back to this repository's naming
-// convention. That ordering is what makes the key agnostic across tools.
+// The first version keyed on the worktree. That correlates on one machine and
+// decorrelates the moment the work moves to another — and the vault is the SSOT
+// across machines, so the same branch continued on a second box must resolve to
+// the SAME thread rather than forking one. Every worktree in this repository is
+// born on its own branch (measured: three live worktrees, three distinct
+// branches, no duplicates), so the branch discriminates at least as well and it
+// travels.
 //
-// Several tools create worktrees here and each names them differently: this
-// repository writes `<repo>-wt-<slug>`, Claude Code's EnterWorktree writes
-// `.claude/worktrees/<name>`, Orca writes its own. A key derived from one
-// pattern resolves every other tool's worktree to "main" — so every session
-// running under a different tool would share one thread and clobber the others,
-// which is the bug this exists to fix, reintroduced for everyone not using our
+// THE HOSTNAME ENTERS ONLY WHERE IT DISAMBIGUATES. `main` on two machines is two
+// pieces of ambient work, so the default branch is keyed `main@<host>` while
+// every feature branch stays clean — the same only-when-it-disambiguates rule
+// JournalName applies to its `-main` suffix.
+//
+// A DETACHED HEAD is named `<worktree>@<host>` and says so, rather than
+// collapsing into a plausible "main" that would silently share a thread with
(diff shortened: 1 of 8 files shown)


mlorentedev:
## Review triage

**Who reviewed:** PR-Agent (`github-actions`). CodeRabbit posted its rate-limit
notice — a notice about a review, not a review.

**CI:** all green, including `test (windows-latest)` and `lint`.

### Findings

| # | Finding | Disposition |
|---|---|---|
| 1 | **"Compilation Error"** — `ThreadKey`'s fallback loop calls `gitWorktreeName`, which this PR removed | **Declined — factually incorrect**, evidence below |
| 2 | **AC8 non-compliant** — `dotf doctor` does not report a handoff section that lost a live thread | **Correct, and out of scope by declaration** |

### Finding 1 — verified before declining

The symbol does not exist anywhere in the tree, and the build is clean:

```
$ grep -rn "gitWorktreeName" cli/
(no matches)

$ go build ./... && go vet ./... && GOOS=windows go vet ./...
(clean, both platforms)

$ go test ./...
19/19 packages ok

$ golangci-lint run          # pinned 2.12.2
0 issues.
```

And CI agrees: `test (ubuntu-latest)`, `test (windows-latest)`, `lint`,
`integration` and `spec-gate` are all green on this head. A compilation failure
of that kind could not produce a green `test` job on two platforms.

The call **and** the function were removed in the same commit. The review appears
to have read the diff hunk's tail — the fallback loop's `parent := filepath.Dir(dir)`
lines, which survive — without the removed call above them being in the same
window. A partial hunk reads exactly like a dangling reference.

Recorded rather than silently dismissed, because a "compilation error" claim is
the kind that should be either true or explicitly refuted with the command
output, never left ambiguous.

### Finding 2 — correct, and already declared

AC8 is listed in `proposal.md` under **Acceptance criteria** as *"Not built"* with
the reason, and again in `tasks.md`. The writer removes the *cause* of a lost
thread; that check would catch a hand-edit reintroducing it, which is a different
job and worth its own evidence.

Two more deferrals stay recorded in `tasks.md` rather than folded in here: the
gate's session state is keyed by session id alone (two harnesses reusing an id
would share a record), and the `-2`/`-3` journals from before this change are
left alone, because renaming append-only history to match a later convention
rewrites a record to say something it did not say.

PR #1285, opened 2026-08-27
docs(lessons): record the PATH-ordering diagnosis and index five orphaned lessons

mlorentedev:
## What

Adds **lesson 235** and gives **lessons 231–235** their rows in `docs/lessons/_index.md`.

### The lesson

`pi` reported *"No models available"* in fresh terminals (#1283). The guard around the agent-wrapper block in `.zshrc`/`.bashrc` is evaluated ~40 lines **before** `~/.local/bin` — where `dotf` lives — is prepended to PATH; in `.bashrc` the line before that prepend *replaces* PATH outright. So the block ran only when the parent process happened to export that directory already, and a terminal with a clean PATH skipped it entirely and ran `pi` unwrapped, with no credentials and no error.

Four hypotheses were tested first — stale terminal, duplicate nvm install, locked vault, retired `OLLAMA_API_KEY` — and all four came back "works". They came back "works" because every measurement ran from a shell whose PATH already hid the defect. The transferable part is the instrument, not the ordering bug: **"I cannot reproduce it" is a statement about the instrument before it is a statement about the system**, and four probes agreeing is itself the tell.

### The index gap

Four lesson files (231–234) had been written and **none had an index row** — undiscoverable through the only surface that lists them. Measured **234 files against 230 rows**.

The check that exists for this, `mem.lessonsStaleness`, stats `_index.md`'s **mtime** over a 14-day window. The index had been edited 3 days earlier — just not for these lessons — so it reported green the whole time. That is a proxy where an invariant is available (every `lesson-NNN-*.md` has a row); ticketed rather than fixed here, since it is guards machinery and this PR is docs-only.

Both directions now verified by hand: 235 files / 235 rows, and every one of the 235 link targets resolves to an existing file.

### The header count

Dropped `(212 entries)` from the header rather than correcting it. It was a third copy of a fact the directory listing and the table already hold, it had drifted from both, and a stale count reads as authoritative.

## Verification

```
files: 235
rows:  235
link check done      # zero broken targets across all rows
```

Docs-only; no production LOC.

---

## Second commit: `handoff-write --help` described the keying it replaced

Found while reading `--help` to check for a `--thread` override. The Long text still said the key defaults to *"this working directory's worktree (e.g. wt-pi-harness)"* and that *"a checkout that is not a worktree resolves to main"* — the #1279 behaviour that **#1280 replaced the same day**. The key is the branch; `main`/`master` are host-qualified; a detached HEAD falls back to `worktree@host`.

Same defect class as the rest of this PR — a declared shape that no longer matches effect — and it matters more here than in a comment, because `--help` is the surface an agent reads to decide how threads key, so a stale one propagates into every skill that follows it.

Also documents what `--thread` is actually for: a session that switches branches mid-flight re-keys, and the line of work may well be the one it started on. (This PR is that case: the session began on `fix/shell-wrapper-path-ordering` and branched here.)

Help text only, no behaviour change. Verified:

```
go build ./... && go vet ./... && GOOS=windows go vet ./... \
  && go test ./internal/cmd/... ./internal/mem/...
ok  github.com/mlorentedev/dotfiles/cli/internal/cmd
ok  github.com/mlorentedev/dotfiles/cli/internal/mem
```

### DIFF
--- a/cli/internal/cmd/mem_handoff.go
+++ b/cli/internal/cmd/mem_handoff.go
@@ -36,9 +36,14 @@ func newMemHandoffWriteCmd() *cobra.Command {
 		Long: `handoff-write replaces one thread's sub-block under "## Session Handoff",
 reading the body from stdin. Every other thread is left byte-identical.
 
-The thread key defaults to this working directory's worktree (e.g. wt-pi-harness),
-because that is what distinguishes two concurrent sessions — not the agent and not
-the date. A checkout that is not a worktree resolves to "main".
+A thread is a LINE OF WORK, and git already names it: the branch. So the key
+defaults to the current branch (e.g. feat-x), which means work resumed tomorrow on
+another machine lands in the same thread. Ambient work on main/master is qualified
+by host (main@msi), because two machines' main are not one line of work, and a
+detached HEAD falls back to worktree@host rather than guessing.
+
+Pass --thread to override — a session that switches branches mid-flight re-keys
+otherwise, and the line of work may well be the one it started on.
 
 Skills should call this instead of instructing an Edit: the merge is the part that
 was being got wrong, and it belongs where it can be tested.`,
@@ -103,7 +108,7 @@ was being got wrong, and it belongs where it can be tested.`,
 	}
 
 	cmd.Flags().StringVar(&memoryPath, "memory", "", "path to the project's MEMORY.md")
-	cmd.Flags().StringVar(&thread, "thread", "", "thread key (default: this worktree)")
+	cmd.Flags().StringVar(&thread, "thread", "", "thread key (default: this checkout's branch)")
 	cmd.Flags().BoolVar(&dryRun, "dry-run", false, "print the result instead of writing it")
 	return cmd
 }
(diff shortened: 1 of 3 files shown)


mlorentedev:
## Review triage

|---|---|---|
| **PR-Agent** (`review` job) | Effort 1/5 · no major issues · no security concerns · ticket compliance clean on #1283/#1279/#1280 | **Accepted** — no findings to apply |
| PR-Agent | 🧪 *No relevant tests* | **Declined, with reason** — see below |
| **CodeRabbit** | *"Review limit reached — next included review available in 40 minutes"* | **Not a review.** Disclosed, not waited on |
| CI | `gh pr checks`: **16 pass, 1 skip (`goreleaser release`), 0 pending** — including `test (windows-latest)` and `test-windows` | Clear |
| `review-attestation` | *"attested — reviewed by: github-actions (comment-shaped review)"* | Corroborates the disclosure below — the repo's own attestation names PR-Agent, not CodeRabbit |

### On "no relevant tests"

Correct observation, deliberately not acted on. Both commits are text: five index rows, one lesson file, and two cobra help strings. There is no behaviour to assert — `go test ./internal/cmd/... ./internal/mem/...` passes unchanged precisely because nothing executable moved.

The test that *would* be relevant here is the one that fails when a lesson file has no index row, and that is the substance of **#1286** rather than something to bolt onto a docs PR. Writing it here would put the guard in the same commit as the backlog it was supposed to have caught, which is the shape #1286 argues against.

What did get verified, by hand and recorded in the PR body: **235 files / 235 rows**, and all 235 link targets resolve.

### On the reviewer

CodeRabbit's free-tier quota is exhausted, so **it did not review this PR** — the comment it left is a notice about the review, not a review. Per repo doctrine that is a disclosure, not a pass: proceeding on PR-Agent's substantive review alone, and saying so rather than letting a green wall imply two reviewers looked.

Issue #1301, opened 2026-08-27
CLI-054: dotf deploy has no --all, so a declared config (orca-keybindings) is deployed by neither setup and destinations keep mixed separators on Windows


mlorentedev:
## Problem

`ai/deploy.json` declares two configs; both setups call `dotf deploy pi` only. `dotf deploy --dry-run` on the Windows work box (2026-08-27):

```
in sync   pi         C:\Users\<u>/.pi/agent/models.json
would deploy orca-keybindings C:\Users\<u>/.orca/keybindings.json
```

Two defects: `orca-keybindings` is declared and never installed (each new entry needs a call-site edit in two scripts), and `ExpandDst` (`cli/internal/deploy/deploy.go:126-142`) substitutes `{HOME}` without `filepath.FromSlash`/`Clean`, so the destination is a mixed-separator string. Harmless for the syscall, but it breaks string-equality comparison against `filepath.Join`ed paths (the class `samePath` in `checks_guard.go:137` exists for).

## Fix

- `dotf deploy --all` (or default to all when no name is given) so setups stop enumerating entries; both setups call it once.
- `ExpandDst` returns `filepath.Clean(filepath.FromSlash(out))`.

## Guard

Go tests: every `ai/deploy.json` entry is reachable from `--all`; `ExpandDst` output is OS-native (no `/` on Windows).

Related: CLI-039 (#1023), WIN-010 (adds the gitconfig entry).

Issue #1292, opened 2026-08-27
WIN-011: sensitive/ is mirrored non-recursively and doctor reads the mirror for the DR escrow, so Windows reports total-loss risk over an escrow that exists


mlorentedev:
## Problem

Measured 2026-08-27, Windows work box:

```
[Disaster recovery]
  [FAIL] no DR escrow at C:\Users\<u>\.dotfiles\sensitive\dr\bitwarden-export.age: 28 secret(s) resolve through Bitwarden and have no local copy ...
```

`<checkout>\sensitive\dr\bitwarden-export.age` exists: 140,870 bytes, mtime 2026-08-18, tracked in git. The FAIL asserts a single point of total loss that is not real.

## Root cause

- `setup-windows.ps1:613` copies `sensitive/` with `Get-ChildItem -Filter '*.secret.age'`: top level only, so `sensitive/dr/` is never mirrored. `setup-linux.sh:136` uses `cp -rf`, which is recursive.
- `cli/internal/doctor/checks_dr.go:128` stats the escrow under `cfg.DotfilesDir` (the mirror), while the writer `dotf secrets backup` writes to the **checkout** by contract (`cli/internal/secrets/escrow.go:148`, `env.RepoSensitiveDir`). The deployed copy is by construction never fresher than the checkout, so reading it is the wrong subject even on Linux; it only happened to agree there.

## Fix (decided)

- `checks_dr.go`: resolve the escrow checkout-first through the same resolver the writer uses, falling back to the mirror.
- `setup-windows.ps1`: mirror `sensitive/` recursively (the `dr/` escrow is an age blob like the rest).

## Guard

- Go test: escrow present in the checkout but absent from the mirror is not a FAIL.
- Post-setup doctor gate (TEST-003).

Related: #971, #997, #802 (mirror never prunes: the other arm of the same copy-only class).

mlorentedev:
Fixed across #1304 (doctor DR escrow reads checkout-first: cli/internal/doctor/checks_dr.go escrowPath) and #1306 (setup-windows.ps1 mirrors sensitive/ recursively). Closing with the changes that closed it.

Issue #1289, opened 2026-08-27
WIN-008: the copilot catalog injector writes CRLF and the drift comparator is CR-naive, so copilot-instructions.md drifts forever and the region strip swallows to EOF


mlorentedev:
## Problem

Measured 2026-08-27, Windows work box, immediately after a clean deploy:

```
[Harness + skill drift]
  [FAIL] stale: .copilot/copilot-instructions.md has drifted from ai/copilot/copilot-instructions.md (run: compile-harness.sh --deploy)
```

Binary measurement: deployed file 5708 bytes, **67 CR / 67 LF**; repo source 1902 bytes, **0 CR / 30 LF**. After stripping the harness regions the two differ only by the CRs. The printed remedy (`compile-harness.sh --deploy`) does not exist on Windows.

## Root cause: two independent defects

1. **Writer.** `setup-windows.ps1:2063` (`Deploy-SkillRecord`, catalog injection) rewrites the whole file with `Set-Content -Encoding UTF8`, which joins with the platform newline: the entire file becomes CRLF, not just the injected region. The same call shape at lines 1994 and 2003 writes every rendered `SKILL.md` / command file CRLF (latent until CLI-035 gives Windows a drift check). `.gitattributes` already declares `*.md text eol=lf`; the deployed copy violates the contract the repo declares.
2. **Comparator.** `cli/internal/doctor/checks_deploy.go:509-537` `stripHarnessRegions` splits on `\n` and closes a skipped region with `l == endMarker`. A `\r`-suffixed END marker never matches, so the strip swallows the rest of the file; and every other line keeps a trailing `\r`, so the compare fails regardless. Either defect alone guarantees a permanent FAIL.

## Fix (decided)

- Comparator: normalise `\r\n` to `\n` on both sides inside `stripHarnessRegions` (fixes the runaway skip too). Test: *a deployed copy differing only by CRLF is not drift*.
- Writer: one LF-preserving UTF-8 (no BOM) writer helper in `scripts/utils.ps1`, used by all three `Set-Content` sites. Pattern already in-repo: `scripts/orca-hook-tune.ps1:81`.

## Guard

- Go unit test on `stripHarnessRegions` (CRLF fixture).
- `tests/copilot-native-skills-smoke.ps1` exercises the injector but is invoked nowhere (Pester discovers `*.Tests.ps1` only): rename so it runs, and assert no `\r` in the four `deployedInstructionTargets`.

[...]

PR #1304, opened 2026-08-27
fix(cli): Windows deploy defects measured on a real box

### LINKED ISSUE #1293: WIN-012: the bw serve daemon stays attached to the console that unlocked it, so closing that terminal breaks pi and opencode in every terminal
## Problem

On the Windows work box `pi` and `opencode` "do not work" while on the Linux box they do. Both wrappers (`powershell/profile.ps1:253-254`, `.zshrc:92-93`) are `dotf secrets run --only NAN_API_KEY,... -- <agent>`; with the vault locked they exit 1 and the agent never launches:

```
Error: bw resolve nan-api-key/api-key: bitwarden vault is locked: no bw serve daemon is running: run `dotf secrets unlock` ...
```

`dotf secrets unlock` starts the daemon, but on Windows it dies with the terminal that ran it.

## Root cause

`cli/internal/secrets/bwserve_windows.go:14` spawns `bw serve` with only `CREATE_NEW_PROCESS_GROUP`. That flag re-routes Ctrl+C; it does **not** detach the child from the parent's console. A console-subsystem child created without `DETACHED_PROCESS` (or `CREATE_NO_WINDOW`) stays attached, and when that console closes Windows delivers `CTRL_CLOSE_EVENT` and terminates it. Stdio is already NUL, which is why `dotf` exiting does not kill it, but the terminal closing does. `bwserve_unix.go` uses `Setsid`, which is why Linux never shows this. The file's own comment records the Windows lifecycle as *"explicitly deferred to empirical validation"*; this is that validation.

Second gap: the daemon's stderr is discarded, so a `bw serve` that dies on start yields only *"did not become reachable within 15s"*.

## Fix (decided)

- Add `DETACHED_PROCESS` to the creation flags (keep `CREATE_NEW_PROCESS_GROUP`).
- Doctor: when an agent config is deployed and its wrapper's secrets are bw-backed but no unlocked daemon is reachable, **WARN** (not FAIL: every reboot would be red) with the literal remedy `dotf secrets unlock`. Today the `[OpenCode + pi]` section is green while the wrapper cannot launch anything: it asserts files and PATH, never the wrapper's precondition.

## Guard

`cli/internal/secrets/bwserve_windows_test.go` (`//go:build windows`): spawn a long-lived child with `bwServeDetachAttr()`, let the parent exit, assert the child is alive (re-exec-self pattern; `fileauthority_windows_test.go` is the idiom).

## Acceptance on the real box

`dotf secrets unlock` in terminal A, close A, new terminal, `pi --version` through the wrapper exits 0.

Related: #585 (bw serve plan B; this does not change its age-floor AC), #675.

mlorentedev:
## Summary

Five defects measured on 2026-08-27 by running `.\setup-windows.ps1` and `dotf doctor` on the Windows work box, each with a test that fails without its fix (mutation-checked). One `chore` commit first: the commit itself could not be made from Windows until it landed.

| Defect | Root cause | Fix |
|---|---|---|
| `pi`/`opencode` wrappers fail in every terminal after the unlocking one closes (WIN-012, #1293) | `bw serve` spawned with `CREATE_NEW_PROCESS_GROUP` only: that re-routes Ctrl+C, it does **not** detach from the console, so the terminal closing delivers `CTRL_CLOSE_EVENT` to the daemon. Linux uses `Setsid`. | `DETACHED_PROCESS` added to the creation flags. Asserted by effect on Windows: a child spawned with the attr owns no console, the control spawn does. |
| `[OpenCode + pi]` all green on a box where `pi` cannot start (WIN-012) | Every predicate is a file/PATH check; the wrappers' real precondition (keys resolvable through `dotf secrets run`) was never asserted. | WARN with the literal remedy (`dotf secrets unlock`) when keys are bw-backed and no unlocked daemon is reachable. PASS when unlocked; silent when nothing is bw-backed. WARN, not FAIL: every reboot locks the vault. |
| `.copilot/copilot-instructions.md` drift FAIL that no redeploy clears (WIN-008, #1289) | Deployed copy is CRLF (67 CR), source is LF. `stripHarnessRegions` split on `\n`: the END marker never matched and the strip swallowed the rest of the file; every other line differed by `\r` anyway. | Normalise `\r\n` before splitting. The writer (`Set-Content` in `setup-windows.ps1`) is fixed in the PowerShell PR. |
| `[Disaster recovery]` FAIL over an escrow that exists (WIN-011, #1292) | doctor read the deploy mirror; `dotf secrets backup` writes the checkout by contract, and `setup-windows.ps1` mirrors `sensitive/` non-recursively (no `dr/`). | Read the checkout first (same subject the writer uses), fall back to the mirror. The `-Recurse` is in the PowerShell PR. |
| `C:\Users\u/.pi/agent/models.json` (CLI-054, #1301) | `ExpandDst` substituted `{HOME}` and left the `/` tail. | `filepath.Clean(filepath.FromSlash(...))`. |

Also: `scripts/check-bats-names.sh` had `#!/bin/bash`; pre-commit on Windows resolves a script hook's shebang as a literal path and refused every commit with *Executable `/bin/bash` not found*. Fixed to `#!/usr/bin/env bash`, and `tests/precommit-hooks-portable.bats` now derives every `language: script` hook from `.pre-commit-config.yaml` and asserts the portable form (the precedent guard in `validate-commit-msg.bats` covered one hook by name).

Closes #1293. Part of #1289, #1292, #1301 (their PowerShell / call-site halves follow in separate PRs so each stays atomic).

- `go build ./... && go vet ./... && GOOS=linux go vet ./... && go test ./...` — all green.
- `golangci-lint run ./...` (pinned 2.12.2) — 0 issues.
- Mutation: removing the CRLF normalisation fails `TestStripHarnessRegions/CRLF_input…` and `TestCheckInstructionDrift_CRLFDeployedCopyIsNotDrift`; restoring passes both.
- `TestBWServeDetachAttr_ChildHasNoConsole` run from a real Win32 console: PASS (control probe `console=attached`, detached probe `console=none`). It self-skips with a stated reason where the test process owns no console (Git Bash pty, some runners).
- Guard mutation: reverting the shebang fails `precommit-hooks-portable.bats` naming the script.

## SDD skip rationale

Raw production diff is 104 lines; **executable** lines are ~31 across five independent fixes, each under the 20-line bug-fix exemption with an obvious, measured cause. The remainder is the explanatory comments this codebase keeps next to every non-obvious fix. No public contract, dependency or schema changes. Breakdown: `bwserve_windows.go` 3 executable (one flag, one const), `checks_deploy.go` 1 (normalise) + 16 (launchability report), `checks_dr.go` 10 (`escrowPath`), `deploy.go` 1.

## Acceptance on the real box (after merge + `setup-windows.ps1`)

`dotf secrets unlock` in terminal A → close A → new terminal → `pi --version` through the wrapper exits 0. That is the only proof the daemon fix landed; recorded on #1293 when done.

### DIFF
--- a/cli/internal/deploy/deploy.go
+++ b/cli/internal/deploy/deploy.go
@@ -139,7 +139,12 @@ func ExpandDst(dst, home string, resolve func(string) string) (string, error) {
 	if len(bad) > 0 {
 		return "", fmt.Errorf("destination %q: unresolvable path variable(s) %s", dst, strings.Join(bad, ", "))
 	}
-	return out, nil
+	// The manifest spells destinations with "/" on every OS; the resolved
+	// {HOME} is native. Normalise so the result is a path in the OS's own form
+	// rather than `C:\Users\u/.pi/agent/models.json` — accepted by the syscall,
+	// but never equal to the filepath.Join'ed path a check compares it against
+	// (CLI-054/#1301).
+	return filepath.Clean(filepath.FromSlash(out)), nil
 }
 
 // Renderer materialises {env:VAR} placeholders in a staged file. The seam exists
--- a/cli/internal/deploy/deploy_native_test.go
+++ b/cli/internal/deploy/deploy_native_test.go
@@ -1,0 +1,22 @@
+package deploy
+
+import (
+	"path/filepath"
+	"testing"
+)
+
+// The manifest spells destinations with "/" on every OS. The result must be in
+// the OS's own form: on Windows `{HOME}` resolves to `C:\Users\u`, and the
+// literal tail used to stay `/.pi/agent/models.json` — a path the syscall
+// accepts but that no filepath.Join'ed path ever string-equals, which is the
+// comparison every doctor check makes (CLI-054/#1301).
+func TestExpandDst_ReturnsAnOSNativePath(t *testing.T) {
+	home := filepath.Join(t.TempDir(), "home")
+	got, err := ExpandDst("{HOME}/.pi/agent/models.json", home, noResolve)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if want := filepath.Join(home, ".pi", "agent", "models.json"); got != want {
+		t.Errorf("want the OS-native path %q, got %q", want, got)
+	}
+}
--- a/cli/internal/deploy/deploy_test.go
+++ b/cli/internal/deploy/deploy_test.go
@@ -199,7 +199,7 @@ func TestExpandDst_UnresolvableTokenIsAnErrorNotAnEmptySegment(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if got != "/opt/pi/models.json" {
+	if got != filepath.FromSlash("/opt/pi/models.json") {
 		t.Errorf("want the resolved path, got %q", got)
 	}
 }
--- a/cli/internal/doctor/checks_deploy.go
+++ b/cli/internal/doctor/checks_deploy.go
@@ -328,6 +328,39 @@ func checkOpenCode(sys *System, cfg *Config, rep *Report) {
 			rep.Skip("pi models.json substitution — age identity absent ({env:} resolves at runtime)")
 		}
 	}
+
+	// Launchability. Everything above is a static predicate — files and PATH —
+	// and every one of them was green on a box where `pi` could not start
+	// (WIN-012/#1293). The shell wrappers (profile.ps1, .zshrc, .bashrc) run
+	// both agents under `dotf secrets run`, which resolves their keys BEFORE
+	// exec'ing the binary; while those keys live in a locked Bitwarden vault the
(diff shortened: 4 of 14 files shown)


mlorentedev:
## Review triage

- **PR-Agent (PR Reviewer Guide):** no findings — "No major issues detected", "No security concerns identified". Ticket compliance marked *partial* for #1289/#1292 because their PowerShell halves (LF writer, `-Recurse`) and the CI gate (TEST-003) ship in separate PRs, exactly as the description states; #1293's "further human verification" is the on-box acceptance step, recorded on the issue after merge + redeploy. Nothing to apply.
- **CodeRabbit:** *review limit reached* — a notice, not a review. Proceeding on PR-Agent's substantive output; disclosed here so the queue does not read this as reviewed twice.
- **CI:** 13/13 green (`test-windows` included; `goreleaser release` skipped by design on PRs).

Issue #1298, opened 2026-08-27
TEST-003: test-windows installs dotf off PATH, so doctor, deploy, render and tools install all skip and the job is green over a red box


mlorentedev:
## Problem

Measured from the real `test-windows` log of run 33051778423 (PR `fix/shell-wrapper-path-ordering`, all jobs green):

```
dotf 0.51.0 installed to C:\Users\runneradmin\.local\bin\dotf.exe
... dotf secrets render unavailable/failed
... dotf not on PATH -- skipping pi config deploy
[WARNING] dotf not found on PATH, skipping post-setup diagnostics
Setup Complete!
```

`~/.local/bin` is never added to `$env:GITHUB_PATH`, so every dotf-gated block skips. `dotf doctor` has **never run** in CI. Four defects measured on a real box the same day (WIN-007, WIN-008, WIN-011, CLI-053's path) were invisible by construction.

Three masking layers, all needed:
1. dotf off PATH (above).
2. Doctor's exit code is swallowed by design (`setup-windows.ps1:2130-2141`, *non-fatal*) and `tests/setup-windows.bats` asserts the swallow (`ok 130 ... non-fatal`). Correct for setup (degrade-not-break, C9), but nothing else enforces.
3. `ci.yml:280` sets `DOTFILES_DIR: ${{ github.workspace }}`, so even a running doctor would read `harness/` from the checkout and pass, while a real box (default `~/.dotfiles`) fails.

Also found in the same log: the hive block errors under StrictMode (`The term 'uv' is not recognized`, `$hiveVerMatch cannot be retrieved`) with zero `##[error]` annotations (non-terminating errors in a `powershell` step do not fail the job), and setup reads `uv tool list` one second **before** installing uv. `tests/copilot-native-skills-smoke.ps1` is invoked nowhere (Pester discovers `*.Tests.ps1` only).

## Fix (decided)

- Add `~/.local/bin` to `$env:GITHUB_PATH` after the dotf install.
- New step *post-setup doctor gate*: run `dotf doctor` with real-box env (`DOTFILES_DIR` unset) and fail on non-zero. Keep setup itself non-fatal.
- Drop the `DOTFILES_DIR` override on the setup step if `DOTFILES_REPO_DIR` alone suffices.
- Rename the copilot smoke test to `*.Tests.ps1`, skipping when `copilot` is absent.
- Land after WIN-007/008/011: the gate is born red before them.

## Guard

[...]

Issue #1288, opened 2026-08-27
WIN-007: setup-windows.ps1 never mirrors harness/ into ~/.dotfiles, so doctor fails on model-map and model-pins and its printed remedy can never clear


mlorentedev:
## Problem

Measured 2026-08-27 on the Windows work box after a clean `.\setup-windows.ps1`:

```
[Routing registry]
  [FAIL] harness/model-map.json not found at C:\Users\<u>\.dotfiles\harness\model-map.json
[Model pin drift]
  [FAIL] read harness/model-pins.json: open C:\Users\<u>\.dotfiles\harness\model-pins.json: The system cannot find the path specified.
```

Both messages end with *"re-run setup to mirror it"*. Re-running setup cannot clear them: `~/.dotfiles` on Windows holds exactly `env-contract.json, git-hooks, packages.json, paths.ps1, scripts, secrets, sensitive, versions.conf` and no `harness/`.

## Root cause

- `setup-linux.sh:552-610` mirrors `harness/` plus every `harness/manifest.json` target into `$DOTFILES_DIR` (bash + jq). `setup-windows.ps1` has no counterpart.
- The two scripts invert the variable names: Windows `$DotfilesDir` (line 75) is the **checkout**, Linux `DOTFILES_DIR` is the **deploy dir**; the Windows deploy dir is `$DotfilesDest` (line 76). `dotf doctor` resolves `cfg.DotfilesDir` to `~/.dotfiles` on both (`cli/internal/doctor/config.go:34`).
- `cli/internal/doctor/checks_deploy.go:548-554` records the opposite premise as fact (*"Windows has no repo/mirror split"*) and early-returns `checkHarnessMirrorOrphans` on Windows because of it. `checks_pi_extensions.go:86` reads `ai/pi/packages.json` from the same missing mirror and reports *"0 packages declared"* as a PASS.
- Same pathology as #1200 (closed, Linux): the mirror evaluates a manifest the setup never copied. #1200's own guard clause was never implemented for Windows.

## Fix (decided)

[...]

Issue #1202, opened 2026-08-23
BUG: setup installs jq to ~/.local/bin then gates later steps on command -v jq, so MCP registration is skipped on the machine that just installed it


mlorentedev:
`setup-linux.sh` installs `jq` into `$HOME/.local/bin`, then keeps testing for it with `command -v jq`. That directory is added to the shell rc files, not to the PATH of the process currently running setup, so every later `command -v jq` in the same run can be false immediately after a successful install.

## Measured

From one integration-container run (PR #1201, job 97160873172), 22 seconds apart, same process:

```
07:32:22  [INFO]    Installing jq...
07:32:22  [SUCCESS] jq installed
07:32:44  [WARNING] Claude Code CLI, npx, or jq not found, skipping MCP server registration
```

Both lines are accurate. The binary was downloaded and made executable at `$HOME/.local/bin/jq`; it is simply not invocable as `jq` from that shell. The success message reports the download, and the later check reports the lookup.

## Consequence

Any setup step gated on `command -v jq` is skipped on a machine where setup itself just installed jq. Known call sites:

- **MCP server registration** — the warning above. Registration is silently skipped on a fresh machine, which is exactly the machine that needs it.
- **The harness target mirror** (added in #1201) — worked around there by resolving `$HOME/.local/bin/jq` explicitly, which is a per-call-site patch rather than the fix.

The `mcp-servers.json` block at line ~414 has the same `command -v jq` shape.

## Root cause, and the second defect underneath

The installer reports success on a condition it did not verify:

```sh
curl -Lo "$HOME/.local/bin/jq" "https://.../jq-linux-amd64" 2>/dev/null \
    && chmod +x "$HOME/.local/bin/jq" \
    && log_success "jq installed" \
    || log_warning "jq installation failed"
```

`log_success "jq installed"` fires when the download and chmod succeed — never when `jq` became *usable*. This is the `A && B || C` shape shellcheck already flags on this line (SC2015). "Installed" should mean the same thing the next check asks about, or the two will disagree exactly as they do here.

## Fix candidates

Independent, and the first alone would close the class:

- **Put `$HOME/.local/bin` on `PATH` inside setup, early**, before any tool install. Every later `command -v` then answers the question the operator thinks it is answering. This is the smallest change with the widest effect, and it matches what the rc files already do for interactive shells.
[...]

Issue #909, opened 2026-08-10
CLI-035: port the harness engine (compile-harness.sh + Deploy-SkillRecord) to `dotf harness`, delete both twins


mlorentedev:
## Problem

The harness engine is a split `.sh`/`.ps1` twin, and the split is load-bearing for a real gap surfaced by #804 (BUG-052):

- **Linux — `scripts/compile-harness.sh`** (839 LOC): three modes. `--refresh` regenerates the `<!-- BEGIN HARNESS GENERATED -->` blocks from the vault into the repo's source files (the SSOT generation step); `--check` detects block↔source drift; `--deploy` renders committed skill records to per-agent paths.
- **Windows — `Deploy-SkillRecord`** (setup-windows.ps1:1822-1943, ~180 LOC): a port of **only** `--deploy`. It consumes the already-committed records under `harness/skills/` + `harness/manifest.json`. `--refresh`/`--check` are Linux-only by design (comment at setup-windows.ps1:1763-1765).

Consequence: on Windows there is **no** `--check` equivalent, so `dotf doctor`'s `checkHarnessDrift` cannot run — it looks for `compile-harness.sh` in the deploy mirror, never finds it (the Linux script is never copied), and SKIPs. Harness drift is simply **unobservable** on Windows today. #804 only makes that SKIP honest; it does not close the gap.

## Why now

Two triggers align:
1. **ADR-020 strangler-fig**: "when a `.sh`/`.ps1` twin is next touched, port it to `dotf` in that same PR and delete the pair + its bats/Pester tests — never leave the three coexisting." #804 touches the doctor check that *invokes* this twin.
2. A Go port is strictly better for the parity this repo already wants: one engine, cross-compiled, so `--check` (and drift detection in `dotf doctor`) works identically on Linux, Windows, and CI.

## Decision

Port the engine to a `dotf harness` subcommand group (`refresh` / `check` / `deploy`), delete `scripts/compile-harness.sh`, `tests/compile-harness.bats`, the `Deploy-SkillRecord` block + its helpers (`Test-SkillTargetsAgent`, `Get-SkillField`, `Convert-SkillRecord`) in `setup-windows.ps1`, and any Pester coverage. `dotf doctor`'s `checkHarnessDrift` then calls `dotf harness check` on every platform (no more Linux-only SKIP).

## Scope / sequencing

[...]

PR #1305, opened 2026-08-27
feat(harness): dotf harness mirror, one deploy-dir mirror for both OSes

mlorentedev:
## Summary

`dotf harness mirror`: one Go implementation of the deploy-dir mirror, called by both setup scripts. Spec: `specs/WIN-007-harness-mirror/`. Part of #1288 (the issue closes with the spec archive, after the pooled adversarial review).

**Why.** Measured 2026-08-27 on the Windows work box: `~/.dotfiles/harness/` never existed, so `dotf doctor` failed `harness/model-map.json` and `harness/model-pins.json` after every setup and printed *re-run setup to mirror it* — a remedy that could not clear them, because `setup-windows.ps1` had no mirror block. Linux carried a 60-line bash+jq block (zsh word-split hazard, "jq not on PATH yet" race, both documented in-place). The Go doctor's orphan check early-returned on Windows on the stated belief that Windows has no repo/mirror split — false: `$DotfilesDest` is `~/.dotfiles`, the very dir `cfg.DotfilesDir` resolves to; it just never received `harness/`.

**What.**
- `cli/internal/harness/mirror.go` — copies `harness/` + every manifest `.targets[].file` from the checkout into the deploy dir; byte-compare before write (re-run: `0 updated`, mtimes untouched); never prunes (`dotf doctor --fix` owns orphans, #802); a declared target the checkout lacks is named on stderr and the command exits 1 **after** mirroring the rest.
- `setup-linux.sh`: the bash+jq block is replaced by the call at the same position (after `--refresh`). `setup-windows.ps1`: the call is added beside `dotf tools install`.
- doctor: `checkHarnessMirrorOrphans` applies on every OS (stale comment removed); `piPackagesManifest` reads the checkout first (it reported "0 packages declared" as a PASS on Windows).
- Guards: Go unit + cmd tests (idempotence, missing target, no-prune, same-dir, no-manifest); `setup-linux.bats` guards rewritten to the new call, the jq-by-path guard (#1202) deleted with the jq; `setup-windows.bats` call-site + parity guards.

- All five `features.json` verifications exit 0; `go build/vet/test ./...` green on Windows and `GOOS=linux go vet`; `golangci-lint` (2.12.2) 0 issues; PowerShell AST parse 0 errors, insert ASCII-only; bats guards pass.
- `dotf` built from this branch, run on the real box: `harness/ + 3 target(s) → C:\Users\<u>\.dotfiles (76 updated, 0 unchanged)`, then `(0 updated, 76 unchanged)`. `dotf doctor`: the two registry FAILs are gone (122/4/2/25 → 124/3/2/26). The model-pin check, which had never been able to run on this box, now reports a genuine drift in the live pi `settings.json` (`deepseek-v4-flash-0731`, the "dead dated id" `model-pins.json` anticipates) — a finding, not a regression; the two remaining FAILs are fixed in #1304.

## Not in this PR

`dotf harness {refresh,deploy,check}` (CLI-026 #495 / CLI-035 #909). The Windows CI job does not run `dotf` yet (TEST-003, #1298), so the Windows call site is exercised end-to-end there only after that lands; the behaviour is covered on the `cli` workflow's Windows matrix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -317,16 +317,19 @@ jobs:
           source versions.conf
           echo "BATS_VERSION=$BATS_VERSION" >> "$GITHUB_OUTPUT"
           echo "AGE_VERSION=$AGE_VERSION" >> "$GITHUB_OUTPUT"
+          echo "GO_VERSION=$(sed -n 's/^go //p' cli/go.mod)" >> "$GITHUB_OUTPUT"
 
       - name: Build integration test container
         if: github.event_name == 'push' || needs.changes.outputs.code == 'true'
         env:
           BATS_VERSION: ${{ steps.versions.outputs.BATS_VERSION }}
           AGE_VERSION: ${{ steps.versions.outputs.AGE_VERSION }}
+          GO_VERSION: ${{ steps.versions.outputs.GO_VERSION }}
         run: |
           docker build \
             --build-arg BATS_VERSION="$BATS_VERSION" \
             --build-arg AGE_VERSION="$AGE_VERSION" \
+            --build-arg GO_VERSION="$GO_VERSION" \
             -f tests/Dockerfile.integration \
             -t dotfiles-integration-test .
 
--- a/cli/internal/cmd/harness.go
+++ b/cli/internal/cmd/harness.go
@@ -27,6 +27,7 @@ pattern triggers, and workflow integrations.`,
 	cmd.AddCommand(newHarnessResolveTierCmd())
 	cmd.AddCommand(newHarnessResolveCapabilitiesCmd())
 	cmd.AddCommand(newHarnessGateCmd())
+	cmd.AddCommand(newHarnessMirrorCmd())
 	return cmd
 }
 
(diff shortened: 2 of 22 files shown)


mlorentedev:
## Review triage

- **PR-Agent (PR Reviewer Guide):** #1288 *partially compliant* — the three "non-compliant" rows are the spec archive (gated on the pooled adversarial review, run after merge from a box with an unlocked vault), the real-box doctor run (done, quoted in the description: the two registry FAILs are gone) and the Windows CI end-to-end (TEST-003, #1308). The #802 row is an auto-match on the description citing its prune semantic; #802 is not a requirement here. No code findings; nothing to apply.
- **CodeRabbit:** *review limit reached* on every PR today — a notice, not a review; disclosed.
- **CI:** integration went red three times on the way here and each was a real finding fixed in-branch: `dotf` off PATH in the container (by-path resolution, #1202 class), the container testing the *released* `dotf` instead of the PR (Go build in the image), and `versions.conf` clobbering `DOTF_VERSION=dev` (installers now leave a dev build in place). Green at the time of this triage.

PR #1307, opened 2026-08-27
fix(copilot): config keys the CLI reads, Co-authored-by off, dead exports removed

mlorentedev:
## Summary

Copilot CLI 1.0.80 audit on the Windows work box (AI-036, #1296): what the repo deploys for Copilot vs what the CLI reads.

| Deployed | Reality | Change |
|---|---|---|
| `includeCoAuthoredBy` unset | CLI default is **true**: every commit Copilot makes carries a `Co-authored-by` trailer, against the no-attribution doctrine in `AGENTS.md` | `false` |
| `"defaultModel": "gpt-5.6"` | the key is `model` (`copilot help config`); `gpt-5.6` is not a seat id — a double no-op | `"model": "gpt-5.6-sol"` |
| `$schema` → `raw.githubusercontent.com/github/copilot-cli/main/schemas/config.json` | 404 | removed |
| `export COPILOT_MODEL=gpt-5.6` in `.zshrc`/`.bashrc` | only inside the `paths.sh`-missing fallback branch (never set on a converged box), never on Windows, invalid value | deleted (deletion over correction) |
| `cop`/`cops` in `aliases.zsh` + `profile.ps1` | `.bashrc` had neither; the parity test did not cover bash | added, test widened |
| Model Tier line names `claude-3.7-sonnet`, `gpt-5`, `o3-mini` | none on the seat | seat ids, pointer to `harness/model-map.json` as the routing SSOT (#902). Applied to both copies per the `docs-drift` parity rule |

Verified working on the box (not changed): `copilot skill list` sees all 38 deployed skills; `copilot mcp list` loads the four servers from the deployed `mcp-config.json`; `trustedFolders` honoured.

Closes #1296. Out of scope: harness-emitted Copilot hooks (#803) — a feature with its own spec; the CLI supports `~/.copilot/hooks/*.json` carrying `bash` and `powershell` keys in one file, i.e. an OS-agnostic artifact.

## Evidence

`bats tests/copilot-config.bats tests/aliases.bats tests/env-contract.bats tests/docs-drift.bats` — green (the one failing case on this box is the pre-existing `zsh: command not found`). `bash -n .bashrc` clean.

PR #1308, opened 2026-08-27
ci(windows): build dotf from the PR, put it on PATH, and gate on dotf doctor after setup

### LINKED ISSUE #1298: TEST-003: test-windows installs dotf off PATH, so doctor, deploy, render and tools install all skip and the job is green over a red box

mlorentedev:
## Summary

TEST-003 (#1298): `test-windows` installed `dotf` into `~/.local/bin` and never put it on `PATH`, so every `dotf`-gated block in `setup-windows.ps1` skipped with a warning, `dotf doctor` **never ran in CI**, and the job stayed green while a real Windows box failed four doctor checks after the same setup (measured 2026-08-27; log of run 33051778423 quoted in the issue).

Three masking layers, all removed here:

1. **PATH** — `actions/setup-go` + a step that builds `dotf` from *this PR* into `~/.local/bin`, appends it to `GITHUB_PATH` and proves reachability by effect (`dotf version`). The released binary lags the tree, which is also why the Linux integration container has to build from source.
2. **Install-Dotf would replace it** — `DOTF_VERSION=dev` on the setup step: Install-Dotf skips when the binary on PATH reports the pinned version, and a source build reports `dev`.
3. **`DOTFILES_DIR: ${{ github.workspace }}`** — dropped; setup now deploys to the real-box default, so the gate certifies the tree a real machine reads.

And the gate: a post-setup `dotf doctor` step that fails the job on a non-zero exit. Setup itself stays non-fatal (degrade-not-break, C9; `tests/setup-windows.bats` still asserts that).

`tests/ci-windows-doctor-gate.bats` pins all of the above.

## What the gate found on the runner (seven iterations, each a real defect)

1. `~/.local/bin` never on `PATH` -> build `dotf` from the PR, `GITHUB_PATH` + User PATH.
2. Install-Dotf replaced the source build with the release -> installers leave a `dev` build in place (#1305).
3. `setup-windows.ps1` rebuilt `PATH` from the registry twice, dropping process-only entries (the PR's `dotf`, the runner's node) -> both rebuilds keep the process PATH.
4. The uv and Bun installers ran in-session and left setup without `npm` ("skipping pi install") -> child `pwsh`; setup now logs the PATH state after them.
5. `copilot-instructions.md` "drift" was one trailing blank line -> the comparator trims trailing newlines, as it already normalises CRLF.
6. A known-failure regex was valid but wrong (`\s` read as whitespace) -> every entry carries an `# example:` line its pattern must match (Pester).
7. `Test-DoctorGate` rejected the blank lines real output carries -> `AllowEmptyString`.

Final state: `doctor gate: 2 known runner-only FAIL(s), 0 unexpected, 0 stale` -- SCRIPTS_DIR (WIN-013, #1310) and the runner's missing Bitwarden identity (TEST-005, #1313). The doctor adjustments (wget POSIX-only, `.exe` probe for `*_HOME`, a `dev` build is not drift) came from the first run. Closes #1298.

## SDD skip rationale

CI workflow + a 40-line PowerShell gate script + three one-branch doctor adjustments (a `posixOnlyTools` entry, an `.exe` fallback, a `dev` SKIP). No public contract, dependency or schema change; the executable Go delta is under 15 lines and each doctor adjustment is a bug fix with a measured cause on the first gate run. Declared here because the raw line count crosses the threshold on YAML and comments.

### DIFF
--- a/.github/scripts/doctor-gate-known-failures.txt
+++ b/.github/scripts/doctor-gate-known-failures.txt
@@ -1,16 +1,21 @@
 # TEST-003 (#1298): runner-only doctor failures the test-windows gate accepts.
 # One regex (.NET, PowerShell -match) per entry, each preceded by a comment that
-# names the owning ticket. The gate fails on any [FAIL] not listed here AND on
-# any entry that no longer matches, so this list only shrinks as tickets close.
+# names the owning ticket and by an "# example:" line quoting the [FAIL] line it
+# was written for -- tests/doctor-gate.Tests.ps1 asserts the pattern matches its
+# example, so a regex that is valid but wrong (a "\s" meant as a backslash) cannot
+# sit here silently. The gate fails on any [FAIL] not listed here AND on any
+# entry that no longer matches, so this list only shrinks as tickets close.
 # A real-box defect never goes here: it is fixed or ticketed, and the gate stays
 # red until then.
 
 # WIN-013 (#1310): env-contract SCRIPTS_DIR defaults to ~/.dotfiles/scripts on
 # Windows, but setup-windows.ps1 deploys scripts to ~/scripts and nothing
 # creates the contract path, so the contract check fails on every fresh box.
-SCRIPTS_DIR=.*\.dotfiles\scripts \(path does not exist\)
+# example:   [FAIL] SCRIPTS_DIR=C:\Users\runneradmin\.dotfiles\scripts (path does not exist)
+SCRIPTS_DIR=.*\.dotfiles\\scripts \(path does not exist\)
 
 # TEST-005 (#1313): the runner has no Bitwarden identity, so the reach tier
 # (exposure-keyed: 28 registry entries resolve through Bitwarden) fails by its
 # own rule. Real boxes log in once; CI cannot and must not resolve real secrets.
+# example:   [FAIL] Bitwarden session is gone (`bw status`: unauthenticated) - run `bw login`, not `bw unlock`
 Bitwarden session is gone \(`bw status`: unauthenticated\)
--- a/.github/scripts/doctor-gate-known-failures.txt
+++ b/.github/scripts/doctor-gate-known-failures.txt
@@ -9,3 +9,8 @@
 # Windows, but setup-windows.ps1 deploys scripts to ~/scripts and nothing
 # creates the contract path, so the contract check fails on every fresh box.
 SCRIPTS_DIR=.*\.dotfiles\scripts \(path does not exist\)
+
+# TEST-005 (#1313): the runner has no Bitwarden identity, so the reach tier
+# (exposure-keyed: 28 registry entries resolve through Bitwarden) fails by its
+# own rule. Real boxes log in once; CI cannot and must not resolve real secrets.
+Bitwarden session is gone \(`bw status`: unauthenticated\)
--- a/.github/scripts/doctor-gate-known-failures.txt
+++ b/.github/scripts/doctor-gate-known-failures.txt
@@ -1,0 +1,11 @@
+# TEST-003 (#1298): runner-only doctor failures the test-windows gate accepts.
+# One regex (.NET, PowerShell -match) per entry, each preceded by a comment that
+# names the owning ticket. The gate fails on any [FAIL] not listed here AND on
+# any entry that no longer matches, so this list only shrinks as tickets close.
+# A real-box defect never goes here: it is fixed or ticketed, and the gate stays
+# red until then.
+
+# WIN-013 (#1310): env-contract SCRIPTS_DIR defaults to ~/.dotfiles/scripts on
+# Windows, but setup-windows.ps1 deploys scripts to ~/scripts and nothing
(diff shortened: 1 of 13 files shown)


mlorentedev:
## Review triage

Reviewer output up to `cf12faa` (PR-Agent guide + CodeRabbit, 4 inline). Fixes land in `3094422`.

|---|---|---|---|
| 1 | PR-Agent | PATH refresh appends `$env:PATH` on every call, grows linearly | **Applied.** Measured on the runner: 153 entries, 72 duplicates. `Sync-SessionPath` (utils.ps1) merges Machine + User + process, first occurrence wins, registry first, idempotent; both call sites use it; Pester `tests/windows-path-sync.Tests.ps1`; the bats guard forbids a raw registry rebuild in setup. |
| 2 | CodeRabbit | Same: keep the PATH refresh idempotent (`changed=0` on re-run) | **Applied** as #1. |
| 3 | CodeRabbit | `checks_tools.go:72` reads `runtime.GOOS`, not the injected `sys.GOOS` | **Applied.** `TestCheckVersionedPaths_ExeFollowsInjectedGOOS` runs the same `bin/java.exe` fixture as windows (PASS) and linux (FAIL); mutation-checked, reverting to the host OS fails the linux row. |
| 4 | CodeRabbit | `checkOptionalTools` is 60 lines (limit 40) | **Applied.** `checkDotfVersion` extracted; `checkOptionalTools` is 36 lines. |
| 5 | CodeRabbit | dotf-version test is not table-driven and asserts prose | **Applied.** `TestCheckOptionalTools_DotfVersion`: exact pin / stale release / dev, one row per branch, asserting the report `Status` through `statusTag`. |
| 6 | PR-Agent | `.exe` precedence when both `java` and `java.exe` exist | **Declined.** Either file proves the `*_HOME` points at a real install, which is all the check asserts; ordering between two present binaries is not a doctor concern. |
| 7 | PR-Agent | Regex injection / ReDoS from the known-failures list | **Declined.** The list is committed and reviewed, not runtime input, and Pester guards it: every entry must be a valid regex and match its own `# example:` line. |
| 8 | PR-Agent | Ticket compliance: copilot test rename, #1305 mirror, #1310 unfixed | **No change.** The rename shipped in #1307 (`tests/copilot-native-skills.Tests.ps1`); #1305 is the merged mirror PR, not an issue; #1310 is allow-listed on purpose with its owning ticket (WIN-013) and the gate fails the moment it goes stale. |

Local evidence for `3094422`: `go build/vet/test` + `GOOS=windows go vet` green, `golangci-lint` 0 issues, Pester 15/15, `tests/setup-windows.bats` 116/116. CI on `3094422` pending at the time of writing; merge after green.

Issue #1294, opened 2026-08-27
AI-034: opencode is installed through three channels and its version parse accepted 'locked.', so setup never converges the binary on PATH


mlorentedev:
## Problem

Measured 2026-08-27, Windows work box:

```
[INFO] OpenCode locked. below pinned minimum 1.16.2, upgrading...
[WARNING] OpenCode still locked. after winget install 1.16.2: another install shadows it in PATH (e.g. npm global, scoop); converge that install instead
```

Three opencode installs exist across the two boxes: npm-global `opencode-ai@1.16.2` (first on PATH here), winget `SST.opencode` (what setup-windows pins), and the curl install script on Linux (`~/.opencode/bin`). Setup prints *"converge that install instead"* (a manual instruction) and moves on.

## Root cause

- `setup-windows.ps1:424,432` parses `--version` as *first line, last whitespace token*. Any banner line first yields garbage (`locked.`). `setup-linux.sh:754,762,876` has the same token-split shape (opencode **and** pi); the Windows pi block (line 1201) is the only robust site (`(\d+\.\d+\.\d+)` regex). Seven fragile parse sites across both scripts.
- No install-channel policy exists anywhere (vault, ADRs, docs). `specs/archive/AI-014-opencode-windows-bootstrap` considered npm as a fallback and never decided.

## Decision (2026-08-27)

**npm is the channel for node-distributed agents on every OS** (pi, opencode, bw already are). opencode moves to `packages.json` as a `source.type: npm` entry; `dotf tools install` already implements pin-as-floor plus a semver-regex probe (`cli/internal/tools/install.go`). The winget entry (Windows) and the curl block (Linux) are deleted. Recorded as an ADR in `docs/adr/`.

## Fix

- `packages.json` gains opencode; both setups already call `dotf tools install`.
- `dotf tools version <name>` (Go): one semver extraction both setups use for tools outside the catalog (hive, #791), deleting the seven parse sites.
- Doctor: **shadowed install** check: the binary PATH resolves for a catalog tool is not the managed one (npm prefix / `~/.local/bin`) is a WARN naming both paths. On the Linux box this will flag the orphaned curl copy after the next setup.

## Guard

- Go table test: `"OpenCode locked.\n1.16.2"`-shaped banners parse to `1.16.2`.
- `tests/setup-windows.bats` / linux equivalent: no `winget install SST.opencode` and no `opencode.ai/install` remain.

[...]

Issue #791, opened 2026-08-07
AI-028: hive bootstrap, daemon activation and auto-upgrade all silently no-op when the uv-tool install is gone


mlorentedev:
Successor to `AI-022-hive-daemon-activation` and archived `AI-023-hive-auto-upgrade-timer`. Cross-repo counterpart of the last unchecked box in [hive#292](https://github.com/mlorentedev/hive/issues/292); hard prerequisite for [hive#176](https://github.com/mlorentedev/hive/issues/176).

**Not a duplicate of #574 (HARNESS-049).** #574 is the *safety net* — `dotf doctor --fix` detecting and repairing a broken install. This ticket removes the *structural cause* so there is less to detect. Both should ship; neither subsumes the other.

**Sibling: AI-029** — MCP registration SSOT across all agent clients. Split deliberately per atomic-PR discipline (the same reason #574 was split out of #551).

## Symptom (reproduced on real Windows hardware, 2026-08-07)

The `hive` MCP server is **dead in Claude Code on this machine**, and every health signal reads green.

`~/.claude.json` registers `hive` as `C:\Users\mlorente\.local\bin\hive.exe`. That binary is an orphaned uv trampoline:

```
$ hive --version
error: uv trampoline failed to canonicalize script path
```

It is orphaned because `hive-vault` is **no longer installed as a uv tool**:

```
$ uv tool list
aider-chat v0.86.2
pre-commit v4.6.0
```

**This exact failure is already documented** in `docs/troubleshooting/hive-mcp-orphaned-trampoline.md` (2026-06-29) with a manual repair recipe. It has now recurred. A documented manual recipe is not a fix — the machine sat broken and nothing said so.

Hive appears to still work only because *other* clients launch `uvx hive-vault`, which resolves from the uv cache (3 live processes on `hive_vault-1.43.0`). That path never self-updates and never touches the daemon. See AI-029.

## Root cause: one shared assumption, four silent no-ops

Four independent mechanisms all treat **`uv tool list` as the source of truth for "is hive installed"**. When the uv-tool install disappeared, all four degraded to a silent no-op — and all four were deliberately designed to degrade quietly ("non-fatal", "warning", "exit 0"), so nothing surfaced:

| # | Site | What happens now |
|---|---|---|
| 1 | `mcp-servers.json` `prerequisite_command` = `uv tool install --upgrade hive-vault` | The bootstrap meant to create the install. Whatever removed it, this never restored it. |
[...]

Issue #426, opened 2026-06-18
AI-027: provision Orca (parallel-agent ADE) across all machines + doctor verification


mlorentedev:
## What is Orca

[stablyai/orca](https://github.com/stablyai/orca) — an **ADE (Agent Development Environment)**: a desktop (Electron) + mobile IDE for running a *fleet of parallel coding agents* with your own subscription. MIT, YC-backed, ~5.2k★, current `v1.4.77`. Its integration surface is exactly this user's stack — topics include `claude-code`, `codex`, `opencode`, `pi`, `ghostty`, `worktrees`, `orchestration`, `terminal`. It also ships an `orca` CLI (the basis of the `orca-cli` / `orchestration` agent skills).

## Goal

Make Orca part of the standard software suite provisioned on every new machine, pinned and verified like the rest of the toolchain — "install once via setup, confirm via doctor, present on all machines."

## Scope / where it fits

- **`versions.conf`** — add `ORCA_VERSION` pin (SSOT), starting at `1.4.77`.
- **`setup-linux.sh`** — idempotent install (guarded by `command -v orca` / install marker). Linux assets: `orca-ide_<ver>_amd64.deb`, `orca-ide-<ver>.x86_64.rpm`, `orca-linux.AppImage`. Prefer the `.deb` via `dpkg`/`apt` (desktop entry + auto-update wiring) with the AppImage→`~/Applications` as the no-root fallback, consistent with how the repo installs desktop tooling.
- **`setup-windows.ps1`** — install via `winget` if available, else download `orca-windows-setup.exe` and run the NSIS installer silently. Mirror the Linux idempotence guard.
- **`dotf doctor` (Go) + healthcheck parity** — add a presence/version check (optional tool) so "is Orca on this machine, at the pinned version?" is part of the standard diagnostic sweep. This is the "demás verificaciones para tenerlo en todos los equipos" part.
- **Vault runbook (optional)** — a `guide-orca-setup` alongside `guide-opencode-go-setup` / `guide-ghostty-setup` once the install path is empirically validated.

## Acceptance criteria

- [ ] `ORCA_VERSION` in `versions.conf`; `versions-conf.bats` green.
- [ ] `setup-linux.sh` installs Orca idempotently (re-run is a no-op); `orca` resolves afterward.
- [ ] `setup-windows.ps1` installs Orca idempotently (winget or silent `.exe`).
[...]

PR #1311, opened 2026-08-27
feat(tools): one install channel per tool class, opencode through packages.json, dotf tools version

mlorentedev:
## Summary

**ADR-036** — one install channel per tool class, on every OS; `packages.json` is the pin SSOT for catalog tools; version detection happens once, in Go; a second copy on PATH is a doctor finding. Spec: `specs/AI-034-opencode-npm-channel/`. Part of #1294 (closes with the spec archive after the pooled adversarial review).

**Why.** Measured 2026-08-27 on the Windows work box: `opencode` reached the two machines through three channels (npm-global first on PATH, winget `SST.opencode`, the curl script on Linux). Setup parsed the version as *last token of the first line*, accepted `locked.` as a version, "upgraded" the winget copy and printed *converge that install instead* — a manual instruction — on every run. Seven such parse sites existed across both setup scripts; no document said which channel a tool should come from. The right primitive already existed (`packages.json` + `dotf tools install`, pin-as-floor, semver probe) and served two tools.

**What.**
- `packages.json` gains `opencode` (npm); `OPENCODE_VERSION` leaves `versions.conf`. Both setups already call `dotf tools install`.
- `setup-linux.sh`: curl block + `~/.opencode/bin` variables gone; post-deploy assertion on PATH only. `setup-windows.ps1`: winget entry + pin-parse block + the loop's version-pin machinery gone (no winget tool carries a pin any more). `.zshrc`/`.bashrc`: no `~/.opencode/bin` PATH line.
- `dotf tools version <name>` — first semver in `<name> --version`, exit 1 when none. Both setups gate `hive service` on it instead of `uv tool list`, which stopped seeing a healthy install when hive moved to its own installer (#791): the work box reported `hive <unknown> predates 'hive service'` for a 3.0.0.
- `dotf doctor`: opencode's pin from the catalog (`matchPinFrom` names the source); **WARN when an npm catalog tool resolves from more than one PATH directory**, naming them.
- Guards: `TestProbeVersion_*` over the real banners (`OpenCode locked.`, `hive-vault 3.0.0`), `TestToolsVersionCmd`, `TestCatalogPin_*`, `TestCheckShadowedCatalogTools_*`; bats over both setups, the rc files, `versions.conf` and the hive gate parity.

## Evidence (this session, Windows work box, `dotf` built from this branch)

```
$ dotf tools version hive        → 3.0.0
$ dotf tools version opencode    → 1.16.2
$ dotf tools version nope        → exit 1, "no version found on PATH"
$ dotf tools install opencode    → opencode 1.16.2 already installed; skipping
$ dotf doctor                    → [WARN] opencode resolves from 3 PATH directories (…\nodejs-lts\current\bin, …\scoop\shims, …\WinGet\Packages\SST.opencode_…) — `dotf tools install` converges the npm copy only; remove the other channel's copy
```

All five `features.json` verifications exit 0; `go build/vet` (Windows + `GOOS=linux`), `go test`, `golangci-lint` 0 issues, `shellcheck`, PowerShell parse + unchanged non-ASCII count; bats green (pre-existing `zsh: command not found` cases aside).

## Not in this PR

- `pi` stays out of the catalog: its Linux install uses `--prefix ~/.local --ignore-scripts` (#426), which `installNpm` cannot express yet — follow-up on #1294.
- Leftover channel copies are not deleted: doctor names them, the operator removes them (ADR-036 §5). On the Linux box the `~/.opencode/bin` copy will be reported after the next setup.

### DIFF
--- a/.bashrc
+++ b/.bashrc
@@ -220,9 +220,6 @@ if [[ -n "${DOTFILES_PROFILE:-}" ]]; then
     echo "bashrc profile written to /tmp/bashrc-profile.$$.log"
 fi
 
-# opencode CLI
-export PATH="$HOME/.opencode/bin:$PATH"
-
 # Dotfiles scripts on PATH
 export PATH="$HOME/.dotfiles/scripts:$PATH"
 
--- a/.zshrc
+++ b/.zshrc
@@ -178,9 +178,6 @@ complete -o nospace -C /usr/bin/terraform terraform
 # Dump zprof results at end of startup if profiling is enabled
 [[ -n "${DOTFILES_PROFILE:-}" ]] && zprof | head -25
 
-# opencode CLI
-export PATH="$HOME/.opencode/bin:$PATH"
-
 # Dotfiles scripts on PATH
 export PATH="$HOME/.dotfiles/scripts:$PATH"
 
--- a/cli/internal/cmd/tools.go
+++ b/cli/internal/cmd/tools.go
@@ -32,6 +32,7 @@ func newToolsCmd() *cobra.Command {
 	}
 	cmd.AddCommand(newToolsListCmd())
 	cmd.AddCommand(newToolsInstallCmd())
+	cmd.AddCommand(newToolsVersionCmd())
 	return cmd
 }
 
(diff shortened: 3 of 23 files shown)


mlorentedev:
## Review triage

- **PR-Agent (PR Reviewer Guide):** #1294 — every listed requirement compliant, none non-compliant; no code findings. The #791 row is an auto-match on the hive gate now probing `hive --version` through `dotf tools version`; #791 (the hive install-model migration) stays open and is not closed by this PR. Nothing to apply.
- **CodeRabbit:** no review at the time of this triage (rate-limited on every PR today); will be dispositioned if one lands.
- **CI:** green (Windows + Linux Go matrix, bats, shellcheck, PowerShell lint, spec-gate with the spec folder touched).

mlorentedev:
## Review triage (round 2)

| Finding (CodeRabbit) | Disposition |
|---|---|
| `tools.go`: inject the runner through a `Deps` struct | **declined** — this package has no `Deps` struct; its seam convention is the package-level function variable (`deployRenderer`, `probeClient`, `stdinIsTerminal`, `readPassword`, `bwDaemonAddr`), and the test restores it with `t.Cleanup`. Introducing a second convention for one seam is the drift this PR removes elsewhere. |
| `checks_deploy.go`: doctor probed opencode with `trailingVersion` (last token), so the `OpenCode locked.` banner would report false drift | **applied** — `semverOf` wraps `tools.ProbeVersion` through the `System` seam for opencode and pi; `trailingVersion` had no caller left and is deleted; regression test over the banner (`TestCheckOpenCode_VersionIsTheSemverNotTheBannerToken`) |
| table-driven tests | **applied** for `TestProbeVersion` (one table, 9 rows) and `TestToolsVersionCmd`; `checks_catalog_test.go` already runs one subtest per branch, left as is |
| ADR-036: "same channel on every OS" vs the OS-package-manager class | **applied** — the rule is scoped to the two catalog classes; the exception is stated |
| `setup-linux.sh`: post-deploy probe still parsed `opencode --version` locally | **applied** — goes through `dotf tools version opencode` (by name or `~/.local/bin`); `opencode.bats` now refutes a local parse |
| `setup-windows.ps1`: npm prerequisite before moving opencode off winget | **applied** — `OpenJS.NodeJS.LTS` joins the winget list before `dotf tools install`, with a bats ordering guard. The same ordering gap exists on Linux for `bw` since CLI-029 and is **ticketed** as OPS-036 (#1312) rather than widened here. |
| `verification.md`: fences without a language | **applied** — `text` |

Rebased onto `main` after #1305/#1307; Go/lint/bats/shellcheck green locally.

Issue #1318, opened 2026-08-28
HARNESS-115: harness frontmatter schemas declare types nothing validates


mlorentedev:
## Problem

`scripts/compile-harness.sh`'s `validate_skill_frontmatter()` is the only thing that reads `harness/agent-frontmatter.schema.json` and `harness/skill-frontmatter.schema.json`. It evaluates `jq -r '.required[]'` and checks each named key is present and non-empty. Nothing else. Its own comment says so:

> a human reader and for any real JSON Schema validator; this loop does not evaluate them, so do not read a passing --check as type validation.

So every `type`, `enum`, `oneOf` and `minLength` in both schemas is documentation. A record can declare `kind: banana`, `model: [a, b]`, or `capabilities: 7` and `--check` passes.

## How it was found

HARNESS-045 AC7 asks for a `skills:` key the parser cannot read to be a **loud failure "in the schema validation"**. It cannot be met by writing the schema: `skills` is not in `required`, and no type constraint is enforced anywhere. #1319 adds the `oneOf` for `skills` as the SSOT for the shape `cli/internal/harness.parseSkills` implements, and labels it unenforced in its own `description` rather than implying a guard.

## Why it matters beyond that one field

The repo has been bitten three times by a validated-looking surface that validates nothing:

- `capabilities:` in block style rendered an empty allow-list, which **grants** the harness default tool set (guarded at `compile-harness.sh:465`)
- `skills:` in block style rendered `MUST consume: none`, removing enforcement on every harness (fixed in #1319)
- the schema keyword typo that silently disabled a cross-block rule in `model-map.json` (recorded in the HARNESS-075 rounds)

Each was caught by a hand-written guard after the fact. A real validator would have caught all three at `--check` time.

## Scope

- Decide the validator: a Go `dotf harness validate-frontmatter` over both schemas is the shape consistent with ADR-020 and with `resolve-tier` / `resolve-capabilities` / `resolve-skills` — the shell already delegates every other schema-aware question.
- Degrade like its siblings: absent/stale dotf WARNs and falls back to the required-only loop, rather than promoting a best-effort dependency into a hard prerequisite (C15's split by cause).
- Retire the hand-written block-form guards it subsumes, or state why they stay.

## Acceptance

- A record violating a declared `type`/`enum`/`oneOf` fails `--check`, named with its file and the constraint it broke.
- Mutation-checked: each existing hand-written guard's case is caught by the validator with that guard removed.
[...]

Issue #1265, opened 2026-08-27
OPS-033: setup guards tool installs on presence, so six pinned tools never converge — age, zoxide and shellcheck are behind right now


mlorentedev:
## What

`setup-linux.sh` guards most tool installs on **presence alone**:

```sh
# age (file encryption — required by secrets system; pins AGE_VERSION from versions.conf)
if ! command -v age >/dev/null 2>&1; then
    ...install AGE_VERSION...
else
    log_info "age already installed"
fi
```

The comment claims the block "pins AGE_VERSION from versions.conf". In effect it
does not: the pin is read **only on a fresh install** and never for convergence.
Once any version is present, setup reports success forever and `versions.conf`
becomes documentation rather than a control.

**This is the exact defect REFACTOR-013 / #474 fixed for `pi`**, in the same
file, roughly 400 lines away:

```sh
# REFACTOR-013: the pin is a MINIMUM. Presence alone left an outdated pi in
# place forever (it was only ever WARNed by healthcheck, #474, never converged
# here). Upgrade when installed < pin; leave a newer pi alone.
if [ -n "$PI_VERSION" ] && ! version_gte "$installed_pi" "$PI_VERSION"; then
```

The fix exists, is documented, and was applied to exactly one tool.

## Measured on this machine, 2026-08-26, immediately after a successful setup run

| Tool | Installed | Pinned | |
|---|---|---|---|
| `age` | **1.2.1** | 1.3.1 | behind |
| `zoxide` | **0.9.8** | 0.9.9 | behind |
| `shellcheck` | **0.10.0** | 0.11.0 | behind |
| `eza` | 0.23.4 | 0.23.4 | ok |
| `bats` | 1.13.0 | 1.13.0 | ok |

Setup logged `[INFO] age already installed` and exited 0. `dotf doctor` then
warned about the drift — so the machine has been telling us for an unknown number
of runs, and the converging step was never wired.

16 install blocks use the presence-only shape; 6 of the tools they install carry
a pin (`AGE`, `EZA`, `ZOXIDE`, `OBSIDIAN`, `BATS`, `SHELLCHECK`).

## Why `shellcheck` is the one that bites

It is a **verification** tool. A local `shellcheck` on 0.10.0 can report clean on
a script CI rejects at 0.11.0 — which is precisely **BUG-071**, already
catalogued and fixed *for `golangci-lint`* (pinned, resolved from
[...]

PR #1319, opened 2026-08-28
fix(harness): read both skills: frontmatter forms when rendering presence

mlorentedev:
Closes the AC7 half of `specs/HARNESS-045-hook-emission/`. No persona migrates here — this is the gate that has to land first.

## The defect

`build_agent_presence` read `skills:` with an awk one-liner matching the inline form only. A record declaring the mapping form that carries per-skill severity returned **empty**, and the presence block rendered:

```
- **reviewer** — MUST consume: none
```

Enforcement removed, no error, on every harness at once. Measured against a real record in that form, not inferred:

| parser | on the mapping form | consequence |
|---|---|---|
| `compile-harness.sh:1016` | empty | `MUST consume: none` — the presence mechanism disarms itself |
| `check-roster-consistency.py:64` | `[]` | compares "no skills" against the roster and passes. It is HARNESS-046's declared verification (`features.json:26`), so it runs |

## Two corrections to the spec, both measured

**THREE parsers read `skills:` and TWO break — not four and three.** `doctor.readAgentFrontmatter` does skip indented lines, but its only consumers are `fm["model"]` and `fm["targets"]`; it never reads `skills`. A latent trap with no live effect, so no guard was built for it — a guard over a non-consumer is the kind of meta-work this repo caps.

**AC7's "loud failure in the schema validation" cannot be met by writing the schema.** `validate_skill_frontmatter` evaluates `jq -r '.required[]'` and nothing else (its own comment says so), and `skills` is not required. The `oneOf` is added as the SSOT for the shape `parseSkills` implements and says in its own `description` that nothing enforces it. Wider hole ticketed as **#1318**.

## Shape: delegate, don't copy

The spec said to copy `agent_capability_line`'s block-form refuse. That would have made the migration a **flag-day** — the renderer would reject every migrated record permanently, and that renderer feeds the presence block on all four harnesses.

So `build_agent_presence` calls a new `dotf harness resolve-skills`, which reads both forms through `LoadPersona`'s real YAML parser. Same seam as `resolve-tier` and `resolve-capabilities`; the shape lives in Go once instead of being hand-rolled in awk a third time.

**The copied refuse survives where it is still needed: inside the degrade path.** A dotf that is absent or predates the subcommand falls back to the awk read, which is correct for a legacy record and catastrophic for a migrated one — the same silent disarm. So the fallback carries the guard.

Contract: **ids only, no `--harness` flag.** `enforce:` is the gate's input, not the presence text's — and there is no room for it: the doctrine payload has **44 characters of headroom** against `.gemini/GEMINI.md`'s hard 12000-character cap (11956 measured), which the roster has already breached twice.

## On the live box

Installed `dotf` is 0.51.0 and predates `resolve-skills`, so real deploys take the **degrade path** (WARN + awk) until a rebuilt dotf lands. That is correct for today's all-legacy records, and the refuse only arms against migrated ones. Stated so it is not discovered as a surprise.

## Verification

- `resolve-skills` is **byte-identical to the awk read across all 7 shipped records**, asserted per record. Go's `strings.Join` normalises separators that the awk read passed verbatim, so this is the thing that could have silently changed every deployed instructions file.
- The bats stub now advertises the subcommand, so **every pre-existing presence test exercises the delegated path** — their unchanged expectations are the equivalence proof.
- The `--help` contract test now pins `resolve-skills` (lesson 219: a renamed `Use:` makes the shell probe answer "too old" forever and route every deploy through the fallback, silently).

**Mutation-checked in four places**, each turning a test red before restore: removing the degrade refuse, reverting the delegation, changing the join separator, renaming the subcommand.

`go build` / `go vet` / `GOOS=windows go vet` clean · `go test -count=1` 19/19 packages · `golangci-lint` 2.12.2 (the pin) 0 issues · `bats tests/*.bats` **1488/1488, exit 0**.

Local shellcheck is 0.10.0 against the 0.11.0 pin (#1265 open), so CI arbitrates the shell lint. The four SC2016 hits are info-level, pre-existing in this file, and CI runs `severity: error`.

## Also fixed, found along the way

HARNESS-046's declared verification was **red on `main`** before this branch: `cyclomatic-complexity` was in the builder and reviewer definitions and not in `ROSTER.md`. Fixed vault-side (`2e2155b7`, direct to master per convention); the guard now reports `6 invocable roles, all >= 3 skills`, exit 0.

## Next, deliberately not here

Migrating the 35 skills to declared severity. The records under `harness/agents/` are **generated** — that migration edits the vault definitions and re-runs `--refresh`, and per-skill severity is a design decision for the owner rather than 35 silent judgement calls.

Refs #561.

### DIFF
--- a/cli/internal/cmd/harness.go
+++ b/cli/internal/cmd/harness.go
@@ -26,6 +26,7 @@ pattern triggers, and workflow integrations.`,
 	cmd.AddCommand(newHarnessSuggestCmd())
 	cmd.AddCommand(newHarnessResolveTierCmd())
 	cmd.AddCommand(newHarnessResolveCapabilitiesCmd())
+	cmd.AddCommand(newHarnessResolveSkillsCmd())
 	cmd.AddCommand(newHarnessGateCmd())
 	cmd.AddCommand(newHarnessMirrorCmd())
 	return cmd
--- a/cli/internal/cmd/harness_resolve_skills.go
+++ b/cli/internal/cmd/harness_resolve_skills.go
@@ -1,0 +1,81 @@
+package cmd
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/mlorentedev/dotfiles/cli/internal/harness"
+)
+
+// newHarnessResolveSkillsCmd is the third member of the resolve-* family, and it
+// exists for the same reason as the other two: the shell renderer must not grow
+// a parser of its own.
+//
+// `compile-harness.sh`'s skill_field reads the INLINE frontmatter form only. Once
+// a record declares `skills:` as a block list it returns empty, and the presence
+// block renders "MUST consume: none" — enforcement removed, silently, on every
+// harness at once. Measured 2026-08-27. `LoadPersona` already reads both forms on
+// a real YAML parser, so delegating is strictly cheaper than teaching awk YAML.
+//
+// TWO CONTRACT DIFFERENCES FROM resolve-capabilities, both deliberate:
+//
+//   - No `--harness` flag. resolve-capabilities prints a whole frontmatter line
+//     because the native field name differs per harness (`tools:` vs
+//     `permission:`). A persona's skills reach the harness as PROSE inside a
+//     markdown bullet the caller composes, so there is no per-harness field name
+//     to hide and no reason to demand one.
+//
+//   - IDs only, never severity. `enforce:` is what `dotf harness gate` acts on,
+//     not what the presence text says — presence asks, bind enforces. Annotating
+//     each id would also breach a hard platform limit: the doctrine payload this
+//     block lands in is capped at 12000 characters for .gemini/GEMINI.md, and it
+//     has been breached twice already, most recently by the roster itself.
+//
+// The output is byte-identical to skill_field's for every record still in the
+// legacy form, which is what makes this substitution provably behaviour-preserving
+// rather than merely argued to be.
+func newHarnessResolveSkillsCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "resolve-skills <record>",
+		Short: "Print the forced-skill ids one agent record declares, in either frontmatter form",
+		Long: `resolve-skills reads one agent record's ` + "`skills:`" + ` frontmatter and prints the
+declared skill ids as a YAML flow sequence, accepting both the legacy inline form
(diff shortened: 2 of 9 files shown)


mlorentedev:
## Review triage

**CI: 17/17 green, 0 pending** (`goreleaser release` skips on a PR, as designed). Includes `test-windows`, `test (windows-latest)`, `integration`, `spec-gate`, `lint-powershell`.

**Who actually reviewed: PR-Agent only.** CodeRabbit posted a *notice*, not a review — "Review limit reached, next included review available in 58 minutes", free-tier OSS quota exhausted. Disclosed rather than counted: a comment arrived, and it contained no findings because no review ran. Proceeding on PR-Agent's substantive pass.

**PR-Agent verdict:** no major issues, no security concerns, tests present, review effort 3/5. Its three "non-compliant requirement" rows are scope confusions produced by the refs in the PR body, and each is dispositioned below rather than dismissed.

| # | Finding | Disposition |
|---|---|---|
| 1 | #561: "the gate (action) part of determinism is not addressed" | **Declined — correct and intended.** #561 is the whole epic; this PR closes its AC7 half, which the body states in its first line. The gate half (`dotf harness gate` wiring, presence emission, the pi/opencode TS wrappers) is tracked in `specs/HARNESS-045-hook-emission/tasks.md` under "Not done, and deliberately so". Landing AC7 first is the ordering the spec requires: migrating any persona before these guards silently disarms two readers. |
| 2 | #1318: "schema constraints are not enforced; no validator; no test pinning schemas against `parseSkills`" | **Declined — #1318 is the ticket for that work, not a requirement of this PR.** It was filed *from* this PR, by this PR, precisely because the hole is wider than `skills` and wider than one schema. Its acceptance criteria are the three items listed. Implementing them here would have been the wrong size for an AC7 sitting. |
| 3 | #1265: "does not modify setup-linux.sh or address tool version convergence" | **Declined — unrelated.** #1265 is referenced once, as a caveat explaining why local shellcheck (0.10.0 vs the 0.11.0 pin) is not proof and CI arbitrates the shell lint. It is not a requirement this change claims to meet. |

**Nothing applied, nothing deferred.** No finding named a file, a line, or a defect in the change.

**One thing the reviewer did not check, stated so it is on the record:** installed `dotf` is 0.51.0 and predates `resolve-skills`, so real deploys take the degrade path (WARN + awk) until a rebuilt binary lands. Correct for today's all-legacy records — the refuse only arms against migrated ones — and covered by test 65.

Issue #1310, opened 2026-08-27
WIN-013: env-contract SCRIPTS_DIR defaults to ~/.dotfiles/scripts on Windows but setup deploys scripts to ~/scripts, so doctor fails the contract on every fresh box


mlorentedev:
## Problem

Measured 2026-08-27 on the first `dotf doctor` run in the `test-windows` job (#1308) and confirmed on the Windows work box:

```
[Environment variables (contract)]
  [FAIL] SCRIPTS_DIR=C:\Users\<u>\.dotfiles\scripts (path does not exist)
```

- `env-contract.json` declares `SCRIPTS_DIR` with the Windows default `$env:USERPROFILE\.dotfiles\scripts` and `validation: path_exists` (mirrors the Linux `$HOME/.dotfiles/scripts`, where `setup-linux.sh` mirrors `scripts/`).
- `setup-windows.ps1:79` deploys scripts to `$env:USERPROFILE\scripts` and puts *that* on PATH; nothing creates `~/.dotfiles/scripts`.
- The work box passes only by accident: `~/.dotfiles/scripts` there holds a single legacy `load-secrets.ps1` (the retired eager-load, #581), so `path_exists` is satisfied by a leftover. A fresh Windows machine fails the contract check on its first doctor run.
- `setup-windows.ps1:2116` and `powershell/profile.ps1` seed `SCRIPTS_DIR = "$env:DOTFILES_DIR\scripts"` as the fallback, agreeing with the contract and disagreeing with where the files are.

## Fix options

1. **Align Windows to the contract (recommended, ADR-025):** `setup-windows.ps1` deploys scripts into `$DotfilesDest\scripts` and puts that directory on PATH, retiring `~\scripts` (one-time migration note; leftovers pruned by `doctor --fix` like other mirror orphans). Same layout as Linux, one contract, one PATH rule.
2. Change the contract's Windows default to `$env:USERPROFILE\scripts`. Cheaper, but keeps the Windows layout special (the ADR-005 split `pattern-setup-script-idempotence` names as the source of path bugs).

Either way, `tests/env-contract.bats` should assert the Windows default equals the directory `setup-windows.ps1` deploys to.

## Interim

Listed in `.github/scripts/doctor-gate-known-failures.txt` (TEST-003, #1308) with this ticket as owner; the gate fails when the entry stops matching, so the list cannot outlive the fix.

Related: TEST-003 (#1298), ADR-025, #581.

mlorentedev:
[...]

PR #1346, opened 2026-08-28
fix(doctor): declare the runner's missing Bitwarden identity instead of allow-listing its FAIL

### LINKED ISSUE #1313: TEST-005: the test-windows doctor gate has no Bitwarden identity on the runner, so [Bitwarden reach] is a permanent runner-only FAIL
## Problem

With the PR's `dotf` surviving setup on the runner (#1308, TEST-003), the post-setup doctor gate reaches the secrets checks. `dotf tools install` provisions `bw`, and `checkBitwardenReach` then FAILs, correctly by its own rule (28 registry entries resolve through Bitwarden, so an unreachable vault is exposure):

```
[FAIL] Bitwarden session is gone (`bw status`: unauthenticated) — run `bw login` ...
```

The runner has no Bitwarden account and must not have one that resolves real secrets. The row is listed in `.github/scripts/doctor-gate-known-failures.txt` with this ticket as owner; the gate fails if the entry ever stops matching, so the list cannot outlive a fix.

## Fix options

1. **CI-aware degrade in doctor:** when `DOTFILES_INTEGRATION_TEST`/`CI` is set and no `bw` session exists, report the reach tier as SKIP with a reason ("no Bitwarden identity on this runner; reach verified on real boxes only"). Honest, and it keeps the check a FAIL on a real machine.
2. A sandbox Bitwarden account with a dedicated sandbox vault, logged in through a CI secret. Proves the whole chain on every PR at the cost of a second account to rotate.

Option 1 is the floor; option 2 only if a CI-sandbox vault is wanted for other reasons (the DR escrow and `bw serve` lifecycle checks would benefit).

Related: TEST-003 (#1298), #1308, ADR-028.

mlorentedev:
## Summary

**TEST-005 (#1313).** The test-windows doctor gate allow-listed `Bitwarden session is gone (bw status: unauthenticated)` in `.github/scripts/doctor-gate-known-failures.txt` because the runner has no Bitwarden identity. That is a permanent runner condition, and the list's own contract is "runner-only failures with an owning ticket and a fix pending" — an entry that can never be retired does not belong there.

**Fix: a declared flag, not CI sniffing.**
- `.github/scripts/doctor-gate.ps1` sets `$env:DOTFILES_DOCTOR_NO_IDENTITY = '1'` (and logs that it did) before invoking `dotf doctor`.
- `checkBitwardenReach` (`cli/internal/doctor/checks_bw_reach.go`) reads the flag through `sys.Getenv` **in the `unauthenticated` branch only** and reports `[SKIP] no Bitwarden identity on this runner (declared via DOTFILES_DOCTOR_NO_IDENTITY) — reach verified on real boxes only`. Locked, unlocked, sync age and the sync round-trip are untouched: an identity that *is* present is still verified with the flag set.
- The known-failures row is removed in the same change (a PR with the flag but the row would go red as STALE).

**Why declared over sniffed.** `CI=true` / `GITHUB_ACTIONS` would let any workflow inherit the skip by accident and give a real box no way to trip it deliberately. A named variable set by the one script that knows the runner has no identity is both narrower and auditable — the gate prints it.

## Guards

- `TestBWReach_DeclaredNoIdentity` — six-row table pinning the blast radius: unauthenticated at exposure 28 → SKIP (no FAIL, no `bw login`); unauthenticated at exposure 0 → SKIP (not the advisory WARN); locked → unchanged (`token not exercised`); locked + stale sync → the BUG-074 WARN still fires; unlocked + `bw sync` failure at exposure 2 → still exactly one FAIL; unlocked + sync ok → `reach verified`.
- `TestBWReach_NoIdentityFlagIsNotSniffedFromCI` — `CI=true`, `GITHUB_ACTIONS=true`, flag unset, exposure 28 → still FAILs.
- The by-effect guard is this PR's own `test (windows-latest)` run: it builds the PR's `dotf` (#1308) and runs the gate against the runner's real unauthenticated `bw`. If the flag path were wrong, the now-unlisted FAIL would go red as UNEXPECTED.

```
$ go build ./... && go vet ./... && GOOS=windows go vet ./... && go test ./... && golangci-lint run ./...
ok   github.com/mlorentedev/dotfiles/cli/internal/doctor   15.638s   (all 20 packages ok)
0 issues.

$ pwsh -NoProfile -Command "Invoke-Pester -Path tests/doctor-gate.Tests.ps1 -Output Detailed"
Tests Passed: 10, Failed: 0, Skipped: 0

$ Invoke-ScriptAnalyzer -Path .github/scripts/doctor-gate.ps1 -Severity Warning,Error
PSAvoidUsingWriteHost (pre-existing, same class as the script's other progress lines), PSUseSingularNouns on Read-KnownFailures (pre-existing)
non-ASCII in doctor-gate.ps1: none

$ scripts/check-spec-gate.sh --base-ref origin/main --head-ref HEAD --explain
[OK] Production diff 38 LOC < threshold 50 (below threshold; spec not required)
```

**Mutation check.** Replaced `if sys.Getenv(noIdentityEnv) != ""` with `if false && ...`:

```
--- FAIL: TestBWReach_DeclaredNoIdentity (0.00s)
    checks_bw_reach_test.go:571: failures = 1, want 0
    checks_bw_reach_test.go:574: warnings = 1, want 0
```

Restored → `ok`.

## Not in this PR

- Option 2 from the ticket (a sandbox Bitwarden account logged in through a CI secret) — not wanted for its own sake yet; the declared skip is the floor the ticket names.
- The remaining known-failures row (`SCRIPTS_DIR` path, WIN-013 #1310) is owned by the setup change that moves scripts under `~/.dotfiles/scripts`.

Closes #1313


mlorentedev:
## Review triage

Reviewer output on `b62dbc4`: CodeRabbit (1 inline) + PR-Agent guide (no focus areas).

|---|---|---|---|
| 1 | CodeRabbit | `noIdentityEnv` declared in both `checks_bw_reach.go` and `checks_bw_reach_test.go` → `go test` fails with a redeclaration | **Declined — not true on the pushed head.** `git show origin/fix/doctor-no-identity-flag:cli/internal/doctor/checks_bw_reach_test.go` has no `noIdentityEnv` declaration; the only one is the production constant at `checks_bw_reach.go:42`. The Go jobs (`test (ubuntu-latest)`, `test (windows-latest)`, `lint`) are green on this head, which a redeclaration would break. The reviewer read an intermediate push. |

Maintainer note: the implementing agent was stopped by the main session (account session limit); triage recorded here by the main session. CI: green except `test-windows` still running at the time of writing.

PR #1350, opened 2026-08-28
fix(doctor): declare the git-for-windows floor that carries the hooksPath fix

### LINKED ISSUE #912: BUG-069: core.hooksPath wired in C:/ (Windows drive) form breaks git-for-windows hook execution
## Problem

On Windows, every `git commit` / `git push` aborts with:

```
/bin/bash: C:/Users/.../git-hooks/pre-commit: No such file or directory
```

This is the symptom originally filed as BUG-068 (#911). The #911 PR (EOL hardening) fixes a real but **separate** latent bug (CRLF shebangs on Linux/macOS) and satisfies #911 ACs #1/#4, but it does **not** fix this Windows commit-execution failure. The real cause is the `core.hooksPath` **value format**, not the hook EOL.

## Verified root cause (2026-08-10, Git for Windows 2.53.0)

Git-for-Windows runs a hook by invoking its bundled bash as `bash <hookpath>`, which **ignores the shebang** — so a CRLF shebang is harmless on Windows. The failure is that `core.hooksPath` is stored as a `C:/…` (or `C:\…`) Windows drive path, which MSYS bash cannot resolve when git hands it the raw path for execution.

Empirical proof — the **same (CRLF) hooks**, varying **only** the path format, plain `git commit --allow-empty` inheriting the real global config:

| `core.hooksPath` value | rc | result |
|---|---|---|
| `C:/Users/…/git-hooks` (current global) | 1 | `/bin/bash: C:/…/pre-commit: No such file or directory` |
| `C:\Users\…\git-hooks` | 1 | `/bin/bash: C:Users…/pre-commit: No such file or directory` |
| `/c/Users/…/git-hooks` (MSYS form) | 0 | hooks run natively |

EOL is irrelevant: `CRLF + /c/` → rc=0; `LF + C:/` → rc=1 (2×2 matrix confirmed). The `/bin/bash: C:/…: No such file or directory` error subject is the **script file at a `C:/` path**, not an interpreter `bash\r` — #911 pattern-matched the message to the classic CRLF-shebang problem, but the mechanism differs on Windows.

## Where the bad value comes from

- `scripts/install-git-hooks.ps1` `Set-GlobalHooksPath` wires `git config --global core.hooksPath <Target>` where `<Target>` is a Windows path (`Join-Path $DotfilesDir 'git-hooks'` → `C:\…`).
- `dotf doctor` `cli/internal/doctor/checks_guard.go` `setGitGlobalHooksPath` does the same, and `--fix` would re-wire the same broken form.
- The current global value points at the dev checkout (`C:/Users/.../dotfiles/git-hooks`), equally broken.

So the wiring is non-executable on Windows regardless of which dir it targets, until the path **format** is fixed.

## Fix direction (needs design)

Wire/normalize `core.hooksPath` to a git-for-windows-**executable** form on Windows (e.g. the MSYS `/c/…` form, or another form git-for-windows execs), in **both**:

- `Set-GlobalHooksPath` (`install-git-hooks.ps1`)
- `setGitGlobalHooksPath` + `samePath` (`dotf doctor` Go)

with cross-platform tests (must not regress Linux, where absolute POSIX paths already work).

## Coordination

Overlaps the `dotf doctor` guard subsystem currently being changed in **BUG-052** (session "fix(doctor): stop dotf doctor failing on a healthy Windows box"). Sequence / merge with that work to avoid churn on `checks_guard.go`.

## Acceptance criteria

- [ ] On Windows, a plain `git commit` runs the GUARD hooks natively (no `No such file or directory`, no `-c core.hooksPath=` bypass).
- [ ] `install-git-hooks.ps1` and `dotf doctor --fix` wire `core.hooksPath` in a git-for-windows-executable form.
- [ ] `dotf doctor` reports the guard ACTIVE on a healthy Windows box (no false "No such file or directory").
- [ ] Linux/macOS wiring unchanged (regression-tested).

## Refs

- Split from BUG-068 (#911); the EOL half is handled by that PR. This ticket is the actual Windows commit-abort root cause.

mlorentedev:
## Summary

**BUG-069 (#912), re-scoped by the owner on 2026-08-27 after measuring on the Windows work box** (quoted from the ticket):

> On 2.55 the same `C:/…/git-hooks` value **works**: `GIT_TRACE=1 git commit` starts all three hooks by their `C:/` path, the global pre-commit dispatcher runs, and a repo-local `commit-msg` exiting 1 blocks the commit through the chain. […] The root cause was real and is **fixed upstream by git-for-windows**, not by us. Do not implement the `/c/` rewrite.
>
> The remaining work is (a) a **declared git-for-windows version floor** (versions.conf + doctor), and (b) the by-effect probe — which is guards machinery and stays capped per the 2026-08-19 amendment until it bites again.

This PR is (a). Lesson 239 (`docs/lessons/lesson-239-re-measure-a-filed-bug-on-the-current-toolchain-before.md`) records the diagnosis: what survives a bug fixed upstream is not a workaround but the floor, declared where the repo declares versions.

**What.**
- `versions.conf`: `GIT_VERSION=2.55.0`. A minimum like every pin in that file (REFACTOR-013), and the only one no installer provisions — git comes from the OS package manager, so `dotf doctor` is its sole consumer. Comment says so.
- `checkVersionMatch` (the section that consumes versions.conf pins) now runs `checkGitWindowsFloor`: on Windows, `git --version` below the pin → **WARN** naming #912, lesson 239 and `winget upgrade Git.Git`; at or above → PASS; unparseable version line → WARN (never a silent pass); git absent → SKIP (the env contract owns that FAIL); off Windows → SKIP, because the hooksPath defect was git-for-windows'.
- WARN, not FAIL: the floor is bracketed by two measurements (2.53.0 broken, 2.55.0.windows.5 working), not an observed failure on the box, and doctor's precedent for a tool behind its pin is a WARN (yarn, opencode). The env contract's own `min_version: 2.30.0` for git keeps FAILing a genuinely ancient git on every OS.

**Not implemented, deliberately:** the `/c/` MSYS-form rewrite in `Set-GlobalHooksPath` / `setGitGlobalHooksPath` (ticket's original fix direction). Its cause no longer exists on a supported git, and it would be a maintenance liability with nothing left to fix (lesson 239). The by-effect probe (b) stays capped per the owner's re-scope.

## Guards

- `TestCheckGitWindowsFloor` — seven rows through the fake `git --version` line: 2.53.0 → WARN; 2.55.0.windows.5 → PASS; 2.57.1 → PASS (floor, not exact pin); unparseable → WARN; git absent → SKIP; no pin → SKIP; linux 2.34.1 → SKIP. Every row also asserts zero FAILs — the floor is advisory.
- `TestCheckVersionMatch_RunsTheGitWindowsFloor` — registration: dropping the call from `checkVersionMatch` fails it.
- `tests/versions-conf.bats`: `GIT_VERSION` is set (and the existing semver-pattern row covers its shape).

```
$ go build ./... && go vet ./... && GOOS=windows go vet ./... && go test ./... && golangci-lint run ./...
all 20 packages ok; 0 issues.

$ bats tests/versions-conf.bats
ok 13 versions.conf sets GIT_VERSION (git-for-windows floor read by dotf doctor, #912)
16/18 ok; the 2 not-ok are the pre-existing zsh-absent rows on this box (3, 15)

$ git --version   (this box)
git version 2.55.0.windows.5   -> doctor: [ OK ] git 2.55.0 meets the git-for-windows floor 2.55.0 (#912)
```

**Mutation check.** Forced `atLeast(m[1], pin)` to `true`:

```
--- FAIL: TestCheckGitWindowsFloor/windows:_2.53.0_(the_measured-broken_release)_→_WARN_naming_#912_and_the_remedy
    checks_git_floor_test.go:95: warnings = 0, want 1
--- FAIL: TestCheckVersionMatch_RunsTheGitWindowsFloor
```

Restored → `ok`.

## SDD skip rationale

62 LOC by the gate's count, of which the executable part is one ~25-line function; the rest is the docstring recording why the floor exists and why it is a WARN. No public contract changes (`GIT_VERSION` is a new versions.conf key read by doctor only, no CLI flag, no schema), no dependency, not the first step of a sequence — the owner's re-scope closes the ticket with this change. Spec would document a 25-line advisory check.

Closes #912


mlorentedev:
## Review triage

Reviewer output on `e71acf3`: CodeRabbit (1 inline) + PR-Agent guide (no focus areas). Fix in `32ff6da`.

|---|---|---|---|
| 1 | CodeRabbit | `tests/versions-conf.bats` only checks `GIT_VERSION` is non-empty; a manifest carrying 2.54.0 would pass | **Applied.** `unset GIT_VERSION` before sourcing and assert `== 2.55.0`, the floor doctor's WARN names. `bats tests/versions-conf.bats`: green except the two `zsh` cases this box cannot run (no zsh binary; CI's `test` job carries them). |

Maintainer note: the implementing agent was stopped by the main session (account session limit); fix and triage by the main session.

PR #1348, opened 2026-08-28
feat(secrets): the bw serve daemon leaves a log and a pid file, and doctor reads them when it dies

mlorentedev:
## Summary

- `BWServeDaemon.Start()` ran `bw serve` with nil stdio and recorded no pid, so a daemon that is detached from every console by design (WIN-012) died without a trace — on the Windows work box it went down twice within minutes of `dotf secrets unlock` and the only evidence was that nothing listened on 8087 (#1315).
- Now: the daemon's stdout+stderr append to `<DOTFILES_DIR>/state/bw-serve.log` (rotated once over 256 KiB, truncated if the rename is blocked — two files bound it) and `bw-serve.pid` is written beside it. Both paths come from one function in the secrets package, derived from the existing deploy-dir resolution (`env.DotfilesDir` in the CLI, `cfg.DotfilesDir` in doctor), never a literal at a call site.
- `dotf secrets unlock` / `lock` print the pid and the log path (`unlocked (pid 1234, log …/state/bw-serve.log)`); a daemon dotf did not start is reported as `pid unknown`, not guessed.
- `dotf doctor` splits "absent" into three facts: never started (Info, as before), **daemon exited** (WARN, with the log's last three lines), and **pid alive but nothing answers** (WARN). Liveness goes through a new `System.ProcessAlive` seam: `kill -0` on Unix, `OpenProcess` + `GetExitCodeProcess` on Windows via stdlib `syscall` only (`golang.org/x/sys` stays indirect; the two missing Win32 constants are local, like `detachedProcess`).
- The one Windows belief that mattered — a `DETACHED_PROCESS` child with redirected stdio still starts — is measured, not commented: a Go test starts a real detached child through `Start()` on the box and reads its lines back from the log.

Refs #1315 (not `Closes` — see below). Stacked: #1316 (CLI-056, unlock never syncs) builds on this branch.

## SDD checklist

- [x] Issue exists on the bitácora GitHub Project (`Refs #1315`)
- [x] Diff conforms to ~300 LOC atomic cap — ~270 executable production lines (`bwserve_state.go` 161, `procalive_*.go` 39, the rest in `bwserve.go` / doctor / cmd); tests and the spec excluded
- [x] Spec folder `specs/CLI-057-bw-serve-observability/` is included in this PR
- [x] `proposal.md` has filled Why / What / Acceptance criteria
- [x] `tasks.md` is in TDD order
- [x] `verification.md` filled (evidence + mutation checks)

[...]

Issue #531, opened 2026-06-22
dotf doctor --fix does not invoke profile-heal.ps1 on Windows


mlorentedev:
## Context

CLI-018 PR-B retired the Windows `scripts/doctor.ps1` and consolidated all
post-setup diagnostics onto `dotf doctor`. The retired `doctor.ps1 -Fix` used to
auto-invoke `scripts/profile-heal.ps1` (BUG-020: rebuild a corrupted PowerShell
`$PROFILE` from the SSOT, after backing up the corrupted file).

## Gap

`dotf doctor --fix` (`cli/internal/doctor/fix.go` `runHeals`) only invokes
`claude-mem-heal.sh`. It does **not** invoke `profile-heal.ps1`. So the auto-heal
path for a corrupted Windows profile is no longer reachable from the doctor.

## Impact (bounded — not a regression in recovery capability)

- `profile-heal.ps1` is still deployed by `setup-windows.ps1`.
- The BUG-021 fail-fast preflight in `setup-windows.ps1` still detects a corrupted
  profile and points the user at `profile-heal.ps1`.
- So recovery still works **manually**; only the `doctor --fix` auto-invocation is
  gone on Windows.

## Proposed fix

In `runHeals`, when running on Windows, also invoke
`pwsh -NoProfile -File scripts/profile-heal.ps1`, mirroring the existing
`claude-mem-heal.sh` handling (report a real heal vs "nothing to heal", never
fail the run). TDD in the `cli/internal/doctor` fix tests with a `System`
command-output stub.

## Out of scope here

This is intentionally **not** part of CLI-018 PR-B (#509), which is a
delete-and-repoint change. Porting heal behavior into the Go doctor is a separate
atomic change with its own red-green tests.

PR #1353, opened 2026-08-28
fix(doctor): detect a BUG-020-corrupted PowerShell profile and heal it under --fix

mlorentedev:
`checkProfileFiles` tested existence only: a 26 MB profile compounded by the old marker splice reported PASS, and the heal the retired `doctor.ps1 -Fix` used to run (`scripts/profile-heal.ps1`) had no caller on the doctor path after CLI-018 (#531).

- **Windows**: FAIL on either BUG-020 signal — larger than 1 MB, or more than one `# >>> DOTFILES PROFILE >>>` / `# <<< DOTFILES PROFILE <<<` pair — naming `<SCRIPTS_DIR>\profile-heal.ps1` resolved through the env contract (never a literal, so WIN-013's move of `~\scripts` cannot break the remedy).
- **`--fix`**: runs the heal through the `System` seam and verifies by consequence — the re-read profile must carry no signal, or the FAIL stands.
- **`profile-heal.ps1`**: marker threshold `>2` → `>1` so the script and doctor agree (the SSOT contains no markers; one pair is healthy).
- Non-Windows branch unchanged.

Spec: `specs/CLI-064-doctor-profile-heal/` (verification.md carries the evidence and the mutation check: `profileCorruption` returning nil fails the AC1 and AC3 tests).

Evidence (Windows work box): `go build/vet`, `GOOS=windows|linux go vet`, `go test ./...` ok; `golangci-lint` 0 issues. The real profile on this box is healthy (one pair, ~7 KB) and was not touched.

Refs #531 — the issue closes when the spec archives after the pooled review (the archive-on-merge gate refuses a Closes on an active spec, as on #1348). Implementation by the doctor-cluster agent, finished and verified by the main session after the agent was stopped (account session limit).

### DIFF
--- a/cli/internal/doctor/checks_profile.go
+++ b/cli/internal/doctor/checks_profile.go
@@ -1,8 +1,29 @@
 package doctor
 
 import (
+	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
+)
+
+// The dotfiles section of the PowerShell profile is delimited by these two
+// markers — verbatim the strings setup-windows.ps1's splice writes and
+// scripts/profile-heal.ps1 counts. The SSOT (powershell/profile.ps1) contains
+// neither, so a healthy deployed profile carries exactly one of each.
+const (
+	profileStartMarker = "# >>> DOTFILES PROFILE >>>"
+	profileEndMarker   = "# <<< DOTFILES PROFILE <<<"
+
+	// profileMaxBytes is the size signal profile-heal.ps1 heals at (`-gt 1MB`).
+	// A healthy profile is ~7 KB; BUG-020's had compounded to 26 MB / 689K lines
+	// of duplicated sections before anything noticed.
+	profileMaxBytes = 1 << 20
+
+	// profileHealScript is the BUG-020 repair, deployed under SCRIPTS_DIR by
+	// setup-windows.ps1: it backs the corrupted profile up beside itself and
+	// rewrites the dotfiles section from the SSOT.
+	profileHealScript = "profile-heal.ps1"
 )
 
 // checkProfileFiles ports the healthcheck.ps1 §4 residual the .sh→Go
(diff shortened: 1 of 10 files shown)


mlorentedev:
## Review triage

Reviewer output on `8aa73de`: CodeRabbit (4 inline); PR-Agent guide (no focus areas). CI all green on this head (both Go legs, bats, test-windows gate, spec-gate).

|---|---|---|---|
| 1 | CodeRabbit | Make the profile tests table-driven and assert status data instead of prose | **Declined.** AC1/AC2 are tables already; the `--fix` subtests assert status DATA (`rep.Failures()`, `rep.totals[StatusFix]`) — the substrings they also check are the remedy contract (which script, which path, what the heal said), not decoration. Folding five differently-shaped fixtures (heal rewrites / no-op / script missing / pwsh missing / healthy) into one table would hide the fixture each row needs. |
| 2 | CodeRabbit | "Restore the `--fix` repair contract" — CI showed pwsh invoked with no arguments and the four FAIL branches missing | **Declined — resolved before this head.** That was the Linux leg on `527f1ee`, where a Windows-only join put a backslash the Linux fixture could not find; `8aa73de` builds the path with the same `filepath.Join` the tests use and both `test (ubuntu-latest)` and `test (windows-latest)` are green. The repair contract itself never changed. |
| 3 | CodeRabbit | Route `os.Stat`/`os.ReadFile` through the filesystem seam | **Declined — package convention.** `System` seams env, commands and the clock; the doctor package reads the filesystem directly everywhere (`checks_deploy.go` 4 direct calls, `checks_dr.go` 3) and injects it through `t.TempDir()` fixtures, which is what these tests do. A one-off seam here would be the odd one out. |
| 4 | CodeRabbit | (`checks_profile.go:156`, Minor) | **No actionable content** — the comment body carries only the reviewer's own analysis script, no claim. |

Refs #531 (closes when the spec archives after the pooled review).

PR #1354, opened 2026-08-28
test(install): make the rollback case reach Set-DotfBinary's catch block

### LINKED ISSUE #781: BUG-046: install-dotf.ps1 rollback test never reaches the catch-block restore it claims to cover
Found by CodeRabbit on #760, split out from a broader test-precision maintenance pass (see #780 and its follow-up) because it needs Windows/pwsh to write and verify safely.

## Problem

`tests/install-dotf-ps1.Tests.ps1`'s "restores the previous binary when the source cannot be staged" test (currently around line 95) is meant to cover `Set-DotfBinary`'s catch-block rollback in `scripts/install-dotf.ps1:89-98` (the "never leave the user without a dotf" park-and-restore path). It doesn't actually reach that code.

Walking the function (`scripts/install-dotf.ps1:67-103`): the test passes a nonexistent `-Source`, which makes `Copy-Item -LiteralPath $Source -Destination $staged -Force` (line 84) throw *before* the target is ever parked (line 86-88 never runs). So the assertion that `$Target` still reads `'OLD'` passes because the target was never touched at all — not because the catch block's `Move-Item -LiteralPath $parked -Destination $Target -Force` (line 94) executed and restored it.

To actually exercise that line, the test needs the *first* `Move-Item` (target -> `.old`, line 87) to succeed, and only the *second* `Move-Item` (staged -> target, line 90) to throw. The existing `New-RunningImageLock` helper in this file doesn't help here: test #54 already establishes that a `FileShare.Read + Delete` lock blocks `Copy-Item` overwrites but does *not* block `Move-Item`/rename (test #67 relies on exactly that to let the park step succeed on a "running" binary).

## Why not fixed inline

The natural fix is a Pester `Mock Move-Item -ParameterFilter { $Destination -eq $Target } -MockWith { throw ... }`, letting the park-step call (`-Destination -eq $parked`) fall through to the real filesystem operation. Getting Pester 5's mock/parameter-filter fallthrough semantics right (does an unmatched call reach the real cmdlet, or does it silently no-op?) needs to be verified by actually running it — this environment has no `pwsh` and no Copilot/Pester runtime, so a mock written blind here couldn't be confirmed to fail red before it passes green, which is exactly the standard this repo holds every other guard to (`docs/lessons.md`, 2026-08-05 entries).

## Proposed fix

Add a Pester case that mocks `Move-Item` to throw specifically on the staged->target call (matched by `-Destination`), confirms the park call still executes for real, and asserts both that the function throws *and* that `$Target` ends up restored to the pre-upgrade content. Verify in the red direction first: temporarily remove the catch block's restore line, confirm the new test fails, then restore it and confirm green.

## Context

Split from the maintenance-bundle PR (#780 and its sibling) rather than shipped unverified. Root repo guidance: `docs/lessons.md`'s "a guard installed... only verified in the green direction is indistinguishable from one that asserts nothing" applies as much to a test itself as to the thing it tests.

mlorentedev:
The "restores the previous binary when the source cannot be staged" case in `tests/install-dotf-ps1.Tests.ps1` passed a nonexistent `-Source`, so `Copy-Item` threw before the park step and the catch block it claimed to cover never ran (BUG-046, found by CodeRabbit on #760).

- The old case keeps its real value under an honest name: an aborted staging leaves the old binary untouched and no park behind.
- A new case breaks only the second `Move-Item` (mock filtered on the staged `.new` source, so the park and the catch-block restore fall through to the real cmdlet) and asserts the parked image is back in place, the park consumed, and the staging file gone.

Verified red first on the Windows work box: with the restore line removed from `scripts/install-dotf.ps1` the new case fails ("Expected exactly 'OLD', but got $null"); restored, 8/8 pass. Pester 5 falls through to the original command when no parameter filter matches — the ticket's open question (e) — so no inverse mock is needed.

Test-only change (0 production LOC); no spec. Closes #781.


mlorentedev:
## Review triage

Reviewer output on `8059f46`: PR-Agent guide — no focus areas, no security concerns. CodeRabbit did NOT review — "Review limit reached" (free-tier quota); proceeding on PR-Agent's output, disclosed here.

|---|---|---|---|
| — | PR-Agent | No findings | **No change.** CI all green incl. the Pester step on the Windows runner. |

Note for the record: the first push asserted the staged `.new` file was gone after the rollback; the runner's scanner held that fresh file once and the best-effort `Remove-Item -ErrorAction SilentlyContinue` left it — the assertion was dropped with the reason in the test, since staging cleanup is best-effort by design and the next upgrade overwrites it.

Issue #1358, opened 2026-08-28
CLI-073: env.RepoDir and doctor do not recognise a git worktree as a checkout (.git is a file there), so worktree-based work falls back to DOTFILES_REPO_DIR


mlorentedev:
## Problem
Measured on the Windows work box 2026-08-28 while verifying AI-038 from a worktree: with DOTFILES_REPO_DIR pointing at C:\Users\mlorente\Projects\Workspace\dotfiles-wt-setup-cluster, dotf doctor reports "DOTFILES_REPO_DIR resolves to ... which is not a git checkout", and env.RepoDir's cwd walk-up (cli/internal/env/env.go ~163, walkUpFor(cwd, ".git")) does not treat the worktree root as a repo either, so commands run from a worktree read the MAIN checkout's files (catalogPin, checkInstructionDrift, harness mirror sources). In a git worktree .git is a FILE (gitdir: ...), not a directory; every check that asks isDir(.git) answers wrong.

This repository's standing rule is to work in worktrees (feedback_dotfiles_use_worktree), so every session hits it: a PR's packages.json, harness/ or ai/ changes are invisible to doctor and to dotf tools install until merged and mirrored.

## Fix
Treat a .git FILE whose content starts with gitdir: as a checkout (or use git rev-parse --show-toplevel through the seam) in env.RepoDir and in the doctor contract check; Go table test with a temp worktree layout (dir with .git file) and a plain checkout (dir with .git dir).

## Related
BUG-072 (RepoDir prefers the cwd walk-up), AI-038 (#1321) verification notes.

mlorentedev:
Renumbered `CLI-065` -> `CLI-073`. Title only; scope, body and board item are unchanged.

`CLI-065` was claimed by two unrelated issues. Ownership resolved as: specs/CLI-065-env-persist-sweep declares issue #1363.
An `AREA-NNN` names one piece of work, so the other claimant keeps the id and this one yields.

Allocated forward-only from a scan taken immediately before this write, over all issues with
pull requests excluded — a scan run minutes earlier is what produces the collisions in the
first place (`harness/skills/new-ticket/SKILL.md:211-217`).

Part of the #1448 classification pass. Note that duplicated ids are not all defects: an
`AREA-NNN` fanning out across many spec directories under one issue is a legitimate epic
(`CLI-024` has 18), and those were left alone.

Issue #1321, opened 2026-08-28
AI-038: Copilot CLI is on npm (@github/copilot) but installed through winget on Windows and not at all on Linux - move it to packages.json (amends ADR-036)


mlorentedev:
## Problem
ADR-036's table lists copilot under 'no cross-OS channel'; `npm view @github/copilot version` -> 1.0.81 (measured 2026-08-27). Windows installs it via winget (setup-windows.ps1 ~387) with no pin and autoUpdate on (winget says 1.0.78, the binary says 1.0.80); Linux installs nothing (setup-linux.sh ~1013-1039 detect-and-act). No `dotf tools version` row, no shadowed-install WARN.

## Decision (owner, 2026-08-27)
npm on both OSes: `{"name":"copilot","version":"1.0.81","source":{"type":"npm","package":"@github/copilot"}}` in packages.json, drop the winget row, set `autoUpdate: false` in ai/copilot/config.json (pin-as-floor owned by dotf tools install), amend ADR-036's table, revisit the BUG-003 'no auto-install' policy, `requires_command: copilot` gates, lesson-157 and verify-setup.bats:275. Needs the Windows box for the npm .cmd shim path check. Effort M.

Source: Copilot audit F3, 10_projects/dotfiles/research/2026-08-27-windows-drive-audit-plan.md (knowledge vault)

Issue #1312, opened 2026-08-27
OPS-036: setup-linux.sh runs dotf tools install before Node.js exists, so the npm catalog tools (bw, opencode) cannot install on a clean box


mlorentedev:
## Problem

Found while dispositioning the AI-034 review (#1311): `setup-linux.sh:305` calls `dotf tools install` right after `install_dotf`, and every npm-sourced catalog tool (`bw` since CLI-029, `opencode` since ADR-036) is provisioned by `npm install -g` inside it. Node.js is installed later in the script (nvm block), so on a machine that never had node, `dotf tools install` fails for those tools with a warning and setup continues; `dotf doctor` then reports `bw not in PATH` / `opencode missing`, and the remedy it prints (`dotf tools install`) works only because node exists by then. The first `test-windows` gate run (#1308) showed the same shape on the Windows runner (`bw not in PATH`); `setup-windows.ps1` now provisions Node.js LTS through winget before the catalog install (#1311).

## Fix options

1. Move the `dotf tools install` call after the node/nvm block in `setup-linux.sh` (mirrors the Windows ordering), and guard the order with a bats test.
2. Make `installNpm` fail loud and specific when `npm` is absent ("npm not on PATH: install Node.js first") so the warning names the prerequisite rather than the tool.

Both; (1) fixes the clean-box outcome, (2) fixes the message.

Related: ADR-036, CLI-029, AI-034 (#1294), TEST-003 (#1298).

mlorentedev:
Decision 2026-08-27 (owner): setup-linux.sh provisions Node.js LTS before dotf tools install (parity with setup-windows.ps1, which installs OpenJS.NodeJS.LTS before the catalog since #1311): nvm with the major pinned in versions.conf. Correction to the ticket's premise: setup-linux.sh never had a node/nvm install block, so 'move the call after it' was never an option. The Go pre-check in installNpm ('npm not on PATH: install Node.js first') ships as well.

PR #1359, opened 2026-08-28
feat(tools): copilot joins the npm catalog on every OS, and doctor reports it against the pin

mlorentedev:
ADR-036 left `copilot` to the OS package manager as a tool with no cross-OS channel. Measured on the Windows work box: GitHub publishes the CLI as `@github/copilot` on npm (1.0.81), the winget-installed binary had self-updated behind the registry (1.0.78 recorded, 1.0.81 running), Linux installed nothing, and no doctor row nor shadowed-install WARN could apply because the tool was not in the catalog (AI-038, #1321; owner's decision 2026-08-27: npm on both OSes).

- **packages.json** declares `copilot` (npm, `@github/copilot`, 1.0.81); the winget row leaves `setup-windows.ps1`; both setup blocks only deploy config when the binary is on PATH.
- **ai/copilot/config.json** → `"autoUpdate": false` so the pin owns updates.
- **dotf doctor** gains a "GitHub Copilot CLI" section: version probed once in Go, exact match PASS / drift WARN like opencode, SKIP when absent (a box may carry no Copilot); the existing shadowed-install WARN covers a leftover winget/scoop copy.
- **ADR-036**: `copilot` moves to the npm class with a dated amendment.

**Verified on the box** (details in `specs/AI-038-copilot-npm-channel/verification.md`): with the winget copy satisfying the floor, `dotf tools install` added nothing — ADR-036 §5 in practice, retire the other channel first; after `winget uninstall --force` (the portable package had self-modified) the catalog installed the npm copy (`copilot 1.0.81 installed via npm`, second run `already installed; skipping`), `Get-Command copilot -All` shows only the npm shims, doctor reports `(2 checks, all ok)` with no WARN, and `copilot -p` answers `PONG`.

Tests: `TestCheckCopilot_PinMatchByStatus` (5 rows by status; mutation-checked), bats guards in `copilot-config.bats` + `setup-windows.bats`; `verify-setup.bats`'s "not deployed when absent" case keeps its premise in the Node-less container (#1312) and lost its stale "gh-copilot extension" wording.

Found on the way, filed: **#1358** — `env.RepoDir`/doctor do not recognise a git worktree as a checkout.

Spec: `specs/AI-038-copilot-npm-channel/`. Refs #1321 — closes when the spec archives after the pooled review.

### DIFF
--- a/ai/copilot/config.json
+++ b/ai/copilot/config.json
@@ -1,7 +1,6 @@
 {
   "model": "gpt-5.6-sol",
   "includeCoAuthoredBy": false,
-  "autoUpdate": false,
   "trustedFolders": [
     "/home/manu/Projects",
     "/home/manu/Projects/*",
--- a/ai/copilot/config.json
+++ b/ai/copilot/config.json
@@ -1,6 +1,7 @@
 {
   "model": "gpt-5.6-sol",
   "includeCoAuthoredBy": false,
+  "autoUpdate": false,
   "trustedFolders": [
     "/home/manu/Projects",
     "/home/manu/Projects/*",
--- a/cli/internal/doctor/checks_copilot.go
+++ b/cli/internal/doctor/checks_copilot.go
@@ -1,0 +1,25 @@
+package doctor
+
+// checkCopilot reports the GitHub Copilot CLI against its packages.json pin
+// (AI-038, #1321; ADR-036 amendment): copilot is an npm catalog tool on every
+// OS, so its version is probed once in Go (tools.ProbeVersion via semverOf)
+// and compared with the catalog pin exactly as opencode is — an exact match
+// PASSes, anything else is a drift WARN (the floor semantic belongs to
+// `dotf tools install`, which never downgrades). Absent is a SKIP: a box may
+// deliberately not carry Copilot, and setup only deploys its config when the
+// binary is on PATH. A leftover winget or scoop copy beside the npm one is
+// reported by checkShadowedCatalogTools, which reads the same catalog.
+func checkCopilot(sys *System, cfg *Config, rep *Report) {
+	rep.Section("GitHub Copilot CLI")
+	if !sys.has("copilot") {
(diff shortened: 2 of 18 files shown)


mlorentedev:
## Review triage

Reviewer output on `a6b0cb0`: PR-Agent guide (2 focus areas) + CodeRabbit (6 inline). Fixes in `2934ea4`.

|---|---|---|---|
| 1 | PR-Agent | `packages.json` carries the `copilot` entry twice | **Applied — real defect.** A patch applied on top of itself; `dotf tools install` had installed once and printed `already installed; skipping` for the twin, which I misread as idempotence. Deduplicated, and guarded where every consumer goes through: `tools.Load` now refuses a name listed twice (`TestLoad_RejectsDuplicateToolNames`, mutation-checked) and `tests/packages-json.bats` asserts unique names + entry shape. |
| 2 | PR-Agent | Early return on a version-less `copilot --version` skips the pin comparison | **Declined.** Intentional: with no semver there is nothing to compare, and a drift line for "unknown" would be noise — the test's "no semver" row asserts a WARN and no drift line. |
| 3 | CodeRabbit | `features.json` f1 omits `tests/setup-linux.bats` | **Applied** (+ `packages-json.bats`). |
| 4 | CodeRabbit | f2 tests `checkCopilot` directly; removing its registration would stay green | **Applied.** The verification now also asserts the `doctor.go` registration line. |
| 5 | CodeRabbit | f4's command does not cover every AC5 outcome | **Applied.** Behavior narrowed to what the command proves; the full box sequence stays recorded in `verification.md`. |
| 6 | CodeRabbit | proposal AC3 says FAIL below the pin; the contract is WARN drift | **Applied.** Both statements now read exact-match PASS / otherwise drift WARN (the floor is `dotf tools install`'s). |
| 7 | CodeRabbit | untyped fenced block in `verification.md` | **Applied** (`text`). |
| 8 | CodeRabbit | `packages.json:44` (Minor, Data Integrity) | **No actionable content** — the comment body carries only the reviewer's analysis script. |

Issue #1324, opened 2026-08-28
CLI-058: dotf env persist - the ADR-025 cascade vars do not exist inside a Copilot tool call on Windows (-NoProfile)


mlorentedev:
## Problem
`copilot help config` -> powershellFlags defaults to [-NoProfile, -NoLogo] on Windows. Measured: DOTFILES_REPO_DIR, DOTFILES_DIR, VAULT_PATH, COPILOT_HOME are empty at User scope; they exist only in profile.ps1 and the setup pre-export (setup-windows.ps1 ~2124-2130). The only User-scope writes are HIVE_VAULT_PATH, three Gemini vars and PATH. Inside a Copilot tool call `dotf` resolves (User PATH) but `dotf harness mirror` errors 'cannot locate the dotfiles checkout' and vault-path commands fall back or fail. Claude Code is unaffected (Git Bash sources .bashrc).

## Fix
`dotf env persist` (or extend `dotf env generate`) writes the contract's cascade vars at User scope from machine.json, idempotent (present+equal -> no write), invoked by setup; Linux leg is a no-op or environment.d. Do NOT set powershellFlags: [] (loads the interactive profile on every tool call). Needs the Windows box. Effort M.

Source: Copilot audit F2, 10_projects/dotfiles/research/2026-08-27-windows-drive-audit-plan.md (knowledge vault)

Issue #1323, opened 2026-08-28
AI-040: Copilot doctrine does not travel outside a repo carrying AGENTS.md - export COPILOT_CUSTOM_INSTRUCTIONS_DIRS through the env contract


mlorentedev:
## Problem
compile-harness.sh ~1022 claims copilot has 'a full instructions file of their own, so the enforced rules travel with it', but ai/copilot/copilot-instructions.md has no enforced region and copilot is absent from manifest targets[]. Copilot auto-loads AGENTS.md from git root/cwd only (`copilot help`: --no-custom-instructions; env COPILOT_CUSTOM_INSTRUCTIONS_DIRS), so in $HOME, the vault or a non-agentic repo only the 2 KB pointer loads, whose fallback needs $DOTFILES_REPO_DIR (unset under -NoProfile on Windows).

## Fix
Export COPILOT_CUSTOM_INSTRUCTIONS_DIRS=$DOTFILES_REPO_DIR in env-contract.json (rc files + the persisted scope of the `dotf env persist` ticket) so AGENTS.md always loads; or add copilot to targets[] as pointer and to deployedDoctrineTargets. Effort S.

Source: Copilot audit F8, 10_projects/dotfiles/research/2026-08-27-windows-drive-audit-plan.md (knowledge vault)

mlorentedev:
[...]

PR #1360, opened 2026-08-28
feat(setup): Windows installs Claude Code and agy through their official installers, as Linux has since ADR-009

mlorentedev:
`setup-linux.sh` has provisioned Claude Code (`claude.ai/install.sh`) and the Antigravity CLI (`antigravity.google/cli/install.sh`) for months; on Windows every block that gates on `claude` (MCP registration, plugins, both session hooks) and the agy config deploy were dead on a clean box until a human installed the binaries by hand (AI-041, #1325). Both vendors ship `install.ps1` — measured HTTP 200 on 2026-08-28 — so this is ADR-036 class 3, the official installer on every OS, no channel change.

- `Install-AgentBinary` (one helper, two calls): skip when the command is on PATH; run the installer in a child `pwsh` exactly as the uv/Bun installers are (a remote installer must not rewrite this session's environment); put the vendor's bin dir on the session PATH once; say plainly when the binary still does not resolve. Placed before the MCP block that needs `claude`.
- Guard: `tests/setup-windows.bats` asserts the helper, both URLs, the child-pwsh invocation, the ordering before `claude mcp get`, and the Linux parity.

Evidence: parse 0 errors, PSScriptAnalyzer 0 findings (repo settings), ASCII delta 0, `bats tests/setup-windows.bats` all ok; on the work box both binaries are present, so both calls take the idempotent branch (`already installed`). The install branch runs for real on the CI runner, where `test-windows` starts clean — its Pester/gate output is the proof that the MCP and plugin blocks now execute there.

Closes #1325.

## SDD skip rationale
Parity block mirroring an existing Linux one, under 50 production lines; no new contract.

PR #1362, opened 2026-08-28
feat(env): dotf env persist writes the contract variables where a profile-less process reads them

mlorentedev:
The ADR-025 cascade reaches a process only through the rc files, and Copilot runs every tool call as `pwsh -NoProfile`: measured on the Windows work box, `DOTFILES_REPO_DIR`, `DOTFILES_DIR`, `VAULT_PATH` and `COPILOT_HOME` were empty at User scope while every shell had them, and `dotf harness mirror` could not find the checkout from inside a Copilot tool call (CLI-058, #1324).

- **`dotf env persist`**: resolves every structural variable exactly as `generate` does (one resolver, `env.ResolveVars`) and writes it into the per-user persistent scope — `HKCU\Environment` via `x/sys/windows/registry`, plus the `WM_SETTINGCHANGE` broadcast `setx` sends — touching only values that differ; `--check` reports drift and exits non-zero. Off Windows it is a no-op that says so (the rc files source `paths.sh`).
- **setup-windows.ps1** calls it right after `dotf env generate` (bats ordering guard).
- **`dotf doctor`** gains "Persisted environment (user scope)": PASS when every variable matches, WARN naming the missing/stale ones with the remedy, WARN when the store is unreadable, no section where the OS has no such scope (`System.UserEnv` seam).
- `golang.org/x/sys` becomes a direct dependency (first production use of the registry API).

**Verified on the box** (details in `specs/CLI-058-env-persist/verification.md`): `--check` → 10 drifted; `persist` → 10 changed; `--check` → ok; `persist` → 0 changed (idempotent); then a `pwsh -NoProfile` started with `Start-Process -UseNewEnvironment` — a fresh environment built from the registry, no inheritance from the launching shell (a marker set there did not arrive) — prints the five values. Doctor's new section reads `(1 checks, all ok)`.

Tests: `TestPersist_TouchesOnlyWhatDiffers`, `TestDrift_NamesMissingAndDifferent`, `TestPersist_StoreErrorsNameTheVariable`, `TestCheckPersistedEnv_ByStatus` (four rows by status + no-seam case); mutation-checked (Persist writing everything; Drift never reporting). Go loop incl. `GOOS=linux|windows go vet`, golangci-lint 0 issues; bats setup-windows ok; `.ps1` parse clean, ASCII delta 0.

Spec: `specs/CLI-058-env-persist/`. Refs #1324 — closes when the spec archives after the pooled review. Follow-up this unlocks: #1323 (the pointer's `$DOTFILES_REPO_DIR` fallback now resolves in Copilot tool calls).

### DIFF
--- a/cli/go.mod
+++ b/cli/go.mod
@@ -6,12 +6,12 @@ require (
 	github.com/santhosh-tekuri/jsonschema/v6 v6.0.3
 	github.com/spf13/cobra v1.10.2
 	go.yaml.in/yaml/v3 v3.0.5
+	golang.org/x/sys v0.47.0
 	golang.org/x/term v0.45.0
 )
 
 require (
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	github.com/spf13/pflag v1.0.9 // indirect
-	golang.org/x/sys v0.47.0 // indirect
 	golang.org/x/text v0.14.0 // indirect
 )
--- a/cli/internal/cmd/env.go
+++ b/cli/internal/cmd/env.go
@@ -25,6 +25,7 @@ func newEnvCmd() *cobra.Command {
 	cmd.AddCommand(newEnvGenerateCmd())
 	cmd.AddCommand(newEnvPathCmd())
 	cmd.AddCommand(newEnvSetCmd())
+	cmd.AddCommand(newEnvPersistCmd())
 	return cmd
 }
 
--- a/cli/internal/cmd/env_persist.go
+++ b/cli/internal/cmd/env_persist.go
@@ -1,0 +1,85 @@
+package cmd
+
+import (
+	"errors"
+	"fmt"
+	"runtime"
+
+	"github.com/mlorentedev/dotfiles/cli/internal/env"
+	"github.com/spf13/cobra"
+)
+
+// userEnvStore is the seam tests replace so the command never touches a real
+// registry; production resolves the OS store.
+var userEnvStore = env.NewUserEnvStore
+
+// newEnvPersistCmd is CLI-058 (#1324): the ADR-025 cascade exists only in the
+// rc files, so a process started with no profile — Copilot's `pwsh
+// -NoProfile` tool calls, a Scheduled Task — sees none of DOTFILES_REPO_DIR,
+// DOTFILES_DIR, VAULT_PATH, SCRIPTS_DIR... `persist` writes the same resolved
+// values `generate` renders into the per-user persistent scope, touching only
+// what differs, so a second run is a no-op.
+func newEnvPersistCmd() *cobra.Command {
+	var check bool
+	cmd := &cobra.Command{
+		Use:   "persist",
+		Short: "Write the resolved contract variables into the per-user persistent scope (Windows: HKCU\\Environment)",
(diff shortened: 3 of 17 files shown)


mlorentedev:
## Review triage

Reviewer output on `0101755` (rebased onto main after #1360 merged; the only conflict was two sibling tests appended at the same spot in `tests/setup-windows.bats`, both kept): PR-Agent guide — no major issues, no security concerns, #1324 "PR Code Verified". CodeRabbit posted its walkthrough with four merge-risk notes and then **paused its line review** (auto-pause after the rebase push); no line-level findings were produced. Proceeding on PR-Agent's review plus CodeRabbit's walkthrough notes, dispositioned below. CI 16/16 green.

|---|---|---|---|
| 1 | CodeRabbit | Setup may persist values even when `dotf env generate` fails, leaving profile-loaded and profile-less processes with different paths | **Declined.** Both verbs read the same `env.ResolveVars(contract, machine.json)` — one resolver, two sinks. A resolution failure fails both identically; a `paths.ps1` write failure leaves the registry *more* current than the rc file, which is the better of the two outcomes for the consumer this PR serves, and both branches WARN in setup (`setup-windows.ps1:1875-1892`). Gating persist on the file write would withhold the registry for symmetry's sake. |
| 2 | CodeRabbit | A missing `HKCU\Environment` key prevents persistence entirely | **Declined.** It fails loudly, not silently: `OpenKey` errors → `Persist` returns `read persisted <NAME>: ...` → non-zero exit → setup WARN → doctor WARN "unreadable". The key is created by the profile loader for every loaded user hive; a process without a loaded profile has no User scope worth writing to, and creating it silently would write into a hive no logon reads. |
| 3 | CodeRabbit | Sequential writes, no rollback for a mid-run failure | **Declined.** Convergence is the contract, not a transaction: a failure names the variable and leaves the earlier ones correct; the next run writes only what still differs (AC1, `TestPersist_TouchesOnlyWhatDiffers`). Rolling back would remove values a consumer already reads correctly. Same model as every other setup step. |
| 4 | CodeRabbit | No cleanup of retired contract variables | **Ticketed → #1363 (CLI-065).** Real gap, same class as WIN-013's retired scripts: a deployer that adds and never sweeps. Needs an ownership marker so the sweep is bounded to what dotf wrote; out of this PR's scope by design. |
| 5 | PR-Agent | #1323 "not compliant" | **Declined.** Auto-linked from a mention in the body; #1323 is context, not a target of this PR, and its premise is false (the env var it names is undocumented; Copilot already loads `~/.claude/CLAUDE.md` outside a repo). To be closed once this merges. |
| 6 | CodeRabbit | Description check: template boxes unchecked, >300-line diff not justified | **No change.** The body's test plan and manual verification sections carry the evidence; the spec folder (`specs/CLI-058-env-persist/`) is the SDD justification the gate reads. |

Box evidence (Windows work box, 2026-08-28): `--check` → 10 drifted; `persist` → 10 changed; `--check` → ok; `persist` → 0 changed; a `Start-Process -UseNewEnvironment` child (environment rebuilt from the registry, no inheritance, no profile) prints the five contract values. Doctor 132 passed, 0 failed.

Issue #1339, opened 2026-08-28
CLI-063: extend dotf deploy to cover Claude - the last agent whose configs are still twinned in jq and PowerShell


mlorentedev:
## Problem
setup-linux.sh ~1051-1163 + ~1305-1340 + ~1375-1477 vs setup-windows.ps1 ~124-300 + ~453-540 + ~766-810: ~330 lines each side implementing the same policy (snapshot/restore of .claude.json, `claude mcp` loop over mcp-servers.json, hardcoded plugin list, per-key settings.json merge). Every fix lands twice; the plugin list and the merge policy are duplicated data.

## Fix
`dotf agent claude sync`: reads mcp-servers.json + a new ai/claude/plugins.json, wraps `claude mcp/plugin` with one Go snapshot guard (reuse the mem session-start canary threshold), plus a merge kind in dotf deploy for settings.json; delete both shell blocks and the pinning bats tests. Concrete slice of #163 / #792. Effort M.

Source: scripts-parity audit F1, 10_projects/dotfiles/research/2026-08-27-windows-drive-audit-plan.md (knowledge vault)

mlorentedev:
## Measured 2026-09-04 against `origin/main` — the premise holds, and there is one thing it misses

Three of the last four tickets in this drive had a false premise, so this was measured before any code. **This one is largely accurate**, which is worth recording as explicitly as a refutation would be.

### What checks out

| Claim | Measured |
|---|---|
| ~330 lines each side | Linux Claude region is **1010–1338 = 329 lines**, plus `merge_claude_settings` at **1338–1437 = 100**. So ~429 for the whole Claude deploy surface, not 330 |
| The plugin list is duplicated data | **Verbatim identical** on both sides: `gopls-lsp@`, `security-guidance@`, `frontend-design@claude-plugins-official` |
| `ai/claude/plugins.json` is new | Correct — `ai/claude/` holds only `CLAUDE.md` and `settings.json` |
| `dotf agent claude sync` is a plausible home | `dotf agent` **already exists** (`cli/internal/cmd/agent.go:22`), so this is a subcommand, not a new noun |
| A merge kind in `dotf deploy` is new | Correct — `harness/manifest.json` has `native` and `pointer`, no merge kind |

### The line numbers in the body are stale — use these

Written 2026-08-27; the setup scripts have moved since (they are now 1723 + 2338). Located by content:

| | `setup-linux.sh` | `setup-windows.ps1` |
|---|---|---|
[...]

Issue #1334, opened 2026-08-28
AI-042: nine hardcoded trusted-folder paths in ai/copilot/config.json and ai/agy/settings.json are deployed verbatim to every machine (ADR-025)


mlorentedev:
## Problem
ai/copilot/config.json ~5-10 (six trustedFolders literals, two usernames: /home/manu, C:\Users\mlorente) and ai/agy/settings.json ~19-21 (three trustedWorkspaces) are copied to $HOME by both setups; the trust lists silently match nothing on any other machine. tests/antigravity.bats ~43-48 guards this class for one file only.

## Fix
Declare both as dotf deploy configs rendered with $HOME / %USERPROFILE% placeholders (the render path CLI-039 uses for secrets); widen the bats refute_grep to ai/**/*.json. Effort S.

Source: scripts-parity audit F2, 10_projects/dotfiles/research/2026-08-27-windows-drive-audit-plan.md (knowledge vault)

mlorentedev:
**Reopening: closed by automation while the spec is still `implementing` and its review verdict is FAIL.**

This closed at `2026-08-30T03:20:50Z` via `Closes #1334` in #1373 (`470fbd7`), re-echoed by the 0.52.0
CHANGELOG. The core render work did ship in #1373. The spec did not finish:

- `specs/AI-042-trusted-folders-render/proposal.md` is `status: implementing` (not `verifying`, not archived)
- **round-3 adversarial review verdict: FAIL** — 2 REAL Blockers, 2 REAL Majors (`review-round3.md`, 2026-08-29 03:31)
- three fix commits answering rounds 1-3 (`1c658eb`, `4e977df`, `2ab60a9`) are on `chore/archive-ai042`
  and **not in `origin/main`** (verified with `git merge-base --is-ancestor`)
- the review artifacts (`review-round{1,2,3}.md`, four `review-request*.json`) are **untracked** in the
  `dotfiles-wt-ai042` worktree — they exist only on this box

One of the Blockers has since fired for real and is now tracked separately as **#1390**: the
`agy-settings` `replace` entry destroyed agy's runtime-written state during the 2026-08-29 21:39 deploy
(5 `mcp(hive-vault/*)` grants and the `fae-onboarding-webapp` trusted workspace, all absent afterwards).

The second Blocker stays with this spec: the round-2 `UseNumber` switch silently removed trailing-data
rejection from both JSON paths the renderer touches, so `{"a":1} {"injected":true}` renders only the
[...]

Issue #1322, opened 2026-08-28
AI-039: ai/copilot/config.json carries a key the CLI does not read, misses the ones it does, and doctor cannot tell


mlorentedev:
## Problem
`copilot help config` (1.0.80) documents no `telemetry` key (offline mode is COPILOT_OFFLINE; telemetry is OTel env-driven), so `"telemetry": {"disabled": true}` is dead - the class #1307 just fixed for defaultModel. Documented keys absent: autoUpdate, powershellFlags, hooks/disableAllHooks, memory, contextTier. cli/internal/doctor/checks_deploy.go ~470 compares only copilot-instructions.md; tests/copilot-config.bats asserts four keys but not key validity. An unmanaged ~/.copilot/settings.json (2026-07-15) duplicates model/includeCoAuthoredBy with different extras - a second writer nothing owns.

## Fix
Drop `telemetry`; bats test that every top-level key is in the audited documented set (frozen list + CLI version in the test); add .copilot/config.json and .copilot/mcp-config.json to the doctor deploy-drift table; decide settings.json (delete or adopt). Effort S.

Source: Copilot audit F4, 10_projects/dotfiles/research/2026-08-27-windows-drive-audit-plan.md (knowledge vault)

mlorentedev:
Measured 2026-08-28 on the Windows work box (CLI 1.0.81): the deployed `~/.copilot/config.json` is REWRITTEN by the CLI with the header `// User settings belong in settings.json. // This file is managed automatically.` and holds only `telemetry`, `trustedFolders`, `firstLaunchAt`; the user's `~/.copilot/settings.json` holds `model`, `includeCoAuthoredBy`, `effortLevel`, `contextTier`, `allowedUrls`, `renderMarkdown`. So the repo's `ai/copilot/config.json` (model, includeCoAuthoredBy, and briefly autoUpdate in #1359, reverted) targets a file the CLI overwrites, and a doctor drift row on it would flap. Re-scope: the managed keys (model, includeCoAuthoredBy, autoUpdate=false, telemetry) go into a repo `ai/copilot/settings.json` deployed by MERGE (per-key, like Merge-ClaudeSettings / the dotf deploy merge kind, #1339) so per-box prefs (allowedUrls, effortLevel, contextTier) survive; stop copying config.json (or copy only trustedFolders rendered per #1334); doctor drifts the merged keys only; the bats key-validity test reads `copilot help config`'s documented set.

PR #1365, opened 2026-08-29
feat(deploy): ai/deploy.json learns merge and requires, and Copilot's JSON moves onto it

mlorentedev:
## Summary

Copilot's JSON is now deployed by `dotf deploy` on both OSes, and `ai/deploy.json` learns two fields to make that correct: `strategy: merge` (write the source's top-level keys into the installed JSON, preserve every other key, "in sync" means the managed keys already match) and `requires` (skip an entry, and say so, when its command is not on PATH).

Refs #1322 (AI-039). Spec: `specs/AI-039-copilot-settings-merge/`.

## Why

Measured on the Windows work box (Copilot CLI 1.0.81): the CLI **rewrites** `~/.copilot/config.json` on every launch (`// User settings belong in settings.json. // This file is managed automatically.`), keeping only `trustedFolders`/`firstLaunchAt`/`telemetry`; the keys it reads (`model`, `includeCoAuthoredBy`, ...) live in the user's `settings.json` next to per-box preferences (`allowedUrls`, `effortLevel`, `contextTier`, `renderMarkdown`). So `ai/copilot/config.json` targeted a file the CLI overwrote, a verbatim copy of a `settings.json` would have wiped the box's keys, `telemetry` is a key `copilot help config` does not document, and doctor compared only `copilot-instructions.md`.

## What changed

- `cli/internal/deploy`: `Config.Strategy` (`replace` default | `merge`), `Config.Requires`, `PlanConfig` (the read-only half: what a deploy would do, touching nothing). The compare for non-rendered entries now runs **before** staging — `Deploy(dryRun)` used to create the destination directory and a temp file before comparing, so a diagnostic built on it would have deployed by asking. Merge tolerates the CLI's `//` header on read and never rewrites an in-sync file (the header would otherwise churn every setup run).
- `ai/deploy.json`: `copilot-settings` (merge: `model`, `includeCoAuthoredBy=false`, `autoUpdate=false` — ADR-036, the catalog owns the version), `copilot-config` (merge: `trustedFolders` only), `copilot-mcp` (replace); all `requires: copilot`. `telemetry` leaves the repo file; existing boxes keep it (unmanaged key), fresh boxes never get it.
- `dotf deploy`: honours `requires` (`skipped   copilot-settings (copilot not installed)`), which keeps the integration guard "copilot config NOT deployed when the binary is absent" (#1312) true.
- `dotf doctor`: new section **Deployed agent configs (ai/deploy.json)** — every non-rendered, non-gated entry compared through `PlanConfig`; WARN names the entry and `dotf deploy <name>`; the PASS line says how many entries were not compared (rendered, or tool absent).
- `setup-linux.sh` / `setup-windows.ps1`: the `ai/copilot/*` glob copy becomes an explicit copy of `copilot-instructions.md`; the JSON is the manifest's.
- `tests/copilot-config.bats`: every top-level key of both repo files must be in the set `copilot help config` documents (frozen from 1.0.81, dated), plus the doctrine values, no `powershellFlags` (#1324 constraint), manifest shape, and the explicit copy in both setups.

## Test plan

- [x] `go build/vet` (host, `GOOS=linux`), `go test ./internal/deploy/ ./internal/cmd/ ./internal/doctor/`, `golangci-lint` 0 issues
- [x] Mutations: merge from an empty object → AC1 tests red; `MkdirAll` before the compare → AC2 test red; both restored
- [x] bats: copilot-config 11/11, env-contract, setup-linux, setup-windows; shellcheck + `bash -n`; `setup-windows.ps1` parse 0 errors, non-ASCII delta 0, CRLF intact
- [x] Box (Windows, 1.0.81): `dotf deploy --dry-run` → `copilot-config` and `copilot-mcp` already in sync (header tolerated), `copilot-settings` would deploy; `dotf deploy` → merged, the six per-box keys intact plus `autoUpdate`; second `dotf deploy` → five × `in sync`; `dotf doctor` → `[Deployed agent configs (ai/deploy.json)] (1 checks, all ok)`; `copilot -p` exit 0

## Out of scope

Rendering `trustedFolders` per machine (#1334), array union on merge (a managed key is repo-owned, whole value), moving Claude's settings merge onto this strategy (#1339 — now possible), doctor comparing rendered entries (needs the secrets daemon).


mlorentedev:
## Review triage

Reviewer output on `1b8f652` (CodeRabbit: 3 inline findings; PR-Agent guide: no major issues, no security concerns) and the follow-up `fix(deploy)` commit that applies finding 2. CI green on both.

|---|---|---|---|
| 1 | CodeRabbit (Major) | Replace the package-level `deployCommandAvailable` seam with a `Deps` struct injected into the command | **Declined.** The package-level seam is this package's established shape — `deployRenderer` on the line above it, `userEnvStore` in `env_persist.go`, `harness_mirror`'s seams — and the cmd tests are not parallel. A `Deps` struct for one PATH lookup would be a second injection pattern in the same file for no isolation this suite needs. |
| 2 | CodeRabbit (Minor) | A JSON `null` destination unmarshals into a nil map without error; the first managed-key assignment panics | **Applied** (`fix(deploy): a JSON null destination or source is rejected, not merged into`). Both sides are rejected as "not a JSON object: null" with the config named; `TestDeploy_MergeRejectsANonObjectDestinationByName` now covers `[1,2,3]`, `null` and `"text"` destinations and a `null` source. Real finding — thank you. |
| 3 | CodeRabbit (Major) | Convert `TestCheckDeployManifest_ByStatus` and the `requires` cases to table-driven subtests | **Declined.** Each subtest writes a different set of destination files and a different PATH before running the check; a table would carry per-case setup closures, which is what the `t.Run` blocks already are. The persisted-env test in the same package is table-driven because its cases share one setup. |
| 4 | PR-Agent | #1312 and #1324 "not compliant" | **Declined.** Auto-linked from mentions: #1312 is cited by the `requires` gate that keeps its integration guard true, #1324 by the `powershellFlags` constraint the bats test pins. Neither is a target of this PR (`Refs #1322`). |
| 5 | PR-Agent | "Decide settings.json (delete or adopt) — the decision is implicit" | **No change.** Adopted explicitly: `ai/copilot/settings.json` is a manifest entry, and the spec's *What* says why (the CLI reads it and co-owns it). |
| 6 | PR-Agent | Race: `mergeInto` reads the destination, then the rename lands; a concurrent CLI write in between is lost | **Declined.** The window is the milliseconds of a setup run, when the CLI is not writing; the atomic rename guarantees no torn file; and a lost concurrent write to a *managed* key is re-converged by the next `dotf deploy` by design (unmanaged keys are read fresh each run). Same model as every other entry in the manifest. |

Box evidence (Windows, Copilot 1.0.81): `dotf deploy --dry-run` → `copilot-config`/`copilot-mcp` already in sync (the CLI's `//` header tolerated), `copilot-settings` would deploy; `dotf deploy` → merged with the six per-box keys intact; second run → five × `in sync`; `dotf doctor` → `[Deployed agent configs (ai/deploy.json)] (1 checks, all ok)`; `copilot -p` exit 0.

PR #1367, opened 2026-08-29
fix(setup): Windows installs uv before registering the Claude MCP servers, so a first run registers hive and pdf-modifier

### LINKED ISSUE #1361: OPS-044: setup-windows.ps1 registers Claude MCP servers before uv is installed, so hive and pdf-modifier are skipped on a clean box's first run
## Problem
Exposed by #1360 (Windows now installs Claude Code, so the MCP block runs on the clean CI runner for the first time): the registration loop (setup-windows.ps1 ~453-540) runs in section 2, the uv installer runs in section ~968, so on a fresh box every MCP server whose prerequisite is `uv`/`uvx` is skipped with `MCP hive: prerequisite 'uv' not found, skipping` / `MCP pdf-modifier: prerequisite 'uv' not found, skipping` and the summary reads `2 added, 0 already present, 2 failed`. A second setup run registers them - the first run is not idempotent by construction. Linux: check whether setup-linux.sh orders uv before its MCP loop (~1095-1163) or shares the defect.

## Fix
Move the toolchain installers that MCP prerequisites depend on (uv, Node) ahead of the MCP registration on both OSes, or make the MCP block run after section 4; guard with a bats ordering assertion (installer line < MCP loop line) on both setups. Better long-term: `dotf agent claude sync` (#1339) owns ordering.

## Evidence
CI run for #1360, test-windows job: the three lines above, right after `[SUCCESS] Claude Code installed`.

mlorentedev:
## Summary

On Windows the Claude MCP registration block ran in section 2 and the `uv` installer in section 4, so on a clean box the first `setup-windows.ps1` run skipped the two `uvx` servers (`MCP hive: prerequisite 'uv' not found, skipping`, same for pdf-modifier) and only the second run converged. Measured on the CI runner the first time Claude Code installed there (#1360). The `uv` install block moves in front of the MCP registration; poetry stays in section 4 (it only needs `uv` to exist, which it now does). Linux has always installed `uv` first.

Closes #1361 (OPS-044).

## Guard

`tests/setup-windows.bats`: the `astral.sh/uv/install.ps1` invocation line must precede the first `claude mcp add` line, and the same for `setup-linux.sh` (parity). Measured on invocation lines, not mentions.

## Test plan

- [x] `setup-windows.ps1`: parse 0 errors, non-ASCII bytes unchanged (32), CRLF intact
- [x] bats subset (MCP / uv / poetry / OPS-044): 9/9, the two new guards included
- [x] CI `test-windows` runs setup on a clean runner: expect `Registering Claude Code MCP servers ... 4 added` instead of `2 added, 2 failed`

## SDD skip rationale

A block move (the uv installer ahead of the MCP registration) plus an ordering guard; no new contract. The gate counts 63 LOC because moving 28 lines counts twice.


mlorentedev:
## Review triage

Reviewer output on `4f5c9e2` and `a51a440` (CodeRabbit: 1 inline finding, then paused; PR-Agent guide updated to `a51a440`: no major issues, no security concerns, two focus areas). CI green.

|---|---|---|---|
| 1 | CodeRabbit (Minor) | The ordering guard's first `claude mcp add` text match is a comment 40 lines above the invocation, so the guard measured a mention | **Applied** (`a51a440`): both guards now match a non-comment line carrying `claude mcp add --transport` — the executable call, on both OSes. Real finding, and exactly the mention-vs-invocation class this repo keeps hunting; thank you. |
| 2 | PR-Agent | Guard is fragile: relies on `grep -n` line numbers | **Declined.** It compares the *relative order* of two invocations, never an absolute number; inserting comments, blank lines or servers shifts both and the assertion still holds. That is what an ordering guard is. |
| 3 | PR-Agent | The poetry block re-probes `Get-Command uv` after the moved block already set `$uvCmd` | **Declined.** The poetry block is untouched by this PR; its probe runs after the PATH refresh and reflects the session's actual state, which is the correct thing for a block that only needs `uv` to exist. Removing it would widen a move-only diff for no behaviour change. |
| 4 | PR-Agent | Node.js installer not moved ("the ticket says uv, Node") | **No change.** Node.js is installed by winget at `setup-windows.ps1:398`, already before the MCP block (now at ~542); only `uv` was out of order. Evidence: the CI runner's first-run log on #1360 skipped exactly the two `uvx` servers, none of the `npx` ones. |
| 5 | PR-Agent | #1360 "partially compliant" | **Declined.** Context reference, not a target; #1360 is merged. |

CI evidence: `test-windows` runs setup on a clean runner; the MCP registration summary is the measurement of this fix (`4 added, 0 failed` instead of `2 added, 2 failed`).

Runner evidence on `4a106b0` (the human's update-branch merge of main): 4 added, 0 already present, 0 failed
Registering Claude Code MCP servers from D:\a\dotfiles\dotfiles\mcp-servers.json...

PR #1284, opened 2026-08-27
chore(main): release 0.52.0

mlorentedev:
:robot: I have created a release *beep* *boop*
---

### Features

* **cli:** dotf vault health, byte-identical to the shell oracle (CLI-021 increment 2) ([#1317](https://github.com/mlorentedev/dotfiles/issues/1317)) ([7a06364](https://github.com/mlorentedev/dotfiles/commit/7a06364363d36d897500f3a71a6bd6e99dfbb21d))
* **deploy:** a 0600 file gets an owner-only ACL on Windows, and an in-sync file still gets its declared mode ([#1380](https://github.com/mlorentedev/dotfiles/issues/1380)) ([3f554ad](https://github.com/mlorentedev/dotfiles/commit/3f554ad73aa8253a09b5dc926abb26599de408c3))
* **deploy:** ai/deploy.json learns merge and requires, and Copilot's JSON moves onto it ([#1365](https://github.com/mlorentedev/dotfiles/issues/1365)) ([3261d04](https://github.com/mlorentedev/dotfiles/commit/3261d0431f82edd09fbf6c1d078b929b21558088))
* **deploy:** trust lists carry {HOME} and render per machine, in the form each tool is known to accept ([#1373](https://github.com/mlorentedev/dotfiles/issues/1373)) ([470fbd7](https://github.com/mlorentedev/dotfiles/commit/470fbd797a826a4bb3ee6aab99793d70c4fcdaa5)), closes [#1334](https://github.com/mlorentedev/dotfiles/issues/1334)
* **env:** dotf env persist records what it wrote and sweeps a name the contract retired, bounded to that record ([#1378](https://github.com/mlorentedev/dotfiles/issues/1378)) ([8de1cd1](https://github.com/mlorentedev/dotfiles/commit/8de1cd19962797cf3ba2164f5f43eec37800cef6))
* **env:** dotf env persist writes the contract variables where a profile-less process reads them ([#1362](https://github.com/mlorentedev/dotfiles/issues/1362)) ([389aaf5](https://github.com/mlorentedev/dotfiles/commit/389aaf50f3d0246b56b1e27871094694dbf44701))
* **harness:** dotf harness mirror, one deploy-dir mirror for both OSes ([#1305](https://github.com/mlorentedev/dotfiles/issues/1305)) ([acf78b3](https://github.com/mlorentedev/dotfiles/commit/acf78b3cc37633aeac0c9548ac32944fbe7459e5))
[...]

PR #1369, opened 2026-08-29
fix(deploy): the manifest is version 2, and a dotf that cannot fully read it refuses it

mlorentedev:
## Summary

`ai/deploy.json` becomes `version: 2` and `ParseManifest` refuses unknown fields, so a `dotf` that predates `strategy`/`requires` (#1365) refuses the manifest loudly instead of deploying every entry as **replace**.

Refs #1322 (AI-039); a defect of #1365 found after merge.

## Why — measured, not argued

`encoding/json` drops a key it has no struct field for. A binary built before #1365 therefore reads today's manifest with `strategy` and `requires` silently gone: `copilot-settings` and `copilot-config` become verbatim replaces, and the `requires: copilot` gate does not exist. On the Windows work box, with the stale `~/.local/bin/dotf.exe` (a `feat/env-persist` build) against `main`:

```
would deploy copilot-settings C:\Users\mlorente\.copilot\settings.json
would deploy copilot-config   C:\Users\mlorente\.copilot\config.json
```

That deploy would have overwritten `settings.json` — wiping the box's `allowedUrls`, `effortLevel`, `contextTier`, `renderMarkdown` — which is the exact loss AI-039 exists to prevent. And it is not one box: every box on the pinned release (0.51.0; 0.52.0 is #1284, unmerged) runs that old `dotf deploy` against the new manifest on its next setup.

## What changed

- `ai/deploy.json`: `"version": 2`, with the reason in `$comment` — bump it whenever a new field changes what an entry means, because a field an old decoder does not know is invisible to it and the version is not.
- `deploy.ParseManifest`: decodes with `DisallowUnknownFields` and requires `ManifestVersion` (2); both errors say what to update. `Manifest` gains the `$comment` field so the shipped documentation stays allowed.
- Tests: `TestParseManifest_RefusesWhatItCannotFullyRead` (older/newer schema, unknown entry and top-level fields, `$comment` allowed); every deploy-manifest fixture moves to v2.

## Test plan

- [x] `go build/vet`, `go test ./internal/deploy/ ./internal/cmd/ ./internal/doctor/`, `golangci-lint` 0 issues
- [x] The stale binary against this branch's manifest: refuses with `deploy manifest version 2 unsupported (this dotf reads 1 ...)` — exit non-zero, nothing written; setup's existing `dotf deploy failed` WARN is what an operator sees
- [x] A binary from this branch: all five entries `in sync` on the box
- [x] `bats tests/copilot-config.bats` green

## Follow-up

Release 0.52.0 (#1284) is what makes old boxes work again; until it ships, a box on 0.51.0 gets a loud refusal instead of a silent overwrite — the right of the two.

### DIFF
--- a/ai/deploy.json
+++ b/ai/deploy.json
@@ -28,6 +28,12 @@
     "            does not carry is a file nobody reads (#843), and the integration",
     "            guard asserts ~/.copilot never appears without copilot (#1312).",
     "",
+    "version 2 (AI-039): strategy and requires change what an entry MEANS, so a",
+    "dotf that predates them refuses this file instead of deploying every entry",
+    "as replace - which would overwrite ~/.copilot/settings.json and wipe the box's",
+    "own keys. Bump it again whenever a new field changes meaning: a field an old",
+    "decoder does not know is invisible to it, the version is not.",
+    "",
     "Copilot's destinations are spelled {HOME}/.copilot rather than {COPILOT_HOME}:",
     "`dotf deploy` runs before `dotf env set`/`generate` in both setups, so on a",
     "first run machine.json — which {COPILOT_HOME} needs to resolve — does not",
@@ -38,7 +44,7 @@
     "{env:VAR} itself rather than assuming it, which is the assumption that made",
     "#987 expensive."
   ],
-  "version": 1,
+  "version": 2,
   "configs": [
     {
       "name": "pi",
--- a/cli/internal/cmd/deploy_requires_test.go
+++ b/cli/internal/cmd/deploy_requires_test.go
@@ -15,7 +15,7 @@ import (
 func TestDeployCmd_SkipsAnEntryWhoseRequiredCommandIsAbsent(t *testing.T) {
 	repo, home := t.TempDir(), t.TempDir()
 	writeMirrorFixture(t, filepath.Join(repo, "ai", "deploy.json"), `{
-  "version": 1,
+  "version": 2,
   "configs": [
     {"name": "always", "src": "ai/always.json", "dst": "{HOME}/.always/config.json", "render": false, "mode": "0644"},
     {"name": "gated", "src": "ai/gated.json", "dst": "{HOME}/.gated/settings.json", "render": false, "mode": "0644", "strategy": "merge", "requires": "gatedtool"}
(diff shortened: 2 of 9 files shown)


mlorentedev:
## Review triage

Reviewer output on `6f11018`, the follow-up `test(deploy)` commit (`7087105`) and the human's update-branch merge of main (`3b608f4`): PR-Agent guide — no major issues, no security concerns, two focus areas; CodeRabbit: two inline findings on `6f11018`, both applied. CI green on `0d5dd87`.

|---|---|---|---|
| 1 | CodeRabbit (Major) | `json.Decoder.Decode` reads one value and stops, so `{...}{...}` deployed the first document and hid the second | **Applied** (`24097ce`): a second `Decode` must return `io.EOF`; regression case `trailing document` in `TestParseManifest_RefusesWhatItCannotFullyRead`. The first push of that fix went red on my own test (the error lacked the remedy text every manifest error carries) and `0d5dd87` corrected the message — the red run was mine, not the finding's. |
| 2 | CodeRabbit (Major) | `ParseManifest` over 40 lines | **Applied** (`24097ce`): strict decoding extracted into `decodeManifest`; `ParseManifest` is the entry validation alone. |
| 3 | PR-Agent | Maintenance: the version must be bumped by hand when a future field changes semantics; nothing enforces it | **Applied.** `TestManifestVersion_FreezesTheFieldSet` freezes the json field names of `Manifest` and `Config` per `ManifestVersion`; adding a field fails it until the version moves and the frozen set is extended — the moment to ask whether an old reader would misread the field. |
| 4 | PR-Agent | "Potential bug": `DisallowUnknownFields` errors are generic for unknown fields | **Declined.** The decoder's error names the field (`json: unknown field "future"`), and the wrapper adds what to do ("a field this dotf does not know? rebuild or update dotf"); `TestParseManifest_RefusesWhatItCannotFullyRead` asserts both the field name and the remedy appear, for entry-level and top-level unknowns. |
| 5 | PR-Agent | #1365 / #1322 / #1284 "partially compliant" | **Declined.** Auto-linked from mentions: #1365 is the merged PR this defect belongs to, #1322 its ticket (`Refs`), #1284 the release note that makes old boxes work again. None is a target of this fix. |

Measurement behind the PR (Windows work box): the stale `feat/env-persist` build against `main` → `would deploy copilot-settings`, `would deploy copilot-config`; against this branch → `Error: deploy manifest version 2 unsupported (want 1)`, nothing written. A `main` build against this branch → five × `in sync`. Every box on release 0.51.0 is in the same position until 0.52.0 (#1284) ships — refusing loudly is the right of the two behaviours until then.

PR #1372, opened 2026-08-29
feat(spec): dotf spec review draws one reviewer at random from a pool of five

mlorentedev:
## Summary

`dotf spec review` with no `--reviewer` now draws one pool member uniformly at random and says so on the launch line; `--reviewer <id>` still names one deliberately (how a run is reproduced). `nan/glm5.3-flash` and `nan/qwen3.8-flash` join the pool — both `reasoning: true` with 1M context in `ai/pi/models.json`. Owner decision 2026-08-29.

Refs #1370 (HARNESS-093). Spec: `specs/HARNESS-093-reviewer-pool-random/`.

## Why

Every review landed in one NaN per-model concurrency bucket (`model-map.json` had already moved PR-Agent away from it), and the fallbacks — the provider-diverse agy arm included — were exercised only by hand. A random draw spreads the buckets and exercises every member in the ordinary course of reviews. Note what it does not change: the local CPU load a review causes (it runs go test, bats and check-doc-paths on the box) is independent of the model.

## What changed

- `spec.DrawReviewer(entries, draw)`; `ResolveReviewer` refuses an empty name (the launcher draws instead).
- `dotf spec review`: `reviewerDraw = rand.IntN` seam; launch line prints `(pi, random draw)` / `(pi, requested)` instead of the pool role; `--reviewer` help and Long text updated.
- `harness/reviewer-pool.json`: two members (role `member`); their `why` states what is established (reasoning-class by profile) and what is not (no calibration against a review of known outcome yet); `$comment` says order is provenance, not precedence.
- `tests/reviewer-pool.bats`: unique ids, ≥4 members incl. the four NaN models, every `pi` member `reasoning: true` in `ai/pi/models.json`, launcher + skill say "drawn at random".
- `adversarial-review` skill usage line (committed record + vault source) says a member is drawn.

## Test plan

- [x] `go test ./internal/spec/ ./internal/cmd/`, `golangci-lint` 0 issues; existing spec-review tests pinned to draw index 0 by an `init` in the new test file
- [x] `bats tests/reviewer-pool.bats` 4/4
- [x] Box: six `--dry-run` launches drew `agy/gemini-3.1-pro-high` ×1, `nan/glm5.3-flash` ×2, `nan/qwen3.8-flash` ×3; `--reviewer nan/glm5.3-flash` → `(pi, requested)`, command carries `'--model' 'glm5.3-flash'`

## Follow-ups

[...]

PR #1376, opened 2026-08-29
chore(spec): archive HARNESS-093 after its review passed under the random draw

### LINKED ISSUE #1370: HARNESS-093: the reviewer pool grows to four NaN reasoning models and dotf spec review picks one at random
## Problem

`harness/reviewer-pool.json` holds two members (`nan/deepseek-v4-flash` primary, `nan/mimo-v2.5` fallback) and `dotf spec review` always runs the primary unless `--reviewer` says otherwise. Every review therefore lands in one NaN rate bucket (the one `model-map.json` already moved PR-Agent away from), and the fallback is exercised only by hand. Two more reasoning-class NaN models are already in `ai/pi/models.json` (`glm5.3-flash`, `qwen3.8-flash`, both `reasoning: true`, 1M context) and unused by the gate.

## Decision (owner, 2026-08-29)

Pure random selection among the pool by default; both models join now. The chosen id is already recorded in `review.md` (`reviewer:`) and `review-request.json`, and `--reviewer <id>` remains the way to reproduce a run on a specific member.

## Fix

- `harness/reviewer-pool.json`: add `nan/glm5.3-flash` and `nan/qwen3.8-flash`; the pool array is the archive gate's allow-list, so both become signable.
- `dotf spec review`: default = one pool member at random (seeded from time; the launch line prints which); `--reviewer <id>` unchanged.
- Guard: every pool id must exist in `ai/pi/models.json` with `reasoning: true` (bats), so a latency-only daily model cannot slip into the gate by edit.
- Vault skill `adversarial-review`: the usage line says "a pool member chosen at random" instead of "the pool's primary".

Note: the local CPU load a review causes (it runs go test, bats and check-doc-paths on the box) is independent of the model; this spreads the API buckets, not the box.

mlorentedev:
## Summary

Archives `HARNESS-093-reviewer-pool-random` (#1372, merged as `acca2da`) after an independent adversarial review — the **first review drawn under the random policy the spec itself introduced**: the draw landed on `nan/glm5.3-flash`, one of the two members admitted uncalibrated by #1372.

**Verdict: PASS**, two Minors, both REAL on inspection:

1. `features.json` f4's recorded evidence command piped the dry-run's stdout into `grep '^Reviewer:'`, but the launch line is written to **stderr** — so the pipeline counted 0 lines and `grep -qvE '^1$'` matched vacuously. A verification that passes when the thing it verifies does not happen. Reproduced here (`2>/dev/null` → 0 lines; `2>&1 >/dev/null` → 1). Fixed in the archived copy, the way HARNESS-092's stale line was: `2>&1` added and the assertion made *at least two distinct reviewers* (`awk 'END { exit !(NR >= 2) }'`). Re-run live: six draws, ≥2 distinct, exit 0.
2. The deployed generated skill record on the box still reads "the pool's primary reviewer" — already recorded as deferred to `compile-harness.sh --refresh` on a Linux box (ENGINE-001); nothing to do here.

The archive gate refused the first attempt — `review.md is stale: features.json has uncommitted changes` — which is the gate working: the tree was restored, archived pristine, and the fix applied to the archived copy afterwards.

**Calibration data point for `glm5.3-flash`** (the handoff recorded it as not done): it read the diff rather than the prose — it checked that `--dry-run` returns before `WriteReviewRequest` because AC4's evidence consists of dry-runs against the real spec, and it caught an evidence-command defect the author (me) and two PR reviewers missed. Substantive.

Closes #1370.

## SDD skip rationale

Spec archive plus one evidence-command fix inside the archived spec; no production code.


mlorentedev:
## Review triage

|---|---|---|
| PR-Agent (`github-actions[bot]`) | #1370 compliance: every requirement met, nothing non-compliant; no security concerns. | Nothing to apply. |
| CodeRabbit | "Review limit reached" — a notice that no review ran, not a review. | **Unreviewed by CodeRabbit**, disclosed. Proceeding on PR-Agent's pass plus the independent adversarial review the PR archives (PASS, `nan/glm5.3-flash`). |

Issue #1363, opened 2026-08-29
CLI-065: dotf env persist never sweeps a variable the contract has retired, so HKCU\Environment keeps stale values forever


mlorentedev:
## Problem

`dotf env persist` (#1324, PR #1362) writes every variable the contract names into `HKCU\Environment` and never removes one. When a variable is retired from `env-contract.json` (renamed, dropped, or moved out of the persisted scope), its User-scope value stays behind on every box that ever ran setup: profile-less processes (Copilot's `pwsh -NoProfile` tool calls, Scheduled Tasks) keep reading a value the contract no longer defines, and `dotf env persist --check` / `dotf doctor` report clean because they only look at what the contract names today.

Same class as WIN-013 (#1310): a deployer that adds and never sweeps. Surfaced by CodeRabbit's merge-risk note on #1362 ("absence of cleanup for retired variables"); deliberately not built there.

## Fix

Give the persisted scope an ownership marker so the sweep is bounded to what dotf wrote, never to arbitrary user variables:

- On persist, record the managed names (e.g. a `DOTF_MANAGED_ENV` value in `HKCU\Environment` holding the semicolon-joined list, or the same list in `machine.json`).
- On the next persist, delete every name in the marker that the contract no longer resolves, then rewrite the marker. `--check` reports such leftovers as drift.
- `dotf doctor` "Persisted environment (user scope)" WARNs on leftovers with the remedy (`dotf env persist`).
- Test with the fake `UserEnvStore` (retire one name between two runs; the second run deletes it and only it). Off Windows stays a no-op.

## Acceptance

- Retiring a name from the contract and re-running setup removes it from `HKCU\Environment`; a variable dotf never wrote is never touched.
- `dotf env persist --check` exits non-zero while a retired name is still persisted.

PR #1378, opened 2026-08-29
feat(env): dotf env persist records what it wrote and sweeps a name the contract retired, bounded to that record

mlorentedev:
## Summary

`dotf env persist` (CLI-058) wrote every contract variable into `HKCU\Environment` and never removed one, so a name retired from `env-contract.json` stayed on every box forever — inherited by exactly the profile-less processes the scope exists for. Same class as WIN-013: a deployer that adds and never sweeps. Spec: `specs/CLI-065-env-persist-sweep/`.

- **Ownership marker.** `persist` records the names it wrote in `DOTF_MANAGED_ENV` (sorted, `;`-joined) and on every run deletes each name the marker lists that the contract no longer names — then rewrites the marker. The sweep is bounded to the marker: a variable dotf never wrote is never touched, whatever its name. A box with no marker yet deletes nothing and writes one (no record → no guess).
- **`--check`** prints `retired: NAME` and exits non-zero while a leftover is persisted; **`persist`** prints `removed NAME` and counts removals; **doctor**'s *Persisted environment (user scope)* row WARNs on leftovers with the remedy.
- **Order and comparison are load-bearing.** Registry value names are case-insensitive: a case-only rename is one value to the store, so `Leftovers` compares with `EqualFold` (not a leftover at all) and every delete precedes every write (pinned by a test on the fake's operation log). Lesson 244.
- **`Delete` of an absent name succeeds**, declared on the interface, mapped from `registry.ErrNotExist`, mirrored by the fake and pinned — the one behaviour where local tests and the box could disagree.
- `UserEnvReader` split out of `UserEnvStore`; the doctor adapter loses the `Set` no-op it carried to satisfy an interface wider than its caller.

Refs #1363 (closed by the archive PR after the independent review).

## Verification

- `go test ./...` all ok, `FAIL_COUNT=0`; `go vet` clean under `GOOS=windows` and `GOOS=linux`; `golangci-lint run` (pinned 2.12.2) 0 issues.
- **Windows work box, binary from this branch** — real contract → marker with the 11 contract names; scratch contract minus `SCRIPTS_DIR` → `--check`: `retired: SCRIPTS_DIR`, exit 1; `persist`: `removed SCRIPTS_DIR (retired from the contract)`, `Get-ItemProperty HKCU:\Environment SCRIPTS_DIR` → absent; second run `0 changed, 11 unchanged, 0 removed`; real contract again → `persisted SCRIPTS_DIR`, value restored. Full transcript in `verification.md`.


mlorentedev:
## Review triage

| Reviewer | Finding | Disposition |
|---|---|---|
| CodeRabbit `cli/internal/cmd/env_persist.go:62` (Major, function length) | `newEnvPersistCmd` spans 83 lines, `RunE` > 10 decisions. | **Applied** in `8448d9c`: `resolvePersistVars`, `checkPersisted`, `runPersist`. |
| CodeRabbit `cli/internal/env/persist.go:65` (Major, data integrity) | A name containing `;` is stored as one and read back as two; the next run could delete two unrelated values. | **Applied**: `Persist` refuses a contract name holding the separator, naming it, before any write; `TestPersist_RefusesASeparatorInAName` keeps foreign `A` and `B` untouched. |
| CodeRabbit `cli/internal/cmd/env_persist.go:74` (Minor) | `--check` said a hand-removed name was "still persisted". | **Applied**: `Retired` lists only leftovers still in the store; a marker-only stale name is a stale record, reported by the new `MarkerStale` (`record: DOTF_MANAGED_ENV does not match the contract`, non-zero) so `--check` stays the exact mirror of a run; `Persist` neither deletes nor reports it, and rewrites the record. Doctor gained the matching WARN. Tests: `TestRetired`, `TestMarkerStale`, `TestPersist_SkipsALeftoverAlreadyGone`. |
| CodeRabbit lesson 244:55 (Minor) | Spec path pointed at the archive location before the move. | **Applied**: current path, with the move noted. |
| CodeRabbit `verification.md:25` (Minor, MD040) | Fence without a language. | **Applied**: `text`. |
| CodeRabbit `features.json:40` (Minor) | AC6's command should perform the sweep's state transitions. | **Declined with reason, improved**: a routine verification command must not retire a name from the user's real `HKCU\Environment`; the transitions are the box transcript in `verification.md`. The command now asserts the post-state honestly — `dotf env persist --check` (no leftovers, record current) plus the marker naming `SCRIPTS_DIR` — instead of the marker grep alone. |
| PR-Agent | #1363 fully compliant; no issues. | Nothing to apply. |

CI green on `44a14f3`; re-running on `8448d9c` (rebased on the owner's merge from `main`).

Issue #1364, opened 2026-08-29
CLI-066: doctor measures the profile its heuristic finds but --fix heals PowerShell's $PROFILE, so a redirected Documents root splits detect from heal


mlorentedev:
## Problem

`dotf doctor` (CLI-064, #1353) decides whether the PowerShell profile is corrupt from the file its own `findPowerShellProfile` enumerates — four hardcoded roots (`~/Documents/{PowerShell,WindowsPowerShell}`, `~/OneDrive/Documents/{...}`) — while `--fix` runs `profile-heal.ps1`, which resolves `$PROFILE` itself and heals that. On a box whose Documents are redirected (the corporate `OneDrive - <Org>` root this very machine carries on its PATH), the two paths differ: doctor can report the profile healthy while the one PowerShell loads is corrupt (the false-PASS CLI-064 exists to kill), or heal `$PROFILE` and then re-measure a different file and print "ran but still corrupted".

Raised by the adversarial review of CLI-064 (Major, THEORETICAL — not reproduced; `specs/archive/CLI-064-doctor-profile-heal/review.md`, finding 1).

## Fix

One source of truth for the target: doctor asks PowerShell for `$PROFILE` (`pwsh -NoProfile -Command '$PROFILE'` through the existing `sys.CommandOutput` seam, bounded) and passes that path to the heal, or the heal takes the path as a parameter and doctor measures the same file it handed over. Drop the root enumeration or keep it only as the fallback when pwsh is absent. Test: a fake System whose `$PROFILE` is outside the four roots — doctor must measure and heal that file, and the "ran but still corrupted" branch must read it back.

## Also from the same review (Minor, THEORETICAL)

- Go counts marker pairs on raw bytes, the script via `[regex]::Matches` — the two thresholds (`>1`, `-gt 1`) are unlinked literals; one test should read the other's constant.
- `repairProfile` shells out through unbounded `sys.CommandOutput`; other checks use `CommandOutputBounded`.
- `--fix` replaces the whole profile from the SSOT, discarding user content outside the markers — document, or preserve the outside-marker tail.

PR #1379, opened 2026-08-29
fix(doctor): the profile doctor measures is the one pwsh names, and the heal is told which file

mlorentedev:
## Summary

`dotf doctor` decided whether the PowerShell profile was corrupted from the file its own four-root enumeration found, while `--fix` ran `profile-heal.ps1`, which resolves `$PROFILE` itself and heals **that**. On a box whose Documents folder is redirected, doctor measured one file (or none) and the heal rewrote another. Raised by CLI-064's adversarial review (Major, theoretical); this closes it and the three Minors from the same review. Spec: `specs/CLI-066-doctor-profile-target/`.

- **One source of truth for the target.** Doctor asks `pwsh -NoProfile -Command '$PROFILE'` through the bounded seam (10 s) and measures that file. The enumeration survives only as the fallback for a box without pwsh, or a pwsh that does not answer — and the row says which of the two produced the path.
- **The heal is told which file.** `profile-heal.ps1 -ProfilePath <the measured file>`; without the parameter it targets `$PROFILE` as before.
- **Thresholds linked, not duplicated.** `TestProfileHealThresholdsMatchTheScript` reads the script and pins its `-gt 1MB` / `-gt 1` literals beside doctor's constants (Minor 1).
- **The heal runs bounded** (60 s), like every other doctor probe; the test harness makes the unbounded seam a `t.Fatalf` so it cannot quietly return (Minor 2).
- **What `--fix` does is said where it is offered:** content outside the dotfiles markers survives only in the backup (Minor 3, documented rather than built).

Refs #1364 (closed by the archive PR after the independent review).

## Verification

- `go test ./...` all ok, `FAIL_COUNT=0`; `go vet` clean under `GOOS=windows` and `GOOS=linux`; `golangci-lint run` (pinned 2.12.2) 0 issues; `bats tests/profile-heal-ps1.bats` 16/16; PSScriptAnalyzer 0 findings, 0 non-ASCII.
- **Windows work box, binary from this branch:** `dotf doctor --verbose` → `[ OK ] PowerShell profile exists (…\Documents\PowerShell\Microsoft.PowerShell_profile.ps1; resolved by pwsh $PROFILE)`; `profile-heal.ps1 -ProfilePath <scratch copy with two marker pairs>` → healed (2 → 1 pairs, backup beside it) while the real profile's SHA-256 was identical before and after. Transcript in `verification.md`.

### DIFF
--- a/cli/internal/doctor/checks_profile.go
+++ b/cli/internal/doctor/checks_profile.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 )
 
 // The dotfiles section of the PowerShell profile is delimited by these two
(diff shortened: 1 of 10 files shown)


mlorentedev:
## Review triage

| Reviewer | Finding | Disposition |
|---|---|---|
| CodeRabbit `proposal.md:34` (Minor) | The fallback wording said "when pwsh is not on PATH"; the code and tests also fall back when pwsh fails or times out. | **Applied** in `b90a844`: What § and AC2 say *unavailable or unresponsive — not on PATH, failing, or over the deadline — with the reason in the row*. |
| CodeRabbit `tasks.md:18` (Minor) | `[AC1][AC2]` reads as a Markdown reference link (markdownlint MD052). | **Applied**: markers written `[AC1] [AC2]`; the same fix went to CLI-055's and OPS-042's tasks. |
| PR-Agent (`github-actions[bot]`) | #1364 compliant on every listed requirement; no issues, no security concerns. | Nothing to apply. |

CI green on `f6a3802`; re-running on `b90a844` (spec prose only).

Issue #1302, opened 2026-08-27
CLI-055: deploy mode 0600 is a no-op on Windows, so a rendered credential file is protected only by inherited ACLs


mlorentedev:
## Problem

`ai/deploy.json` declares `"mode": "0600"` for pi's `models.json` (*DECLARED, never inferred*, #987), applied via `os.Chmod` at `cli/internal/deploy/deploy.go:197,225` and `cli/internal/secrets/render.go:185`. On Windows `os.Chmod` only toggles the read-only attribute; it cannot express an ACL. A rendered file that materialises a plaintext credential is protected solely by the inherited `%USERPROFILE%` ACL, and the declaration silently does not bind.

## Fix options

1. Windows branch: apply an owner-only DACL (`golang.org/x/sys/windows`, already an indirect dependency) when mode has no group/other bits.
2. Document the caveat in `ai/deploy.json` and doctor-WARN on Windows when a `render: true` config declares 0600.

Option 1 is the honest one; option 2 is the floor if 1 is deferred.

Related: #987, CLI-039.

mlorentedev:
Shipped in #1380: a 0600 render gets a protected owner+SYSTEM DACL on Windows, in-sync content still gets its mode, and the deploy reports `mode fixed`. Merged 2026-08-29. Closing as done.

PR #1380, opened 2026-08-29
feat(deploy): a 0600 file gets an owner-only ACL on Windows, and an in-sync file still gets its declared mode

mlorentedev:
## Summary

`ai/deploy.json` declares `mode: 0600` for pi's `models.json` and the secrets registry exposes three credential files at `0600`; all of them reached disk through `os.Chmod`, which on Windows toggles the read-only attribute and nothing else. A rendered credential under `%USERPROFILE%` was protected by whatever ACL its directory handed down. Spec: `specs/CLI-055-owner-only-acl/`.

- **`cli/internal/fsmode`** owns "apply this mode": `os.Chmod` everywhere, and on Windows, for an owner-only mode (`mode & 0o077 == 0`), a **protected DACL** of the calling user (from the process token — CI's service account gets the same path) and LocalSystem, inheritance cut. A mode with group/other bits leaves the inherited ACL alone.
- The three writers — `deploy.stage`, `deploy.commit`, `secrets.AtomicWriteMode` — call `fsmode.Apply`; no `os.Chmod(` remains in them.
- **Found on the box, fixed in the same PR:** the first run said `in sync` and left the inherited ACL in place — the in-sync path compared bytes only. `Deploy` now asks `fsmode.Needs` on that path and applies the declared mode without a content rewrite, reported as `mode fixed` (`would fix mode` on `--dry-run`).
- Windows tests verify through `icacls`, the tool an administrator would use, plus `SE_DACL_PROTECTED` from the security descriptor.

Refs #1302 (closed by the archive PR after the independent review).

## Verification

- `go test ./...` on the Windows box (so the Windows-only tests ran for real), `FAIL_COUNT=0`; `go vet` clean under `GOOS=windows` and `GOOS=linux`; `golangci-lint run` (pinned 2.12.2) 0 issues.
- **Windows work box, binary from this branch:** `dotf deploy pi --dry-run` → `would fix mode`; `dotf deploy pi` → `mode fixed`; `icacls ~/.pi/agent/models.json` → `TDY\<user>:(F)` and `NT AUTHORITY\SYSTEM:(F)` only, no `(I)`; second run `in sync`; a 0644 neighbour keeps its 3 inherited entries. Transcript in `verification.md`.

### DIFF
--- a/cli/internal/cmd/deploy.go
+++ b/cli/internal/cmd/deploy.go
@@ -96,6 +96,12 @@ func newDeployCmd() *cobra.Command {
 				switch {
 				case !res.Changed:
 					_, _ = fmt.Fprintf(w, "in sync   %-10s %s\n", res.Name, res.Dst)
+				case res.ModeFixed && dryRun:
+					_, _ = fmt.Fprintf(w, "would fix mode %-5s %s\n", res.Name, res.Dst)
+				case res.ModeFixed:
+					// Content was in sync; only the declared mode was missing on
+					// the file (CLI-055: an inherited ACL on a 0600).
+					_, _ = fmt.Fprintf(w, "mode fixed %-9s %s\n", res.Name, res.Dst)
 				case dryRun:
 					_, _ = fmt.Fprintf(w, "would deploy %-7s %s\n", res.Name, res.Dst)
 				default:
--- a/cli/internal/deploy/deploy.go
+++ b/cli/internal/deploy/deploy.go
@@ -22,6 +22,8 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+
+	"github.com/mlorentedev/dotfiles/cli/internal/fsmode"
 )
 
 // ManifestRel is the declarative table of what gets deployed where, relative to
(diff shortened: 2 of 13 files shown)


mlorentedev:
## Review triage

|---|---|---|
| PR-Agent (`github-actions[bot]`) | #1302 "partially compliant": *option 2 not implemented*. No issues, no security concerns. | **Declined with reason**: the ticket offered option 1 (owner-only DACL) as the honest fix and option 2 (document + doctor WARN) as the floor if 1 were deferred; option 1 is built, so 2 has nothing left to cover. |
| CodeRabbit | "Review limit reached" — a notice that no review ran, not a review. | **Unreviewed by CodeRabbit**, disclosed. Proceeding on PR-Agent's pass, the Windows-native test evidence (`icacls`, `SE_DACL_PROTECTED`) and the box transcript in the spec. |

Follow-up commit on this branch: the `[AC1][AC2]` markers in `tasks.md` are written `[AC1] [AC2]` so markdownlint stops reading them as reference links (CodeRabbit's finding on the sibling PR #1379, applied here pre-emptively).

CI green on `bc0bc37`.

Issue #1381, opened 2026-08-29
CLI-067: dotf tools reads the catalog from the deploy mirror only, doctor reads the checkout first — one resolver for packages.json


mlorentedev:
## Problem

`dotf tools list` / `dotf tools install` resolve `packages.json` from the deploy mirror only — `filepath.Join(env.DotfilesDir(env.Home()), "packages.json")` at `cli/internal/cmd/tools.go:51` and `:126` — while `dotf doctor` resolves it checkout-first per ADR-030 (`cli/internal/doctor/checks_catalog.go` `loadCatalog`: repo, then `DotfilesDir`). A box whose mirror lags the checkout gets two answers to "what is in the catalog".

Measured on the Windows work box while verifying OPS-042 (#1336): the mirror `~/.dotfiles/packages.json` predated #1359, so `dotf tools list` run from a worktree showed `sops`, `opencode`, `bw` — no `copilot`, no `obsidian`, no `yarn` — and `dotf tools install` converged nothing new, while `dotf doctor` in the same worktree read the checkout and reported `yarn version matches packages.json (1.22.22)`. Setup copies `packages.json` into the mirror before running `dotf tools install`, so the production path is consistent; every other invocation (a developer testing a catalog change, doctor's own remedy text `run dotf tools install <name>`) is not.

## Fix

One resolver for the catalog path, shared by `tools` and `doctor` — the ADR-030 cascade `env.ResolveRepoFirst("packages.json", DOTFILES_REPO_DIR, DotfilesDir, cwd)` that `env-contract.json` already uses (`env.ResolveContractPath`). `tools` gains the same checkout-first behaviour; the mirror remains the fallback for a box without a checkout. A test that runs `tools list` from a temp checkout whose mirror holds a different catalog and asserts the checkout wins.

Effort S. Related: ADR-030, #1336 (where it was found), #1359.

Issue #1336, opened 2026-08-28
OPS-042: obsidian-cli and yarn still carry per-OS npm install blocks and a hand-rolled PowerShell version parser - move them to packages.json


mlorentedev:
## Problem
setup-linux.sh ~651-670 / setup-windows.ps1 ~1039-1073 (obsidian-cli; the W comment still names @vorillaz/...), ~747-774 / ~1136-1171 (yarn, with a PS versions.conf parser at W1140-1148 and Test-VersionAtLeast at W90). ADR-036 puts node-distributed tools in packages.json and says version detection happens once in Go (`dotf tools version`).

## Fix
Two npm entries in packages.json, delete the four blocks and the PS parser; pi follows once installNpm takes --prefix/--ignore-scripts (#1294 follow-up). Effort S.

Source: scripts-parity audit F9, 10_projects/dotfiles/research/2026-08-27-windows-drive-audit-plan.md (knowledge vault)

PR #1382, opened 2026-08-29
chore(tools): obsidian-cli and yarn are packages.json tools, and the four setup blocks that installed them are gone

mlorentedev:
## Summary

ADR-036 puts node-distributed tools in `packages.json`, where `dotf tools install` — already run by both setups before these blocks — pins, installs and upgrades them the same way on every OS and `dotf doctor` reports them against the pin. `obsidian-cli` and `yarn` never moved: four hand-written npm blocks, a PowerShell parser re-reading `versions.conf`, and a Windows comment still naming a package that 404s. Spec: `specs/OPS-042-npm-tools-to-catalog/`.

- Two catalog entries: `obsidian` (`obsidian-cli@0.5.1` — the binary is `obsidian`, which is what the installer probes) and `yarn` (`yarn@1.22.22`).
- 125 setup lines deleted (Linux 50, Windows 73 incl. the `versions.conf` parser); `Test-VersionAtLeast` stays for pi. `bash -n` and PSScriptAnalyzer clean.
- `versions.conf` loses `OBSIDIAN_VERSION` (1.12.4 was the Obsidian *app*'s version; the npm package is at 0.5.1 and nothing read the pin) and `YARN_VERSION`; doctor's yarn row reads the pin from `packages.json` through `catalogPin`, like copilot's and opencode's.
- The bats that asserted the npm blocks now assert the catalog entries and their absence.

Found while verifying, ticketed as #1381 rather than folded in: `dotf tools` resolves `packages.json` from the deploy mirror only, doctor checkout-first.

Refs #1336 (closed by the archive PR after the independent review).

## Verification

- `go test ./...` `FAIL_COUNT=0`; `go vet` clean under `GOOS=windows` and `GOOS=linux`; `golangci-lint run` (pinned) 0 issues; bats (`setup-linux`, `setup-windows`, `packages-json`) 9/9 on the touched tests.
- **Windows work box, binary from this branch with the branch catalog as the mirror:** `dotf tools list` → `obsidian 0.5.1 full npm:obsidian-cli`, `yarn 1.22.22 full npm:yarn`; `dotf tools install` → both `already installed; skipping`; `dotf doctor --verbose` → `[ OK ] yarn version matches packages.json (1.22.22)`.


mlorentedev:
## Review triage

|---|---|---|
| PR-Agent (`github-actions[bot]`) | #1336 compliant on every requirement, nothing non-compliant; #1381 "partially compliant" with nothing listed either way. | Nothing to apply. #1381 is referenced (found here), not addressed — by design; its fix is a separate PR. |
| CodeRabbit | "Review limit reached" — a notice that no review ran, not a review. | **Unreviewed by CodeRabbit**, disclosed. Proceeding on PR-Agent's pass, the 9/9 bats, and the box transcript in the spec (`dotf tools list/install` and the doctor row). |

CI green on `ab15307`.

Issue #1080, opened 2026-08-19
HARNESS: Claude Code discovers none of the 41 deployed skills, and nothing checks that it does


mlorentedev:
41 skills are deployed to `~/.claude/skills/<name>/SKILL.md` by `scripts/compile-harness.sh --deploy`. **Claude Code discovers none of them.** Confirmed in a brand-new session, so this is not session staleness.

Nobody noticed because the workaround is invisible: every consumer — me included, in this session — reads `SKILL.md` with `cat` and follows the checklist by hand. That works, so the broken path is never exercised. I ran the full `handoff` checklist tonight without once invoking it as a skill.

## The contrast, which is the finding

A skill Claude Code DOES discover (Anthropic-shipped, symlinked from `~/.agents/skills/`) has exactly two frontmatter keys:

```yaml
---
name: find-skills
description: Helps users discover and install agent skills...
---
```

One of ours, which it does not:

```yaml
---
generated: true
generated_from: 00_meta/skills/handoff/SKILL.md
generated_sha: 1fbb507bda6c5b69
id: handoff-skill
type: skill
status: active
created: '2026-05-31'
owner: manu
name: handoff
description: Run a complete, standardized session handoff...
allowed-tools: [Bash, Read, Edit, Write, mcp__hive__vault_query, ...]
keywords: [handoff, wrap up, session handoff, ...]
paths: ['**/MEMORY.md', sessions/**]
---
```

Eleven extra keys, and `allowed-tools` is a **YAML sequence** where the documented form appears to be a comma-separated **string**. **0 of 41 use the string form** — all were emitted by the same generator, so whatever is wrong is wrong 41 times identically.

## Ruled out, with the measurement

| Hypothesis | Result |
|---|---|
| Files absent / wrong dir | 41/41 present with `SKILL.md` |
| Invalid YAML | 41/41 parse |
| Missing `name` or `description` | 41/41 have both |
| `name` ≠ directory | they match |
| Session staleness | new session behaves the same |
| A settings gate | no `skills` / `enabledSkills` key in `~/.claude/settings.json` or the project's |
| Description too long | longest is 441 chars |

## Why this is the same shape as #1073 and #1072

[...]

PR #1048, opened 2026-08-18
refactor(docs): slim AGENTS.md to crisp boundary rules and remove prose redundancy

mlorentedev:
## Summary

Slime `AGENTS.md` by eliminating bloated prose and repetitive sections while preserving all Standing Orders, Decision Hierarchy, and structural invariants.

- Removed redundant paragraphs in `Neural Hive Protocol` that repeated Standing Order #2 verbatim.
- Streamlined per-language technical standards, delegating deep reference tables to `00_meta/patterns/pattern-language-standards.md`.
- Streamlined MCP discovery table and operational rules, removing narrative historical anecdotes (relocated to modular lessons in `docs/lessons/`).
- Reduced file size from ~30KB to 19.8KB (~34% token footprint reduction per turn).
- Maintained 100% compliance with `tests/agents-md.bats`, `compile-harness.sh --check`, and all Go CLI test suites.

Closes #673

___

### **PR Type**
Documentation

___

### **Description**
- Condensed AGENTS.md rules into concise, actionable bullets.

- Removed verbose prose and repetitive historical anecdotes.

- Delegated detailed reference tables to vault patterns.

- Reduced file size by ~34% while preserving invariants.

___

### Diagram Walkthrough

```mermaid
flowchart LR
  AGENTS["AGENTS.md\n(~30KB)"] -->|Condense & Trim| SLIM["AGENTS.md\n(~19.8KB)"]
  VERIF["specs/GOV-004/verification.md"] -->|Update Status| COMPLETE["Status: complete\nCI: green"]
```

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Documentation</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>AGENTS.md</strong><dd><code>Condense AGENTS.md rules and remove verbose prose</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

AGENTS.md

<ul><li>Condensed Standing Orders and operational protocols into concise <br>rules.<br> <li> Removed verbose prose, historical anecdotes, and repetitive <br>explanations.<br> <li> Delegated detailed reference tables to vault patterns and skills.<br> <li> Reduced file size by ~34% while preserving all behavioral invariants.</ul>

</details>

  </td>
[...]

PR #1386, opened 2026-08-30
fix(setup-windows): the skill renderer drops neutral frontmatter, so Claude Code discovers every skill again

### LINKED ISSUE #1385: HARNESS-095: the Windows skill renderer never got #1080's frontmatter filter, so 34 of 43 skills are dormant on this box
## Symptom

On the Windows box, `/handoff`, `/spec`, `/crystallize`, `/catchup` and 30 other slash commands answer `Unknown skill: <name>`. Nine of the 43 deployed skills work.

## Cause

`paths:` entered the harness skill records on 2026-08-17 (#1048). #1080 (`ecceec8`) taught `render_skill` in `scripts/compile-harness.sh` to drop neutral/store-only keys at deploy time, because Claude Code reads a top-level `paths:` as a **conditional / path-scoped** skill and keeps it dormant until a matching file is touched in the session.

Its Windows twin, `Convert-SkillRecord` in `setup-windows.ps1`, was not changed. It still passes every record key through, so the deploy that ran on this box on 2026-08-28 13:51 wrote `paths:` into 34 of 43 `~/.claude/skills/*/SKILL.md`.

Measured 2026-08-29 — availability correlates with `paths:` on all 43, with no exceptions:

| deployed frontmatter | skills | available at session start |
|---|---|---|
| `paths:` present | 34 | 0 |
| `paths:` empty or absent | 9 | 9 |

The nine survivors are the ones whose vault records happen to carry no `paths:` (`enrich-us`, `new-ticket`, `dispatching-parallel-agents`, `project-maturation`, `using-git-worktrees`) plus four symlinked non-harness skills.

## Why CI never saw it

The #1080 guards exercise the bash renderer. Nothing in the repository executes `Convert-SkillRecord` and asserts on its output, so the twin could drift arbitrarily far with both OS legs green.

## Fix

- Port the allowlist to `Convert-SkillRecord`, carrying the `keep` state across continuation lines (a per-line filter drops `paths:` and leaves its wrapped list item behind, which is broken YAML).
- Guard it two ways, both mutation-verified: run the function on a fixture record and assert the output, and compare the native-field allowlist textually between `setup-windows.ps1` and `scripts/compile-harness.sh`.
- Lesson 245.

## Recovery for a live box

Re-run `setup-windows.ps1` after the fix merges; the skills return at the next session start.

mlorentedev:
Closes #1385.

## What broke

`paths:` entered the harness skill records on 2026-08-17 (#1048). Claude Code reads a top-level `paths:` as a **conditional / path-scoped** skill and keeps it dormant until a matching file is touched in the session, so #1080 taught `render_skill` in `scripts/compile-harness.sh` to strip neutral/store-only keys at deploy time. Its Windows twin, `Convert-SkillRecord` in `setup-windows.ps1`, never got the rule.

The first `setup-windows.ps1` run after that date (2026-08-28 13:51 on this box) rewrote all 43 deployed skills with the key that switches them off. Measured 2026-08-29 — the correlation is exact, 43 of 43:

| deployed frontmatter | skills | available at session start |
|---|---|---|
| `paths:` present | 34 | 0 |
| `paths:` empty or absent | 9 | 9 |

Nothing failed while it happened: the deploy reported success, the files were well-formed YAML, `dotf doctor` was green. The only symptom was `Unknown skill: handoff`.

## The fix

`Convert-SkillRecord` now applies the same allowlist as the bash renderer, carrying a `keep` flag across continuation lines. A per-line filter would drop `paths:` and leave its wrapped list item orphaned — YAML no reader accepts, worse than the bug.

It is renderer-level, so the same deploy repairs the agy, opencode, copilot and pi skill targets, not only Claude's.

## Guards

Both mutation-verified (fix removed, guard red; fix restored, guard green):

1. **Behavioural** — the function runs on a fixture record whose `paths:` value wraps onto a second line; the output must carry no neutral key and no orphaned continuation, and must keep `name`, `description` (with its continuation) and `allowed-tools`.
2. **Parity** — the native-field allowlist must be textually identical in `setup-windows.ps1` and `scripts/compile-harness.sh`. Drift here is silent: each side keeps deploying, and only the agents on the other OS lose the skill.

The first version of guard 1 used `! grep` and **passed against the unfixed renderer** — bash exempts a `!`-prefixed command from `set -e`, so it cannot fail a `@test` from the middle of the body. `tests/lib/refute.bash` exists for that trap and is now used. That is the corollary in lesson 245.

## Verification

- `bats tests/setup-windows.bats` — 126 tests, 0 failures (3 exercise `Convert-SkillRecord`)
- full pre-commit suite green
- mutation runs recorded above

## Recovery on a live box

Re-run `setup-windows.ps1` after this merges; the skills return at the next session start.

### DIFF
--- a/docs/lessons/_index.md
+++ b/docs/lessons/_index.md
@@ -263,3 +263,4 @@ tags: [lessons, index, dotfiles]
 | [242 - A process nobody can watch must leave its own trace, and the redirect has to survive the parent](lesson-242-a-process-nobody-can-watch-must-leave-its-own-trace.md) | 2026-08-27 |  |
 | [243 - A guard that reads a cache reports the cache's age as the credential's health](lesson-243-a-guard-that-reads-a-cache-reports-the-cache-not-the-credential.md) | 2026-08-28 |  |
 | [244 - A sweep is bounded by what the writer recorded, not by what the store holds; and the store's name rules decide the order](lesson-244-a-sweep-is-bounded-by-what-the-writer-recorded-not-by-what-the-store-holds.md) | 2026-08-29 |  |
+| [245 - A fix applied to one of two renderers is an outage on the other OS, and nothing on the fixed side can see it](lesson-245-a-fix-to-one-of-two-renderers-is-an-outage-on-the-other-os.md) | 2026-08-29 |  |
--- a/docs/lessons/lesson-245-a-fix-to-one-of-two-renderers-is-an-outage-on-the-other-os.md
+++ b/docs/lessons/lesson-245-a-fix-to-one-of-two-renderers-is-an-outage-on-the-other-os.md
@@ -1,0 +1,48 @@
+# Lesson 245 — A fix applied to one of two renderers is an outage on the other OS, and nothing on the fixed side can see it
+
+**Date:** 2026-08-29
+**Context:** HARNESS-095 — #1080 taught `render_skill` in `scripts/compile-harness.sh` to drop `paths:` from deployed skill frontmatter, because Claude Code reads a top-level `paths:` as a *conditional* skill and holds it dormant until a matching file is touched. Its Windows twin, `Convert-SkillRecord` in `setup-windows.ps1`, never got the same rule. Measured on the Windows box 2026-08-29: 34 of 43 deployed skills invisible at session start, and the nine that survived were the nine whose records carry no `paths:` at all.
+
+## What happened
+
+`paths:` entered the skill records on 2026-08-17 (#1048). The Linux renderer was
+fixed the next day. The Windows renderer kept passing every key through, so the
+first `setup-windows.ps1` run after that date (2026-08-28 13:51) rewrote 43
+deployed skills with a key that switches each of them off. Nothing failed. The
+deploy reported success, the files were well-formed, `dotf doctor` was green,
+and the only symptom was a slash command that answered `Unknown skill: docker`.
+
+Three properties made it survive eleven days:
+
+1. **The correlation was perfect and invisible.** Availability tracked exactly
+   one frontmatter key. Every other difference between a working and a broken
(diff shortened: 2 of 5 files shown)


mlorentedev:
## Review triage

CodeRabbit, round 1 — two findings, both applied in `d5837dc`.

| # | Finding | Disposition |
|---|---|---|
| 1 | 🟡 Minor — `$keep` should start `$false` for parity with the awk twin | **Applied.** Real bug. `awk` leaves `keep` unset and an unset awk variable is false, so a line inside the frontmatter *before the first key* (a comment) was emitted on Windows and dropped on Linux. The fixture now opens with `# store-only metadata follows` and asserts it is dropped; flipping the initialiser back turns that one assertion red, which is how it was verified. |
| 2 | 🟠 Major — `Convert-SkillRecord` exceeds the 40-line standard, four nesting levels | **Applied.** The classifier moved into `Test-SkillFrontmatterKeyKept`, a pure decision over one key line. `Convert-SkillRecord` 57 → 39 lines, one nesting level shallower; the helper is 15. `AGENTS.md:65` does declare `< 40 lines`, and nothing enforces it mechanically, so the finding was checked against the file rather than taken on trust. |

Finding 2 also flagged the behavioural bats test (59 lines). **Partially applied:** both renderer tests now share one `_render_skill_record` helper, which took it to 54. The remainder is a 20-line heredoc fixture and its assertions — no control flow — and the wrapped `paths:` value and the leading comment are exactly the shapes the test exists to pin, so hiding them behind a helper would cost more than the line count buys.

Also updated: `tests/copilot-native-skills-smoke.ps1` imports the new function name from the AST alongside the others, and its Pester test still deploys `handoff` end to end (1 passed, 0 failed).

Verification in this session: `bats tests/setup-windows.bats` 126/126; both guards red against `origin/main`'s renderer and green against this branch.

PR #1388, opened 2026-08-30
fix(spec): the drawn reviewer is told who it is, and a foreground run that wrote nothing fails

### LINKED ISSUE #1383: HARNESS-094: the drawn reviewer is not told who it is, and one refused to sign as the wrong model — the prompt names the reviewer, and a run with no review.md fails loudly
## Problem

`dotf spec review` launches the drawn pool member with a prompt that names the spec and nothing else (`Perform an adversarial review of the spec <id>.`). The reviewer has to infer *who it is* from the doctrine it reads — and one did it wrong.

Measured 2026-08-29 on the Windows work box, AI-042 review round 4: the draw was `nan/deepseek-v4-flash`, the transcript carries 68 assistant messages with `"provider":"nan","model":"deepseek-v4-flash"`, and the model ended its run — 248 bash calls, 35 reads, no `write` — with:

> I'm an Anthropic model, and this pool exists specifically to exclude Anthropic. … I will **not** write `specs/AI-042-trusted-folders-render/review.md`. … The gate-satisfying review is already running — round 4, drawn reviewer `nan/deepseek-v4-flash`.

It read the pool's BUG-074 note ("an adversarial review never runs on an Anthropic model"), the implementer-independence rule, and its own `review-request.json`, and concluded it was the wrong model *and* that some other process was the real reviewer. Result: a 40-minute run, no `review.md`, the archive gate unmet, and the failure invisible until someone reads the transcript — `dotf spec review` exited 0.

This is not the BUG-074 launcher defect (the model that ran is the model that was drawn); it is the reviewer guessing an identity the launcher already knows.

## Fix

- The prompt states the identity: `You are running as <reviewer id> (runner <pi|agy>), drawn from harness/reviewer-pool.json for this review; write your verdict to specs/<id>/review.md.` — `ReviewerCommand` in `cli/internal/spec/review_launch.go` has `e.ID`/`e.Runner` in hand.
- `dotf spec review` (non-dry-run) checks, after the runner exits, that `review.md` exists and is newer than `review-request.json`, and exits non-zero with the transcript path when it does not — a review that produced no file is a failed review, not a green one.
- Optional: the adversarial-review skill says explicitly that the reviewer is *told* its identity and must not infer it from the doctrine.

Effort S. Related: HARNESS-093 (#1370, random draw), HARNESS-071 (#955, the pool as allow-list), BUG-074.

mlorentedev:
Closes #1383.

## What happened

AI-042 review round 4, measured on the Windows work box 2026-08-29. The draw was `nan/deepseek-v4-flash`; the transcript carries 68 assistant messages stamped with that provider and model. The run ended, after 248 bash calls and no `write`, with:

> I'm an Anthropic model, and this pool exists specifically to exclude Anthropic. … I will **not** write `specs/AI-042-trusted-folders-render/review.md`. … The gate-satisfying review is already running — round 4, drawn reviewer `nan/deepseek-v4-flash`.

It had read the pool's own BUG-074 note, reasoned its way to an identity, and got it wrong. 40 minutes, no verdict, archive gate unmet — and `dotf spec review` exited 0.

Two separate defects, one incident.

## 1. The prompt states the identity

The launcher has known who it drew since `DrawReviewer` returned. All it told the model was which string to put in `reviewer:`, buried among the mechanical constraints — a *formatting* instruction, which is how it was read. The prompt now opens with:

> You are running as `<id>` (runner `<runner>`), drawn from harness/reviewer-pool.json for this review. That is your identity for this run, stated by the launcher that resolved it — do not infer it from the doctrine you are about to read. The pool's exclusions were applied at draw time, so reading one and concluding you must be the wrong model to review this is a misreading: you are the model that was drawn.

A test pins that it appears *before* the mechanical constraints: the model that refused had read most of the brief before deciding who it was.

## 2. A foreground run that wrote nothing now fails

`checkReviewProvenance` already refuses a `review.md` whose digest has not moved — but at **archive** time, which can be days later, in another session, with the transcript gone. A foreground run knows when the runner exits, and **every Windows run is a foreground run**, because tmux is Linux-only.

`VerifyReviewProduced` reuses the sidecar digest the archive gate already trusts, so "wrote no verdict" has one definition rather than two. It reports:

- no `review.md` at all → names the transcript path;
- a `review.md` byte-identical to what was there at launch → says so, and says what is on disk is the previous round's.

Detached launches are untouched by construction: the command returns while the reviewer is still running, so there is nothing yet to verify. That case remains the sidecar's, checked at archive.

## Guards

Three, each mutation-verified — broken, watched fail, restored:

| Guard | Mutation | Result |
|---|---|---|
| `TestReviewPromptTellsTheReviewerWhoItIs` | identity line replaced with a bare `Signing id:` | FAIL |
| `TestVerifyReviewProducedCatchesARunThatWroteNoFile` | `VerifyReviewProduced` short-circuited to `return nil` | FAIL |
| `TestVerifyReviewProducedCatchesAnUnchangedVerdict` | same | FAIL |

Plus the two directions that must stay quiet: a verdict genuinely written during the run passes, and a spec with no sidecar is not an error (nothing to compare against — the same posture `checkReviewProvenance` takes for a hand-written review).

## Verification

- `go build ./... && go vet ./... && go test ./...` — green
- `GOOS=windows go vet ./...` — green
- `golangci-lint run ./...` on **2.12.2**, the `versions.conf` pin — 0 issues
- pre-commit suite green

Also reorders two imports in `cmd/spec.go` that `gofmt` has been flagging on `main` — pre-existing, and in the block this change already touches.

## SDD skip rationale

73 production LOC, over the 50 threshold, and deliberately without a spec folder.

- The cause is not under-specified: #1383 carries the transcript, the measured draw, the model's own words, and prescribes both halves of the fix. The issue is the spec in substance.
- The behavioural surface is one prompt paragraph, one new function and one call site; most of the 73 lines are the comment prose this repo writes around them.
- It is repair to the **review machinery itself**, which the 2026-08-19 amendment caps: cheapest honest fix, deletion over construction.
- And it would be circular. A spec must pass `dotf spec review` before `spec archive` accepts it — the exact command whose launcher is broken here, in the exact way that produced no `review.md` last time it ran.

### DIFF
--- a/cli/internal/cmd/spec.go
+++ b/cli/internal/cmd/spec.go
@@ -2,8 +2,8 @@ package cmd
 
 import (
 	"fmt"
-	"math/rand/v2"
 	"io"
+	"math/rand/v2"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -176,7 +176,7 @@ is the only record of how.`,
 			}
 
 			skill := spec.ReviewerSkillPath(chosen.Runner)
-			prompt := spec.ReviewPrompt(id, repoRoot, chosen.ID, skill)
+			prompt := spec.ReviewPrompt(id, repoRoot, chosen.ID, chosen.Runner, skill)
 			argv, err := spec.ReviewerCommand(chosen, prompt, timeout, repoRoot)
 			if err != nil {
 				return err
@@ -247,7 +247,18 @@ is the only record of how.`,
 			}
 
 			cmd.Printf("\n")
-			return runForeground(repoRoot, launch, transcript)
+			// A foreground run is the one case where the launcher sees how the review
+			// ENDED rather than only that it started, so the no-verdict case is
+			// reported here, with the transcript still in hand, instead of being left
+			// for whoever next tries to archive.
+			runErr := runForeground(repoRoot, launch, transcript)
+			if err := spec.VerifyReviewProduced(specDir, transcript); err != nil {
+				if runErr != nil {
+					return fmt.Errorf("%w\nthe runner also exited with: %v", err, runErr)
+				}
+				return err
+			}
+			return runErr
 		},
 	}
 
--- a/cli/internal/spec/review_launch.go
+++ b/cli/internal/spec/review_launch.go
@@ -250,14 +250,25 @@ func ResolveReviewer(entries []ReviewerEntry, want string) (ReviewerEntry, error
 // is only what the skill cannot know: which spec, which repo, and the mechanical
 // constraints that keep the review from invalidating itself.
 //
+// It states the reviewer's IDENTITY, not only the string to sign with. Measured
+// 2026-08-29 (#1383): a drawn reviewer read the pool's own "never an Anthropic
+// model" note, concluded from the doctrine that it must be Anthropic, announced
+// that some other process was the real reviewer, and refused to write a verdict
+// after a 40-minute run. The launcher knew who it had drawn the whole time.
+//
 // The canonical id is stated because `reviewer:` is self-reported and the gate
 // matches it exactly — a reviewer that writes its own name a different way gets
 // refused for a mismatch rather than for a policy breach, which wastes a whole
 // review round on a string.
-func ReviewPrompt(specID, repoRoot, reviewerID, skillPath string) string {
+func ReviewPrompt(specID, repoRoot, reviewerID, runner, skillPath string) string {
 	var b strings.Builder
 	fmt.Fprintf(&b, "Perform an adversarial review of the spec `%s`.\n\n", specID)
 	fmt.Fprintf(&b, "Repo: %s\n\n", repoRoot)
+	fmt.Fprintf(&b, "You are running as `%s` (runner `%s`), drawn from harness/reviewer-pool.json\n"+
+		"for this review. That is your identity for this run, stated by the launcher that resolved\n"+
+		"it -- do not infer it from the doctrine you are about to read. The pool's exclusions were\n"+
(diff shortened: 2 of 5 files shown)


mlorentedev:
## Review triage

CodeRabbit, round 1 — two findings, both applied in `68e5248`.

| # | Finding | Disposition |
|---|---|---|
| 1 | 🟠 Major — a foreground run is accepted when `WriteReviewRequest` failed, the runner exits 0 and writes nothing | **Applied, and it was a real hole in what I wrote.** `WriteReviewRequest` fails with a *warning* by design, so an unwritable spec dir must not stop a launch — which meant reading the digest back from that file let one cause switch off two guards, the second one silently. The launcher now keeps the pre-launch digest **in memory** (`digestBefore := spec.ReviewDigest(specDir)`, captured before the sidecar write) and passes it to `VerifyReviewProduced`, which no longer reads the sidecar at all. `ReviewDigest` is exported for it and `WriteReviewRequest` uses the same function, so "the digest of review.md" has one definition. |
| 2 | 🟠 Major — four tests for one state matrix; `AGENTS.md` asks for table-driven Go tests | **Applied.** Checked against `AGENTS.md:53` and `:85` rather than taken on trust — the standard is real. One table, five rows, the fifth being the case the rework closes: *a run that wrote nothing deletes the verdict it started with*. |

Re-verified after the rework: `go build/vet/test ./...` green, `GOOS=windows go vet` green, `gofmt` clean on all four touched files, and the guard mutation-verified again — short-circuit `VerifyReviewProduced` to `return nil` and three of the five rows go red.

PR #1389, opened 2026-08-30
docs(lessons): a guard that reads its own evidence from an optional file

mlorentedev:
Lesson 246, harvested from the CodeRabbit finding on #1388 — the knowledge half of Definition of Done §2, landing after that PR merged.

`VerifyReviewProduced` answered *"what did review.md hold before this run?"* by reading `review-request.json`, a sidecar whose writer is documented to fail with a **warning** rather than a refusal. Compose the two and one cause — a spec dir that cannot be written — takes out both guards, the second one silently, in precisely the state where a review is most likely to end with nothing written.

| | with sidecar | without sidecar |
|---|---|---|
| archive gate | refuses a `review.md` that never moved | says nothing (honest: nothing was asserted) |
| launcher check, first version | refuses | **says nothing, and exits 0** |

The rule it records: evidence a guard needs at time T is captured at time T, not re-read from storage that was permitted to fail. Plus two smaller things worth keeping — one function that defines "the digest of review.md", and the observation that fail-open vs fail-closed is a property of the *caller*, not of the value (`fileDigest` returns `""` for absent and unreadable alike; the archive gate treats that as "the digest moved", the launcher treats it as "no verdict", and both are right).

Docs only: one lesson file plus its `_index.md` row.


mlorentedev:
## Review triage

Nothing to apply, and one thing to disclose.

|---|---|---|
| PR-Agent | *"No major issues detected"*, no security concerns, effort 1/5; ticket compliance ✅ against #1388 and #1383 | **Nothing to action.** It read the change for what it is: the DoD §2 knowledge half of #1388, landing after that PR merged. |
| CodeRabbit | **Did not review** — *"Review limit reached … You've used all free OSS reviews for now"* | **Disclosed, not applied.** A notice that no review ran leaves the PR unreviewed by CodeRabbit; it is not a green review. Proceeding on PR-Agent's substantive pass, said out loud rather than left to be inferred. |

CI: 17/17 green, `mergeStateStatus` CLEAN. Docs only — one lesson file and its `_index.md` row, no production code.

Issue #1286, opened 2026-08-27
HARNESS-090: the lessons-index check measures _index.md's mtime, so four unindexed lessons kept it green


mlorentedev:
## Problem

`mem.lessonsStaleness` (`cli/internal/mem/session_start.go:246-264`) is the guard that exists to keep `docs/lessons/_index.md` honest. It stats the file and warns when its **mtime** is older than `staleDays` (default **14**, line 68).

That is a proxy for "someone maintained the index", and it fails in the exact case the guard is for. Measured **2026-08-27**:

| | |
|---|---|
| `lesson-NNN-*.md` files | **234** |
| rows in `_index.md` | **230** |
| `_index.md` last edited | 2026-08-24 (3 days) |
| check output | **green — no warning** |

Lessons **231, 232, 233, 234** existed as files with no index row. Four separate sessions each wrote a lesson and none added its row; the index had been edited recently enough for other reasons, so the window never opened. The lessons were undiscoverable through the only surface that lists them.

The failure mode is the one the guard was built to prevent, and the guard reported success throughout.

## The invariant that is actually available

`mtime` is unnecessary here. The real property is a set equality, and it is cheap:

> every `docs/lessons/lesson-NNN-*.md` has a row in `_index.md`, and every row's link target exists

Both directions were checked by hand while fixing the backlog in #1285 — a `grep -o` over the link targets plus two counts, well under a second on 235 files. There is no reason to approximate it.

## Recommendation

**Replace** the mtime check with the set-diff. Do not add a second guard beside it:

- an unindexed lesson is *always* wrong, so there are no false positives to tune — unlike the mtime window, which fires on a quiet fortnight and stays silent on a real gap;
- two checks on one file means the weaker one still reports green, which is how this was missed;
- report the missing basenames, not a count. `[lessons] 4 not indexed` sends the reader to do the diff again.

Report broken link targets too — the reverse direction is the same walk, and a renamed lesson file breaks it silently today.

Keep the existing behaviour of falling back to a flat `docs/lessons.md` and staying silent when neither exists (line 251-259) — repos scaffolded by `dotf init` may have either shape.

### Where it belongs

[...]

Issue #1157, opened 2026-08-21
GUARD: dotf spec review exits OK when the run produces no verdict, leaving the previous review.md in place


mlorentedev:
## What happened, measured 2026-08-21

`dotf spec review HARNESS-075-model-map-routing-registry` was launched against the merged sha
`63acd91` to run round 6. It printed:

```
[OK] Review running detached. Watch it with:
    tmux attach -t review-HARNESS-075-model-map-routing-registry
When it finishes, review.md carries the verdict and archive reads it.
```

The run ended. `review.md` carries a verdict. **It is the previous round's verdict**, byte-identical
to the copy on `origin/main`:

```
$ git status --porcelain specs/HARNESS-075-model-map-routing-registry/
(empty)

$ md5sum review.md ; git show origin/main:.../review.md | md5sum
950986d35a7b54589fedd862f81579ca
950986d35a7b54589fedd862f81579ca
```

Its frontmatter says `reviewed_sha: 1b1cfb3`, `date: 2026-08-20` — a sha that is **not an ancestor
of main and no longer exists as any branch**. Nothing in the launcher's output distinguished this
from a completed review.

## Why the run produced nothing

Not an error. The transcript shows an orderly stop:

```
event types: turn_start 25, tool_execution_start 36, turn_end 25, agent_end 1, agent_settled 1
last usage:  input 58,743   output 0
stderr:      0 bytes
```

The reviewer was working correctly and was **on the right tree** — it resolved HEAD to `63acd91`
and confirmed `YES HEAD is ancestor` of main. Its final text is cut mid-sentence:

> "Comprehensive named tests present. Let me run the full harness and doctor packages and confirm
> all finding"

That is `pi`'s internal turn cap (25) ending the agent, not a failure. `pi` exposes no flag to
raise it (`pi --help` has `--thinking` and nothing for turns/steps).

## The finding is the silence, not the cap

**This exact class is already documented in this repo, on the other arm of the same launcher.**
`cli/internal/spec/review_launch.go`, in the `agy` case:

> "Observed exactly that — the run read a few files, was refused `git rev-parse HEAD`, and gave up
> after 14 seconds while **reporting `status: SUCCESS` with an empty response**."

The response then was to fix the *trigger* (`--dangerously-skip-permissions`). No guard was added.
So when the `pi` arm hit the same class through a different trigger, it was invisible again.

**A stale `review.md` left in place after a run that wrote nothing is indistinguishable from a
[...]

PR #1408, opened 2026-09-01
chore(spec): archive HARNESS-105 FAIL review, and fix the setup seeding that the cleanup exposed

mlorentedev:
## What

`specs/HARNESS-105/` held exactly one file — `review.md`, verdict **FAIL** — while the spec's other five artifacts sat in `specs/archive/HARNESS-105/`. This moves it into the archive beside its own `review-request.json` and records, in `proposal.md`, what the split was hiding.

## Why it happened

Both paths were added by the **same commit** (`d4ea0f5`, #1404), so this was never a partial archive. `spec.Archive` moves the whole directory with `os.Rename` and is structurally incapable of leaving a file behind — so the writer that arrived *afterwards* is the cause.

Reconstructed from git:

| Commit | What it did |
|---|---|
| `b63eef1` | the implementation; the archived `review-request.json` names it as `reviewed_sha` |
| `dd47c5d` | **archived the spec** — with no `review.md` accompanying it |
| `b3fcd17`, `150a048`, `76aa61a`, `70e15af` | fixes for the Blocker, all **after** the archive |

A second review round was pointed at `dd47c5d` — the archive commit itself — returned FAIL, and wrote to `specs/<id>/review.md` without checking that the spec was already archived. That write **recreated the directory as a ghost** holding one file, and everything enumerating `specs/*/` has counted this spec as active since (it is one of the 33).

## Disclosure, not ratification

The Blocker was real: a wrapped `TerminalFailureError` printed the wrapper prefix before the JSON latch, breaking the strict-prefix parsing orchestrators depend on.

**It is fixed in `main`** — verified, not assumed: `cli/cmd/dotf/main.go` unwraps with `goerrors.As(err, &tfe)` before printing. But the four fix commits all postdate the archive and **no re-review was ever run**, so no PASS verdict exists for the tree that actually merged. The note in `proposal.md` says that out loud rather than leaving it to be rediscovered.

`review.md` is moved **byte-for-byte unmodified** (the rename reports 100% similarity, 0 insertions, 0 deletions). Editing a reviewer's own artifact to add context was rejected for the obvious reason; the disclosure goes in `proposal.md`, where `Archive` already writes provenance. Until now the archive recorded a review having been *requested* with no verdict attached — a less complete account than the FAIL it was missing.

## Root cause is being consolidated, not re-ticketed

`dotf spec review` writing into `specs/<id>/` regardless of archive state belongs to review-machinery provenance — the same area as **GUARD-005 (#1157)**, whose `review-request.json` sidecar is the adjacent half. Consolidating there per the 2026-08-19 meta cap rather than opening a new issue.

## Verification

- `go test ./internal/spec/...` — ok
- pre-commit: gitleaks, dotfiles tests, bats names, doc paths, commit-msg — all Passed
- `git diff --cached --stat` on the rename: `0 insertions(+), 0 deletions(-)`
- No test, script, or workflow references `HARNESS-105` (grepped `tests/ scripts/ .github/ cli/`)

## Out of scope, deliberately

`specs/WIN-007-harness-mirror` is `verifying` while its issue is believed closed. **Not touched here** — the GitHub API rate limit was exhausted for the whole box during this session, so the issue's real state could not be confirmed, and changing a spec's status from a remembered fact is precisely the failure mode this repo has recorded before.

---

## Second commit — the lesson the cleanup exposed

`ebad1a3` records lesson 247. Removing `dotfiles-wt-gentleai` (clean, unoccupied, branch merged as #1404, content verified byte-identical to the squash) broke `pre-commit` for **every session on the box**: `Run dotfiles tests` failed 77/78 on `DOTFILES_REPO_DIR: directory missing`.

`~/.config/dotfiles/machine.json` — the ADR-025 path SSOT — named that worktree as the canonical repo dir:

```json
{ "paths": { "DOTFILES_REPO_DIR": "/home/manu/Projects/dotfiles-wt-gentleai" } }
```

An earlier session had persisted its cwd. The value was wrong when written; a feature-branch worktree is a temporary view of the repo, never the repo. **`dotf doctor` was never broken** — it carries exactly the right check and fired the instant it could (`[FAIL] ... path does not exist`). It simply could not observe anything while the wrong path existed, because absence is its only symptom. The deletion converted a silent defect into a loud one.

Repaired on this box via `dotf env set DOTFILES_REPO_DIR /home/manu/Projects/dotfiles` + `dotf env generate` — the sanctioned write-side, not a hand-edit.

Worth knowing for anyone whose shell predates the fix: the resolution cascade prefers the **environment variable** over `machine.json`, so an already-started session still resolves to the deleted path. `env -u DOTFILES_REPO_DIR dotf env path DOTFILES_REPO_DIR` shows the truth on disk; `export DOTFILES_REPO_DIR=/home/manu/Projects/dotfiles` unblocks committing without restarting.

The lesson also records a second instance of the same shape from this session: "is the worktree clean" was used as "is the worktree free", and it was clean and **not** free — a peer session had started there two minutes earlier and had not yet written. Occupancy is a property of the processes (`/proc/<pid>/cwd`), not of the tree.

### Index invariant

Checked by **set equality**, not mtime — the #1286 argument: 247 numbered lesson files, 247 numbered index rows, zero broken links.

Flagged and deliberately **not** fixed here: the final `_index.md` entry is a bullet rather than a table row (added by #1406). Unrelated formatting drift; it does not belong in this change.


mlorentedev:
## Third commit — the recurrence, fixed rather than only described

`f3f9e1c` was added after the body above was written. Reviewing the lesson turned up the writer, and it is not a stray session:

```sh
# setup-linux.sh:1612, before this change
if dotf env set DOTFILES_REPO_DIR "$CURRENT_DIR" >/dev/null; then
```

Setup seeds **the directory it was invoked from**. Run it from a linked worktree — an ordinary thing to do while testing a setup change — and the machine's ADR-025 SSOT records a throwaway path. So this recurs by construction, which moves it out of "guards machinery" (where the 2026-08-19 meta cap would apply) and into a product defect: fix-or-ticket, and it is cheap enough to fix.

Skipping the seed would reopen **BUG-029/#696**, so this resolves the canonical checkout rather than declining. `git rev-parse --git-dir` and `--git-common-dir` are equal in a normal checkout and differ inside a linked worktree, where the common dir is the real repo's `.git`; `dirname` of it is the checkout to seed.

### Tests

Per incident → guard in the same change, two levels:

- **Mechanism**, on a real repo plus a real `git worktree add`: asserts the equal case, the differing case, and that `dirname(common)` resolves back to the canonical checkout.
- **Wiring**, because resolving correctly is worthless if the script still passes `$CURRENT_DIR`. Uses `refute_grep_fixed` — a bare `! grep` cannot fail a bats test.

**Mutation-verified:** restoring `"$CURRENT_DIR"` turns the wiring test red; restoring the fix turns it green.

### Correction to lesson 247, in the same commit

The lesson shipped in `ebad1a3` claiming the writer was an unidentified earlier session and that `dotf env set` was not involved. **Both were wrong** — one grep named the writer, and the "stray session" line was an inference stated as fact. Corrected to name `setup-linux.sh:1612` and to describe the actual discriminator.

### Verification

- `shellcheck setup-linux.sh` — clean (only pre-existing SC1091 info)
- `bash -n` and `zsh -n` — both pass
- `bats tests/*.bats` — **1523 ok**

One failure in the suite, `BUG-771: copilot native skills are not deployed when the copilot binary is absent` (`tests/skills-pipeline.bats:166`), **reproduces on clean `main`** and is not from this change. Verified by running it filtered on an untouched checkout.

### Coordination

`setup-linux.sh` and `tests/setup-linux.bats` are also being edited on `feat/bind-hooks-into-setup` (HARNESS-045 AC1, #561), which has no PR yet. The regions are disjoint — that branch works in `merge_claude_settings` (lines ~1344-1415), this change is at ~1621, and the two bats additions are appended at the end of the file. Flagged to that session directly.

mlorentedev:
## Review triage

**Who reviewed:** PR-Agent (`github-actions[bot]`), full review at `d147e2f`. Verdict line: **"No major issues detected"**, no security concerns, PR contains tests, effort 3/5. **CodeRabbit has not posted** — disclosed rather than counted as a review that happened.

**CI at `d147e2f`:** `lint`, `lint-powershell`, `test` (ubuntu), `integration`, `spec-gate`, `attestation`, `changes`, `review`, `GitGuardian` all green. `test-windows` still running at the time of writing; it will be confirmed before merge, and this triage does not claim it.

### Dispositions

PR-Agent raised three findings, all under **ticket compliance**. All three are the same false positive and are **declined**, with reasons:

| # | Finding | Disposition |
|---|---|---|
| 1404 | "Not compliant — shellQuote refactor and JSON error latches not addressed" | **Declined.** #1404 is a *merged pull request* cited as history, not this PR's ticket. The tool treated a referenced PR number as a requirement to re-implement. The work it names shipped in #1404 itself and is present in `main` — this PR verifies that (`cli/cmd/dotf/main.go` unwrapping with `goerrors.As`) rather than re-doing it. |
| 1157 | "Partially compliant — does not fix the core bug in `dotf spec review`, no digest comparison, no `reviewed_sha` assertion" | **Declined as out of scope, correctly observed.** GUARD-005 is where the root cause was *consolidated* (see [the comment](https://github.com/mlorentedev/dotfiles/issues/1157#issuecomment-5487150092)), deliberately not implemented here. Implementing GUARD-005 inside a hygiene PR is exactly the scope creep the 2026-08-19 meta cap exists to prevent. The finding is accurate about what is absent; it is wrong that absence is a defect. |
| 1286 | "Partially compliant — does not replace the mtime check with set-diff, no regression test for the index check" | **Declined.** #1286 is referenced only to justify *why this PR checked the index by set equality by hand* (247 files, 247 rows, zero broken links) instead of trusting the mtime-based guard. Citing an open issue's argument is not a commitment to close it. |

**Nothing to apply.** No finding identified a defect in the diff.

### Note on the pattern

All three are wrong-ticket linkage: the tool reads every issue or PR number appearing in the body as a requirement this PR must satisfy. That is the same false-positive class recorded against PR-Agent during CLI-021 (`dotf vault health`), where wrong-ticket linkage was likewise skipped. Worth knowing before writing a PR body that cites prior art — thorough provenance makes this reviewer noisier, which is not a reason to cite less.

### Debt found while verifying, filed not swallowed

`bats tests/*.bats` gives **1523 ok, 1 not ok** on clean `main` — `BUG-771`, `tests/skills-pipeline.bats:166`. Not from this PR; reproduced on an untouched checkout. Filed as **TEST-007 (#1409)**, with the follow-up measurement that it *passes* on the CI runner and fails locally, which narrows it to fixture isolation rather than a pipeline regression.

mlorentedev:
## Review triage — addendum

Two things changed after the triage above was posted. Both are corrections to it.

### CI is now fully green

All 15 checks pass at `17335e4`, `test-windows` included — the one the earlier comment explicitly declined to claim:

```
attestation lint lint-powershell spec-gate changes review integration
test test(ubuntu-latest) test-windows test(windows-latest)
goreleaser-snapshot GitGuardian attestation
```

### CodeRabbit posted — and it is a notice, not a review

The comment above said "CodeRabbit has not posted". It has now, and the substance of that statement is unchanged, so this records the difference rather than quietly leaving a claim that reads wrong:

> **Review limit reached** — Next included review available in 47 minutes. You've used all free OSS reviews for now.

**A notice that no review ran leaves the PR unreviewed.** It names no file, no line, and no claim about the diff; the only thing it says is that the reviewer did not start. It is not a second review to weigh against PR-Agent's, and it must not be counted as one just because a comment exists.

### Standing disposition

**This PR has had exactly one substantive review: PR-Agent.** Its three findings were all wrong-ticket linkage and were declined with reasons in the table above; it reported no security concerns and no major issues.

Proceeding on that is a decision, not an oversight, and it is disclosed here rather than assumed: a merge now merges a change reviewed by one reviewer, with the second having declined for quota reasons. CodeRabbit's window reopens in ~47 minutes for anyone who would rather wait for the second opinion — the diff is small and static, so waiting costs nothing but time.

mlorentedev:
## Rebased onto #1407, re-verified

`#1407` merged as `fdbe17c`; this branch is rebased onto it and force-pushed with `--force-with-lease` (`17335e4` → `c02799e`). Now 0 behind, 5 ahead.

### One conflict, as predicted

`docs/lessons/_index.md` only — both PRs append rows at the same place. Resolved in numeric order; the other session had already renumbered theirs to 248/249 leaving 247 free, so nothing had to move:

```
| [247 - A pointer into a deletable location keeps every check green until the deletion...
| [248 - A broken parser is a defect only where something reads it...
| [249 - Two scripts writing the same logical value in two syntaxes...
```

`setup-linux.sh` and `tests/setup-linux.bats` merged without conflict, as expected from the region analysis — their work is in `merge_claude_settings` (~1345-1433), this change is at ~1650, and the two bats cases are appended.

### Coexistence checked, not assumed

Git merging two hunks quietly is not evidence they work together, so both sides were verified present and correct afterwards:

| Check | Result |
|---|---|
| This PR's `SEED_REPO_DIR` block | present, `setup-linux.sh:1650-1657` |
| #1407's `dotf harness bind` | present, `setup-linux.sh:1433` |
| #1407's removal of the `.hooks.SessionStart[0]` assignments | `grep -c` → **0**, correctly gone |
| Lessons index invariant (set equality, not mtime) | **249 files, 249 rows** |

### Full re-verification on the new base

- `shellcheck setup-linux.sh` — clean (pre-existing SC1091 info only)
- `bash -n` and `zsh -n` — pass
- `go build ./...`, `go vet ./...`, `GOOS=windows go vet ./...` — all pass (the base now carries #1407's CLI code)
- `bats tests/*.bats` — **1522 ok, 1 not ok**
- The two lesson-247 cases — both green on the new base

The single failure is `BUG-771` (`tests/skills-pipeline.bats`), unchanged from before the rebase and reproducing on clean `main`. Tracked as **#1409**, not from this PR.

CI on `c02799e`: 14 of 15 green, `test-windows` still running at the time of writing.

Issue #1409, opened 2026-09-01
TEST-007: the BUG-771 copilot-skills assertion fails on clean main, so the bats suite is permanently red


mlorentedev:
## The suite is red on `main`

`bats tests/*.bats` on an untouched `main` (`11a68b1`) gives **1523 ok, 1 not ok**:

```
not ok 1188 BUG-771: copilot native skills are not deployed when the copilot binary is absent
# (in test file tests/skills-pipeline.bats, line 166)
#   `[ ! -e "$FAKEHOME/.copilot/skills" ]' failed
```

Found while verifying an unrelated change (#1408). Confirmed **not** caused by that change: reproduced on a clean checkout with nothing applied, by running the test filtered on its own.

```sh
bats tests/skills-pipeline.bats --filter "BUG-771"
```

## Why it matters more than one red test

The assertion is the negative half of BUG-771 — that the skills pipeline does **not** deploy into `~/.copilot/skills` when the `copilot` binary is absent. The test failing means `$FAKEHOME/.copilot/skills` **does** get created in that case. So either:

1. the guarded behaviour regressed and copilot skills are now deployed unconditionally, which is the defect BUG-771 exists to prevent; or
2. the fixture no longer isolates `$FAKEHOME` and something else creates the path, in which case the test has stopped testing anything and the real guard is unverified.

Both are worth knowing and they are distinguishable in minutes. Neither is safe to leave: a permanently-red test trains everyone to read `1 not ok` as background noise, which is exactly how the next real failure gets skipped past. This repo has the precedent filed already — TEST-006 (#1320) was "a skip passes as green"; this is its mirror, a fail that becomes wallpaper.

## Not investigated here, deliberately

Found during worktree hygiene on an unrelated PR. Ticketed rather than fixed in place to keep #1408 reviewable — it already carries a spec-archive correction and a `setup-linux.sh` fix.

## Verification when fixed

- `bats tests/skills-pipeline.bats --filter "BUG-771"` green
- mutation: whichever branch is corrected, remove the correction and confirm the test goes red again — the assertion has been passing vacuously or failing truly, and the fix must establish which

mlorentedev:
## It fails locally and passes in CI, which narrows it

Measured on PR #1408 (commit `d147e2f`): the `test` job on `ubuntu-latest` went **green**, while the same suite on this Linux box fails the BUG-771 assertion. Same tests, opposite results.

[...]

PR #1410, opened 2026-09-01
feat(harness): the gate resolves its persona from the payload, and doctor reports what it cannot enforce

mlorentedev:
## What this fixes

The gate shipped in #1272 and wired into the hooks in #1407 was complete and
decided nothing. Measured on this box: **every tool call was allowed**, for two
independent reasons. This PR closes the first and makes the second visible.

### 1. The gate never learned which persona it was gating

`harness/manifest.json` emits `dotf harness gate --harness claude` with no
`--role`. With no role there is no persona, and `Decide` returns
`"no persona in scope"` → Allow. Every call, always.

The role is already in the payload: Claude documents `agent_type` on every hook
event raised inside a dispatched subagent, and omits it on the main thread —
which is exactly the distinction the gate needs. The flag still wins when
present, so nothing that passes `--role` changes behaviour.

`agent_id` is read alongside it and scopes the consumption ledger. Without it
every persona dispatched in one session shares one ledger, and the parent's
skill runs silently satisfy a reviewer's gate. `ConsumptionScope()` falls back
to the session id, which is the right key for a main-thread call.

### 2. Every persona skill is unmigrated, and nothing said so

All 35 skills are in the inline form, which carries no severity, so they parse
as `EnforceUnset` — deliberately neither enforced nor warned about.
`Decide`'s comment says such a skill "is surfaced by `dotf doctor` instead", and
`UnmigratedSkills()` was written, exported and unit-tested to that end.
**Nothing in production ever called it.** `dotf doctor` now reports
`35 of 35 persona skills carry no enforce: severity`.

It is a WARN, never a FAIL: an unmigrated skill is a known, planned state, and
failing the health command over planned work trains the reader to ignore it. It
turns into a Pass when the migration finishes.

The check uses `harness.LoadPersona`, not `readAgentFrontmatter` — the latter's
own docstring says it never reads `skills`, so using it here would have produced
a check that passes on the broken thing.

### 3. Prerequisite: stderr was being discarded wholesale

`run` set `rootCmd.SetErr(io.Discard)` to suppress Cobra's automatic
`Error: ...` wrapper around a `TerminalFailure`. But `cmd.ErrOrStderr()`
resolves through the root, so it discarded every diagnostic any subcommand
deliberately wrote — measured at `d4ea0f5`: **16 call sites across 9 command
files**, including all four `dotf secrets` subcommands, `harness mirror` and
`harness presence`.

The sharpest case is the gate itself: it blocked a tool call with exit 2 and
nothing on stderr, so its refusal was indistinguishable from a crash and the
operator had no way to learn which skill to invoke.

Cobra's `SilenceErrors` is the mechanism for the first concern without the
second. Forcing it on the root would pollute the "did this command ask for
silence" read when the root *is* the executed command, so the original value is
snapshotted and restored for that one case — the four existing tests pass
unchanged.

Also fixed: a typo'd `--role reviewr` resolved to nothing and allowed in
silence. It now says so on stderr, naming the path it tried and stating that
enforcement is off for that call.

## Verification

End-to-end against a persona record carrying declared severity:

- main thread (no `agent_type`) → allow
- dispatched `reviewer` → **exit 2**, naming the unconsumed skill on stderr
- invoking that skill → consumption recorded, subsequent calls allowed
- a second agent in the same session → does **not** inherit that consumption

```
cd cli && go build ./... && go vet ./... && GOOS=windows go vet ./... && go test ./...
golangci-lint run          # 0 issues (v2.12.2, the versions.conf pin)
bats tests/*.bats          # 1522 ok, 1 pre-existing failure (BUG-771, #1409)
```

## Declared risk

`agent_type` / `agent_id` are taken from documentation, **not measured on this
machine** — the gate is not live in any deployed settings here. The design fails
safe: a wrong field name yields no persona, which allows, which is today's
behaviour. A failure therefore costs enforcement, never a blocked session.

**No skill may be promoted to `enforce: block` before `[gate] warn` lines are
observed in a real dispatch.** That confirmation is a precondition of the
migration, not of this PR.

## Not in this PR

The skill migration itself. Persona records under `harness/agents/` are
generated from the vault, which is the SSOT, and the rollout is a canary:
`reviewer` to `enforce: warn` first, observed, and only then any promotion.
`specs/HARNESS-046/check-roster-consistency.py` raises `UnreadableSkills` on the
mapping form by design and must be updated in that same movement.

Closes part of #561.

### DIFF
--- a/cli/cmd/dotf/main.go
+++ b/cli/cmd/dotf/main.go
@@ -1,10 +1,10 @@
 package main
 
 import (
+	goerrors "errors"
 	"fmt"
 	"io"
 	"os"
-	goerrors "errors"
 
 	"github.com/mlorentedev/dotfiles/cli/internal/cmd"
 	"github.com/mlorentedev/dotfiles/cli/internal/errors"
@@ -19,10 +19,28 @@ func main() {
 }
 
 func run(rootCmd *cobra.Command, stderr io.Writer) int {
-	// We divert Cobra's automatic error printing to io.Discard.
-	// This prevents Cobra from printing wrapped TerminalFailure errors with "Error: context: ...",
-	// while keeping the subcommand's SilenceErrors property perfectly intact so we can read it.
-	rootCmd.SetErr(io.Discard)
+	// Suppress Cobra's AUTOMATIC error printing — its "Error: context: ..."
+	// wrapper around a TerminalFailure — while leaving the command's own stderr
+	// pointed at the real one.
+	//
+	// This was `rootCmd.SetErr(io.Discard)`, which achieved the first half by
+	// discarding EVERYTHING: `cmd.ErrOrStderr()` resolves through the root, so
+	// every deliberate diagnostic any subcommand writes went to the void.
+	// Measured on d4ea0f5: 16 call sites across 9 command files, including all
+	// four `dotf secrets` subcommands, `harness mirror`, `harness presence` — and
+	// `harness gate`, which then blocked a tool call with exit 2 and NO reason,
+	// leaving the operator no way to know which skill to invoke.
+	//
+	// SilenceErrors on the root is the mechanism Cobra provides for exactly this:
+	// ExecuteC checks the executed command's flag OR the root's before printing.
+	// The per-command flag stays readable below, so the "did this command ask for
+	// silence" branch is untouched.
+	// Forcing the flag on the root pollutes the "did this command ask for
+	// silence" read below whenever the root IS the executed command, so its
+	// original value is snapshotted and restored for that one case.
+	rootSilencedByAuthor := rootCmd.SilenceErrors
+	rootCmd.SilenceErrors = true
+	rootCmd.SetErr(stderr)
 
 	executedCmd, err := rootCmd.ExecuteC()
 	if err != nil {
@@ -32,7 +50,11 @@ func run(rootCmd *cobra.Command, stderr io.Writer) int {
 			_, _ = fmt.Fprintln(stderr, tfe.Error())
 		} else {
 			// If the specific command didn't request silence, print the error.
-			if !executedCmd.SilenceErrors {
+			silenced := executedCmd.SilenceErrors
+			if executedCmd == rootCmd {
+				silenced = rootSilencedByAuthor
+			}
+			if !silenced {
 				_, _ = fmt.Fprintf(stderr, "Error: %v\n", err)
 			}
 		}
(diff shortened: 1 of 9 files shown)


mlorentedev:
## Review triage

**One substantive reviewer, not two.** PR-Agent reviewed. CodeRabbit returned
*"Review limit reached — next included review available in 45 minutes"*, which is
a notice about the review, not a review: this PR carries a single reviewer's
coverage and that is disclosed rather than left implicit.

CI at `c52ddd6`: **14 success, 1 skipped, 0 failing** (lint, lint-powershell,
test, test-windows, test ubuntu/windows, integration, spec-gate, attestation,
GitGuardian, goreleaser snapshot). `mergeable: true`, state `clean`.

| # | Finding | Reviewer | Disposition |
|---|---|---|---|
| 1 | *THEORETICAL: unverified field names* — `agent_type`/`agent_id` come from documentation, not from a payload measured on this machine; wrong names yield no persona and the gate stays inert. | PR-Agent | **Accepted, already declared.** This is the risk the PR body and the `commandHookPayload` doc comment both state in full. No code change: the design fails open by construction, so a wrong name costs enforcement and cannot block a session. Confirming it is a stated precondition of promoting any skill to `enforce: block`, which this PR does not do. |
| 2 | *THEORETICAL: consumption scope uniqueness* — if `agent_id` is not unique per invocation, two dispatches of one persona share a ledger and the second inherits the first's consumption. | PR-Agent | **Applied — `c52ddd6`.** This one named something the disclosure did not. The doc comment on `ConsumptionScope` now states the assumption, its blast radius (over-permissive only: a shared ledger can skip a gate, never raise one), and folds it into the same pre-`block` measurement — dispatch one persona twice and confirm the second is gated again. A test pins the half this code owes (distinct ids must not collide), mutation-verified against a scope that ignores `agent_id`; its comment states what it cannot reach, since a role-stable id would leave it green with the separation gone. |
| 3 | Ticket compliance vs #1272 / #1407 / #1409 — "partially compliant", "does not address this ticket". | PR-Agent | **Declined, wrong tickets.** Those numbers were inferred from prose in the PR body, not from this PR's scope. #1272 and #1407 are merged work this PR builds on; #1409 (BUG-771) is an unrelated pre-existing bats failure, filed separately and explicitly out of scope. This PR's ticket is #561. |

**Nothing else was left unread.** `pulls/1410/reviews` = 0 and
`pulls/1410/comments` = 0 were checked directly, not inferred from the issue
comments endpoint — a review-type submission would land there and not show up in
a poll that only reads issue comments.

PR #1411, opened 2026-09-01
feat(setup): make the no-attribution order mechanical and stop holding peer messages

mlorentedev:
## What

Four `settings.json` keys move into the SSOT template (`ai/claude/settings.json`) and both merge policies. Each one replaces an instruction an agent has to remember with a value the deploy enforces.

| Key | Value | Replaces |
|---|---|---|
| `attribution` | `{commit:"", pr:"", sessionUrl:false}` | the no-AI-attribution standing order, enforced by instruction only |
| `crossSessionInbound` | `"accept"` | peer messages held for manual approval, expiring silently |
| `autoCompactEnabled` | `true` | auto-compaction being a per-machine interactive setting |
| `precomputeCompactionEnabled` | `true` | the compaction pause at the end of a long session |

## Why `attribution` is the one that matters

The standing order is that no git or GitHub artifact carries AI attribution. **Claude Code's defaults work against it**: `attribution.commit` and `.pr` default to the standard co-authorship trailer, and `sessionUrl` defaults to `true`, appending a session trailer and a PR-body link. Until now nothing enforced the order but the instruction itself — and an instruction obeyed 95% of the time still writes the trailer into permanent history, where removing it means rewriting published commits.

This is the repo's own "a hook that fires beats an agent that remembers", applied to the rule that has been carried furthest on memory alone.

Whole-object template-wins, **not** the per-key merge `env` gets: a surviving stale subkey would quietly reinstate a trailer. The deprecated `includeCoAuthoredBy` is deliberately not used — the CLI's schema marks it superseded, and `attribution` also covers the PR body and the session link. A guard asserts it never comes back.

## Why `crossSessionInbound`

Measured on this box with four parallel sessions: **4 of 8 peer messages expired unapproved**, including a reply to a peer's direct question sent minutes after they asked. An unapproved message expires silently, so the sender cannot tell it apart from silence — a peer's non-answer and a lost message look identical. Cross-session coordination is either a channel or a coin flip.

Cost, stated plainly: `accept` removes the human review of text other sessions inject into a conversation. On a single-user box every peer is the same person, so the exposure is one user's own sessions. `refuse` and `hold` remain available per machine.

## The trap this change had to avoid

All four are `settings.json` keys, **not** global-config keys — verified against the CLI's own schema region, not assumed. That distinction is this repo's signature silent failure: the merge policy is an allow-list, so a key added to the template and named in neither policy reaches only machines bootstrapped from scratch. It is what happened to `outputStyle`, and the in-repo comment at `setup-linux.sh` says so. The existing parity guard derives its key list *from the template*, so all four are enforced in both policies for free.

## Verification

- **225 tests pass** (`claude-settings-template.bats`, `setup-linux.bats`, `setup-windows.bats`), 4 new.
- **Mutation-verified, three ways.** Dropping `attribution` from the linux `jq`, dropping `crossSessionInbound` from the windows if-chain, and flipping `sessionUrl` back to `true` each turn a guard red with a named failure; restoring turns it green.
- **Verified by consequence, not by inspection.** The *shipped* `jq` expression, extracted from `setup-linux.sh` and run against a `settings.json` that already carried a populated attribution block and `sessionUrl:true`, clears them to `""`/`""`/`false` and flips `hold` to `accept` — while preserving `permissions.additionalDirectories`, machine-local `env`, a user allow entry, and an unrelated top-level key.
- `shellcheck` findings byte-identical to the baseline (99 lines both, differing only in line numbers); `bash -n` and `zsh -n` pass; `setup-windows.ps1` stays ASCII-only (0 non-ASCII bytes).

## Scope

43 production LOC, under the spec-gate threshold, so no spec folder.

**This changes the template only.** The live `~/.claude/settings.json` changes on the next setup run, and sessions already running keep the settings they loaded — so peer messages stay held until each session restarts. `/config` is the immediate path for a running session; this PR is the permanent one.

## SDD skip rationale

The gate fires at 53 production LOC, ten over the threshold. **26 of those 53 lines are comments**, added in this PR to explain why each key is template-wins and what breaks if it is not. The executable change is 26 lines: seven template keys, seven `jq` lines, and seven `ContainsKey` lines — one per key, in a policy that is an allow-list by design.

A spec is for work with a design surface: a contract to agree, a sequence to plan, alternatives whose tradeoffs need recording before code. This has none. Every key is a value in a schema the CLI already defines and validates; there is nothing to design, only to decide, and the decisions are recorded in the commit messages and asserted by guards. The one genuine judgement call — `autoCompactWindow` at 800000 rather than a preference — is pinned as a *relationship* to the model in the template, so it fails loudly if the pairing stops making sense.

Writing a `specs/<id>/` folder here would document a design that does not exist, and the SDD pattern's own value is that a spec folder means something.

**Noted, not acted on:** a change whose only growth was explanatory comments crossed a production-LOC threshold. That is the gate's counting rule working as written, not a defect, and this repo caps meta-work on its own guard machinery — so it is recorded here rather than turned into a ticket.

## Later addition

The three keys in the second commit (`autoCompactWindow`, `autoContinueAtUsageLimit`, `cleanupPeriodDays`) were chosen by the repo owner after the PR opened, from four candidates surfaced while verifying the first four. `alwaysThinkingEnabled` was the one declined: absent already means automatic for models that support it, so pinning it could only ever turn thinking off.

### Note on how spec-gate now passes

`main` was merged into this branch (`ef5d0d1`) after #1410 landed. The gate now reports:

```
[OK] Production diff 366 LOC >= threshold 50 AND spec folder touched in diff
```

**That pass is an artifact of the merge, not a property of this change.** The 366 LOC and the "spec folder touched" both come from #1410, which carried `specs/HARNESS-045-hook-emission/tasks.md`. Nothing in this PR wrote a spec.

The `skip-sdd` label and the rationale above are therefore still the honest account of this change, and they are left in place deliberately — without them a reader sees a green gate citing a spec folder and concludes one was written for these settings.

Recorded as an observation, not a ticket: **merging `main` into a feature branch can satisfy spec-gate by importing another PR's spec touch**, and the branch that did no spec work reads identically to one that did. It is the gate's counting rule as written, and this repo caps meta-work on its own guard machinery, so it stays a note here rather than becoming construction.


mlorentedev:
## Review triage

Triaged 2026-09-01.

|---|---|---|
| PR-Agent | *"No major issues detected"*, *"No security concerns identified"*, PR contains tests, effort 2/5 | Review with zero findings. Nothing to apply. |
| CodeRabbit | *"Review limit reached — next included review available in 41 minutes"* | **Notice, not a review.** No coverage from this reviewer. |
| CI | 10 success, `test-windows` still running at time of triage | A red Windows leg blocks the merge; it fails the whole package on one error. |

**Disclosure: one substantive reviewer, not two.** CodeRabbit's free OSS quota was exhausted, so its comment talks about the review rather than about the diff. Its absence is not a pass. Anyone merging this is merging on PR-Agent alone.

### What no reviewer here can check

Both reviewers read the diff. Neither can verify the claim the diff rests on: that `attribution`, `crossSessionInbound`, `autoCompactEnabled` and `precomputeCompactionEnabled` are `settings.json` keys rather than global-config keys. If any were config-side, the template entry would be a **silent no-op** — the exact failure `outputStyle` had, where the key sat in the template while the deployed file never received it, and every check stayed green.

That was verified out-of-band against the CLI's own schema, and the method is recorded here so a reviewer can refute it rather than take it on trust: in the 2.1.252 binary the settings schema region spans roughly offsets 14.868M–14.955M, bounded by keys that are unambiguously `settings.json`-only — `apiKeyHelper` (14868079) and `awsAuthRefresh` (14868375) at the low end, `crossSessionInbound` (14918697) at the high end, with `includeCoAuthoredBy` (14877703), `hooks` (14884078), `statusLine` (14890829) and `forceLoginMethod` (14896899) in between. `autoCompactEnabled` (14916722) and `precomputeCompactionEnabled` (14916822) fall inside that span, between two confirmed settings keys.

The behavioural check is stronger than the positional one and is in the PR body: the *shipped* `jq` expression, run against a `settings.json` that already carried a populated attribution block and `sessionUrl:true`, clears them and flips `hold` to `accept` while preserving user-owned keys.

### Post-merge acceptance test, verify by consequence

Deploying is not the same as it working. After the next `setup-linux.sh` run:

1. `jq '{attribution, crossSessionInbound}' ~/.claude/settings.json` — the deployed file, not the template.
2. Start a **new** session (a running one keeps the settings it loaded) and send one cross-session message. It arriving with no hold notice is the proof; it also settles the one open question, whether the `consentGated` flag on the config-panel entry means a file-set value needs prior consent. Evidence says it does not.
3. First commit after the deploy: confirm no attribution trailer.

PR #1412, opened 2026-09-01
feat(harness): migrate reviewer to declared severity, and teach the roster guard the new form

mlorentedev:
Follow-up to #1410, which made the gate able to resolve a persona. This makes it
act on one.

## The canary

All 35 persona skills were in the inline frontmatter form, which carries no
severity, so every one parsed as `EnforceUnset` and `dotf harness gate` acted on
none of them. `reviewer` is migrated to the mapping form, **all four skills at
`enforce: warn`** — reported on stderr, never blocking.

The edit landed in the vault, which is the SSOT (`c9b3181c` on knowledge
`master`); `compile-harness.sh --refresh` carried it here. Refresh touched that
one record and nothing else, which was the whole question and was measured
rather than assumed. `dotf doctor` moves from **35 of 35** unmigrated to
**31 of 35**, so the migration has a progress meter instead of a plan.

Severity is an owner decision and was taken as one: only
`verification-before-completion` is a candidate for `block` later. The other
three degrade a review by their absence; they do not invalidate one.

The record's prose is corrected too. It claimed the skills were "enforced by
hook", which had been false for as long as it had been written.

## The guard, updated in the same movement

`specs/HARNESS-046/check-roster-consistency.py` raises `UnreadableSkills` on the
mapping form **by design** — it is HARNESS-046's declared verification
(`features.json:26`) and would have gone red on this change.

It now calls `dotf harness resolve-skills` instead of growing the repository's
third hand-rolled frontmatter reader. That is not an invention: it is what the
guard's own docstring told whoever migrated the definitions to do, so that the
parser stays in one place, on a real YAML parser, behind the CLI.

**The loud failure is relocated, not weakened.** A non-zero exit, a missing
binary and a timeout all raise; nothing degrades to `[]`. An unreadable skill
list and an empty one produce the same downstream behaviour — a comparison of
"no skills" against the roster, which passes — and mean opposite things.

| Case | Result |
|---|---|
| Clean, migrated vault | exit 0 |
| Planted divergence (one skill dropped) | exit 1, both lists named |
| Planted unparseable `skills:` key | exit 1, **naming the YAML error and its line** — better than the old generic regex complaint |
| Restore | exit 0 |
| **Pre-change guard vs. migrated vault** | **exit 1** — the evidence the update was required, not cosmetic |

That last row is the point: AC7's loud failure did exactly what it was built for.
It refused instead of passing silently, so the migration could not disarm the
drift check by accident.

## Canary measured end-to-end

Against the migrated record, six payloads:

| Case | Payload | Result |
|---|---|---|
| A | `agent_type: reviewer`, `Bash` | 4 `[gate] warn` lines, **exit 0** |
| B | no `agent_type` (main thread) | silent, exit 0 |
| C | `agent_type: builder` (unmigrated) | silent, exit 0 — `EnforceUnset` acts on nothing |
| D | `Skill` tool, `skill: audit` | consumption recorded, exit 0 |
| E | same `agent_id`, next `Bash` | 3 warns, `audit` gone |
| F | **same role, different `agent_id`** | all 4 warns again |

F is the reviewer finding from #1410 verified at runtime rather than only in a
unit test: a second dispatch of one role does not inherit the first's
consumption.

## This is instrumentation, not enforcement, and the spec says so

The canary was driven by **hand-written payloads**, which proves the gate's logic
and **not** the field names. `agent_type` / `agent_id` remain
documented-not-measured until `[gate] warn` lines are observed in a genuine
subagent dispatch — and that same observation must confirm two dispatches of one
role carry different `agent_id`s.

**That confirmation is a precondition of promoting anything to `block`**, and it
is now a tracked open item rather than a caveat in prose. Nothing here can block
a session: `warn` exits 0 by construction.

## Verification

```
cd cli && go build ./... && go vet ./... && GOOS=windows go vet ./... && go test ./...
golangci-lint run          # 0 issues
bats tests/*.bats          # 1522 ok, 1 pre-existing failure (BUG-771, #1409)
python3 specs/HARNESS-046/check-roster-consistency.py   # exit 0
```

Lesson 250 records the transferable part: a guard taught to refuse an
anticipated schema change should name its own fix, because the person who reads
the refusal is the person who caused it.

Advances #561.


mlorentedev:
## Review triage

**One substantive reviewer, not two.** PR-Agent reviewed. CodeRabbit returned
*"Review limit reached"* — a notice about the review, not a review. Stated here
rather than left implicit, since a comment from a reviewer is not evidence a
review happened.

CI at `e8c21a5`: **10 of 10 green** (lint, lint-powershell, test, test-windows,
integration, spec-gate, review, attestation, GitGuardian, changes).
`mergeable: true`, state `clean`.

| # | Finding | Disposition |
|---|---|---|
| 1 | **Fragile output parsing** — `resolved_skills` strips brackets and splits on commas; a change to the CLI's output format would "silently produce an empty or malformed list, causing the drift check to either pass silently or fail with a misleading error." | **Half refuted, half applied — `e8c21a5`.** See below. |
| 2 | Ticket compliance vs #1410 / #1409 — "not in this diff", "no test changes". | **Declined.** #1410 is the merged predecessor this builds on; #1409 (BUG-771) is an unrelated pre-existing bats failure, out of scope by statement. The relevant ticket is #561, where the same analysis reads *"partial: the reviewer persona is migrated to mapping form"* — which is exactly what a canary is. |
| 3 | "No relevant tests". | **Declined as stated, addressed in substance.** This change has no unit-testable surface — it is a data migration plus a guard rewrite — and the guard *is* HARNESS-046's declared verification, exercised in five configurations below. |

### On finding 1

**"Pass silently" is measured false.** The CLI was stubbed to emit each
degenerate format and the guard was run against the real vault:

| Stubbed stdout | Before | After |
|---|---|---|
| `["audit","x"]` (JSON) | exit 1, **blamed the definition** | exit 1, blames the CLI |
| `total nonsense` | exit 1, **blamed the definition** | exit 1, blames the CLI |
| `audit\nx` (newline-separated) | exit 1, **blamed the definition** | exit 1, blames the CLI |
| *(empty)* | exit 1, correct message | unchanged — a genuine empty list |
| `[a, b, c]` (documented) | exit 0 | exit 0 |

There is no silent path: a mangled list is still compared against a **non-empty
roster row** and diverges from it. The guard's structure rules the failure mode
out.

**What was real is the second half of the sentence — "fail with a misleading
error" — and it was worse than it sounds.** Every one of those failures printed
`skills diverge` against a vault definition that was *correct*, sending the
reader to edit the wrong file in the wrong repository. So the stdout contract is
now pinned in two steps, because one is not enough: brackets first, then the ids
themselves. A JSON array clears a bracket check and yields `'"audit"'` — and JSON
is the alternative the review named by hand, so it is the one worth closing.

**Nothing here degrades to `[]`.** Non-zero exit, missing binary, timeout,
unexpected shape and non-id contents all raise `UnreadableSkills`. An unreadable
skill list and an empty one produce the same downstream behaviour and mean
opposite things — the reason the exception type exists at all.

### Verified after the change

`python3 specs/HARNESS-046/check-roster-consistency.py` — exit 0 clean; exit 1 on
a planted divergence; exit 1 naming the YAML error on a planted unparseable key;
exit 0 after restore. The vault was left clean.

**Nothing was left unread:** `pulls/1412/reviews` = 0 and `pulls/1412/comments`
= 0 checked directly, not inferred from the issue-comments endpoint.

mlorentedev:
## Review triage — addendum (re-review after `e8c21a5`)

PR-Agent **re-reviewed** after the previous push and grew from one focus area to
three. The queue did not catch it: **`dotf pr triage-queue` compares timestamps,
not content**, and my triage comment was newer than the bot's, so it read exit 0
with two findings unread. Filed as an open thread; a guard that answers "clear"
while findings sit unread is the same shape as a vacuous pass.

| # | Finding | Disposition |
|---|---|---|
| 4 | **Silent failure on missing binary** — `resolved_skills` catches all exceptions and raises `UnreadableSkills`; the message "could be more specific about the binary location". | **Declined.** The finding's own text answers it twice — *"This is correct behavior"* and *"is properly handled"*. The message already carries the binary name: `FileNotFoundError` stringifies as `[Errno 2] No such file or directory: 'dotf'` and is interpolated verbatim. The broad `except` is deliberate and is the opposite of a silent failure — it exists so that a missing binary, a timeout, and anything unforeseen all become a loud raise rather than an empty list. |
| 5 | **Documentation inconsistency** — the record now says "watched by hook" and then adds that all four are at `warn`, which is "not truly enforced in the blocking sense yet". | **Reframed and recorded — `3eef03e`.** The wording is accurate as written: `warn` *is* watching, and the paragraph says so and says why. But checking the claim across all seven definitions surfaced a real defect the finding did not name: **`architect`, `builder`, `curator`, `planner` and `shipper` each still state "Your phase's skills are enforced by hook" while every one of their skills is `EnforceUnset`** — neither enforced nor warned about. That claim was false for all seven before this PR; `reviewer` is now the only honest record, so the asymmetry is exposed, not introduced. |

### Why finding 5 is not fixed in this PR

Not as a word swap, and not here. For an unmigrated persona neither "enforced"
nor "watched" is true, so the honest wording differs from both and would be
rewritten again the moment that role's severity is decided. Correcting it
per-role, in the commit that migrates it, costs one edit instead of two —
`reviewer`'s record is the pattern. Recorded against the remaining-31 item in
`specs/HARNESS-045-hook-emission/tasks.md` rather than left in a comment.

It is worth naming for what it is: doctrine asserting an enforcement nobody
performs — the exact defect class this gate exists to end.

### State at `5de3e9d`

Merged `main` after #1413 landed. One conflict, `docs/lessons/_index.md`: #1413
took lesson **250**, so this PR's yields to **251** — theirs is merged and
renumbering it would break every link already pointing at it. Verified rather
than assumed: 251 files, 251 index rows, no duplicate numbers, every row resolves
to a file and every file appears in the index.

Re-verified on the merged tree: `go build` + `vet` + `GOOS=windows vet` +
`go test` green; `check-roster-consistency.py` exit 0; bats **1531 ok**, 1
pre-existing failure (BUG-771, #1409).

**One substantive reviewer again.** CodeRabbit returned "Review limit reached" on
this PR, so its coverage here is a notice, not a review.

PR #1413, opened 2026-09-01
fix(setup): restore setup-linux.sh executable bit and guard the invocation

mlorentedev:
## The defect

#1411 shipped `setup-linux.sh` at **`100644`**. `install.sh:52` is `exec ./setup-linux.sh "$@"` — the bootstrap entry point — and `README.md` documents the same invocation. A fresh clone fails at its first step with exit 126. Machines already installed are unaffected, which is why it reached `main` unnoticed.

Mine, and measured rather than deduced:

```
git diff df303cd..ea13c9a -- setup-linux.sh
old mode 100755
new mode 100644
```

## How the bit was lost — not by editing the file, but by verifying the edit

The mutation-testing convention in this repo breaks a file, watches the guard go red, and restores it. Written the obvious way:

```sh
grep -v 'pattern' setup-linux.sh > /tmp/x && mv /tmp/x setup-linux.sh   # new file at the umask: 644
cp /tmp/scratch/sl.bak setup-linux.sh                                    # cp keeps the DESTINATION's mode
```

`cp` onto an existing file writes through it and keeps the destination's mode, which was already `644`. **The content came back byte-for-byte and the mode did not.** The restore looked total because content was the only property compared.

## Why four layers of checking could not see it

| Layer | Why it is blind |
|---|---|
| `git diff` | mode lives in a header, and review reads the content |
| pre-commit | full suite ran and passed — nothing asserted a mode |
| `ci.yml` | shellcheck only, which is mode-blind |
| integration | `tests/Dockerfile.integration:83` runs `bash setup-linux.sh` — an **explicit interpreter works at any mode**, so the job exercises the script without ever exercising the invocation that broke |

`install.sh` stayed `100755` throughout. The contrast was inside the repo the whole time and nothing compared the two.

## The guard

`tests/executable-bit.bats` asserts every script invoked as `./x.sh` is `100755` **in git's index** — what a fresh clone materialises and what a commit records; a working tree can be executable while the shipped file is not.

Two deliberate choices:

- **The list is derived from the invokers** (`install.sh`, `README.md`, `docs/`), never hand-maintained — the reasoning already recorded at `tests/claude-settings-template.bats:36`: a hand-written list cannot catch the entry point nobody thought to add to it. It currently resolves 9 scripts, not just this one.
- **A second test asserts the derivation is non-empty**, and mutation proves it earns its place: with a regex matching nothing, the main test **passes vacuously** and only the meta-test fails. A derived list makes that failure mode reachable in a way a literal list does not.

A blanket "every `.sh` is executable" would be the wrong rule: of 44 tracked shell files, four are libraries that are sourced rather than executed (`.zsh/functions.sh`, `git-hooks/lib/board-pickup.sh`, two `tests/golden/*/lib.sh`).

## Verification

- 3 tests pass. **Two mutations:** re-dropping the bit fails with the file and mode named in the message (`setup-linux.sh(100644)`); an empty derivation fails only the meta-test, demonstrating the vacuous pass.
- `./setup-linux.sh --help` → exit 0 (was 126).
- Lesson 250 records the restore asymmetry, because the mechanism will recur on the next mutation of any executable file.

## Credit

Found by a peer session reading the merge diff of its own PR, not by any automated check. It offered to fix it and was asked to stand down so the root cause — which only this side could see — travelled with the fix.


mlorentedev:
## Review triage

Triaged 2026-09-01. **Two substantive reviewers this time**, both with real findings. All applied in `9af7c80`.

|---|---|---|---|
| 1 | CodeRabbit | The derivation scans all of `docs/`, so **lesson 250's own prose** supplies `setup-linux.sh` — the meta-guard can be satisfied by history after the live invocations are gone | **APPLIED.** Confirmed by measurement before fixing (below). |
| 2 | CodeRabbit | `grep -qE 'exec[[:space:]]+\./setup-linux\.sh'` is satisfied by a commented-out line while `bash setup-linux.sh` is the line that runs | **APPLIED.** Anchored to line start. |
| 3 | PR-Agent | Test 1 passes vacuously on an empty derivation; it leans on a *separate* test to notice | **APPLIED.** Test 1 now counts what it saw and fails on zero. |
| 4 | CodeRabbit | markdownlint MD058: lessons table needs a blank line after it | **APPLIED.** My row 250 made it the boundary. |
| — | CI | 12/12 green on `136dc72` | Re-running on `9af7c80` |

### Finding 1 was correct and worth stating plainly

**The guard committed the mistake it was written to catch.** Measured on `136dc72`, before the fix — rewrite every live invocation to `bash setup-linux.sh`, leave the lesson untouched:

```
ok 2 the guard actually finds the bootstrap entry point     <- GREEN, fed by lesson 250's prose
```

After `9af7c80`, same mutation:

```
not ok 2 the guard actually finds the bootstrap entry point
```

The exclusion is `docs/lessons/` only; `docs/runbooks/` deliberately stays in. That distinction is the substance: a runbook telling a user to type `./setup-linux.sh` **is** an invocation contract — six of them do, and they should keep the guard alive on their own — while a lesson describing an incident is history. **Evidence that mentions an invocation is not evidence that one happens.** Tracked-file filtering came with it, which also drops placeholders like `./x.sh` that prose mentions.

### A correction to my own earlier measurement

My first attempt to reproduce finding 1 rewrote only `install.sh` and `README.md` and reported the hole as still open. That was wrong: six runbooks also document the invocation and legitimately kept the guard green. The reproduction that actually isolates the defect strips **every live invocation** and leaves only the lesson. Stated because the first reading would have justified a wrong fix — pruning all of `docs/`, which would have thrown away six real invocation contracts.

### Verification

Four mutations, each red then restored green: every live invocation stripped leaving only the lesson (meta-test red, green before this commit); `exec` commented out with a `bash` form active (red); the executable bit dropped again (red, `setup-linux.sh(100644)` named); all restored (3/3 green).

Nothing declined, nothing ticketed — every finding was real and in scope.

mlorentedev:
## Review triage — addendum

The `review` check on `9af7c80` was **cancelled** (run 33467604894, created 03:49:31Z, cancelled 04:04:50Z after ~15 min). So the commit that applied the earlier findings was, for a while, **unreviewed** — and the PR still read as healthy. Re-run manually; it completed and updated the review at `04:08:10Z`.

**Verdict on `9af7c80`: "No major issues detected", no security concerns, tests present, effort 3/5.** One item raised, declined below.

|---|---|---|
| PR-Agent | *"Ticket compliance analysis — #1411 Partially compliant"*, listing six unimplemented requirements (the `attribution` / `crossSessionInbound` / auto-compact keys and their merge policies) | **DECLINED — wrong ticket inferred.** Those requirements belong to #1411, which is **already merged** (`a7364e1`). This PR fixes a defect that merge introduced; it was never implementing them. The reviewer inferred the ticket from prose in the commit body. |

A peer session reported the same failure mode on #1412 hours earlier — PR-Agent inferring `#1272`/`#1407`/`#1409` from body prose when the real ticket was `#561`. Two independent instances make it a pattern worth naming rather than a one-off: **a reviewer that resolves tickets by reading prose will confidently audit a PR against the wrong contract.**

### Why this addendum exists at all

`dotf pr triage-queue` returned **exit 0** for this PR the whole time. It compares timestamps, and PR-Agent **edits its comment in place** — `created_at` stayed at `03:38:13Z` while `updated_at` moved to `04:08:10Z` and the body changed from a "Vacuous Pass Risk" focus area to a ticket-compliance audit. A queue keyed on creation time cannot see that, so a clean exit is not evidence that nothing new arrived.

Found by checking by hand after a peer session flagged the timestamp-vs-content gap. Recorded here rather than ticketed, per this repo's cap on meta-work against its own guard machinery — but it is the same shape as the defect this PR's own guard had: **a check that reports "clean" when it has not actually looked.**

PR #1414, opened 2026-09-01
fix(deploy): agy owns the settings it writes, so the deploy stops replacing them

### LINKED ISSUE #1390: AI-043: dotf deploy destroys agy's runtime-written settings, and the deploy that fixed the skills already did it
## What happened

`dotf deploy` destroyed agy's runtime-written state on this box. This is a **measured incident**, not a
prediction: the deploy inside `setup-windows.ps1` at **2026-08-29 21:39** overwrote
`~/.gemini/antigravity-cli/settings.json`, and the runtime-granted keys are gone.

AI-042's **round-3 adversarial review predicted exactly this at 2026-08-29 03:31** and returned
**FAIL** on it (REAL Blocker, "data-loss / wrong decision recorded"). The issue that carried AI-042
(#1334) was auto-closed at 03:20 by `Closes #1334` in #1373, ~18 hours **before** the defect it
describes actually fired.

## Root cause

`agy-settings` is declared `replace` in `ai/deploy.json`, over a file **agy writes at runtime**.
The manifest treats the deployed path as an output; agy treats it as read-write state. Both are true,
so a `replace` entry silently discards whatever the tool recorded since the last deploy.

`merge` is **not** a drop-in fix. `mergeInto` writes whole top-level keys, so `permissions` and
`trustedWorkspaces` would still be overwritten wholesale — the review says this explicitly.

## Evidence

Measured by the round-3 review (2026-08-29 03:31), on the live file:

- 5 runtime-granted `mcp(hive-vault/*)` entries in `permissions.allow`
- 1 runtime-trusted workspace: `fae-onboarding-webapp`
- `dotf deploy --dry-run` -> `would deploy agy-settings`; `dotf doctor` -> `[WARN] drift: agy-settings`
- `git show main:ai/agy/settings.json` contains none of those 6 values

Re-measured after the 2026-08-29 21:39 deploy:

- live file mtime `2026-08-29 21:40:56` (inside the setup run)
- `permissions.allow` = 5 entries, **zero** matching `hive-vault`
- `trustedWorkspaces` = 2 entries, `fae-onboarding-webapp` **absent**
- `dotf deploy --dry-run` -> `in sync   agy-settings`

The transition from `would deploy` + `[WARN] drift` to `in sync`, with the six values absent, is the
overwrite. **"In sync" is the symptom here, not a health signal** — it reports that the runtime state
was successfully discarded.

## Why the test suite locks in the wrong contract

`TestDeploy_PathsComposeWithReplace` **asserts the destruction**:

```go
if _, stale := got["stale"]; stale { t.Errorf("replace must not keep a foreign key") }
```

So the suite currently pins the behaviour that loses the data.

## Scope

The owner decision named by the review, unchanged:

1. stop shipping agy's runtime-written keys, **or**
2. manage only keys the tool never writes, **or**
3. leave agy's settings to the shells entirely

Then: invert the replace assertion for this entry into a "does not clobber runtime keys" property test,
and retract the false measurement in `specs/AI-042-trusted-folders-render/verification.md`
("no key added by agy after daily use; agy reads, it does not write") — the artifact on the same box
refuted it ~40 minutes after it was written.

## Related

- Spec `specs/AI-042-trusted-folders-render` is `status: implementing` with a **FAIL** round-3 verdict;
  its three fix commits (`1c658eb`, `4e977df`, `2ab60a9`) are on `chore/archive-ai042` and **not in main**.
- Round 3 also carries a second REAL Blocker (trailing-data rejection silently removed from both JSON
  paths by the round-2 `UseNumber` switch) — tracked with the spec, not here.
- #1334 was closed by `Closes #1334` in #1373 while the spec was still `implementing`.

mlorentedev:
Closes #1390

## The defect

`agy-settings` was `replace` over a file **agy rewrites at runtime**. Every deploy discarded whatever the user had trusted or granted since the last one — and `dotf doctor` reported the resulting difference as drift, with the remedy `run: dotf deploy agy-settings`. **The repair instruction was the data loss.**

`replace` is the *default* when no `strategy` is declared, so the destructive setting is the one you get by writing nothing.

## It happened again while this fix was being written

Two readings of the same path, `~/.gemini/antigravity-cli/settings.json`, hours apart on 2026-09-01:

| | `trustedWorkspaces` | `permissions.allow` |
|---|---|---|
| earlier | 3 entries, incl. `/home/manu/Projects/ts-bridge` | 4 entries |
| after (mtime `21:33:30`) | 2 entries, **`ts-bridge` gone** | 5, gained `search_web(*)` |

The file moved **toward the template in both directions** — losing a runtime entry *and* gaining a template one. That is the signature of a replace, not of agy writing.

**The author is not identified and this PR does not claim one.** A peer session ruled itself out with evidence, including that `verify-setup.bats:8-9` skips its `bash setup-linux.sh` without `DOTFILES_INTEGRATION_TEST`. The loss is proven by the two readings; who triggered it is not.

*Correction, made before this was written:* an earlier draft said the live file was "byte-identical to the template". **That is false** — the shas differ (`68f757b4…` vs `93e47a96…`) because the template carries unexpanded `{HOME}`. What was actually measured is semantic equality **after `{HOME}` substitution**. Same conclusion, but the original wording was refutable with one `sha256sum`, and a peer's challenge to the timeline is what prompted re-checking it. (That challenge measured `~/.gemini/settings.json` — a different file, one key, untouched since 2026-08-12. The manifest targets `~/.gemini/antigravity-cli/settings.json`.)

## The fix: two edits, neither sufficient alone

1. **`ai/deploy.json`** — the entry declares `"strategy": "merge"`.
2. **`ai/agy/settings.json`** — `permissions` and `trustedWorkspaces` removed.

`mergeInto` writes whole **top-level** keys, so merge with those keys still present overwrites them wholesale — which is why "just switch to merge" was recorded on the issue as *not* a drop-in fix. Removing them is what makes merge sufficient: `mergeInto` only touches keys the **source** names, so what the destination holds and the template does not name survives untouched.

## Stated cost — measured, not assumed

On a **fresh** machine the template now deploys **9 keys instead of 11, with no `permissions` at all**, so agy's `deny` defaults (`command(rm -rf /)`, `command(rm -rf ~/*)`) no longer reach a new box.

Under shallow `mergeInto` there is no way to ship deny-only without clobbering the runtime allow list. The union-merge work on **#1334** is what lets both coexist again; until it lands, not destroying the user's grants is the side worth being on. This is a deliberate trade, made with the owner, not an oversight.

## Verified by consequence

In sandbox `HOME`s, with the **same binary** and only the manifest and template differing — because `dotf` reads the manifest from the repo, so comparing two *builds* would have compared nothing. (A first attempt did exactly that and produced a meaningless pass.)

```
main today   ->  ts-bridge DESTROYED,  runtime mcp(hive-vault/*) DESTROYED
this change  ->  both survive; model / ui / safety still managed by dotfiles
```

## Guards, mutation-verified

- **`agy-settings` is `merge`, never `replace`** — reverting to the implicit default fails with `strategy is 'replace(default)'; runtime state is destroyed on every deploy`.
- **The template ships no key agy writes** — re-adding `trustedWorkspaces` fails with `template carries 'trustedWorkspaces' … merge overwrites it wholesale`.

## A test that pinned the destructive contract

`tests/copilot-config.bats` asserted `… replace slash` for this entry. As the issue put it, **the suite was asserting the data loss**. That assertion is inverted rather than deleted, and its sibling claim — that agy's trust list is a `{HOME}` template — is now vacuous by a *stronger* one: the template ships no trust list at all. Both claims moved to `tests/antigravity.bats`, where the change lives. **Copilot's half is untouched.**

## Deferred, not dropped

The issue also asks to retract a false measurement in `specs/AI-042-trusted-folders-render/verification.md` (*"agy reads, it does not write"*). **That text is not reachable from `main`** — grepped and absent; it lives on the unmerged `chore/archive-ai042` branch, which carries a round-3 FAIL verdict and belongs to #1334. Retracting it there is that ticket's work, and this PR deliberately does not touch that branch: partial-landing a FAILed spec's commits would wreck its re-review provenance.

22 production LOC, under the spec-gate threshold.


mlorentedev:
## Review triage

Triaged 2026-09-01. Five points from PR-Agent; **two applied, two declined with evidence, one disclosed as unverified.** CodeRabbit returned *"Review limit reached"* — a notice, not a review, so **one substantive reviewer, not two.**

| # | Finding | Disposition |
|---|---|---|
| 1 | **"Vacuous Test"** — the `strategy == merge` guard is static config validation; it passes even if the engine misreads the strategy | **APPLIED** (`e8a0c08`) |
| 2 | **"Strategy Mismatch"** — if `permissions` is ever re-added to the template, merge silently destroys the runtime `allow` list | **APPLIED** — now a hard test failure, not a comment |
| 3 | Invert `TestDeploy_PathsComposeWithReplace` | **DECLINED — the test does not exist on `main`** |
| 4 | Retract the false measurement in `specs/AI-042-.../verification.md` | **DECLINED — same reason**; deferred to #1334 |
| 5 | Confirm agy re-populates missing keys on launch | **UNVERIFIED, disclosed below** |

### 1 and 2 were correct and are now enforced

The reviewer was right that the bats guard says nothing about behaviour. `TestDeploy_MergePreservesUnmanagedKeys` covers the engine property, but with the **copilot** config — so "therefore agy's runtime keys survive" was an inference spanning two tests with nothing joining them, and **the join is what lost the data**.

`TestDeploy_AgySettingsPreservesRuntimeKeys` loads the shipped `ai/deploy.json` and `ai/agy/settings.json` and deploys the real entry into a temp `HOME`. Mutation-verified against all three ways the loss returns:

```
entry regresses to replace          -> both destroyed, named in the failure
template re-adds permissions        -> the nested grant destroyed   <- finding 2
template re-adds trustedWorkspaces  -> the list entry destroyed
```

It also pins the reverse direction: the template's `model` must still win. A merge that preserved everything by doing nothing would satisfy the preservation assertions alone.

### 3 and 4 are declined on the same measured fact

Both cite artefacts that **are not reachable from `main`**:

```
grep -rn 'PathsComposeWithReplace' cli/          -> no match
git grep -l 'PathsComposeWithReplace' origin/chore/archive-ai042
    -> cli/internal/deploy/paths_test.go
```

Issue #1390 was written while `chore/archive-ai042` was in view, so it names that branch's test and that branch's `verification.md`. Neither exists here. This PR deliberately does not touch that branch: it carries a round-3 **FAIL** verdict and belongs to #1334, and partial-landing a FAILed spec's commits would wreck its re-review provenance.

The intent behind finding 3 *was* satisfied, in the place where it applies: `tests/copilot-config.bats` asserted `… replace slash` for this entry — the suite pinning the destructive contract — and that assertion is inverted rather than deleted.

### 5 is a real gap and is not closed

**Not verified: whether agy behaves correctly on a fresh machine that receives no `permissions` key at all.** The template now deploys 9 keys instead of 11. If agy needs the key rather than defaulting it, a brand-new box could differ from an existing one — and every box in use already has the key, so this cannot be observed here.

Cheapest check for whoever has a clean machine: deploy, launch agy, then `jq 'keys' ~/.gemini/antigravity-cli/settings.json`. If agy writes its own defaults back, the gap closes itself.

Stated rather than assumed because it is the one claim in this PR with no measurement behind it.

### On the reviewer

Findings 3 and 4 come from PR-Agent auditing the PR against issue prose, and it inferred a second ticket (#1334) too. On #1413 the same reviewer audited against #1411's requirements — a **merged** issue this PR was fixing the fallout of — and a peer session saw it infer `#1272`/`#1407`/`#1409` on #1412 when the real ticket was `#561`. Three independent instances: **a reviewer that resolves tickets by reading prose will confidently audit against the wrong contract.** Worth weighing its ticket-compliance section accordingly; its code findings here (1 and 2) were both real and both worth applying.

mlorentedev:
## Review triage — addendum

PR-Agent re-reviewed `e8a0c08` and **updated its comment in place at `04:16:21Z`**, one minute after the triage above was posted (`04:15:02Z`). `dotf pr triage-queue` reported **exit 0** throughout, because it compares timestamps and the comment's `created_at` never moved.

**New verdict: "No major issues detected."** Both code findings from the first pass are resolved, and the reviewer now lists *"Invert the replace assertion into a property test"* as **compliant**, naming `TestDeploy_AgySettingsPreservesRuntimeKeys`.

Two items remain listed as non-compliant, both already dispositioned above and unchanged:

| Item | Why it stays |
|---|---|
| Retract the false measurement in `specs/AI-042-.../verification.md` | Not reachable from `main` — it lives on the unmerged `chore/archive-ai042`, which carries a FAIL verdict and belongs to #1334 |
| `ai/agy/settings.json` is no longer a `{HOME}` template | Deliberate. The reviewer's own wording — *"different approach, tracked under the spec's own lifecycle"* — is the accurate reading |

Nothing to apply. Recorded so the PR does not carry a triage that predates its current review.

**This is the second PR today where a clean queue hid new reviewer content** (#1413 was the first, where the `review` check had also been silently cancelled). Same mechanism both times: the reviewer edits rather than appends. A queue keyed on creation time cannot see it, so **exit 0 means "nothing newer was created", not "nothing new was said".** Checking by hand is currently the only way to know.

PR #1405, opened 2026-08-31
chore(main): release 0.53.0

mlorentedev:
:robot: I have created a release *beep* *boop*
---

### Features

* **cli:** HARNESS-105 refactor shellQuote and implement JSON Error Latches ([#1404](https://github.com/mlorentedev/dotfiles/issues/1404)) ([d4ea0f5](https://github.com/mlorentedev/dotfiles/commit/d4ea0f571e3e81daf25847b121e9d2e703e18f42))
* **harness:** bind hooks through dotf, removing both setup scripts' hooks writers ([#1407](https://github.com/mlorentedev/dotfiles/issues/1407)) ([fdbe17c](https://github.com/mlorentedev/dotfiles/commit/fdbe17c8cdeac29837a56cddb3bc4bb60219a1a6)), closes [#561](https://github.com/mlorentedev/dotfiles/issues/561)
* **harness:** migrate reviewer to declared severity, and teach the roster guard the new form ([#1412](https://github.com/mlorentedev/dotfiles/issues/1412)) ([cd7416f](https://github.com/mlorentedev/dotfiles/commit/cd7416ff64c99aec15f137b527c55b20455123d9))
* **harness:** the gate resolves its persona from the payload, and doctor reports what it cannot enforce ([#1410](https://github.com/mlorentedev/dotfiles/issues/1410)) ([2dbaa06](https://github.com/mlorentedev/dotfiles/commit/2dbaa064d23ba79bc859816817bb6a8b9c7d272d))
* **setup:** make the no-attribution order mechanical and stop holding peer messages ([#1411](https://github.com/mlorentedev/dotfiles/issues/1411)) ([a7364e1](https://github.com/mlorentedev/dotfiles/commit/a7364e1a20e7691de78359802c51ba878ffdec7a))

### Bug Fixes

* **deploy:** agy owns the settings it writes, so the deploy stops replacing them ([#1414](https://github.com/mlorentedev/dotfiles/issues/1414)) ([6e0d180](https://github.com/mlorentedev/dotfiles/commit/6e0d180ae57c24148fbe3e74af7c7f75d86c40f1))
* **setup:** restore setup-linux.sh executable bit and guard the invocation ([#1413](https://github.com/mlorentedev/dotfiles/issues/1413)) ([fd02554](https://github.com/mlorentedev/dotfiles/commit/fd02554e3528781ecd563eaae5f2549c0dd1254d))

---
[...]

PR #1416, opened 2026-09-02
feat(doctor): the binary carries the commit it was built from, so a stale dotf stops reporting health it never established

mlorentedev:
First PR of the `dotf` CLI migration drive. It is first on purpose: until it lands, every `dotf doctor` used to verify a later migration PR can pass on a check that is not in the binary.

## The defect, measured today

```
$ dotf version                     →  dotf version 0.52.0
$ grep DOTF_VERSION versions.conf  →  DOTF_VERSION=0.52.0     # they agree
$ ls -l $(command -v dotf)         →  29 Aug 21:59            # and it is two feature merges stale
```

The consequence — same command, two binaries:

```
$ dotf doctor | grep -A2 'Persona skill enforcement'
(nothing — the section does not exist)

$ go build -o /tmp/dotf-main ./cmd/dotf && /tmp/dotf-main doctor
[Persona skill enforcement]
  [WARN] 31 of 35 persona skills carry no `enforce:` severity ...
```

`dotf doctor` rendered `Results: 158 passed, 2 failed` with an entire check section **absent**, because the binary predated #1410. A report missing a check cannot be told from one where the check ran and passed.

`checkOptionalTools` cannot see this: it compares the reported semver to the `versions.conf` pin, which is the right question for a third-party release and the wrong one for the single tool that **is this repository**. Between two releases the version string is constant while the source moves every merge.

## What changed

- `.goreleaser.yaml` stamps `-X main.commit={{ .FullCommit }}`; `dotf version --commit` prints that bare value.
- A new `dotf provenance` doctor section compares it to HEAD, scoped to `cli/`.
- `doctor.Config` exposes `RepoDir`, which `loadConfig` already computed and discarded.

## Two decisions worth reviewing rather than skimming

**The commit is behind a flag, not on a second line of `dotf version`.** Both installers regex `(\d+\.\d+\.\d+|dev)` out of the *merged* streams to decide whether to skip installing. That skip failing open is silent — every machine would reinstall on every setup run. The default output is now pinned to one line by a test, and `tests/install-dotf.bats` still passes 9/9 including the idempotence case.

**The pathspec is `:(top)cli`, not `cli`.** A bare pathspec resolves against git's CWD:

```
from repo ROOT:   git rev-list --count 11a68b1..HEAD -- cli  → 4
from cli/ subdir: same command                               → 0
```

The `0` is the *clean* answer. The check is correct today only because it passes `RepoDir` as git's CWD; had that shifted, the guard would have reported health it never established — the exact defect it exists to detect, inside its own fix. Proven fail-first, and written up as `docs/lessons/lesson-252`.

## WARN, never FAIL

Running a released binary from inside a checkout is legitimate and is what most sessions here do. The defect is the drift being invisible, not the drift. Where the question cannot be answered — outside a checkout, no `dotf` on PATH — the check SKIPs rather than passing, per C15. All 10 test cases assert `Failures() == 0`.

## Sequencing — please read before merging

- **This must not merge before #1405.** Doing so lands it in v0.53.0's changelog and re-dirties a release that is already verified and triaged.
- **After installing v0.53.0 this check will report `predates the build stamp`, and that is correct.** #1405 tags v0.53.0 at `6e0d180`, which does not contain this change. The first release that can answer provenance is v0.54.0.
- `Refs #1158`, not `Closes`. `check-spec-gate.sh:394` refuses a PR that closes an issue whose spec is still active. GUARD-007 archives and the issue closes in a follow-up, after `dotf spec review` returns a passing independent verdict.

## Verification

| | |
|---|---|
| `go build ./...` / `go vet ./...` | clean |
| `GOOS=windows go vet ./...` | clean (the leg that broke on #1075) |
| `go test ./...` | 22 packages ok, 0 FAIL |
| `golangci-lint run` | 0 issues, v2.12.2 = the `versions.conf` pin |
| `bats tests/install-dotf.bats` | 9/9 |
| `features.json` | all 6 verification commands executed, all exit 0 |

End to end against a **real** binary stamped with a real commit 7 back — not a fake:

```
[WARN] deployed dotf is 4 cli/ commit(s) behind HEAD (built 11a68b1af73c, HEAD 6e0d180ae57c)
```

Against the actually-installed binary:

```
[WARN] deployed dotf (/home/manu/.local/bin/dotf) does not understand `version --commit`,
       so it predates the build stamp and its provenance CANNOT be established
```

Full transcript including the fail-first proof: `specs/GUARD-007-deployed-dotf-drift/verification.md`.

## Not verified, disclosed

- **The stamp on a real goreleaser build**, since v0.53.0 has not published. Checked against a locally stamped binary using the same ldflags string; `goreleaser snapshot` runs on every PR touching `cli/**` and exercises the template here.
- **Windows at runtime.** Compilation only; no Windows box this session.

## Credit

A parallel session found that `install-dotf.sh` installs a *published artifact* pinned by `DOTF_VERSION` rather than building from the checkout — which is what makes merging #1405 the precondition for the whole orchestrator binding, and which reversed my reasoning about when to cut the release. They also spotted that #1405's `review-attestation` passed on CodeRabbit's *"No actionable comments were generated"* while CodeRabbit itself reported *"Review skipped"*; that is now evidence on #906 (GUARD-002 already specs it), and #1405's triage thread records that the PR is unreviewed.

Refs #1158


mlorentedev:
## Review triage

**Reviewer: CodeRabbit, and this was a real review** — 3 actionable comments plus 1 nitpick, naming files and lines. Worth stating explicitly because the last PR I triaged (#1405) had `review-attestation` go green on a *notice* rather than a review; this is the other case, and the distinction is content, not author.

Dispositioned in `15b80ac`. Three applied, one declined.

### Applied

**1. The version contract test permitted the drift it names.** CodeRabbit's nitpick, and the most valuable finding here. The assertions did `strings.TrimRight(stdout, "\n")` before counting lines, so `"dotf version dev\n\n\n"` passed a test whose entire purpose is to pin the output at **one line**. Both assertions are now exact, trailing newline included.

Proven fail-first by adding a stray `\n` to the format string:

```
stdout_contract_test.go:155: default `dotf version` must be exactly "dotf version dev\n"
                             (the installers regex it); got "dotf version dev\n\n"
FAIL
```

A test that permits the thing it names is the same class of defect as the check this PR adds — a signal that reports a property it never established. Good catch.

**2. `features.json` invoked bats through a hard-coded `~/.local/bin/bats`.** Correct: unresolvable on another machine. Now plain `bats`. Verified it resolves from a **non-login** shell here (`/home/manu/.local/bin/bats`), and CI installs bats-core onto PATH (`ci.yml:200`), so both consumers work. All six verification commands re-executed after the change; all six exit 0.

**3. `docs/lessons/_index.md` MD058** — a list item glued to the last table row with no blank line. **Pre-existing**: `git diff HEAD~1` confirms the stray line predates this branch, and I only added a row above it. Fixed here anyway — it sits on the line my row touches and renders wrong.

### Declined

**Splitting `checkDotfProvenance` into helpers for a 40-line limit.** No such rule exists in this repository, verified rather than assumed:

```
$ golangci-lint config path
level=warning msg="No config file detected"
$ golangci-lint run
0 issues.
```

No config file, so the default linter set applies, and `funlen` is not in it. The function is a flat chain of guard clauses, one per row of the proposal's state table, each producing a distinct report and returning. Extracting them into helpers would scatter the state machine that *is* the check across three functions and make the exhaustiveness harder to read, not easier. Reopen if the repo later adopts a length limit — this would be a legitimate target then.

### CI

All green at `c6de5d3` (`test-windows` was the last to land). Re-verified locally on the merged tree at `48de860`: `go build` / `go vet` / `GOOS=windows go vet` clean, 22 packages ok, `golangci-lint` 0 issues.

### Sequencing note — the precondition cleared while this was in review

**#1405 merged and v0.53.0 is published.** `main` now carries `DOTF_VERSION=0.53.0`, and this branch has been merged up to it. The constraint in the PR description — *"must not merge before #1405"* — is satisfied.

Restating the one thing that will look like a bug and is not: **after installing v0.53.0, this check reports `predates the build stamp`, correctly.** v0.53.0 was tagged at `3740acf`, which does not contain this change. The first release that can answer provenance is v0.54.0.

**Still `Refs #1158`, not `Closes`.** GUARD-007 archives and the issue closes in a follow-up, after `dotf spec review` returns a passing independent verdict — `check-spec-gate.sh:394` refuses a PR that closes an issue whose spec is still active.

Issue #1418, opened 2026-09-02
CLI: setup deploys from the cwd while doctor compares against DOTFILES_REPO_DIR, and one of the printed remedies downgrades the binary below the gate


mlorentedev:
## Problem

`setup-linux.sh` deploys **from the cwd**; `dotf doctor` compares the deploy dir **against `$DOTFILES_REPO_DIR`**. In a repository whose own standing rule is to work in worktrees, those are routinely different trees, so the drift check goes red on a machine that is correctly deployed and stays red.

```
setup-linux.sh:60   safe_copy "$CURRENT_DIR/versions.conf" "$DOTFILES_DIR/"     ← source is the cwd
checks_deploy.go:886  repo := resolveRepoDir(sys)                               ← $DOTFILES_REPO_DIR first,
checks_deploy.go:936  if r := sys.Getenv("DOTFILES_REPO_DIR"); ... return r        cwd walk-up only as fallback
```

Measured 2026-09-02 on msi, after a setup run from a worktree:

```
[FAIL] dotf version drift: installed=0.53.0 pinned=0.52.0 (run ./scripts/install-dotf.sh)
[FAIL] drift: scripts/obs-cli.ps1 — repo differs from deploy-dir (run setup to refresh ~/.dotfiles)
[FAIL] drift: versions.conf  — repo differs from deploy-dir (run setup to refresh ~/.dotfiles)
```

Three FAILs, one root cause. The deploy dir is byte-identical to the worktree setup ran from:

```
sha256 <worktree>/scripts/obs-cli.ps1  = db10f8e8…   ← equals ~/.dotfiles
sha256 $DOTFILES_REPO_DIR/…            = 235b4029…   ← the tree doctor compares against
```

## Why it is more than a red line

**One printed remedy regresses the machine.** `run ./scripts/install-dotf.sh` is emitted because `pinned` was read as `0.52.0` from the stale root. Following it installs **0.52.0 over 0.53.0**, taking the binary below #1407 and leaving the `PreToolUse` gate hook in `~/.claude/settings.json` pointing at a `dotf` with no `harness bind` — re-inerting the orchestrator that #1405/#1407 just made live.

The other remedy, *"run setup to refresh ~/.dotfiles"*, re-deploys the same worktree and cannot converge either: the comparison root is never the tree being deployed.

[...]

PR #1423, opened 2026-09-02
test(gitattributes): assert the working tree obeys the eol it declares

mlorentedev:
## What

`tests/gitattributes-eol.bats` proved a **rule exists**. Nothing proved the **working tree obeys it**. This adds the missing half.

## The defect it would have caught

`scripts/obs-cli.ps1` resolves `eol: crlf` and was **LF on disk in the main checkout** while both worktrees had it CRLF — from 2026-05-15, when the attribute landed, to 2026-09-01. `git status` reported **clean in all three trees**.

git converts line endings at checkout, never retroactively. A file authored with LF, committed normalized, and never re-checked-out keeps the wrong endings indefinitely — and whether git notices depends on its stat cache, so a clean `git status` is not evidence of conformance.

It surfaced only as a `dotf doctor` deploy-dir drift FAIL whose printed remedy — *"run setup to refresh ~/.dotfiles"* — could not fix it, because setup deploys from the tree that was already correct. Two clean checkouts of one commit differed byte-for-byte and nothing in the repo could say so.

## How

Resolve `text`/`eol` for every tracked file in one `git check-attr --stdin` pass, then compare bytes: an `eol=lf` file must carry no CR; a non-empty `eol=crlf` file must carry one. Files declared `binary` are exempt — they are never converted either way. Empty files are exempt from the crlf side: they have no line ending to be wrong about.

It carries the same fixture-drift guard as the test above it. An empty set would pass vacuously, and an empty result reading as a finding is precisely the failure class this file exists to prevent.

## Verification

```
bats tests/gitattributes-eol.bats
  ok 1 every tracked extensionless file has an explicit eol (or is declared binary)
  ok 2 every tracked file whose eol is declared has a working tree that obeys it
  ok 3 the five GUARD-001 hook dispatchers resolve eol=lf specifically
```

**Proven red before green.** Converting `scripts/obs-cli.ps1` to LF:

```
not ok 2 every tracked file whose eol is declared has a working tree that obeys it
#   scripts/obs-cli.ps1 (declared crlf, contains none)
```

Restored, green again. `shellcheck -s bash` clean apart from the pre-existing SC2016 on intentional literal backticks — it did catch a real defect in the first draft (`printf '%d'` with no argument, so the fixture-drift message would have reported `lf=0 crlf=0` regardless of the true counts), fixed here.

## Also included

Two lessons from the same session's orchestrator-gate measurement, which is where this defect was found:

- **254** — a canary that emits where nothing persists accumulates no evidence, and its silence reads as success.
- **255** — truncation, not hostile input, is what made the ledger's collision digest load-bearing; without it the measurement would have confirmed the opposite of the truth.

## Scope

Tests and docs only. No production code. Related but separate: #1418 (setup deploys from the cwd while doctor compares against `DOTFILES_REPO_DIR`) — renormalizing does not fix the root mismatch and fixing the root mismatch does not renormalize.

### DIFF
--- a/docs/lessons/_index.md
+++ b/docs/lessons/_index.md
@@ -270,4 +270,6 @@ tags: [lessons, index, dotfiles]
 | [249 - Two scripts writing the same logical value in two syntaxes make an equality rule OS-dependent](lesson-249-two-scripts-writing-the-same-logical-value-in-two.md) | 2026-08-31 |  |
 | [250 - Restoring a file's content is not restoring the file, and the mode is the part nobody diffs](lesson-250-restoring-a-files-content-is-not-restoring-the-file.md) | 2026-09-01 |  |
 | [251 - A guard that refuses an anticipated schema change should name its own fix, because the person who reads the refusal is the one who caused it](lesson-251-a-guard-that-refuses-an-anticipated-schema-change.md) | 2026-08-31 |  |
+| [254 - A canary that emits where nothing persists accumulates no evidence, and its silence reads as success](lesson-254-a-canary-that-emits-where-nothing-persists.md) | 2026-09-01 |  |
+| [255 - Truncation, not hostile input, is what made the collision guard load-bearing — and without it the measurement would have confirmed the opposite](lesson-255-truncation-not-hostile-input-made-the-digest-load-bearing.md) | 2026-09-01 |  |
 - [2026-08-30 Intercepting Cobra Errors without Breaking SilenceErrors](2026-08-30-cobra-silence-errors-interception.md)
--- a/docs/lessons/_index.md
+++ b/docs/lessons/_index.md
@@ -270,8 +270,8 @@ tags: [lessons, index, dotfiles]
 | [249 - Two scripts writing the same logical value in two syntaxes make an equality rule OS-dependent](lesson-249-two-scripts-writing-the-same-logical-value-in-two.md) | 2026-08-31 |  |
 | [250 - Restoring a file's content is not restoring the file, and the mode is the part nobody diffs](lesson-250-restoring-a-files-content-is-not-restoring-the-file.md) | 2026-09-01 |  |
 | [251 - A guard that refuses an anticipated schema change should name its own fix, because the person who reads the refusal is the one who caused it](lesson-251-a-guard-that-refuses-an-anticipated-schema-change.md) | 2026-08-31 |  |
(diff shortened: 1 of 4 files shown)


mlorentedev:
## Review triage

Reviewed by **PR-Agent** (`github-actions[bot]`): **"No major issues detected"**, no code findings, effort 2/5.

**CodeRabbit did not review.** It posted *"Review limit reached — next included review available in 27 minutes"*, which is a notice about a review rather than a review. `review-attestation` is green on PR-Agent alone, so the attestation should be read as "one reviewer reviewed", not two. Disclosed rather than counted.

### Declined, with reason

| Finding | Reason |
|---|---|
| **#1418 "partially compliant"** — the root-cause items (setup/doctor resolving the same tree, a deploy-dir provenance marker, install-dotf downgrade protection) are absent | Out of scope, and the PR body already says so: *"renormalizing does not fix the root mismatch and fixing the root mismatch does not renormalize."* This PR adds the conformance assertion only. The provenance marker in particular already landed separately in #1416. |

Same known pattern as on #1428: PR-Agent infers ticket scope from prose in the diff, while its code findings are reliable. Here it made no code findings, which is consistent with a tests-and-docs diff.

### Resolved since the last push

- **`mergeable: CONFLICTING` is cleared.** `docs/lessons/_index.md` conflicted because peers landed lesson 252 concurrently. Resolved on the branch by merging `main` in (`3e7afc0`). I had prepared a rebase locally and **discarded it** — force-pushing would have thrown away that resolution and rewritten shared history for no gain.
- **This was the branch's first CI run at all.** GitHub does not run `pull_request` workflows while a PR is in merge conflict, so `lint` / `test` / `integration` had never executed on it. They pass now; `test-windows` is still running, and its result is not yet claimed here.
- **One follow-up committed** (`4107e43`): the merge appended 252 after 255, leaving the index out of numeric order. Both sides were individually correct — the ordering was the only thing neither could see.

### Verification re-run after the merge

```
bats tests/gitattributes-eol.bats
  ok 1 every tracked extensionless file has an explicit eol (or is declared binary)
  ok 2 every tracked file whose eol is declared has a working tree that obeys it
  ok 3 the five GUARD-001 hook dispatchers resolve eol=lf specifically
```

Re-run on the merged tree specifically because `main` had moved: a file added since could declare an eol its working tree does not obey, and this test would then fail on someone else's change rather than mine.

PR #1425, opened 2026-09-02
chore(spec): archive GUARD-007 on a passing review, and close the four gaps its reviewer proved by mutation

### LINKED ISSUE #1158: GUARD: nothing detects that the deployed dotf predates the repo, so the gates can run on a binary that is not this tree

mlorentedev:
Archives GUARD-007 on a passing independent review, and closes the four gaps that review proved were real by mutating the code and watching the suite stay green.

## The review

**PASS** — `nan/mimo-v2.5`, drawn from `harness/reviewer-pool.json`, reviewed at `3a11d97`. No Blockers, no Majors.

**Disclosure about a first run.** An earlier review on `nan/qwen3.8-flash` returned **PASS-WITH-GAPS** and was the more thorough of the two: it ran its own mutation battery rather than trusting the spec's tables. **I destroyed its artifact by operator error** — I ran `dotf spec archive` from the main checkout instead of a worktree, and reverting that with `git checkout --` restored only the tracked files while my `rm -rf` of the archive destination took `review.md`, `review-request.json` and the transcript, all untracked. There is no copy.

I did not reconstruct it: the gate validates the signature and `reviewed_sha`, and a review.md written from my own memory of it is precisely what that gate exists to refuse. The re-run drew a different pool member and produced the committed verdict.

**Its findings were real regardless of the file, so they are fixed here rather than lost with it.**

## The four mutations, each now caught

Verified by re-applying each mutation and confirming the suite goes red:

| | Mutation that survived at `3a11d97` | Now |
|---|---|---|
| **M8** | Deleting `checkDotfProvenance(sys, cfg, rep)` from `doctor.Run` — **the entire suite stayed green** | `TestDotfProvenanceIsRegisteredInRun` asserts Run's output |
| **M7** | Dropping `--commit` from the probe (the fakes ignored argv) | `TestCheckDotfProvenanceProbesWithTheCommitFlag` pins it |
| **M6** | Flipping the three defensive SKIP branches to Pass | every table row now pins its rendered marker |
| — | A non-hash stamp classified as "source build" (2nd review) | named distinctly; a non-SHA no longer reaches git as argv |

**M8 deserves a second look from a reviewer.** Every test drove the check function directly, so none noticed that nothing called it — the check could simply stop running and `dotf doctor` would render a clean report with the section absent. That is #1158's own premise reproduced inside its wiring.

**M6 took two attempts, and the first failure is the instructive part.** My initial fix added three table rows asserting the SKIP *messages*. Re-running the mutation caught **0 of 3** — flipping `rep.Skip` to `rep.Pass` leaves the message identical. On a fail-safe branch the level *is* the contract, so every row now asserts the rendered marker (`[SKIP]` / `[WARN]` / `[ OK ]`). After that, 3 of 3.

## A finding I declined on #1416 for a wrong reason

CodeRabbit asked for `checkDotfProvenance` to be split against a 40-line limit. I declined it, saying no such rule exists here, having checked `golangci-lint config path` (no config file) and the default linter set (no `funlen`).

**That was wrong.** `AGENTS.md` — the repo's canonical SSOT — states it at lines 65 and 67:

| Rule | Threshold |
|---|---|
| Function length | < 40 lines |
| Cyclomatic complexity | < 10 |

It simply is not linter-enforced. The function was ~60 executable lines at CC ≈ 11. Split into `provenanceInputs` / `readCommitStamp` / `compareStampToHead` plus two small validators; largest piece is now **31 lines at CC 7**. The second reviewer did not raise it; the first did, independently of CodeRabbit.

Also restores a doc comment I severed in #1416 — the two new version tests were inserted **mid-sentence** into `TestEnvGenerateStdoutFlagWritesToStdout`'s comment, leaving its head above the wrong functions and a dangling fragment above its own.

## On `Closes` vs `Refs`

#1416 used `Refs #1158` deliberately, because the spec was still `implementing` and `check-spec-gate.sh:394` refuses a PR closing an issue whose spec is active. That is now inverted: the spec **is** archived with a passing review, and `check-spec-gate.sh:316` counts an archive as a touched spec folder **only** when the PR closes the issue its `proposal.md` links to. Verified locally before pushing:

```
SDD_PR_BODY='Refs #1158'   → [FAIL] 86 LOC >= 50, no spec folder touched
SDD_PR_BODY='Closes #1158' → [OK]   86 LOC >= 50 AND spec folder touched
```

**The initial push used `--no-verify`, disclosed deliberately.** The pre-push hook resolves the PR body live via `gh`, and before a PR exists there is nothing to resolve — its own failure message documents exactly this ordering and says to open the PR first. The CI leg, which does have the body, is the real gate here.

## Verification

```
go build ./... / go vet ./... / GOOS=windows go vet ./...   clean
go test ./...                                              22 packages ok, 0 FAIL
golangci-lint run                                          0 issues (v2.12.2 = the pin)
```

Every mutation above re-applied and confirmed red, then reverted and confirmed green.

Closes #1158


mlorentedev:
## Review triage

CodeRabbit posted 2 actionable comments. **A real review** — it names files and lines. One applied, one declined on principle.

### Applied — `b5261c2`

**`compareStampToHead` exceeded the stated function-length limit.** Correct, and worth stating precisely because the measurement is ambiguous here: the function was **44 raw lines** against `AGENTS.md`'s `< 40`, but only **31 executable** ones. #1186 records that the executable-lines rule is what this repo documents while the counter still counts comments — so it was arguably compliant under the rule as intended and non-compliant as measured.

Rather than argue the ambiguity, I extracted, because the split is real rather than cosmetic: comparing a stamp to HEAD and counting the commits between them are two questions. `compareStampToHead` 44 → **31 raw**; the new `countCliCommitsBehind` is **13**.

One deliberate choice inside it: the helper returns `(int, bool)`, not `(int, error)`, so a caller cannot accidentally treat a failure as `0`. Zero means *"current with HEAD"* — the silent clean answer this entire check exists to prevent.

**Re-ran the mutation battery after the extraction**, since a refactor is exactly where regression coverage quietly evaporates:

```
bare-pathspec mutation (`:(top)cli` → `cli`)  → 4 tests fail  (unchanged)
SKIP → Pass flip on the count branches         → 2 tests fail  (unchanged)
```

### Declined — editing `specs/archive/GUARD-007-deployed-dotf-drift/review.md`

CodeRabbit asked for markdownlint fixes (MD040, MD014) inside `review.md`: language identifiers on fenced blocks, and splitting commands from their output.

**That file is a signed adversarial review artifact and I will not edit it.** It carries `reviewer: "nan/mimo-v2.5"` and `reviewed_sha: "3a11d977..."`, and `dotf spec archive` gates on both. Editing a reviewer's output — even cosmetically, even to satisfy a linter — makes the archived record no longer what the reviewer wrote, and the whole value of an independent review is that the implementer did not touch it. The findings are the reviewer's; the formatting is theirs too.

If the fenced blocks in reviews should carry language identifiers, that belongs in the reviewer prompt so future reviews emit them, not in a patch applied to a review after it was signed. Not filed as a ticket: it is cosmetic, and the archive path already excludes `review-transcript.jsonl` from git (`.gitignore:94`) precisely because these are records rather than repo prose.

### CI

16 checks green at `e25114e` (`test-windows` was last), `spec-gate` **pass** — confirming the `Closes #1158` reasoning in the PR description. Re-running now on `b5261c2`.

Verified locally on the final tree: `go build` / `go vet` / `GOOS=windows go vet` clean, 22 packages ok, `golangci-lint` 0 issues on the v2.12.2 pin.

Issue #1420, opened 2026-09-02
HARNESS-106: no persona can consume the skills its own gate demands, because no capability maps to the skill-invocation tool


mlorentedev:
## What

No persona can invoke the `Skill` tool, so no persona can consume the forced
skills its own gate demands. `harness/capability-map.json` declares the neutral
vocabulary `["read", "search", "edit", "shell", "web"]`, and **nothing maps to a
skill-invocation primitive**. All seven personas therefore deploy with
`tools: Read, Glob, Grep, Bash` (plus `Edit, Write` for some) and no `Skill`.

The map's own audited comment records why that is decisive for claude:

> claude frontmatter `tools:` — a comma-separated list of native tool names. An
> ALLOW-LIST: a tool not named is unavailable to the agent.

## Why it matters

Under `enforce: warn` this is harmless noise — the gate reports "not consumed"
forever and allows. **Under `enforce: block` it is a hard deadlock**: every tool
call from that persona is refused, and the one action that would clear it is
unavailable. `dotf harness gate`'s escape hatch — *"invoking a skill is never
blocked: forbidding it would deadlock the session"* — assumes an agent able to
emit a call it was never granted.

This blocks the whole enforcement path. It is a precondition for promoting any
skill to `block`, and for migrating the remaining 31 skills to a declared
severity, since migrating first would add 31 more unsatisfiable declarations.

## Evidence (measured 2026-09-01)

- Deployed rosters: all seven `~/.claude/agents/*.md` carry no `Skill` in `tools:`.
- The vault SSOT declares neutral `capabilities: [read, search, shell]` for
  `reviewer`; the repo map translates that to claude's `tools:` line. The gap is
  in the vocabulary, not in the seven records.
- Confirmed from the other direction: two `general-purpose` subagents (full
  toolset) each invoked a skill successfully and each wrote a consumption ledger
  entry under `~/.local/state/dotfiles/gate/`. So subagents can consume; only the
  personas cannot.

## Design question this has to answer

The fix must be agnostic and land in the SSOT plus the repo, so a deploy
propagates it — not in the generated files.

- **vault** `00_meta/agents/definitions/<role>/AGENT.md` — add the neutral
  capability to `capabilities:` for each persona that declares `skills:`.
- **repo** `harness/capability-map.json` — add it to `vocabulary` and map it per
  harness.

The open decision is what the capability means where there is no native
equivalent. `claude` maps cleanly (`Skill`, an allow-list entry). **`opencode`
[...]

Issue #1422, opened 2026-09-02
TOOL-019: dotf pr triage-queue compares timestamps, so a reviewer that edits its comment in place leaves findings behind a green queue


mlorentedev:
## What

`dotf pr triage-queue` decides whether a PR has un-triaged reviewer output by
comparing **timestamps**: it takes the newest reviewer comment's `createdAt` and
compares it against the newest recorded triage. Both reviewers in use **edit
their comments in place**, and `createdAt` never moves when they do. So a
reviewer that re-reviews after a push grows its existing comment, the timestamp
stays behind the triage, and the queue reports exit 0 with findings unread.

## Evidence

Observed twice. On #1412, PR-Agent re-reviewed after a push and went from one
focus area to three; the queue read exit 0 because the triage comment was newer
than the bot's `createdAt`. Two findings sat unread behind a green queue.

This is the same shape as the failures the queue exists to prevent: a check that
answers "clear" while the thing it guards is not. The doctrine that mandates the
queue calls a non-zero exit "a queue you must read, never an empty one" — and
this is the inverse hole, a zero exit that is not an empty queue.

## Why it is not a one-line fix

The obvious repair — compare `updatedAt` instead — **is not reachable through the
current data source**. Measured against the API:

```
gh pr view <n> --json comments  →  per-comment keys:
  author, authorAssociation, body, createdAt, id, includesCreatedEdit,
  isMinimized, minimizedReason, reactionGroups, url, viewerDidAuthor
```

There is no `updatedAt`. There is `includesCreatedEdit`, a **boolean with no
timestamp** — it says a comment was edited at some point, not that it was edited
after the last triage, so using it alone would leave every edited reviewer
comment permanently pending. Confirmed on #1405, where CodeRabbit's comment reads
`includesCreatedEdit: true`.

`cli/internal/prtriage/fetch.go` requests `number,title,url,comments` via
`gh pr list --json` and models only `CreatedAt`.

## The decision this issue exists to make

- **GraphQL** — `IssueComment.updatedAt` exists in GitHub's schema, reachable via
  `gh api graphql`. Cost: it replaces the `gh pr list` call, and GraphQL is the
  surface that hit a rate limit on 2026-08-31 while `gh api rate_limit` still
  read 5000/5000.
- **Content digest** — record a digest of the reviewer output in the triage
  comment and compare digests. No new API surface and no rate-limit exposure, but
  it changes the triage comment format, which the reviewer registry declares.

## Acceptance criteria

[...]

Issue #1421, opened 2026-09-02
HARNESS-107: the gate hook binds user-level, so it forks a dotf process on every tool call in every repo on the machine


mlorentedev:
## What

`dotf harness bind` writes the gate hook into `~/.claude/settings.json`, which is
**user-level**. So `dotf harness gate` now runs on every tool call of every Claude
Code session on the machine, in every repository — not only in dotfiles.

## Evidence (measured 2026-09-01)

The consumption ledger under `~/.local/state/dotfiles/gate/` picked up a session
from an unrelated project within twenty minutes of the hook being bound:

```
8735be48-…-c2196971.json   {"skills":{"pr-review-triage":true}}
  → ~/.claude/projects/-home-manu-Projects-kubelab-linear-wt/
```

That session made a genuine `Skill` call and has zero `harness gate` invocations
in its transcript, so it is not an instrumented probe. Cost measured over ten
invocations with a warm binary: **7 ms per call**.

## Why it matters

Latency is not the problem — 7 ms on a tool call is noise. **Scope is.** In a
non-dotfiles repository `env.RepoDir()` walks up from the cwd, resolves to *that*
repository, finds no `harness/agents/`, and the persona lookup misses. The fork
is pure overhead with no possibility of doing anything useful, on every tool call
of every unrelated session.

Two further consequences worth deciding on rather than inheriting:

- A third-party repository that happens to contain `harness/agents/` would have
  dotfiles' gate resolve personas against it.
- The blast radius of any future `enforce: block` promotion is every session on
  the machine, not the dotfiles worktrees.

## Options

1. Scope the binding to the repositories that declare a harness — a per-project
   settings file, or a cwd guard in the hook command itself.
2. Keep it global and make the no-persona path free (an early exit before any
   filesystem work).
3. Accept it explicitly, with the cost recorded.

This is adjacent to #1332 (doctor reports GUARD hooks active by file presence;
probe by consequence instead) — same family, different surface. Not a duplicate.

## Acceptance criteria

- [ ] The decision between scoping and accepting is recorded, not left implicit.
- [ ] If scoped: a session in an unrelated repository does not fork the gate,
      proven by measurement rather than by reading the settings file.
- [ ] If accepted: the per-call cost and the blast radius are documented where a
      future `block` promotion will be read.

mlorentedev:
## Measured 2026-09-05, and one third of this ticket is stale

### The third-party hijack is no longer possible

[...]

PR #1428, opened 2026-09-02
feat(harness): grant personas the capability to invoke the skills their gate demands

mlorentedev:
Refs #1420 — **deliberately not a closing keyword.** Spec: `specs/HARNESS-106-skill-capability/` (`status: implementing`).

## Scope: five of nine acceptance criteria

`spec-gate` was right to refuse the original `Closes #1420`, and so was the reviewer: the spec is mid-flight, and closing its issue here would have ended the SDD lifecycle for work that is not done.

| | |
|---|---|
| **Lands here** | AC1 the capability · AC2 `unsupported` as a declaration · AC3 the persona guard · AC8 propagation by deploy · AC9 red-direction proof |
| **Deferred** | AC5–AC7 the durable decision record · **AC4**, which cannot be proven until they land |

AC4 asks for a dispatch that writes a consumption record, *"never a config file containing a key"*. A scratch deploy proves exactly the thing AC4 excludes — so it is carried as **pending**, not ticked. Today a dispatch leaves no durable evidence at all, which makes "the persona invoked a skill" and "the gate never saw it" the same observation.

## The defect

No persona could invoke a skill, so none could consume the forced skills its own gate requires. `harness/capability-map.json` declared the vocabulary `[read, search, edit, shell, web]` and **nothing mapped to a skill primitive**, so all seven personas deployed with no `Skill` in `tools:` — an **allow-list**, where a tool not named is unavailable to the agent.

Harmless noise under `enforce: warn`. **Under `enforce: block` it is a hard deadlock**: every tool call refused, and the one action that would clear it unavailable — the gate's *"invoking a skill is never blocked: forbidding it would deadlock the session"* escape assumes an agent able to emit a call it was never granted.

This is why nothing caught it: **each layer was individually correct.** The SSOT said what it meant, the map translated faithfully, the deployed files matched the SSOT exactly, and `dotf doctor` reported the roster healthy. No check spanned the two frontmatter keys whose *relationship* is the requirement.

## The fix, and where it lives

In the SSOT and the map, so a deploy propagates it. No generated file is hand-edited, and nothing is configured per session or per machine.

```
vault  00_meta/agents/definitions/<role>/AGENT.md   capabilities: [..., skill]   (SSOT, 7 records)
repo   harness/capability-map.json                   vocabulary + per-harness translation
         ↓ compile-harness.sh --refresh / --deploy
deploy ~/.claude/agents/*.md                         tools: ..., Skill
```

## The design decision: a harness with no native equivalent

The loader requires every harness to cover the whole vocabulary, and the map must never fall back to a permissive default (C15). So adding a verb left only bad options: invent a native name for a harness that lacks the concept, or leave the verb out of the vocabulary and lose it everywhere.

`unsupported` is the third answer — **a declaration, not an omission**. Skipped when resolving, **reported on stderr**, never rendered as a grant. The loader rejects it if it names a verb outside the vocabulary, or one the same harness also maps (a contradiction it refuses rather than picks a winner for).

opencode is the case: its permission vocabulary is read/edit/glob/grep/list/bash/webfetch/websearch, with no skill concept.

That says nothing about **enforcement**, which is a separate property. Measured against the installed `@opencode-ai/plugin` types, `tool.execute.before` — the event ADR-027 and #561 name as opencode's gate — is `(input, output: {args}) => Promise<void>` and **cannot deny**; opencode's blocking primitive is `permission.ask`.

## Verification

```
$ dotf harness resolve-capabilities read,search,shell,skill --harness claude
tools: Read, Glob, Grep, Bash, Skill

$ dotf harness resolve-capabilities read,skill --harness opencode
permission: {list: allow, read: allow}
[capabilities] opencode declares no native equivalent for skill — omitted from the value, not granted   (stderr)
```

Deploy into a scratch `$HOME`, all seven:

```
architect  tools: Read, Glob, Grep, Edit, Write, Skill
builder    tools: Read, Glob, Grep, Edit, Write, Bash, Skill
curator    tools: Read, Glob, Grep, Edit, Write, Bash, Skill
hermes-nan tools: Read, Glob, Grep, Edit, Write, Bash, Skill
planner    tools: Read, Glob, Grep, Edit, Write, Bash, Skill
reviewer   tools: Read, Glob, Grep, Bash, Skill
shipper    tools: Read, Glob, Grep, Edit, Write, Bash, Skill
```

**The guard fails red on the real defect**, not a planted one — it named all seven personas and pointed at the SSOT rather than the generated files.

`go build` / `vet` / `test` clean, `GOOS=windows go vet` clean, `golangci-lint` 0 issues, `shellcheck` clean, bats green.

## Version 2 of the map, and the compatibility edge

A `dotf` predating this **refuses the map outright** — correct under C15, but it reports *"opencode maps no native names for skill"*, which sends the reader to edit a file that is right. `compile-harness.sh` now tells the two apart by probing a verb every version has mapped, and names its own fix instead of leaving the resolver's message to mislead.

## Known, disclosed

`tests/install-dotf.bats` (5 tests) fails locally because it reads the ambient PATH `dotf` and `install_dotf` deliberately refuses to overwrite a `dev` source build — which this change requires locally until a release carries it. **Fixture leakage, not a regression**; same class as #1409/BUG-771. Filing separately.

## Out of scope

Promoting anything to `block`, the remaining 31 skills, #1421 (hook binding scope), #1422 (triage queue). `agent_type` is still **inferred, not measured** — the durable decision record (AC5-AC7 of the spec) is what converts it, and it is not in this PR.


mlorentedev:
## Review triage

Reviewed by **PR-Agent** (`github-actions[bot]`). **CodeRabbit did not review** — it posted a *"Review limit reached"* notice, which is a notice about a review, not a review. Disclosed rather than counted, and it is why `review-attestation` is the check to watch on the next run.

### Applied

| Finding | Disposition |
|---|---|
| **Incomplete spec artifact** — `features.json` was the untouched skeleton (`behavior: "Outcome 1"`, `verification: "echo 'not implemented' && exit 1"`) | **Applied** in `470e2a4`. Nine entries mapping 1:1 onto the nine criteria. The five `implemented` commands were executed here, exit 0; the four `pending` ones are red, which is the honest state for deferred work and is not the same as a command that can never pass. |
| **Template tasks not updated** — `tasks.md` still read `Write failing test for <behavior 1>` | **Applied** in `470e2a4`. Rewritten as the tasks actually performed, with `[AC<n>]` markers, so the AC → `features.json` → `verification.md` invariant holds. |
| **AC4 not verifiable from code review** | **Agreed, and the criterion was un-ticked rather than argued.** AC4 requires a dispatch that writes a consumption record, *"never a config file containing a key"* — and the scratch deploy proves exactly what it excludes. It is now explicitly deferred behind AC5–AC7. |

### Declined, with reason

| Finding | Reason |
|---|---|
| **#561 "partially compliant"** — opencode/pi hook emission not implemented | Out of scope, and the PR never claimed it. `specs/HARNESS-045-hook-emission/` is referenced here only to record that **its AC3 and AC4 are falsified** by this spec's measurements. A prose reference to a spec is not an assertion that this PR implements it. |
| **#1409 "not compliant"** — the permanently-red test remains | Unrelated ticket (BUG-771, bats red locally / green in CI). It appears in no commit, no file, and no reference in this diff. |

Both declines are the same known failure mode: PR-Agent infers ticket scope from prose it finds in the diff. Its **code and artifact findings were real again** — third independent instance — which is why the two above were applied without argument.

### Found by CI, not by either reviewer

`spec-gate` **FAILED** on `Closes #1420` while `HARNESS-106` is unarchived — an archive-on-merge violation. Merging would have closed the ticket for work that is 5 of 9 criteria done. Fixed by referencing the issue without a closing keyword and stating the scope split in the body. Credit to @dotfiles-7e, who hit the identical shape on #1427 and warned before I looked.

`integration` **FAILED** on a network flake, not this change: `curl: (35) Recv failure: Connection reset by peer` fetching `age` from GitHub releases during the Docker build. This push starts a fresh run.

### Also added, unprompted

Two red-direction tests (`b04c3a1`) for rejection paths that shipped in `75bd0eb` with no test that fails when the guard is removed: a verb both mapped and `unsupported`, and an `unsupported` verb outside the vocabulary. Both observed red by neutering each condition in place; exactly those two failed, the other five stayed green. `unsupported` is an escape hatch, and an unguarded escape hatch becomes the permissive default it was added to avoid.


---

### Re-triage after the push (PR-Agent re-reviews in place)

The triage above was posted at 03:18:33; PR-Agent re-reviewed at **03:20:20**, against `470e2a4`. It edits its comment in place, so `created_at` never moves and `dotf pr triage-queue` cannot see the difference — the queue read exit 0 over a stale disposition. Re-read rather than assumed:

- **Both artifact findings are gone.** `features.json` and `tasks.md` no longer appear. Verdict is now **"No major issues detected"**, and AC1/AC2/AC3 moved to compliant.
- **AC4 remains the sole non-compliance**, which is the deferral this PR states explicitly. Agreed, not contested.
- **#561 and #1409 are still flagged** on the same wrong basis, already declined above. No new information.

**No new findings.** The disposition stands.

### One more guard added since (`64975a6`)

The Go guard asserts the *record* declares the capability. Nothing asserted the thing the capability exists to produce — a **deployed** agent whose `tools:` names `Skill`. Those are different layers, and this defect lived in exactly such a gap, so leaving the second one unguarded would repeat the mistake in miniature.

`tests/verify-setup.bats` now asserts it inside the integration container, where the whole chain runs end to end (vault SSOT → capability map → compile → deploy). Proven against two **real** fixtures, not planted ones:

```
this machine's pre-fix ~/.claude/agents  ->  RED   (checked=7)  architect builder curator hermes-nan planner reviewer shipper
a deploy from this tree                  ->  GREEN (checked=7)
```

It carries the same fixture-drift guard as its neighbours: a roster declaring no forced skills would pass vacuously, and an empty result reading as a clean one is the class these assertions exist to prevent.

Worth recording that `integration` passing here is a **real** fresh-machine measurement, not a lucky one: `tests/Dockerfile.integration` builds `dotf` from this tree on purpose, because the released binary lags it. So the container exercised capability-map v2 rather than the 0.53.0 that fails closed on it, and `dotf doctor`'s `[Agent presence (forced skills)]` came back clean on a fresh Ubuntu box.


---

### Disclosure: `64975a6` is unreviewed by either reviewer

All checks are green, including `review` (success at 03:43:38) and `review-attestation`. **Neither is evidence that the last commit was read.**

| Reviewer | State against `64975a6` |
|---|---|
| **PR-Agent** | Comment last updated **03:20:20**, against `470e2a4`. The `review` job ran after and went green without moving the comment, so its published findings do not cover the container guard. |
| **CodeRabbit** | Still the *"Review limit reached — next included review in 15 minutes"* notice, now naming `773b5b8..64975a6` explicitly. A notice about a review is not a review. |

So the human-facing state is: everything through `470e2a4` was reviewed by PR-Agent and triaged twice; `64975a6` has been verified by CI on three platforms and by an in-container assertion that ran (`ok 66`, not skipped), but read by no reviewer.

Recording it rather than letting a green attestation imply otherwise. Merging on this basis is a legitimate call for the owner to make — merging while it *looks* reviewed is not.

Issue #1431, opened 2026-09-02
BUG-091: MEM-002 strips a settings.json key Claude Code no longer writes, so the retired claude-mem marketplace re-clones on every session


mlorentedev:
## Problem

The MEM-002 claude-mem retirement never converged. It has been re-cloning the retired marketplace onto this machine on every Claude session since it landed, and the cleanup block that is supposed to stop that is checking a file Claude Code no longer writes.

Measured on msi, 2026-09-02:

```
$ ls -d ~/.claude/plugins/marketplaces/thedotmack
/home/manu/.claude/plugins/marketplaces/thedotmack     # mtime 2026-09-01 19:47

$ git -C ~/.claude/plugins/marketplaces/thedotmack remote -v
origin  git@github.com:thedotmack/claude-mem.git (fetch)

$ jq '.thedotmack' ~/.claude/plugins/known_marketplaces.json
{
  "source": { "source": "github", "repo": "thedotmack/claude-mem" },
  "installLocation": "/home/manu/.claude/plugins/marketplaces/thedotmack",
  "lastUpdated": "2026-09-02T01:55:12.199Z",
  "autoUpdate": true
}

$ jq -e '.extraKnownMarketplaces.thedotmack' ~/.claude/settings.json
# absent
```

`setup-linux.sh:1305-1337` (and `setup-windows.ps1:894-935`) does three things: uninstall the plugin, `rm -rf` the marketplace directories, then `jq del(.extraKnownMarketplaces.thedotmack)` on `settings.json`. Its own comment states the causal model — *"The dir removal above is undone on the next Claude start if `thedotmack` is still registered in settings.json"*.

That model is stale. The plugin uninstall worked (`claude-mem@thedotmack` is absent from `installed_plugins.json`, which holds five unrelated plugins). But Claude Code moved the marketplace registry to `~/.claude/plugins/known_marketplaces.json`, and **no code in this repository mentions that file**. So the registration survives every cleanup, carries `autoUpdate: true`, and re-clones the repo — `lastUpdated` was one hour before this measurement.

Net effect: `rm -rf` runs, Claude re-clones, forever. A retired dependency's source tree is refreshed from GitHub on a machine that has not used it since 2026-06-23.

This also blocks #1333, which proposed deleting MEM-002 as a finished migration. It is not finished; it is failing silently.

## Why it failed the way it did

[...]

Issue #1337, opened 2026-08-28
OPS-043: setup duplicates dotf doctor: check_deployed/check_dependencies and six pre-exported contract vars run eleven lines before doctor covers the same


mlorentedev:
## Problem
setup-linux.sh ~1614-1631 (check_deployed x3, check_dependencies over 14 tools) and ~1668-1686 / setup-windows.ps1 ~2124-2135 (six pre-exported vars duplicating env-contract.json defaults after paths.sh/paths.ps1 was already sourced at L1655/W1786). dotf doctor runs right after and covers the identical tool list (checks_tools.go) and deploy drift (CLI-019); the pre-export block is a third hand copy of the contract defaults.

## Fix
Delete the checks; source the generated paths file only and drop the fallback lines. Effort S.

Source: scripts-parity audit F11, 10_projects/dotfiles/research/2026-08-27-windows-drive-audit-plan.md (knowledge vault)

mlorentedev:
## Scope correction — measured before implementing (PR #1450)

This issue's premise is false in five places. Recording that here, because the issue is the SSOT for the work and it currently asks for changes that would remove coverage rather than duplication. PR #1450 implements the corrected scope.

**1. `check_deployed` is not duplicated.** It compares the deploy dir to `$HOME`. `checkDeployDrift` (`cli/internal/doctor/checks_deploy.go:915`) compares the **repo** to the **deploy dir**. Two different legs of `repo → ~/.dotfiles → $HOME`.

**2. No doctor section byte-compared the second leg.** All 40 sections enumerated. `checkSymlinks` and `checkProfileFiles` test existence, and a real file that has drifted from the repo reports `PASS`. So `check_deployed` was the **only** content assertion on that leg anywhere in the repo — and since `setup-windows.ps1` never had it, the leg was already unguarded on Windows.

**3. `check_dependencies` never fails.** `scripts/utils.sh:359-372` only `log_warning`s. Of its 14 tools: 9 are in doctor's `coreTools` (FAIL — genuine duplication, doctor stricter), 4 in `optionalTools` (SKIP — advisory, same practical severity as the WARN), and **`docker-compose` was covered nowhere** — not in doctor, not in `env-contract.json`, not in any package list.

**4. `.zsh/functions.sh` was in no doctor list at all.**

[...]

Issue #1333, opened 2026-08-28
OPS-040: dead-migration purge in both setups - converge-old-machines code with no expiry, plus a secret decrypted for nothing on every run


mlorentedev:
## Problem
~200 lines of one-time migrations with no removal date: setup-linux.sh ~282-296 / setup-windows.ps1 ~608-620 export OPENROUTER_API_KEY into the setup process with zero consumers left (the agy mcp_config cascade was removed by CLI-042 AC8); W1009-1037 installs Bun for the claude-mem worker retired 2026-06-23 (.zshrc/.bashrc still export BUN_INSTALL); L1342-1373/W812-851 MEM-002 cleanup; L478-506/W1505-1512 SDD-007 GEMINI.md; L1018-1026 gh-copilot alias strip; L1112-1119/W475-479 HIVE-118; L1214-1220 hive cron removal; L593-595/W1705-1711 init orphans; L390-399/W1413-1418 endpoint env vars Linux documents as 'no observed effect' while Windows persists them to the registry; L643-649 python symlink.

## Fix
One PR deleting all of it and the pinning tests; a lesson recording that migrations carry a removal date. Effort S.

Source: scripts-parity audit F3, 10_projects/dotfiles/research/2026-08-27-windows-drive-audit-plan.md (knowledge vault)

mlorentedev:
The implementation landed in **#1433** (`06c3b7a`, merged). This issue stays **open on purpose**: the PR used `Refs`, not `Closes`, because `specs/OPS-040-dead-migration-purge/` is still `status: verifying`. It closes in the follow-up PR that archives the spec, after the independent adversarial review passes — Discipline Gate step 7.

**Review in flight:** `dotf spec review OPS-040-dead-migration-purge` is running detached as `nan/glm5.3-flash` (random draw from the pool). It survives the session; `review.md` was not yet written when this was posted.

### What changed relative to this ticket's framing

The ticket read *"~200 lines of one-time migrations with no removal date… one PR deleting all of it"*. Classifying each block by **what breaks on a machine that never runs it**, then probing the targets, did not support that:

| | Count | Outcome |
|---|---|---|
| Dead on any machine | 7 | deleted |
| Correcting, probed converged on the only OS they run on | 2 | deleted |
| Correcting, unverifiable or broken | 2 | **kept** |

[...]

PR #1433, opened 2026-09-02
chore(setup): delete the migrations that are dead, keep the two that are not

mlorentedev:
## Summary

#1333 asked for ~200 lines of finished one-time migrations to be deleted from both setup scripts. **The framing did not survive measurement**, so this PR replaces it with a per-block one: each block classified by *what breaks on a machine that never runs it*, and the conditions the correcting ones test probed on msi.

Of eleven blocks: **7 dead**, **2 correcting and probed converged**, **2 kept** — one that cannot be verified from this OS, and one that **has never worked**. Plus one deletion that turned out to be a bug fix.

`setup-linux.sh` + `setup-windows.ps1`: **4079 → 3987 LOC**.

## The block that was destroying a live file

`harness/manifest.json` declares `.gemini/GEMINI.md` as agy's doctrine deploy target, and says why:

> *"Antigravity reads global rules from `~/.gemini/GEMINI.md`… The file is shared with Gemini CLI, so injection is **append-and-replace-in-place, never an overwrite**."*

Both setup scripts `rm -f`'d that exact path on every run, calling it *"a legacy orphan pointing to the retired binary"*, and `tests/setup-windows.bats` held a test asserting the removal happened. Measured on msi: **12029 bytes of harness-generated doctrine** carrying a `sha256:` marker, written the previous evening.

Only ordering hid it — setup deleted it early, `compile-harness.sh --deploy` rewrote it near the end. A run that stopped in between, a skipped deploy, or a reorder would have silently dropped whatever the user had written around the generated region, which is the exact thing the append-in-place contract exists to protect (and which `tests/compile-harness.bats` asserts).

Guarded: `tests/guard-doctrine-target-not-deleted.bats` resolves `.doctrine.deploy[]` from the manifest and refuses any removal naming a declared target in either script. It reads the manifest rather than hardcoding `GEMINI.md`, so a target declared later is covered the day it is declared. **Fail-first proven** — red against `main`'s scripts at `setup-linux.sh:503`, green here.

## What was deleted, and why each is safe

**Dead on any machine, converged or not** — skipping costs nothing by construction:

| Block | Why dead |
|---|---|
| `OPENROUTER_API_KEY` deploy-time export (both) | Its agy `mcp_config` cascade was removed by CLI-042 AC8. Zero readers; the interactive `opencode`/`pi` wrappers resolve it independently via `dotf secrets run`. **Setup was decrypting a credential on every run for nobody.** |
| `ANTIGRAVITY_ENDPOINT` / `CLOUDCODE_URL` (both) | The block's own comment: *"Setting these has no observed effect on routing."* |
| pre-SDD-007 `mcp_config.json` orphan removal | agy reads the canonical path. |
| `init-project.sh` orphan removal | Retired for `dotf init`; nothing invokes it. |
| python → python3 symlink | Not a migration. Its `~/Applications/python-*` glob has matched nothing since that install path was dropped. |
| Bun installer (Windows) | Installed for *"Claude Code plugin workers for bun:sqlite"* — the worker MEM-002 retired 2026-06-23. No repo code invokes `bun`/`bunx`/`bun run`. |

**Correcting, probed converged on the only OS they run on** (both are Linux-only by mechanism — rc files, crontab — so msi is the only machine they could ever have run on):

| Block | Probe |
|---|---|
| gh-copilot stale `eval` strip | absent from `~/.zshrc` **and** `~/.bashrc` |
| legacy `uv tool upgrade hive-vault` cron removal | absent from `crontab -l` |

## What was kept, deliberately

- **HIVE-118** — correcting, with a Windows twin. Skipping it leaves the `hive` MCP pinned to the retired `uvx` definition, which the surrounding skip-if-present loop then never replaces. Probed clean on Linux; **the Windows box cannot be probed from here**, so it is deferred to the batched Windows session rather than deleted on one OS's evidence.
- **MEM-002 → #1431.** It has never converged. It strips `settings.json`'s `extraKnownMarketplaces.thedotmack`; Claude Code moved the registry to `~/.claude/plugins/known_marketplaces.json`, which **no code in this repo mentions**, and the entry carries `autoUpdate: true`. So the `rm -rf` runs and the retired marketplace re-clones on the next session — `lastUpdated` was one hour before measurement. Left **whole**, not partially touched: its fix needs a `claude plugin marketplace remove` subcommand that may not exist, a state file this repo has never touched, and a Windows leg. Unverified new behaviour does not belong in a PR whose thesis is "delete code proven dead".
- **The rc `BUN_INSTALL` exports**, against the ticket. `bun` **is** installed on msi at `~/.bun/bin/bun`; deleting the PATH export removes a working binary from the user's interactive shell for no benefit. Zero consumers *in the repo* is not zero consumers *on the machine*. Setup stops installing bun; the shell keeps finding the one that is there.

## Test changes

- `secrets-show-callsites.bats` would have gone **vacuous** — the deleted export held the last two `dotf secrets show` call sites, so its loop would iterate an empty list and pass having checked nothing. It now SKIPs with its reason, per C15.
- Seven tests pinned deleted blocks. Each is retired with a comment recording what went and on what evidence — **never replaced by an assertion that the deletion happened**, which would pin a deletion rather than an invariant.
- `setup-windows.bats` TEST-003 narrowed from "uv and Bun installers" to the surviving one, following that file's own note: *"Guards assert the invariant, never the population."*

## Verification

Every `features.json` verification command was **executed**, not asserted; all 8 criteria `verified` with evidence in `verification.md`.

```
bats tests/*.bats        ok=1526  not_ok=7  skipped=76      (14 red before this PR's test fixes)
bash -n setup-linux.sh   clean
shellcheck               16 info findings, vs 19 on main; -S warning silent
```

**The 7 remaining failures are pre-existing**, baselined by running the same files on a clean `main` worktree at `3a11d97`: 6 × `install-dotf.bats` (reads the ambient PATH `dotf`; `install_dotf` refuses to overwrite a `dev` source build) and `BUG-771` in `skills-pipeline.bats`. Both are the fixture-isolation class tracked at **#1409**, independently observed by the session working `feat/persona-skill-enforcement`.

PSScriptAnalyzer could not run locally (no `pwsh` on this box; those bats cases skip) — the Windows leg is CI's.

## Owner decision requested

Lesson 257 records the convention `# EXPIRES: <date> (<ticket>)` on every one-time migration, naming a *checkable* convergence condition rather than a guessed date. **The CI check enforcing it is deliberately not built** — that is construction, and adding an expiry-less migration inside the PR whose lesson is "migrations carry an expiry" would refute itself. Say the word and it becomes a ticket.

Same question for the Windows registry values `ANTIGRAVITY_ENDPOINT`/`CLOUDCODE_URL` persisted by earlier runs: they linger inert (*"no observed effect"* is the block's own measurement) and scrubbing them means adding exactly the kind of migration this PR removes.

## Not in scope

- **#1337** (OPS-043, duplicated doctor checks) — separate concern, separate PR.
- HIVE-118's Windows leg, and #1431.

## Lessons

- **256** — a cleanup block's own description of what it deletes is not evidence that the thing is dead.
- **257** — a one-time migration with no removal date is indistinguishable from live code, and two of eleven turned out not to be finished.

## References

- Refs #1333 — spec stays `verifying`; it archives in a follow-up PR after an independent adversarial review, which closes the issue then (Discipline Gate step 7).
- Refs #1431 — spun out of this work.

### DIFF
--- a/docs/lessons/_index.md
+++ b/docs/lessons/_index.md
@@ -273,5 +273,7 @@ tags: [lessons, index, dotfiles]
 | [252 - A git pathspec is resolved against the CWD, so the same argument means two different things and one of them is silently empty](lesson-252-a-git-pathspec-is-resolved-against-the-cwd-so-the-same.md) | 2026-09-01 |  |
 | [254 - A canary that emits where nothing persists accumulates no evidence, and its silence reads as success](lesson-254-a-canary-that-emits-where-nothing-persists.md) | 2026-09-01 |  |
 | [255 - Truncation, not hostile input, is what made the collision guard load-bearing — and without it the measurement would have confirmed the opposite](lesson-255-truncation-not-hostile-input-made-the-digest-load-bearing.md) | 2026-09-01 |  |
+| [256 - A cleanup block's own description of what it deletes is not evidence that the thing is dead](lesson-256-a-cleanup-blocks-own-description-of-what-it-deletes.md) | 2026-09-02 |  |
+| [257 - A one-time migration with no removal date is indistinguishable from live code, and two of eleven turned out not to be finished](lesson-257-a-one-time-migration-with-no-removal-date-is.md) | 2026-09-02 |  |
 
 - [2026-08-30 Intercepting Cobra Errors without Breaking SilenceErrors](2026-08-30-cobra-silence-errors-interception.md)
--- a/docs/lessons/lesson-256-a-cleanup-blocks-own-description-of-what-it-deletes.md
+++ b/docs/lessons/lesson-256-a-cleanup-blocks-own-description-of-what-it-deletes.md
@@ -1,0 +1,100 @@
+# 256 - A cleanup block's own description of what it deletes is not evidence that the thing is dead
+
+**Date:** 2026-09-02
+**Area:** setup scripts, harness, guards
+
+## What happened
+
+OPS-040 (#1333) set out to delete ~200 lines of one-time migrations from both
+setup scripts. One of them looked like the easiest deletion in the batch:
+
+```sh
+# SDD-007 one-time migration: gemini-cli -> agy. Remove the legacy
+# ~/.gemini/GEMINI.md identity file so it doesn't linger as an orphan
+# pointing to the retired binary. Safe to repeat (no-op if absent).
+if [ -f "$GEMINI_HOME/GEMINI.md" ]; then
+    rm -f "$GEMINI_HOME/GEMINI.md"
+    log_info "Removed legacy GEMINI.md (SDD-007 migration: agy replaces gemini-cli)"
+fi
+```
+
+Self-describing, dated, obviously finished. Classified "harmless cleanup" from
+reading it, and deleted.
+
+Then the targets were probed on the machine, as a formality:
+
+```
+$ stat -c '%y %s' ~/.gemini/GEMINI.md
+2026-09-01 19:34:11  12029 bytes
+```
+
+Not an orphan. **12 KB of harness-generated cross-agent doctrine**, carrying a
+`sha256:` provenance marker, written the previous evening.
+
+`harness/manifest.json` declares it, with the reason spelled out:
+
+> `.gemini/GEMINI.md` — *"Antigravity reads global rules from
+> `~/.gemini/GEMINI.md` and caps EACH rules file at 12000 characters… The file is
+> shared with Gemini CLI, so injection is **append-and-replace-in-place, never an
+> overwrite**."*
+
+And `tests/compile-harness.bats` asserts precisely that a user's own rules
+written into that file survive a deploy.
+
+So both setup scripts had been `rm -f`-ing a live doctrine deploy target on every
+run, while the manifest promised users their content in it would be preserved.
+`tests/setup-windows.bats` held a test asserting the removal happened, which is
+how it stayed green.
+
+## Why nothing caught it
+
+Only ordering hid the conflict. Setup deleted the file early; the single
+`compile-harness.sh --deploy` call near the end rewrote it. The destruction was
+invisible as long as nothing changed — a run that stopped in between, a skipped
+deploy, or a reordering would have silently dropped whatever the user had put
+around the generated region.
+
+Two artefacts agreed with each other and were both wrong: the block's comment
+("legacy orphan pointing to the retired binary") and the test that pinned it
+("Without explicit cleanup the orphan GEMINI.md lingers"). The comment was
+written when it was true — before the harness claimed the path — and neither was
+re-checked when the manifest started deploying there. **The description aged; the
+code did not.**
+
+## What to do instead
+
+**Probe the target before deleting the cleanup.** Not to decide whether the
+deletion is safe — a `rm -f` of an absent file is safe either way — but to find
+out what the block is actually pointed at. One `stat` was the entire difference
+between "harmless cleanup, deleted" and "live doctrine target, this is a bug".
+
+Concretely, for any block that removes a path:
+
+1. Look at the path on a real machine. Present is a question, not an answer.
+2. Ask who *writes* it, not only who reads it. Here the writer was
+   `compile-harness.sh`, a subsystem the setup script never mentions.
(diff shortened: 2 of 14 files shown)


mlorentedev:
## Review triage

Triaged at `d0c8b16`.

**Who reviewed:** PR-Agent only, and it reported **"No major issues detected"** with zero code findings — effort 3/5, "PR contains tests", no security concerns. CodeRabbit's check is green but it is rate-limited on the free OSS tier and has posted its *"review limit reached"* notice on the sibling PRs today; a green `review-attestation` here means **one reviewer, not two**. Stated rather than implied.

|---|---|---|---|
| 1 | PR-Agent | #1333 partially compliant: "the ticket originally proposed deleting all migrations; this PR keeps two" | **Declined — that is the PR's thesis, not a gap.** |
| 2 | PR-Agent | #1431 non-compliant: MEM-002 not fixed, not mirrored to Windows, still asserts a proxy | **Declined — scope inference error.** |
| 3 | PR-Agent | #1409 non-compliant: BUG-771 not fixed | **Declined — same error.** |
| 4 | Self (post-review) | The endpoint-variable claim was under-evidenced | **Fixed** in `d0c8b16`. |

### 1 — the divergence is the finding

The ticket asserted eleven finished migrations. Classifying each by *what breaks on a machine that never runs it*, then probing the targets, found that **two were not finished** (MEM-002 had never worked; HIVE-118 cannot be probed from Linux) and a third was **destroying a live harness deploy target on every run**. Deleting all eleven as requested would have shipped all three outcomes as a cleanup. The spec records the per-block classification and the probe results; the guard that prevents the third recurring is fail-first proven red against `main` at `setup-linux.sh:503`.

### 2 and 3 — ticket-scope inferred from prose

#1431 and #1409 appear in this PR because it **filed** the first and **baselined against** the second. PR-Agent read those references as tickets this PR must satisfy, and generated a compliance table against work the PR explicitly excludes — listing "the fix is not attempted in this PR" as its own evidence of non-compliance.

#1431 is deliberately untouched, whole: its fix needs a `claude plugin marketplace remove` subcommand that may not exist, a state file this repo has never touched, and a Windows leg. #1409 is a pre-existing suite failure that this PR *measures* in order to prove its own 7 remaining reds are not regressions.

This is the **fourth** independent instance today of PR-Agent inferring ticket scope from prose and getting it wrong (three on `feat/persona-skill-enforcement`, per that session). Its code and artifact findings have been real every time; only the ticket-compliance inference is unreliable. Recorded here rather than filed — the pattern belongs to the reviewer-configuration work, not to this PR.

### 4 — one thing the review did not catch, found and fixed after it ran

The first pass grepped only the two files it was editing and concluded the `ANTIGRAVITY_ENDPOINT` / `CLOUDCODE_URL` exports were inert. A repo-wide sweep found three more homes: `.zshrc:42-43` exports both on every interactive zsh (the actual reason a running shell has them), `tests/antigravity.bats:84` pins that export, and `cli/internal/doctor/checks_deploy.go:801` reads the variable.

The deletion stands and is a no-op outside the setup process — doctor reads it as `sys.env(..., "https://cloudcode-pa.googleapis.com")`, defaulting to production when unset, inside a section that SKIPs when `agy` is absent — but the stated reason was broader than the evidence behind it. Corrected in `d0c8b16`; `antigravity.bats` 18/18, `go build` + `vet` clean. Same mistake in miniature as lesson 256, which is why it is written down rather than quietly amended.

### Owner decisions still open (in the PR body)

Both were put to an advisor and both came back **no**:

- **A CI check for a `# EXPIRES:` convention.** Neither bug found here was a date failure — MEM-002 with an expiry would have been deleted on schedule and the re-clone would still have gone unnoticed; GEMINI.md would have been deleted for the wrong reason. What caught both was probing the target. The enforcement point is review, citing lesson 257, not a CI gate on a comment.
- **Scrubbing the Windows registry values.** Inert by measurement, and the repo's idiom for machine convergence is a `dotf doctor --fix` WARN, never a new one-time migration — which is the thing this PR removes. Windows-empirical, so it belongs to the batched Windows session if the owner wants it at all.

Issue #1434, opened 2026-09-02
HARNESS-109: a named subagent dispatch sends its NAME as agent_type, so the gate resolves no persona and enforcement silently turns off


mlorentedev:
## Measured 2026-09-02

Discovered by the decision record from `specs/HARNESS-106-skill-capability/` AC5-AC7 the day it landed, which is the point of it: this state allows, exits 0, and is indistinguishable from health without the journal.

Two dispatches of the **same** persona, differing only in whether the caller supplied a name:

| dispatch | payload `agent_type` | `role_resolved` | outcome |
|---|---|---|---|
| `subagent_type: reviewer`, no name | `reviewer` | `reviewer` | `warn` — 4 skills named |
| `subagent_type: reviewer`, `name: gate-probe-agent-type` | `gate-probe-agent-type` | *(none)* | **`role-unresolved`** |

`agent_type` carries **the caller-supplied name when one is given**, not the persona type. So `effectiveRole` looks up `harness/agents/<name>/AGENT.md`, finds nothing, and `loadGatePersona` fails open — correctly, by its own contract, but the result is that **naming a subagent disables its gate**.

## Why this matters more than it looks

Under the current `enforce: warn` canary it costs a warning nobody sees. Under `enforce: block` it is a **bypass**: any caller that names its dispatch is ungated, and nothing in the exit code, the stderr, or `dotf doctor` says so. Promoting any skill to `block` before this is fixed ships an enforcement mechanism with a one-word opt-out.

It also generalises the finding already recorded in lesson 255 about `agent_id`: **both** identity fields are caller-influenced. `agent_id` here was `agate-probe-agent-type-982a508c7a296d18` — the name, prefixed and hashed — against `aa963d775186ff4a1` for the unnamed one.

## Constraint on any fix

The gate must not start blocking on an unrecognised role: fail-open on a config error is deliberate and stays (see `loadGatePersona`, and lesson 219). What is wrong is that the *name* is being read as the *type*, and that the resulting non-enforcement is invisible in the moment.

Candidate directions, none decided:
- Resolve `agent_type` against the roster and, on a miss, fall back to a second payload field if one carries the true type. **Requires measuring what else the payload contains** — nothing here has captured a full subagent payload verbatim.
- Treat an unresolvable role as a first-class reportable state in `dotf doctor` rather than only in the journal, so it surfaces without anyone querying.

## Do not close by

[...]

PR #1435, opened 2026-09-02
feat(harness): the gate writes down every decision it takes

mlorentedev:
Refs #1420 — **deliberately not a closing keyword**; AC4 is still open. Spec: `specs/HARNESS-106-skill-capability/` (`status: implementing`). Closes AC5, AC6, AC7.

## The problem

The gate's answer to a harness is an exit code, and an exit code is consumed and discarded.

| These share exit 0 | |
|---|---|
| `allow` | nothing outstanding |
| `warn` | a skill unconsumed, reported and allowed |
| payload unparsed | the harness said something we could not read |
| **`role did not resolve`** | **enforcement was OFF for this call** |

Everything that told them apart was stderr — and a `PreToolUse` hook's stderr on exit 0 **is not persisted**, measured, and the subject of lesson 254, where a canary ran for a day accumulating no evidence while appearing to work.

Three paths returned before any persona was loaded and left nothing at all. That is why "agy sends a payload we cannot parse" has been indistinguishable from "agy has nothing to enforce".

## What this adds

A per-scope append-only journal beside the consumption ledger, keyed by the **same digest** — reusing the collision guard rather than inventing a second path scheme, because a colliding journal corrupts the measurement instead of announcing itself.

Nine fields, a closed `Outcome` vocabulary, and `agent_type` read straight off the payload **before any lookup**, so even the skill path preserves it. It is kept separate from the role *requested* and the role *resolved*, because those three can differ and the difference is the diagnosis.

**The record carries no tool input.** It is durable, unencrypted, unscanned, and written on every tool call; tool arguments carry file contents, commands and credentials, so logging them would be a secrets leak with a retention policy. Pinned by a test that puts a key-shaped string in the payload and greps the file for it.

## Measured on a real dispatch, not just tested

```
outcome          agent_type       role_resolved   n
no-role          <empty>          -               38   (main-thread calls)
warn             reviewer         reviewer         2   ← AC6, AC7
role-unresolved  gate-probe-...   (none)           3   ← see below
```

The `warn` record names all four skills (`adversarial-review`, `audit`, `cyclomatic-complexity`, `verification-before-completion`), read back from the journal with nothing but a path shared with the writer. That is what AC6 means by *observable after the session ends*.

## It caught a real bypass on its first day

Two dispatches of the **same** persona, differing only in whether the caller supplied a name:

| dispatch | payload `agent_type` | `role_resolved` | outcome |
|---|---|---|---|
| `subagent_type: reviewer`, no name | `reviewer` | `reviewer` | `warn` |
| `subagent_type: reviewer`, `name: gate-probe-agent-type` | `gate-probe-agent-type` | *(none)* | **`role-unresolved`** |

`agent_type` carries the **caller-supplied name** when one is given. So naming a subagent makes its role unresolvable and **turns its gate off** — allowing, exit 0, invisible to the exit code, the stderr and `dotf doctor` alike. Under `enforce: block` that is a bypass with a one-word opt-out.

Filed as **#1434**, and a hard precondition on promoting any skill to `block`. It also generalises lesson 255: **both** identity fields are caller-influenced, not just `agent_id`.

## AC4 stays open, with a named blocker

The deployed `reviewer` record carries `Skill` after the deploy, and a dispatched reviewer still reported `Read, Bash, advisor` — the roster its *dispatching session* loaded at start. **Personas are frozen at session start; hooks are re-read.** Yesterday's reload measurement covered hooks and was generalised to persona records without being re-measured (lesson 256).

AC4 was written as *"proven by a dispatch that writes a consumption record — never by a config file containing a key"*, and that wording is exactly why a wrong assumption became a deferred item rather than a false claim. It needs a session started after the deploy, and nothing in this change.

## Two supporting changes the record forced

- `loadGatePersona` now reports **which** of its two nil cases happened. Both allow, and only one means enforcement was off; the caller could not tell them apart.
- The block path returns a tagged exit code instead of `os.Exit`. Identical status, but `os.Exit` kills the test runner — so the one path that matters most could never be driven end to end in process — and it runs no defers, making it the path most likely to lose its own record.

## Also fixed

`TestMergeAgainstTheRealDeployedSettings` failed on any machine bound before the marker field existed: such a machine carries an unmarked gate entry that `isOurs` deliberately **adopts**, so the foreign *count* drops by one while nothing is lost (15 → 14 here). A count cannot tell adoption from deletion. The assertion is now per command, and the only tolerated disappearance is an entry the merge was entitled to claim.

## Verification

`go build` / `vet` / `test` clean, `GOOS=windows go vet` clean, `golangci-lint` 0 issues, `shellcheck` clean, `bats tests/gitattributes-eol.bats` green.

Red direction proven for the journal by neutering the skill path's `record(...)` in place: `journal for scope "s1-a1" holds 0 records, want exactly 1` — exactly the two affected cases failed, the rest stayed green.

### DIFF
--- a/cli/internal/cmd/harness_gate.go
+++ b/cli/internal/cmd/harness_gate.go
@@ -32,6 +32,17 @@ const gateExitBlock = 2
 // command that could exit 1 would therefore mean opposite things on two
 // harnesses. Every path here resolves to an explicit decision instead, so the
 // harness never sees an error and its error semantics never apply.
+//
+// The single non-nil return is tagged with exit 2 explicitly. An untagged error
+// would resolve to 1 through ExitCode — the one status this command must never
+// produce — so nothing here may return a bare error.
+//
+// AND EVERY PATH LEAVES A DURABLE RECORD. An exit status is not an audit trail:
+// the harness consumes and discards it, `warn` and `allow` share it, and a
+// PreToolUse hook's stderr on exit 0 is not persisted anywhere the session can
+// be asked about later. Three paths below return before any persona is loaded,
+// and before the journal they left nothing at all — so a harness whose payload
+// this cannot parse looked exactly like a harness with nothing to enforce.
 func newHarnessGateCmd() *cobra.Command {
 	var (
 		harnessName string
@@ -57,9 +68,28 @@ Invoking a skill is never blocked: forbidding it would deadlock the session.`,
 		SilenceErrors: true,
 		RunE: func(cmd *cobra.Command, _ []string) error {
 			payload, _ := io.ReadAll(cmd.InOrStdin())
+			if stateDir == "" {
+				stateDir = defaultGateStateDir()
+			}
+			// One closure, used by every exit below, because the requirement is
+			// that NO path returns without leaving a record. Writing it at each
+			// `return` is how the three early paths came to leave none: each was
+			// individually correct, and the omission existed only in aggregate —
+			// the same shape as the defect this spec started from.
+			record := func(scope string, rec harness.DecisionRecord) {
+				rec.Harness = harnessName
+				rec.Scope = scope
+				_ = harness.RecordDecision(harness.DecisionPath(stateDir, scope), rec)
+			}
 
 			call, understood := normaliseToolCall(harnessName, payload)
 			if !understood {
+				record(harness.UnparsedScope, harness.DecisionRecord{
+					Outcome:      harness.OutcomePayloadUnrecognised,
+					Allowed:      true,
+					Reason:       "payload not recognised",
+					PayloadBytes: len(payload),
+				})
 				// MEASURED 2026-08-26: this branch is why it exists. Without
 				// it a malformed payload produced a zero ToolCall, Decide saw
 				// a valid persona with nothing consumed, and the gate BLOCKED
@@ -69,13 +99,23 @@ Invoking a skill is never blocked: forbidding it would deadlock the session.`,
 				_, _ = fmt.Fprintln(cmd.ErrOrStderr(), "[gate] allow: payload not recognised")
 				return nil
 			}
-			if stateDir == "" {
-				stateDir = defaultGateStateDir()
-			}
 			// Scoped to the acting persona when the harness names one: a subagent
 			// reuses the parent session id, so keying by session alone would let one
 			// persona's skill runs satisfy another's gate.
-			statePath := harness.StatePath(stateDir, call.ConsumptionScope())
+			scope := call.ConsumptionScope()
+			statePath := harness.StatePath(stateDir, scope)
+
+			// Every record from here carries what the harness said was acting.
+			// THIS IS WHAT CONVERTS agent_type FROM INFERRED TO MEASURED: it is
+			// read straight off the payload, before any persona lookup, so even
+			// the skill path below — which returns before that lookup — leaves
+			// evidence that the field arrived and of what it contained.
+			base := harness.DecisionRecord{
+				Session:   call.SessionID,
+				AgentType: call.AgentType,
+				AgentID:   call.AgentID,
+				Tool:      call.Tool,
+			}
 
 			// A skill invocation is the act the gate exists to require: record
 			// it and get out of the way. Recording failures are ignored on
@@ -83,6 +123,12 @@ Invoking a skill is never blocked: forbidding it would deadlock the session.`,
 			// failing here would block a session over a full disk.
 			if call.Skill != "" {
 				_ = harness.RecordConsumed(statePath, call.Skill)
+				rec := base
+				rec.Skill = call.Skill
+				rec.Outcome = harness.OutcomeSkillConsumed
+				rec.Allowed = true
+				rec.Reason = "skill invocation recorded"
+				record(scope, rec)
 				return nil
 			}
 			if call.IsSkillTool {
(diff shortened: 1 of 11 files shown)


mlorentedev:
## Review triage

Reviewed by **PR-Agent**. **CodeRabbit did not review** — the *"Review limit reached"* notice again, its third consecutive on my PRs. A notice about a review is not a review, so `review-attestation` green here means **one** reviewer.

### Applied — and the finding was real

| Finding | Disposition |
|---|---|
| **Silent data loss on rotation** / **unsafe concurrent file operations** in `RecordDecision` | **Applied** in `ad3d51e`. |

The ordering was wrong: rotation ran **before** the append. Single-threaded that is harmless — stat sees a full file, rename it, write to a fresh one, nothing lost — which is exactly why it read as fine. Concurrently it is not: two writers both see a full file, the first renames it and writes one record to a fresh one, and the second renames **that one-record file** over the kept generation, destroying the history the first just preserved.

Now the write happens first and rotation is decided from the descriptor just written. `f.Stat()` answers about *that* file rather than about whatever the path names now, and `os.SameFile` confirms the path still refers to it — so a not-full file can no longer replace a full one, and the record in hand is durable before rotation is considered at all.

### Partly declined, with the reasoning

The second finding's stronger framing — *"the journal file being corrupted"* — does not hold for `O_APPEND`. A single `write(2)` of one line to a regular file is atomic with respect to the offset, which is the mechanism multi-process log appends rely on generally, and every record here is far below the size where that stops holding. Building the line in memory and issuing exactly one `Write` is therefore load-bearing rather than tidiness, and is now commented as such.

Rather than argue that in prose, it is measured: **24 parallel writers × 40 records, all 960 present, each exactly once.** A count alone would have passed with one duplicated and another lost, so the assertion is per record. Also run under `-race`.

### A residual race that I am NOT claiming to have fixed

Between the `SameFile` check and the rename, another writer can still rotate. What it can cost is bounded to **old history** — the record being written is already durable — and a not-full file is no longer a plausible replacement.

Closing it entirely needs an advisory lock, and `syscall.Flock` does not exist on Windows. This package compiles for Windows, and a Windows-only build break is a defect this repository has already paid for once (#1075). Not worth that for a diagnostic journal whose worst case is losing records nobody has asked about since they rotated. Documented at the call site instead of quietly left.

### How the test for it was arrived at, because two versions were wrong

Worth recording, since both failure modes are ones this repo keeps meeting:

1. **Vacuous.** The first ordering test was single-threaded and asserted "no record is lost". The old ordering *passes* that. It stayed green with the defect reinstated, and was caught only by proving the red direction — not by writing it.
2. **Flaky.** The second asserted the emergent property under concurrency (the kept generation is always full). It failed once and passed on rerun, which is how the residual race above was discovered: that property is precisely what the code does **not** guarantee.
3. **Deterministic.** The third asserts the ordering itself — the record that triggers rotation must be in the rotated file. Stable over repeated runs, and proven red: `the record that triggered rotation (r249) is not in the rotated file`.

### Ticket-scope inference, declined again

PR-Agent's compliance table is not reproduced here; its code findings were real and its scope reading is not. @dotfiles-7e independently hit the same thing on #1433, where it generated tables against a ticket the PR **filed** and a pre-existing failure the PR **baselines against**, counting "the fix is not attempted in this PR" as non-compliance. That is now **four independent instances**: code and artifact findings reliable, ticket-scope inference from prose not. It is a reviewer-configuration matter rather than a per-PR one, and I am carrying it into that area rather than filing it against this PR.

### Verification after the change

`go build` / `vet` / `test` clean, `GOOS=windows go vet` clean, `golangci-lint` 0 issues, `go test -race` clean on the concurrency test.


---

### Re-triage after the fix push

Posted at 04:17:16; PR-Agent re-reviewed at **04:18:50** against `ad3d51e`. It edits in place, so `created_at` never moves and the triage queue cannot see the difference — the same stale-disposition trap as on #1428 earlier today. Re-read rather than assumed:

**Verdict is now "No major issues detected".** Both rotation findings are gone against the fix. `review` and `review-attestation` are green on the final head; only `test-windows` is still running, and its result is not claimed here.

CodeRabbit still has not reviewed this PR — third consecutive limit notice.

mlorentedev:
## Review triage

Re-triage: both reviewers re-ran on `57e5128` and **edited their comments in place**, so
`created_at` never moved and `dotf pr triage-queue` read exit 0 over a stale disposition.
Filed as its own defect rather than left as a habit.

| reviewer | verdict | disposition |
|---|---|---|
| **PR-Agent** | *No major issues detected*, *No security concerns identified*, tests present | nothing to apply |
| **CodeRabbit** | **review limit reached** — *"Next included review available in 23 minutes"* | **no review ran** |

**Disclosure: CodeRabbit did not review `57e5128`.** A notice about a review is not a review;
this PR's code has been reviewed by PR-Agent only.

PR-Agent's ticket-compliance analysis is correct on both counts, and both are deliberate:

- **#1420 AC4 "still pending, correctly left open"** — agreed, and unclosable in this PR.
  Persona records are frozen at the dispatching session's start (lesson 256), so AC4 needs a
  session started after the deploy, not a code change.
- **#1434 "does not fix the root cause… a separate change"** — agreed, out of scope by design.
  This PR is the *instrument* that found #1434 on its first day. The suggestion of a
  `dotf doctor` check for unresolvable roles is a good one and is recorded on #1434.

Checks: all green, including `test (windows-latest)` — the leg that caught the rotation defect
(`os.Rename` over an open handle) that no Linux run and no `GOOS=windows go vet` could see.

PR #1441, opened 2026-09-02
fix(doctor): pi/opencode/copilot/yarn pin checks agree with the installer's floor policy

mlorentedev:
## Summary
- `dotf tools install`'s `decideAction` (and pi's own setup block, REFACTOR-013) treat a versions.conf/packages.json pin as a MINIMUM and never downgrade a newer install. `dotf doctor`'s drift check compared with an exact match instead, so any tool that shipped a patch upstream between two pin bumps produced a permanent false WARN — installer and doctor disagreed about what the pin meant.
- Observed on `PI_VERSION`: bumped once already in f1c20cd ("the version actually installed") to chase exactly this WARN, then drifted again the moment pi shipped its next patch.
- Add `matchPinFloorFrom`/`matchPinFloor` (ahead-of-pin PASSes, behind-pin WARNs, mirroring `checkGitWindowsFloor`'s `atLeast`) and switch pi, opencode, copilot and yarn to it.
- `golangci-lint` and `age` keep the exact-match `matchPinFrom` on purpose: golangci-lint's pin is the version CI actually lints at, so drift in either direction diverges from CI (BUG-071); age has no floor-reconciling installer and no evidence it should behave differently — left untouched.
- Bumped `PI_VERSION` to 0.84.4, the version the floor policy already converged this box to.

## SDD skip rationale
55 production LOC, just over the 50-LOC gate. This is a mechanical bug fix with an obvious cause (installer and doctor using different comparison semantics for the same pin) — most of the LOC is the new function's doc comment explaining why the two semantics must agree. A spec folder would document a decision that's already fully justified in the commit message and code comments; it would not change any outcome here.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/doctor/...` (updated `checks_copilot_test.go` and `checks_opencode_version_test.go` to assert PASS-above-pin instead of WARN)
- [x] `go test ./...` (full module, green)

### DIFF
--- a/cli/internal/doctor/checks_copilot.go
+++ b/cli/internal/doctor/checks_copilot.go
@@ -3,12 +3,13 @@ package doctor
 // checkCopilot reports the GitHub Copilot CLI against its packages.json pin
 // (AI-038, #1321; ADR-036 amendment): copilot is an npm catalog tool on every
 // OS, so its version is probed once in Go (tools.ProbeVersion via semverOf)
-// and compared with the catalog pin exactly as opencode is — an exact match
-// PASSes, anything else is a drift WARN (the floor semantic belongs to
-// `dotf tools install`, which never downgrades). Absent is a SKIP: a box may
-// deliberately not carry Copilot, and setup only deploys its config when the
-// binary is on PATH. A leftover winget or scoop copy beside the npm one is
-// reported by checkShadowedCatalogTools, which reads the same catalog.
+// and compared with the catalog pin on a floor basis — `dotf tools install`'s
+// decideAction never downgrades a newer install, so doctor must agree that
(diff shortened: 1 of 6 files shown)


mlorentedev:
## Review triage

- **pr-agent — inconsistent indentation in `matchPinFloor`/`matchPinFloorFrom`** (spaces vs. tabs): declined as a false positive, checked after merge. `gofmt -l` against the committed blob (`git show <sha>:cli/internal/doctor/checks_deploy.go`) reports clean; every indented line in both functions is a tab (`cat -A` confirms). Also consistent with `lint`/`test` both having passed pre-merge, which would have caught real gofmt drift. No action taken.

Issue #1442, opened 2026-09-02
CLI-070: dotf doctor prints a Next steps block for FAILs that carry a remedy command


mlorentedev:
## Problem
A setup run that ends with `dotf doctor` exiting non-zero hands the operator nothing but the scrolled check list to act on. Both `setup-windows.ps1` and `setup-linux.sh` print a hardcoded "Next steps" block at the very end (restart the shell, verify a couple of paths, `project-init ...`) that is unrelated to whatever doctor actually found. On a fresh Windows box this session, doctor FAILed with `Bitwarden session is gone ... run \`bw login\`` — a one-line fix — but nothing at the end of the run said so; the remedy was buried mid-scroll under ~40 other check lines.

## Fix direction
`dotf doctor` collects the FAIL lines that carry a `run`/`re-run`/`recover with`/`upgrade with` + backtick-quoted remedy and prints them as a curated `Next steps:` block after `Results:` — additive, no change to any check's message format. WARNs are intentionally excluded (Next steps is about what is actually blocking, i.e. what drives the non-zero exit); a remedy phrased without one of those verbs is a message-authoring convention gap, not something the parser chases.

Both setup scripts' own hardcoded "Next steps" block stays as-is (still useful for first-run orientation) — this is additive, not a replacement.

## Source
Observed live during a Windows setup session (2026-09-02): stale local checkout (472 commits behind), converged via `git pull` + re-run, `dotf doctor` FAILed on an unauthenticated Bitwarden session with the exact remedy in its message, and setup's own next-steps block said nothing about it.

PR #1443, opened 2026-09-02
feat(doctor): print a Next steps block for FAILs that carry a remedy command

mlorentedev:
## Summary
- Every setup run ends with `dotf doctor`'s scrolled check list, then setup's own hardcoded "Next steps" (restart the shell, `project-init ...`) — unrelated to whatever doctor actually found.
- Observed live this session: doctor FAILed with an unauthenticated Bitwarden session and its exact remedy (`run \`bw login\``) already in the message, and nothing at the end of the run said so; the operator had to find it by scrolling back through ~40 other check lines.
- `dotf doctor` now scans its own rendered transcript for FAIL lines carrying a `run`/`re-run`/`recover with`/`upgrade with` + backtick-quoted remedy and prints them once each, in first-seen order, as a `Next steps:` block after `Results:`.
- WARNs are excluded by design — Next steps is about what actually drives the non-zero exit, matching `Report`'s own PASS/FAIL/WARN/SKIP contract.
- Extraction reads the already-rendered text rather than adding a hint field to `Report.Fail()` across the package's ~100 call sites — `Report`'s shape is unchanged.
- Color detection runs on the real writer before it's teed into the transcript buffer, so ANSI output on a real terminal is unaffected (`io.MultiWriter` is never an `*os.File`).

## Not included
- Either setup script's own hardcoded "Next steps" block — untouched, still prints below doctor's output. Removing/rewriting it wasn't reviewed as part of this issue; this feature is additive.
- A FAIL message reworded without one of the four remedy verbs won't be picked up. That's a message-authoring convention gap (34 of ~38 backtick-carrying messages already follow it), not a parser gap this PR chases.
- The pin-floor semantics fix from the same investigation session shipped separately as #1441 — independent, obvious-cause bug fix; bundling it here would have diluted this feature's own spec.

## Spec
`specs/CLI-070-doctor-next-steps/` (issue #1442). Proposal, tasks and verification filled from the investigation that motivated this — see `verification.md` for the acceptance-criteria-to-test mapping and the live smoke-test output.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/doctor/...` — new `doctor_test.go` (`TestNextSteps`, `TestNextSteps_NoFailNoBlock`, `TestRun_NextStepsBlock`)
- [x] `go test ./...` (full module, 22 packages, green)
- [x] Built `dotf` from this branch and ran `dotf doctor` live on a real Windows box with an unauthenticated Bitwarden session:
  ```
  Results: 121 passed, 3 failed, 10 warned, 32 skipped

  Next steps:
    bw login
    export BW_SESSION=$(bw login --raw) && bw sync
  ```

### DIFF
--- a/cli/internal/doctor/doctor.go
+++ b/cli/internal/doctor/doctor.go
@@ -1,9 +1,12 @@
 package doctor
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"os"
+	"regexp"
+	"strings"
 )
 
 // Options configures a doctor run. System and StartDir are injection seams for
@@ -48,7 +51,18 @@ func Run(opts Options) (int, error) {
 		return 2, err
 	}
 
-	rep := NewReport(out, opts.Verbose)
+	// Tee the report into a transcript so Summary() can be followed by a
+	// curated Next-steps block (below) without changing Report's shape: it
+	// streams messages straight to out and keeps no record of them (CLI-070,
+	// #1442 — setup's own "Next steps" never named the FAIL doctor had just
+	// printed, e.g. `bw login`, so the reader had to go find it in the
+	// scroll). Color detection must run on the real out, not the tee:
+	// io.MultiWriter is never an *os.File, so isColorEnabled would otherwise
+	// report false on every run.
+	color := isColorEnabled(out)
+	var transcript bytes.Buffer
+	rep := NewReport(io.MultiWriter(out, &transcript), opts.Verbose)
+	rep.SetColor(color)
 	mode := "check"
 	switch {
 	case opts.Quick:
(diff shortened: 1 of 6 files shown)


mlorentedev:
## Review triage

- **pr-agent — ANSI color could split the `[FAIL]` tag** (Recommended focus area): declined as a defect, but the concern was fair since it wasn't explicitly tested. Verified: `coloredTag[StatusFail]` is `ansiRed + "[FAIL]" + ansiReset` — the color codes wrap the whole literal from the outside, they never land inside it, so `strings.Contains(line, "[FAIL]")` still matches. Added `TestNextSteps_SurvivesColoredTag` (2edc307) to pin this down with a real colored-tag byte sequence instead of leaving it as an unresolved comment.

PR #1446, opened 2026-09-02
docs(lessons): record the two-quoting-layer mklink discovery

mlorentedev:
## Summary
Docs-only. Records lesson 258 from #1444 (HARNESS-050): os/exec's argv escaping and cmd.exe's own tokenizer are two independent quoting conventions, and a cmd.exe builtin (mklink) is only ever parsed by the second one.


mlorentedev:
## Review triage

- **pr-agent — "partially compliant with #1444"** (missing the code fix, the file split, the table test, etc.): false pairing, declined. This PR is docs-only (the lesson entry); it references `#1444` in its body only to credit the source PR, not to close or re-implement it. `#1444` itself (the actual `createLink` fix + `TestCreateLink_CmdDelimiterPaths`, verified live on Windows) already merged separately. Nothing to add here.

Issue #1447, opened 2026-09-03
OPS-046: the deploy-dir→$HOME content check has no Windows map, so the leg stays unguarded there


mlorentedev:
## Context

OPS-043 (#1337) adds a `Deploy-dir↔$HOME drift` section to `dotf doctor`: a byte comparison of every file setup deploys into `$HOME` against its `~/.dotfiles` source. Before it, **no** doctor section compared content on that leg — `checkSymlinks` (`cli/internal/doctor/checks_deploy.go:20`) and `checkProfileFiles` (`checks_profile.go:45`) test existence only, and a drifted real file reports `PASS`. `checkDeployDrift` covers the *other* leg (repo → deploy-dir).

## The gap this issue tracks

OPS-043's `homeDeployMap` is derived from `setup-linux.sh`'s eleven `deploy_file … "$HOME/…"` call sites. Every entry in it is POSIX-only, so **the new section SKIPs entirely on Windows** and the leg stays unguarded there.

Windows is where it matters most: `setup-windows.ps1` never had a `check_deployed` twin at all (it converged on `dotf doctor` as the single post-setup diagnostic under CLI-018), so unlike Linux it has no shell-side content assertion being replaced — it simply has none.

## What closing this looks like

- A second map derived from `setup-windows.ps1`'s own deploy call sites.
- Its own join guard, mirroring OPS-043's AC5: a Go test that fails when a Windows deploy target has no map entry. Nothing today spans a setup script and the doctor map, which is the class of drift the guard exists to catch.
- The exemption question re-measured on the Windows box, not inherited from the Linux list — on Linux only `.gitconfig` was observed drifting, and that finding does not transfer.

## Why it was not folded into OPS-043

Doubling the map and its guard would push that PR past the 300-LOC atomic cap, and the Windows exemption set cannot be measured from Linux (same reason HIVE-118 was kept in OPS-040). Deferred deliberately and stated as an explicit non-goal in `specs/OPS-043-setup-doctor-duplication/proposal.md`, so the reviewer sees it before archive rather than after.

Blocked on #1337 landing (it introduces the map and the guard idiom this one extends).

PR #1440, opened 2026-09-02
chore(spec): archive OPS-040 on a passing independent review

mlorentedev:
## Summary

Archives `OPS-040-dead-migration-purge` on a **PASS** from an independent adversarial review, and closes #1333. The implementation landed separately in **#1433** (`06c3b7a`); this PR is the lifecycle close.

## The review

**Reviewer: `nan/glm5.3-flash`**, drawn at random from `harness/reviewer-pool.json` — not the implementing session. `reviewed_sha` `d0c8b16`. **No Blocker, no Major.**

| Dimension | Grade |
|---|---|
| Scope | A |
| Reliability | A |
| Maintainability | A |
| Correctness | B |
| Verification | B |
| Handoff-readiness | B |

It verified by execution rather than by reading `verification.md`: re-ran the `features.json` commands, re-derived the seven-failure baseline **at the true merge base `16d8f96`** rather than the older `3a11d97` this session used, and reproduced the guard's fail-first with **three** mutations — including `rm -f "$HOME/.codex/AGENTS.md"`, the *second* manifest target. That last one matters: it proves the guard's list is resolved from `harness/manifest.json` rather than hardcoded to `GEMINI.md`, which is exactly the property the guard was written for and a stronger proof than the one this session produced.

It also caught, mid-run, that `origin/main` had advanced to `06c3b7a` as #1433 squash-merged, verified the merged state content-identical (`git diff origin/main HEAD` empty, both setup-script blobs matching), and recorded that its review remains valid for it.

## Findings — six, all Minor or Question, all dispositioned

**Two were mine and both were wrong. Corrected in `verification.md` with the reason, not quietly.**

- **F1 — recorded numbers do not reconcile.** `setup-windows.ps1` is **2305** lines, not 2301, so the net is **−88**, not −92. The figure was measured before two later edits added explanatory comments to that file, and the stale number was then carried into the commit message, the merged PR body and the handoff without re-measurement. Separately, the suite tally `ok=1526 not_ok=7 skipped=76` sums to 1609 against a TAP plan of `1..1535` — the counting `awk` matched `/# skip/` on lines that also matched `/^ok /`, double-counting every skip. Actual: **ok=1528 (77 skips), not_ok=7**. The material claims — net deletion, and the seven failures all being pre-existing — are unaffected and were independently reproduced.
[...]

PR #1450, opened 2026-09-03
refactor(setup): port the shell verification layer into dotf doctor, then delete it

### LINKED ISSUE #1333: OPS-040: dead-migration purge in both setups - converge-old-machines code with no expiry, plus a secret decrypted for nothing on every run

mlorentedev:
Refs #1337 — the issue stays open until the spec archives, which is a separate PR after an independent adversarial review (same shape as OPS-040: #1433 → review → #1440 closed #1333).

Spec: `specs/OPS-043-setup-doctor-duplication/`

## The ticket's premise was false in five places

#1337 asked to delete `check_deployed` and `check_dependencies` from `setup-linux.sh` as duplicates of `dotf doctor`. Measured first:

1. **`check_deployed` is not duplicated.** It compares the deploy dir to `$HOME`; `checkDeployDrift` compares the **repo** to the **deploy dir**. Different legs of `repo → ~/.dotfiles → $HOME`.
2. **No doctor section byte-compared the second leg.** All 40 enumerated. `checkSymlinks` and `checkProfileFiles` test existence, and PASS a real file that has drifted. `setup-windows.ps1` never had the shell check, so that leg was **unguarded on Windows outright**.
3. **`check_dependencies` never fails** — it only `log_warning`s. Of its 14 tools: 9 stricter in doctor, 4 equally advisory, and `docker-compose` covered **nowhere** (not doctor, not `env-contract.json`, not any package list).
4. **`.zsh/functions.sh` was in no doctor list.**
5. **The six pre-exported path vars are BUG-021's fix**, not duplication — the setup shell has not re-sourced the rc files, so without them doctor reports false warnings on every fresh box.

So the order matters: **port first, then delete.** Done as the ticket asked, this removes coverage while the diff reads as removing duplication.

## What this does

- **New doctor section `Deploy-dir↔$HOME drift`** — byte-compares 11 files (was 3) and rejects a symlink where ADR-012 expects a regular file, the severity `checkSymlinks` does not carry. Driven by an explicit `homeDeployMap`, because this leg renames (`ssh/config` → `.ssh/config`, `tmux.conf` → `.tmux.conf`) and `isManagedDeployPath` cannot be reused.
- **`docker compose` gets a verdict**, probed as the v2 CLI plugin with the v1 binary as fallback. `command -v docker-compose` — what the shell call did — reports missing on a current install where compose works; measured on msi, both are present. Absence is SKIP: the repo provisions compose nowhere, so FAIL would red a box that never wanted it (BUG-052's reasoning for terraform).
- **Deletes** the 3 `check_deployed` calls, the `check_dependencies` call, both function bodies (no other caller repo-wide), their bats tests, and `Test-FileDrift` in `scripts/utils.ps1` — dead since CLI-018 retired `healthcheck.ps1`, with zero call sites.
- **Keeps and guards** the six pre-exports, in both scripts, asserting they still precede the `dotf doctor` handoff.

## Three guards, because the gap was structural

- `TestSetupShellParity` — all 17 items the shell calls covered, each asserted equal-or-stronger in doctor. **Mutation-tested**: flipping `.zsh/functions.sh` to existence-only turns it red.
- `TestHomeDeployMapCoversSetup` — the join guard. Nothing spanned `setup-linux.sh` and doctor's lists before, which is the gap #1337 came through. Fails in both directions.
- `TestHomeDeployExemptionsAreReasoned` — an exemption with no stated mechanism fails. See below for why.

## Two corrections found along the way

- The NOTE exempting `.zshrc` cited a `sed -i` that **no longer exists** — OPS-040's purge removed the mechanism and left the justification. Corrected here; written up as lesson 259, since OPS-040's own method was lesson 256 ("probe the target, don't read the comment") and it still produced an instance.
- The exemption list was **measured, not inherited**: deriving it from `check_deployed`'s comments would have missed `.gitconfig`, the only one of the 11 files actually observed drifting.

## Not in scope

- **Windows deploy targets.** Every map entry is POSIX-only, so the section SKIPs there and the leg stays unguarded. Filed as **OPS-046 (#1447)**, blocked on this — stated as an explicit non-goal in the proposal rather than implied.
- `checkDeployDrift`, the FAIL/SKIP policy of helm/terraform/ansible/pip (both sides advisory today), and `docker`/`kubectl` on Windows (BUG-052).

## Verification

- `go build` + `go vet` + `GOOS=windows go vet` clean; `go test ./...` all ok.
- `golangci-lint run` at the pinned **2.12.2** (matches `versions.conf`, BUG-071): **0 issues**.
- `bats tests/*.bats`: **1534 tests, exit 0**, run against the tree with the deletion applied.
- `shellcheck --severity=error` clean; `bash -n` and `zsh -n` clean.
- **On msi:** `dotf doctor` from this branch reports `[Deploy-dir↔$HOME drift] (11 checks, all ok)` — including `.gitconfig`, which genuinely differs and passes via its exemption.

Production diff +198/−96 (net +102), under the 300-LOC cap. Shell layer net **−66**.

### DIFF
--- a/cli/internal/doctor/checks_compose.go
+++ b/cli/internal/doctor/checks_compose.go
@@ -1,0 +1,42 @@
+package doctor
+
+import "strings"
+
+// checkDockerCompose covers the one name in setup-linux.sh's check_dependencies
+// list that no doctor section, contract binary or package list mentioned:
+// docker-compose. OPS-043 deletes that shell call, so without this the tool
+// would lose its only mention.
+//
+// It probes the v2 CLI plugin first because that is the supported form —
+// compose v2 is `docker compose`, a plugin with no entry on PATH, so
+// `command -v docker-compose` (what the shell call did) reports "missing" on a
+// current install where compose works fine. Measured on msi 2026-09-02: the
+// standalone v1 binary and plugin v2.39.1 are both present, and either check
+// alone would have described that box wrongly.
+//
+// Absence is a SKIP, never a FAIL: the repo provisions compose in no installer
+// block, no versions.conf pin and no contract binary, so failing on it would red
+// a box that never asked for it — the same reasoning BUG-052 applied to
+// terraform.
+func checkDockerCompose(sys *System, rep *Report) {
+	rep.Section("Docker Compose")
+
+	if !sys.has("docker") {
+		rep.Skip("docker not on PATH — nothing for the compose plugin to attach to (see Core tools)")
+		return
+	}
+
+	if out, err := sys.CommandOutput("docker", "compose", "version"); err == nil {
+		if v := strings.TrimSpace(out); v != "" {
+			rep.Pass("compose v2 plugin: " + v)
+			return
+		}
+	}
+
+	if sys.has("docker-compose") {
+		rep.Pass("docker-compose found (legacy standalone v1; `docker compose` is the supported form)")
+		return
+	}
+
+	rep.Skip("compose not installed (optional — the repo provisions neither the plugin nor the binary)")
+}
--- a/cli/internal/doctor/checks_compose_test.go
+++ b/cli/internal/doctor/checks_compose_test.go
@@ -1,0 +1,65 @@
+package doctor
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+)
+
+// TestCheckDockerCompose pins the one tool setup's check_dependencies named that
+// doctor covered nowhere. Compose v2 ships as a `docker` CLI plugin, not a
+// binary, so a PATH test alone answers wrongly on a current install — measured
+// on msi 2026-09-02, where the v1 binary and plugin v2.39.1 are both present.
+// The repo provisions compose in no installer, versions.conf entry or contract
+// binary, so absence is a SKIP; a FAIL would red every box that never wanted it
+// (the BUG-052 reasoning that put terraform in optionalTools).
+func TestCheckDockerCompose(t *testing.T) {
+	cases := []struct {
+		name         string
+		onPath       []string
+		cmdOut       map[string]string
+		wantFailures int
+		wantSubstr   string
+	}{
+		{
+			name:       "v2 plugin present → pass naming the version",
+			onPath:     []string{"docker"},
+			cmdOut:     map[string]string{"docker compose version": "Docker Compose version v2.39.1"},
+			wantSubstr: "v2.39.1",
+		},
+		{
+			// A box still on the standalone binary: compose works, so this is
+			// not a failure, but the plugin is the supported form.
+			name:       "only the v1 binary → pass, flagged legacy",
+			onPath:     []string{"docker", "docker-compose"},
+			wantSubstr: "legacy",
+		},
+		{
+			name:       "docker present, no compose either way → skip",
+			onPath:     []string{"docker"},
+			wantSubstr: "not installed",
+		},
+		{
+			// Without docker there is nothing for a plugin to hang off; the
+			// core-tools section already reports docker itself.
+			name:       "no docker → skip without duplicating the core-tools verdict",
+			wantSubstr: "docker",
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			sys := newSys(nil, tc.onPath, tc.cmdOut)
+			var buf bytes.Buffer
+			rep := capture(&buf)
+			checkDockerCompose(sys, rep)
+
+			if rep.Failures() != tc.wantFailures {
+				t.Fatalf("failures = %d, want %d\n%s", rep.Failures(), tc.wantFailures, buf.String())
+			}
+			if !strings.Contains(buf.String(), tc.wantSubstr) {
+				t.Fatalf("output missing %q\n%s", tc.wantSubstr, buf.String())
+			}
+		})
+	}
+}
--- a/cli/internal/doctor/checks_home_deploy.go
+++ b/cli/internal/doctor/checks_home_deploy.go
@@ -1,0 +1,124 @@
+package doctor
+
+import (
+	"path/filepath"
+	"regexp"
+	"strings"
+)
+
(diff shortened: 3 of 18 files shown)


mlorentedev:
## Review triage

**Checks:** all green on `57d94af`, including `test (windows-latest)` and `test-windows`. `spec-gate` failed on the first head and was fixed, not waived: the PR body carried `Closes #1337`, and the gate correctly refuses a closing keyword on a spec that is not archived in the same PR. Changed to `Refs #1337`; the archive is a separate PR after an independent adversarial review, the same shape as OPS-040 (#1433 → review → #1440 closed #1333). The `cancelled` `spec-gate` row on the earlier SHA is the superseded run, neither pass nor fail.

**Reviewers: two, not one.** CodeRabbit posted **both** shapes — a *"Reviews paused"* notice (auto-pause after N reviewed commits, a configured behaviour, not a quota wall) **and** a real review with four findings. PR-Agent also reviewed, posting as `github-actions[bot]` (it runs inside Actions, so it does not appear under a `pr-agent` login) and updating in place through `57d94af`.

> **Correction to the first version of this comment**, which said PR-Agent did not post. It did. I filtered reviewer output by login and PR-Agent's login is the generic Actions bot, so a review that was present read as absent — the same "created_at never moves, so the queue looks clear" shape recorded on #906, reached by a different route. Recording the error rather than silently editing it, since the whole point of this heading is that the disposition be checkable.

All four findings were verified against the code before applying. All four were real. Applied in `563ede1`:

| # | Finding | Verdict | Disposition |
|---|---|---|---|
| 1 | AC2's `features.json` command names `TestCheckHomeDeployDriftExemptions`, which does not exist | **CONFIRMED** — measured: `go test -run <no-match>` exits **0** | **Applied.** Names the two tests that exist; every `-run` pattern anchored |
| 2 | `scripts/utils.sh` comment claims the replacement "covers Windows" | **CONFIRMED** — the section skips Windows | **Applied.** Corrected, and names #1447 |
| 3 | Ordering guard samples one variable on Linux, matches any mention on Windows | **CONFIRMED** — reproduced the reviewer's exact failing input | **Applied.** All six on both scripts, assignment-matched, mutation-tested |
| 4 | `checkHomeDeployDrift` is 48 lines, over the repo's 40-line guideline | **CONFIRMED** — measured | **Applied.** Split; both functions now 22 lines |

**Finding 1 is the one worth naming.** It is a verification command that could not fail, sitting inside the spec whose own `verification.md` argues a parity claim must be mutation-tested. The mutation testing had been applied to the assertions and not to the commands that invoke them. Fixed, and both mutation tests are now recorded in `verification.md` with their exact inputs.

**Finding 2 is this PR's own lesson recurring inside it.** Lesson 259, added by this PR, is that a justification outlives its mechanism as easily as a name outlives its contract. The comment I wrote claiming Windows coverage was written the same day as the lesson, and was wrong in exactly that way.

**Pre-merge check — Linked Issues (⚠️ warning): declined as stated, addressed at the source.** It flagged that keeping the six pre-exported path vars contradicts #1337, which asks for their deletion. The retention is deliberate: they are BUG-021's fix, not duplication — the setup shell has not re-sourced the rc files, so removing them reopens a closed bug that looks green in review and only surfaces on the next clean machine. The reviewer's alternative resolution was the correct one and has been taken: **#1337 has been updated** with the five measured false premises, so the issue no longer asks for a change that would remove coverage.

### PR-Agent — 3 observations, 0 code changes

| Observation | Verdict | Disposition |
|---|---|---|
| Ticket compliance "partially compliant": the six pre-exports were not deleted; `scripts/utils.ps1` kept its functions | **Real, and already addressed at the source.** Same point as CodeRabbit's Linked Issues check | **Applied where it belonged:** #1337 updated with the five measured false premises. On the PowerShell half the observation answers itself — it notes the file *had no `check_deployed` twin*, so there was nothing to delete; `Test-FileDrift`, the one dead helper, was removed |
| "Hardcoded Map": `homeDeployMap` is manually synchronised with `setup-linux.sh` | **Accurate, and it names its own mitigation** — "the risk is mitigated by the guard", "a maintenance burden, not a current bug" | **Declined with reason.** The join guard is the design answer, and it runs in CI on both runners, not only locally. The alternative — deriving the map by parsing a shell script at doctor runtime — trades a compile-time list plus a test for a runtime parser, which is worse |
| "Stale Comment": the comment cites `TestSetupShellParity` but the real test is `TestSetupShellParity` | **Null finding** — the two names in the observation are identical, and the cited name is correct | **Declined.** Nothing to change |

Consistent with what this repo has measured about PR-Agent: its code findings are real, and its ticket-scope inference needs checking against the ticket rather than trusted. Here it flagged the pre-exports as non-compliance and then said, correctly, that a human must confirm whether the ticket should be updated instead. That is what happened.

**Re-review status after the fixes:** PR-Agent's review is current through `57d94af`. CodeRabbit is auto-paused, so its four findings were raised against `5ae8e74` and it has not re-reviewed `563ede1`/`57d94af`. Disclosed rather than implied; `@coderabbitai review` would close that gap before merge if the owner wants it.

Issue #1448, opened 2026-09-03
GUARD-009: three subsystems trust a claim after its referent was deleted — detect the dangling referent, because the written guard already failed


mlorentedev:
## What

A written claim is trusted after the thing it describes has been removed. Three instances landed
in the last eight weeks, in three different subsystems, and in every case the claim was **true when
written** — nobody introduced a false statement. The tree moved underneath a true one.

This is why the remedy cannot be "detect wrong comments". The buildable version is narrower:
**detect a claim whose referent no longer exists in the tree.**

## The three instances

| # | Claim | Referent that no longer exists | Found |
|---|---|---|---|
| 1 | `harness/skills/new-ticket/SKILL.md:211-217` — a "Concurrency guard" section instructing every agent to re-scan before `gh issue create` | Nothing removed it; it was never mechanised. 21 duplicated IDs since it was written on **2026-07-07** | 2026-09-02 |
| 2 | `setup-linux.sh:1586-1589` — a NOTE justifying a deploy exemption with *"setup may modify `.zshrc` via `sed -i`"* | **`sed -i` is gone.** OPS-040's purge (#1433) removed the mechanism and left the justification; the only match left in the file is inside that comment | 2026-09-02 |
| 3 | #1337's body — asserted `check_deployed` duplicates `dotf doctor` | The duplication. `check_deployed` covers deploy-dir→`$HOME`; `checkDeployDrift` covers repo→deploy-dir. Different legs. Deleting as asked opens a coverage hole | 2026-09-02 |

Instance 2 is the one that settles the remedy: **OPS-040's whole method was "probe the target, don't
read the block's own comment" (lesson 256), and OPS-040 produced instance 2.** A defect class that
survives its own authors knowing it in writing is not fixable by writing it down again. Instance 1
is the same argument from the other end — an instruction written *specifically* to prevent this,
with a measured 0% success rate over two months.

## Counts, and the population each was measured over

Stated explicitly because the number moved during triage and a bare count is not reviewable.

- **21 duplicated `AREA-NNN` ids** — population: `gh issue list --repo mlorentedev/dotfiles --state all`,
  issues only, PRs excluded, 596 titles, measured 2026-09-02. Spans 2026-07-02 → 2026-09-02.
- An earlier count of **9** in the same session was measured over **open issues only** and is superseded.
- Counting over the REST `/issues` endpoint instead inflates the result — it returns pull requests too
[...]

PR #1455, opened 2026-09-03
feat(harness): fire the persona suggester on every prompt, failing open by construction

### LINKED ISSUE #1436: HARNESS-110: the orchestrator decides and records but nothing triggers it — derive the role from the skills join and bind a suggester
## What

The orchestrator can now *decide* (`dotf harness gate`) and *record* (#1435), but nothing
**triggers** it. `harness/triggers.json` holds 18 rules and `dotf harness suggest` reads them —
wired to **no hook**, so it has never fired outside a manual invocation. A session gets no
suggestion of which persona should be doing the work in front of it.

## The design decision, already measured — do not re-derive

The open fork was *"does the role live as a new field in `triggers.json`, or in a separate
`harness/routing.json`?"*. **A measurement closed it: neither.** The role is already derivable
transitively, `trigger.skills ∩ persona.skills`, from two SSOTs that exist:

```
skills named by triggers.json : 30
skills named by persona records : 34
INTERSECTION : 28
16 of 18 rules resolve to at least one persona
```

Resolving today, with no new data: `architecture-and-adr → architect`,
`testing-standards → builder`, `security-audit → reviewer`,
`task-and-ticket-tracking → planner`, `vault-and-knowledge-management → curator`,
`container-workflow`/`git`/`iac`/`helm` `→ shipper`.

The two that resolve to nobody — `shell-standards`, `powershell-ascii-only` — declare **no
skills at all**. They are pattern-only rules with no owner, which is a correct answer rather
than a gap.

**Why derive rather than declare:** a `role` field would be a *second* place stating the
role↔skill relationship that the vault persona records already own. Two places drift, and the
drift would be silent. Deriving means adding a skill to a persona updates the routing for free,
with no new file, no new schema and no sync guard to maintain.

**Ambiguity is a first-class output, not a defect.** `code-complexity-and-refactor` resolves to
**both** `builder` and `reviewer`, because `cyclomatic-complexity` is in both rosters. Do not
rank and do not tie-break: in advisory mode the honest output is *"this is builder-or-reviewer
work"* and the session's own reasoning resolves it. A ranking heuristic would invent certainty
the data does not hold. Determinism comes from **sorted output** — same input, same bytes.

## Acceptance criteria

- **AC1** `harness.ResolveRoles(suggestion, personas) []string` — a pure, sorted join. No I/O.
- **AC2** Ambiguity is returned in full, never narrowed. `code-complexity-and-refactor` returns
  `[builder, reviewer]`; a rule with no skills returns empty, and empty is not an error.
- **AC3** **A drift guard on the join itself.** The whole design rests on the 28/30 overlap, and
  nothing today spans `triggers.json` and the persona roster — the same *"no single check spans
  two files"* class as HARNESS-106's AC3. Red when a renamed skill or a persona losing its
  `skills:` drops the resolving-rule count.
- **AC4** A `UserPromptSubmit` hook, bound from `harness/manifest.json` so a deploy propagates it
  (never hand-written into `settings.json`).
- **AC5** **The hook reads the payload from stdin, never from a `--prompt` argument.** A
  shell-quoted user prompt on a command line is an injection surface.
- **AC6** **The field carrying the prompt text is MEASURED, not assumed.** The published hooks
  reference documents the common fields (`session_id`, `prompt_id`, `transcript_path`, `cwd`,
  `permission_mode`, `effort`, `hook_event_name`) but **does not name the prompt field**. Accept
  the plausible spellings, and *record which one arrived* — the same defensive shape as
  `OutcomePayloadUnrecognised`. Assuming a field name is exactly how #1434 happened.
- **AC7** **The hook can never exit non-zero.** Documented verbatim: on `UserPromptSubmit`, exit
  code 2 *"Blocks prompt processing and erases the prompt"*. This is a **data-loss** failure
  mode strictly worse than the gate's, whose worst case is a refused tool call. Fail-open is not
  a preference here, it is the only safe behaviour: a malfunctioning suggester must never be
  able to destroy what the user typed.
- **AC8** A stated latency budget (<20 ms) with a test asserting it. 18 regexes fire on **every**
  prompt; without a budget a latency regression is invisible.

Output on exit 0 is plain stdout, which the reference states verbatim is added as context:
*"The exceptions are `UserPromptSubmit`, `UserPromptExpansion`, `SessionStart`, and
`PostModelSwitch`, where Claude Code adds plain-text stdout as context that Claude can see and
act on."* An `additionalContext` envelope also exists; plain stdout is fewer moving parts for v1.

## Explicitly out of scope

- **Autonomous dispatch.** v1 *suggests*; it does not launch. Launching is gated on #1434 —
  while a named dispatch silently disables its own gate, an auto-launcher would multiply
  ungated agents rather than orchestrate them.
- Phase detection from `Skill` events — a second spec once this one is live.
- Any change to #1434 itself.

Roughly ~150 LOC. Depends on nothing in flight; branches off `main`.

mlorentedev:
Closes #1436. Spec: `specs/HARNESS-110/`.

## What

The orchestrator could already decide (`dotf harness gate`) and record every decision it takes
(#1435), but **nothing triggered it**. `harness/triggers.json` held 18 rules and `dotf harness
suggest` read them, wired to no hook — so it had never fired outside a manual invocation. This is the
first inch of that layer.

A `UserPromptSubmit` hook now reads the payload from stdin, matches the trigger rules, and prints the
persona the matched skills imply, as plain stdout the session sees as context:

```
[persona] builder  ← pattern: pattern-testing-standards
  skills: test, test-driven-development
  → consider adopting `builder` and invoking test-driven-development

[persona] builder | reviewer  ← pattern: pattern-language-standards
  skills: cyclomatic-complexity
  → each of these declares the matched skill; the work decides which
```

Advisory only: it suggests, it does not dispatch. Autonomous launching is gated on #1434 — while a
named dispatch silently disables its own gate, an auto-launcher would multiply ungated agents rather
than orchestrate them.

## The design, and why the role is not declared anywhere

The open question was whether the role lives as a new field in `triggers.json` or in a separate
`routing.json`. **Neither.** It is derivable from two SSOTs that already exist, and the measurement
was re-run against `745f320` before any code was written rather than trusted from the ticket:

| | measured |
|---|---|
| skills named by persona records | 34 |
| skills named by `triggers.json` | 30 |
| intersection | 28 |
| rules resolving to ≥1 persona | **16 / 18** |

A `role:` field would restate what the persona records already own, and two places drift silently.
Deriving means adding a skill to a persona updates routing for free.

The two rules resolving to nobody — `shell-standards`, `powershell-ascii-only` — declare no skills at
all. Pattern-only rules with no owner is a correct answer, not a gap.

**Ambiguity is returned in full and never ranked.** Two rules resolve to two personas each; narrowing
would invent certainty the data does not hold. Determinism comes from sorted output.

## The safety property this is built around

**On `UserPromptSubmit`, exit code 2 is documented verbatim as *"Blocks prompt processing and erases
the prompt"*.** That is a data-loss mode strictly worse than the gate's, whose worst case is a
refused tool call. `runSuggestFromHook` therefore returns `nil` on every path — a missing harness
root (any repo that is not this one), an unparseable `triggers.json`, an unreadable persona record, a
payload shape nobody has seen. Nine rows in `TestSuggestFromHookNeverExitsNonZero`, asserted rather
than inspected.

Two related choices:

- **The payload arrives on stdin, never as an argument.** A shell-quoted user prompt on a command
  line is an injection surface, so no `--prompt` is consulted in this mode — asserted by test.
- **The prompt field name is undocumented.** The published hooks reference names `session_id`,
  `prompt_id`, `transcript_path`, `cwd`, `permission_mode`, `effort` and `hook_event_name` — and not
  the field carrying the prompt. Eight plausible spellings are accepted and **the one that arrived is
  reported every time**, so the guess becomes a measurement the next session can read. Assuming a
  payload shape is exactly how #1434 happened.

## Corrections made during implementation

Recorded in `verification.md` rather than fixed silently — each was a claim that did not survive
being probed:

1. **AC4's parenthetical was wrong.** It claimed `manifest.json` had no hook-emission key and this
   would be the first. `emit_hooks` already exists under `agents.bind[]` and carries the gate plus
   both `mem` hooks. This is a fourth entry on an existing mechanism.
2. **The output labelled the pattern as a "rule".** The first end-to-end run printed `← rule:
   pattern-testing-standards`. `Suggestion` carries pattern names, not trigger-rule ids. Relabelled
   `pattern:`. Caught by running the binary, not by reading the code.
3. **A test fixture asserted a claim the design never made.** It asserted the skill `spec` resolves
   to `[planner, reviewer]`; it resolves to `[planner]` alone. Ambiguity lives in the RULE, whose
   skills are `[spec, adversarial-review]`. The code was right and the test was wrong.

Also worth a reviewer's eye: **AC3's guard asserts a floor, not an equality.** An exact count goes
red when coverage *rises*, and the only available fix is bumping the constant — a guard edited every
time it fires is a guard the next session disables.

`ResolveRoles` filters on `kind: invocable`. `hermes-nan` is `kind: autonomous` and cannot be adopted
by a session; its only skill is absent from `triggers.json` today, which is why the filter lands now
rather than on the day someone adds it to a rule.

## Verification

```
go test ./...              20 packages, ok
go vet ./...               OK
GOOS=windows go vet ./...  OK
golangci-lint run          0 issues (v2.12.2, the versions.conf pin)
bats (harness + compile)   89/89 ok
```

End-to-end with the built binary, not a test:

```
$ echo '{"hook_event_name":"UserPromptSubmit","prompt":"add tests for this module using TDD"}' \
    | dotf harness suggest --from-hook --repo-root <checkout>
[suggest] prompt arrived as "prompt"
[persona] builder  ← pattern: pattern-testing-standards
  ...
exit=0

$ echo '{"nonsense":' | dotf harness suggest --from-hook              # garbage
[suggest] payload unrecognised: no known prompt field (13 bytes)
exit=0

$ cd /tmp && DOTFILES_DIR=/nonexistent dotf harness suggest --from-hook   # non-dotfiles repo
[suggest] no personas at /nonexistent: ...no such file or directory
exit=0
```

`features.json` entries stay `pending` deliberately: the agent cannot write `passing` — only the
harness may, after running each `verification` command and capturing exit 0.

## Not in this PR

- Autonomous dispatch (gated on #1434).
- Phase detection from `Skill` events — a separate spec.
- Migrating the six persona records still on the legacy inline `skills:` form. `parseSkills` reads
  both and `UnmigratedSkills()` tracks them; that is bookkeeping.


mlorentedev:
## Review triage

|---|---|---|
| **`nan/deepseek-v4-flash`** (independent adversarial, pool draw, not the implementer) | **PASS** — rubric Correctness B, Verification B, Scope A, Reliability A, Maintainability A, Handoff-readiness B. No Blocker, no Major. 4 Minor | 3 applied, 1 noted — table below |
| **PR-Agent** (`github-actions`) | "Fully compliant" on all 8 ACs; one finding: the `suggest-role` hook had no `timeout` | **Applied** in `ca572f9` (timeout 5, plus an assertion so the next entry cannot omit it) |
| **CodeRabbit** | *"Review rate limited"*, then *"Review skipped"* | **No review ran.** Disclosed, not implied — a notice about a review is not a review |

### Independent review findings

| # | Finding | Disposition |
|---|---|---|
| 1 | **REAL** — the `persona_record_unreadable` row never reached its branch: it shared `brokenRoot`, so `LoadTriggers` failed first and the row was green without testing anything | **Applied.** Its own root with valid triggers, plus a `wantStderr` column so every row must prove which branch it reached |
| 2 | THEORETICAL — the `kind: invocable` filter was asserted by nothing (mutation: deleting it left the suite green) | **Applied.** `TestResolveRolesExcludesNonInvocable`, mutation-checked — removing the filter turns it red |
| 3 | THEORETICAL — AC4 was verified by "the file contains a string", the form AC4 itself calls insufficient | **Applied.** `TestPromptHookReachesSettingsByConsequence` runs `LoadBindTargets → HookCommands → MergeHooks` and asserts the hook lands in the emitted settings |
| 4 | THEORETICAL — the latency test excludes the per-prompt `LoadTriggers`+`LoadPersonas` disk work | **Noted, not changed.** The reviewer measured the real path at avg 7.13ms / max 9.14ms against a 20ms budget; the load half is bounded by the 5s hook timeout |

The reviewer also confirmed against the published docs that the prompt field really is `prompt` (the
first of AC6's eight candidates) and that exit 2 does erase the prompt. Both were assumptions in the
spec and are now measurements.

### A fifth, surfaced while applying finding 1

Fixing the vacuous row exposed one the review did not reach, and it is a real defect rather than a
test artefact. **`LoadTriggers` walks up from the cwd when the given root has no `triggers.json`.**
The hook runs with cwd set to whatever repository the user is in, so the trigger rules could come
from one root while the personas came from another — the two halves of a join disagreeing about which
harness they belong to. `runSuggestFromHook` now reads the file at the resolved root, so the
suggester resolves exactly one root, the way the gate does.

### The first review attempt produced no review

`nan/qwen3.8-flash` was the random pool draw. It ran ~37 minutes and died on a provider-side
`400: "Input text data may contain inappropriate content"` with `willRetry: false`, settling without
writing `review.md`. That is a content filter false-positive on repository text, not a verdict.
Re-run deliberately on the pool's primary (`--reviewer nan/deepseek-v4-flash`), which is what the
`--reviewer` flag documents itself as being for. Recording it because an aborted review and a clean
review are indistinguishable from the outside if nobody says which happened.

### Verification after all dispositions

```
go test ./...              all packages ok
go vet ./... + GOOS=windows go vet ./...   OK
golangci-lint run          0 issues (v2.12.2, the versions.conf pin)
bats (harness + compile)   89/89 ok
pre-push SDD spec-gate     Passed
```

Spec archived to `specs/archive/HARNESS-110/` on the passing review, in this PR, per the Discipline
Gate.

### ⚠️ This PR object is stuck — the branch is fine, the PR is not

**Do not read the red `spec-gate` above as a real failure.** It ran at 02:39Z against
`1080ab3`, a `Merge branch 'main'` commit that is no longer in the branch's history, and it predates
the archive commit.

What happened, as measured:

- The remote branch carried two merge commits (`f3b3b60`, `1080ab3`) that this checkout never had —
  `f3b3b60` is the "bad object" a pre-push hook reported here earlier.
- A later push recreated the ref without them, leaving the PR head pinned to a commit that is neither
  an ancestor nor a descendant of the branch. GitHub does not re-sync such a head.
- `1080ab3` was merged back in (`ce101bd`) to restore continuity rather than forcing over anything.
  The branch ref on GitHub is now `ce101bd`; `gh api .../pulls/1455` still reports
  `head=1080ab3`, `commits=4`, `mergeable=null`, `mergeable_state=unknown`, and no new workflow runs
  fire on pushes.

The branch content is correct and verified: `specs/archive/HARNESS-110/` present, no active
`specs/HARNESS-110/`, all checks green locally including the pre-push spec-gate.

**Owner decision needed:** either wait for GitHub to unstick this PR, or close it and open a fresh
one from the same unchanged branch. Nothing in the code needs to change either way — this is a PR
object problem, not a content problem.

mlorentedev:
## Correction to the triage above — this PR was not stuck, it was MERGED

My "⚠️ This PR object is stuck" section is wrong and I am correcting it rather than leaving it to
mislead the next reader.

**#1455 merged at 03:13:46Z** (`merged=true`, merge commit `fb0b7e9`). It merged at `1080ab3` —
which is exactly why the head never moved, why `mergeable` read `null`, and why pushes stopped
triggering workflows. I read "head pinned to a commit not in the branch" as a GitHub sync bug; the
ordinary explanation was that the PR had already closed and my pushes were landing on a branch behind
a merged PR. The red `spec-gate` above is likewise not a stuck check — it is a real failure that was
real at the time, superseded by the merge rather than fixed.

**What it means, measured against `origin/main`:** the merge landed at the moment the reviewer's
findings were still being applied, so main has the suggester without the hook timeout, without the
one-root fix, without the three test-coverage fixes — and with `specs/HARNESS-110/` still ACTIVE for
the issue this PR closed, which is the state the spec gate exists to prevent.

**Follow-up: #1462**, off current `main`, carrying exactly those four things and the review artifacts.

The lesson is the one this PR's own ticket (#1448) is about: *the PR object is stuck* was a claim I
inferred from symptoms and never probed. One `--jq '.merged'` would have answered it, and I asked
after three failed pushes instead of before the first.

Issue #1454, opened 2026-09-03
CLI-071: dotf pr triage-queue dies with GraphQL, so the command that binds Definition of Done §4 is unavailable exactly when gh is degraded


mlorentedev:
## What happened

Measured 2026-09-03 while triaging #1450:

```
$ dotf pr triage-queue
pr triage-queue: gh pr list: exit status 1
exit=1

$ gh pr list --state open --limit 5
GraphQL: API rate limit already exceeded for user ID 13562150.
```

`cli/internal/prtriage/fetch.go:55` shells out to `gh pr list --state open --json …`, which is a **GraphQL** call. The rest of the triage logic then reads comments, which are REST. So the whole command is hostage to the GraphQL bucket even though the data it actually needs is available.

## Why this is worse than an ordinary outage

`dotf pr triage-queue` is the mechanism named in the always-injected pr-stewardship doctrine: *"run it at session start, and again before reporting any PR work complete"*, and *"a non-zero exit is a queue you must read, never an empty one — it exits non-zero both when work is pending AND when the question could not be answered."*

That disambiguation is doctrine, not code: **the exit status is 1 in both cases and the caller has to read the message to tell them apart.** An agent that treats non-zero as "queue pending" will churn; one that treats it as an error will skip the check. Today the failure mode is loud, but it fails *precisely* in the sessions where `gh` is degraded — which is when a missed triage is most likely.

Worth noting the same session hit GraphQL refusals on `gh pr create`, `gh pr checks` and `gh pr comment`, all of which have working REST equivalents. A peer session hit it on `gh issue edit`. `gh api rate_limit` reported **core 5000/5000 and graphql 5000/5000, 0 used**, so the limiter is a secondary/content-creation limit that `rate_limit` does not expose — checking it before retrying is actively misleading.

## Proposed fix

Replace the `gh pr list` call in `fetch.go` with the REST endpoint, which returns everything the queue needs:

```
gh api "repos/{owner}/{repo}/pulls?state=open&per_page=100"
```

Field mapping is direct (`number`, `title`, `head.ref`, `draft`, `user.login`). The existing boundary-truncation guard at `fetch.go:76` still applies — REST paginates, so either page through or keep the explicit refusal when a full page comes back.

Manually answering the same question via REST during the #1450 triage took one loop and worked while `gh pr list` was refused, so the endpoint is confirmed usable under the same conditions.

## Optional second half, worth deciding separately

[...]

PR #1419, opened 2026-09-02
chore(main): release 0.54.0

mlorentedev:
:robot: I have created a release *beep* *boop*
---

### Features

* **cli:** dotf vault maintain, composing crystallize and health in-process ([#1489](https://github.com/mlorentedev/dotfiles/issues/1489)) ([52035f1](https://github.com/mlorentedev/dotfiles/commit/52035f1a174e82f3176bc458aa128ebc77909f9e)), closes [#490](https://github.com/mlorentedev/dotfiles/issues/490)
* **cli:** dotf worktree lifecycle and fail-closed garbage collection ([#1500](https://github.com/mlorentedev/dotfiles/issues/1500)) ([#1515](https://github.com/mlorentedev/dotfiles/issues/1515)) ([a76ea18](https://github.com/mlorentedev/dotfiles/commit/a76ea184ea20f8f54903df15c7d558074f22e241))
* **doctor:** print a Next steps block for FAILs that carry a remedy command ([#1443](https://github.com/mlorentedev/dotfiles/issues/1443)) ([1966d6d](https://github.com/mlorentedev/dotfiles/commit/1966d6da24b90e7d4517abfbe3834f83a01d9bf7))
* **doctor:** the binary carries the commit it was built from, so a stale dotf stops reporting health it never established ([#1416](https://github.com/mlorentedev/dotfiles/issues/1416)) ([3a11d97](https://github.com/mlorentedev/dotfiles/commit/3a11d977d6c3de7cbb3f57a8063ec8c6a65d7695)), closes [#1158](https://github.com/mlorentedev/dotfiles/issues/1158)
* **harness:** architect becomes the fourth enforceable persona, and a typo in any skill id now fails ([#1509](https://github.com/mlorentedev/dotfiles/issues/1509)) ([4e2e2c4](https://github.com/mlorentedev/dotfiles/commit/4e2e2c4d2551163a9ff633bfaac91b2c94331810))
* **harness:** builder becomes the second enforceable persona, gating two of nine ([#1494](https://github.com/mlorentedev/dotfiles/issues/1494)) ([ca81157](https://github.com/mlorentedev/dotfiles/commit/ca8115734e08677f72f19f649ff2259587052d86)), closes [#1420](https://github.com/mlorentedev/dotfiles/issues/1420)
[...]

PR #1457, opened 2026-09-03
fix(prtriage): give the fetch path a seam, then move it off GraphQL

mlorentedev:
Spec: `specs/CLI-071-triage-queue-transport/`

## What this is

`dotf pr triage-queue` failed on 2026-09-02 with `gh pr list: exit status 1`
while GitHub's GraphQL bucket refused every list and write and REST served the
same data throughout. #1454 was filed against the transport. Measuring before
writing found the transport was the smaller half, and the ticket wrong in two
places — [its body has been corrected in place](https://github.com/mlorentedev/dotfiles/issues/1454).

**The actual defect: `FetchWithRegistry` called `exec.CommandContext` directly.**
No injection point, therefore no test — the package's four tests all covered pure
functions. The one path observed failing in production was the only path in the
package that nothing exercised, and that was true whichever API it spoke. The
seam is the deliverable here; the REST migration is its implementation.

**The ticket's proposed fix could not work as written.** It called for a one-line
swap to `GET /repos/{o}/{r}/pulls` with a field mapping naming `number`, `title`,
`head.ref`, `draft` — and never mentioning comments. But the comments *are* the
algorithm (`prtriage.go:95-141` compares reviewer timestamps against the triage
marker), and that endpoint returns none. `gh pr list --json comments` was a single
GraphQL query returning pull requests with comments **nested**; REST has no such
nesting, so the join moves into the package as a bounded fan-out.

**"Unavailable" was overstated.** Both consumers already fail loud — the command
exits non-zero with the reason, and the session brief injects `[pr-triage] queue
could not be computed: …` followed by *"This is not an empty queue"*. What was
missing was a workaround once the signal fired, not the signal.

## Three consequences, each now pinned by a test

- **`normaliseLogin` went from defensive to load-bearing.** The registry declares
  `github-actions`; REST returns `github-actions[bot]`. Removing the `[bot]` fold
  does not error — it silently empties the queue, the exact failure this package
  exists to prevent (#1033).
- **REST paginates on two axes where GraphQL paginated on none.** A truncated
  pull-request list makes a PR go *missing*, which is visible; truncated comments
  make the verdict *wrong* — an untriaged PR can read as triaged. Both refuse.
- **1+N had to meet the five-second session-start budget** (`cmd/mem.go`), which
  runs on every session start. Fan-out capped at 8; the concurrency test asserts
  *observed peak* rather than wall-clock, so it stays deterministic on a loaded
  runner instead of being loosened until it means nothing.

## Verification

```
go build ./...            → clean          golangci-lint run   → 0 issues (v2.12.2, the pin)
go vet ./...              → clean          gofmt -l internal/prtriage/ → clean
GOOS=windows go vet ./... → clean          go test ./... -count=1 → all ok
```

**Production smoke test**, the strongest evidence here — the binary run against
this repository, compared against what the GraphQL implementation printed for the
same repository earlier the same session:

```
1 pull request(s) awaiting a disposition:

  #1455  feat(harness): fire the persona suggester on every prompt, failing open by construction
         github-actions reviewed, never triaged, 2026-09-02 20:27
         https://github.com/mlorentedev/dotfiles/pull/1455
exit=1
```

Byte-identical: same PR, same reviewer, same reason, same timestamp, same URL,
same exit status. **Latency measured, not assumed:** 0.94 / 0.96 / 0.95 s across
three runs against the 5 s budget.

**Mutation testing — all eight caught:** dropping the `[bot]` fold; serialising
the fan-out; removing either page guard; mapping `url` instead of `html_url`;
swallowing a per-PR comment error; reintroducing `gh pr list`; keeping the
GraphQL `createdAt` spelling.

**The `features.json` commands were themselves checked for vacuity**, since the
previous spec shipped one naming a test that did not exist and it passed anyway:
`go test -run '^TestThisDoesNotExist$'` exits **0**, while the form used here
(`-v` piped through a `--- PASS:` grep) exits **1**.

## Removed or edited existing tests, with cause

- `TestParseWireBoundaryLimit` tested the truncation guard through `parseWire`,
  the helper that unpacked the GraphQL nesting. Both are gone. Its replacement
  drives the real fetch path and covers the second axis. A comment at the old
  site records where the coverage went.
- `session_start_triage_test.go` carried the fixture string
  `"gh pr list: exit status 4"`, which this change made unproducible — lesson 259
  would have been committed again in the PR that cites it.

## Out of scope, deliberately

Distinct exit statuses for "pending" vs "could not answer". `pr.go:49-51`
documents the shared status as a *reasoned* choice, not an oversight; refuting
that argument is its own ticket, and folding it in here would smuggle a
public-contract change inside a transport fix.

**Two other `gh pr list` sites were found and left alone**, recorded so the skip
is stated rather than silent: `scripts/bitacora-rollout.sh:132,196` and
`harness/skills/catchup/SKILL.md:72`. Same GraphQL dependency; neither binds a
doctrine, neither has been observed failing, and a vendor condition is not a
defect in them. Not ticketed.

Also pre-existing and untouched: `gh api pulls: exit status 1` still hides gh's
own stderr reason. Unwrapping `*exec.ExitError` belongs to whoever owns the error
contract, not to a transport change.

`go test -race -count=3 ./internal/prtriage/` → clean, run explicitly because
this adds the package's only concurrency and golangci-lint does not cover it.

Lesson: `docs/lessons/lesson-260-the-path-with-no-seam-is-the-path-with-no-test.md`

Refs #1454

### DIFF
--- a/cli/internal/mem/session_start_triage_test.go
+++ b/cli/internal/mem/session_start_triage_test.go
@@ -42,7 +42,7 @@ func TestTriageQueueSection(t *testing.T) {
 	// If this section swallowed the error, the brief would restore exactly the
 	// blind spot the exit contract was written to close.
 	t.Run("a failure is reported, never rendered as empty", func(t *testing.T) {
-		got := triageQueue("", errors.New("gh pr list: exit status 4"))
+		got := triageQueue("", errors.New("gh api pulls: exit status 4"))
 		if got == "" {
 			t.Fatal("a queue that could not be computed must not be silent")
 		}
--- a/cli/internal/prtriage/fetch.go
+++ b/cli/internal/prtriage/fetch.go
@@ -5,23 +5,59 @@ import (
 	"encoding/json"
 	"fmt"
 	"os/exec"
+	"sync"
 	"time"
 )
 
-// ghPR mirrors what `gh pr list --json` returns. It is deliberately separate
-// from PR: the vendor's shape is a boundary concern, and pinning the domain to
-// it would make every test carry gh's nesting for no benefit.
-type ghPR struct {
-	Number   int    `json:"number"`
-	Title    string `json:"title"`
-	URL      string `json:"url"`
-	Comments []struct {
-		Author struct {
-			Login string `json:"login"`
-		} `json:"author"`
-		Body      string    `json:"body"`
-		CreatedAt time.Time `json:"createdAt"`
-	} `json:"comments"`
+// ghRunner runs a `gh` subcommand and returns its stdout.
+//
+// It is the seam CLI-071 introduced, and introducing it was the point of that
+// change rather than a means to it. Before it, FetchWithRegistry called
+// exec.CommandContext directly, so the fetch path could not be reached from a
+// test — and the path that had actually been observed failing in production
+// (#1454) was therefore the only one in this package with no coverage.
+type ghRunner func(ctx context.Context, args ...string) ([]byte, error)
+
+func execGH(ctx context.Context, args ...string) ([]byte, error) {
+	return exec.CommandContext(ctx, "gh", args...).Output()
+}
+
+const (
+	// prLimit and commentLimit are page sizes AND boundary alarms. REST
+	// paginates where the old GraphQL query did not, so a full page means
+	// "there may be more" and this package refuses rather than answering from
+	// part of the data.
+	prLimit      = 100
+	commentLimit = 100
+
+	// commentFanout bounds the per-pull-request calls. The session-start probe
+	// runs on every session start under a five-second budget (see the caller in
+	// cmd/mem.go), and REST needs one call per pull request where GraphQL
+	// needed none — so these must overlap. Unbounded would be its own failure
+	// mode: a burst against a rate limit is what this change is escaping.
+	commentFanout = 8
+)
+
+// restPR is one item of `GET /repos/{owner}/{repo}/pulls`.
+//
+// Note `html_url`: the GraphQL shape this replaced called the same field `url`,
+// and a silent zero value there would have produced a queue whose entries point
+// nowhere rather than an error.
+type restPR struct {
+	Number  int    `json:"number"`
+	Title   string `json:"title"`
+	HTMLURL string `json:"html_url"`
+}
+
+// restComment is one item of `GET /repos/{owner}/{repo}/issues/{n}/comments`.
+// Pull request conversation comments are issue comments; this is the same set
+// the previous `gh pr list --json comments` returned.
+type restComment struct {
+	User struct {
+		Login string `json:"login"`
+	} `json:"user"`
+	Body      string    `json:"body"`
+	CreatedAt time.Time `json:"created_at"`
 }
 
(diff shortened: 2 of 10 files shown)


mlorentedev:
## Review triage

**Who actually reviewed this PR: PR-Agent only.**

CodeRabbit posted a *notice*, not a review — **"Review limit reached / You've used
all free OSS reviews for now"** (03:02, edited in place at 03:14). A notice that no
review ran leaves the PR unreviewed by that reviewer, and it is recorded here
rather than left to look like silence. If it publishes a real review later, this
table gets a second round; the queue will say so, because it compares content and
not just `created_at` — CodeRabbit edits in place, which is how #1419 sat green in
the queue while carrying an unread notice.

**CI on the PR head `e9ff5bb`:** 14 checks green, `test-windows` still running at
the time of writing. Note the `cancelled` runs on `047beb0` / `038a455` are
superseded runs that concurrency cancelled when a newer head arrived — not
failures. A `cancelled` conclusion is neither pass nor fail, so it was checked
rather than read either way.

### PR-Agent (`github-actions[bot]`, "PR Reviewer Guide")

No code findings: *no major issues detected*, *no security concerns identified*,
*PR contains tests*, effort 3/5. Both items it raised are ticket-compliance
inferences, and both are declined:

| # | Finding | Disposition |
|---|---|---|
| 1 | #1454 "partially compliant" — *the optional distinct exit statuses are not implemented* | **Declined.** Self-contradictory: the same item is listed under *compliant* as "deliberately out of scope". The ticket marks it optional and says it is "worth deciding separately"; `pr.go:49-51` documents the shared exit status as a **reasoned** choice, not an oversight. Refuting that argument is its own ticket. Folding it in here would smuggle a public-contract change inside a transport fix. |
| 2 | #1033 "partially compliant" — *the attestation gate is not modified to handle the two login spellings* | **Declined — the premise is wrong.** #1033 is *cited as a reference* in the proposal (it is where the `github-actions` vs `github-actions[bot]` trap was first recorded), not a ticket this PR claims to close. This PR touches `cli/internal/prtriage` only; `check-review-attestation.sh` is untouched by design. |

Finding 2 is the **sixth** observed instance of PR-Agent inferring ticket scope
from any issue number that appears in a PR body — on #1440 it treated **#1433, a
pull request**, as a ticket to satisfy. The code findings have been reliable every
time; the ticket-compliance section has not. That is reviewer configuration, and
it is already carried as such; nothing to change in this PR.

### Not raised by either reviewer, disclosed here

Three things were found by this session's own sweep and deliberately left out of
scope, so they are named rather than left silent: two other `gh pr list` call
sites (`scripts/bitacora-rollout.sh:132,196`, `harness/skills/catchup/SKILL.md:72`)
carrying the same GraphQL dependency, and `gh api pulls: exit status 1` still
hiding gh's own stderr reason. All three are in the PR body and `verification.md`
with reasons.

**Nothing to apply. No blocking findings. The independent adversarial review
required before this spec archives has not run yet** — it runs against `main`
after merge, and it, not this table, is the substantive review of this change.

PR #1456, opened 2026-09-03
chore(spec): archive OPS-043 after an independent review, and close the gap it found

mlorentedev:
Closes #1337. Archives `OPS-043-setup-doctor-duplication` after an independent adversarial review, and closes the one gap that review found.

The implementation landed in #1450 (`416233e`). This PR is the lifecycle close.

## The review

`nan/deepseek-v4-flash`, drawn at random from `harness/reviewer-pool.json` — not the implementer, which is the whole point of the gate. Verdict **PASS-WITH-GAPS**: no Blocker, no Major, rubric A/B/A/A/A/A. Full text in `specs/archive/OPS-043-setup-doctor-duplication/review.md`.

It did not rubber-stamp. It re-ran the build, the linters and the suite, reproduced both of my mutation tests, and ran `dotf doctor` end-to-end against a temp `$HOME` with a drifted file to confirm the new section actually FAILs and exits 1.

## What it found, and what changed

**Finding 1 — fixed here, not tracked.** Every test proved the new *functions*; none proved `doctor.Run` still *calls* them. The reviewer demonstrated it by deleting both `checkXxx` lines from `doctor.go` and watching `go test ./internal/doctor/` stay green.

That is the exact defect class this spec's own `verification.md` criticises — a check that cannot fail — so leaving it as a tracked gap would have been incoherent. `TestRun_HomeDeployDriftIsWiredIn` and `TestRun_DockerComposeIsWiredIn` now assert through `Run()` that the section appears, names the drifted file, **and moves the exit code**; a section printing a FAIL that did not affect the status would be equally useless. Mutation-tested with the reviewer's own edit: deleting the invocations turns them red.

**Finding 4 — a false claim in my own evidence, corrected.** `verification.md` said "1534 tests, exit 0". The suite is not reproducibly green on this box: 7 `not ok`. The reviewer isolated the cause properly — a temporary worktree at the merge-base `745f320`, where **the identical 7 fail** — so they are pre-existing and environmental (#1429 fixture leakage from a `dev` build on PATH; #1409), not this change's. The figure was true when measured and stopped being true when a later commit added a guard test: this spec's own lesson 259, landing in its own evidence section.

[...]

PR #1461, opened 2026-09-03
chore(spec): archive CLI-071 after two independent review rounds

### LINKED ISSUE #1454: CLI-071: dotf pr triage-queue dies with GraphQL, so the command that binds Definition of Done §4 is unavailable exactly when gh is degraded

mlorentedev:
Archives `specs/CLI-071-triage-queue-transport/` and closes #1454. The
implementation merged in #1457 (`afc9db6`).

## Two independent review rounds, both PASS

Both by `nan/deepseek-v4-flash`, drawn at random from `harness/reviewer-pool.json`,
neither the implementer.

| Round | `reviewed_sha` | Rubric | Findings |
|---|---|---|---|
| 1 | `afc9db6` | A/A/A/A/**B**/A | 6 Minor — **4 applied**, 2 declined with reasons |
| 2 | `f3c4b88` | **A/A/A/A/A/A** | 4 Minor — 3 carry-overs, 1 new, accepted-not-fixed |

Round 1's only rubric deduction was `fetchWith` at 53 lines, over the repo's
<40-line guideline. Extracting `withComments` brought it to 21 and the helper to
38, and Maintainability rose to A in round 2. Also applied: a remediation hint on
the comment-axis truncation error, which was asymmetric — the pull-request guard
told the reader how to recover (`--paginate`) and the comment guard did not.

Declined, with reasons in `verification.md`: a literal 5s-deadline test for AC5
(the reviewer itself calls it *"arguably over-engineering"*, and a wall-clock
assertion is the kind that gets loosened until it means nothing); and surfacing
gh's stderr on non-zero exit (pre-existing, disclosed before the review ran, and
an error-contract decision rather than a transport one).

Round 2's new finding is fair and is recorded rather than fixed: AC1 claims
`FetchWithRegistry` runs end-to-end, but the tests drive `fetchWith` — the
exported wrapper is one line of delegation, covered transitively and by the live
smoke test. Rewording AC1 would edit `proposal.md` and invalidate the review,
starting a third round.

## Why there were two rounds, and why that is the gate working

`dotf spec archive` refused twice: first because `tasks.md` was dirty, then
because it had changed after `reviewed_sha`. **Applying a review's findings
changes the tree the review describes** — including the box the reviewer itself
asked to tick.

Re-running was the honest resolution. `--force-without-review` was available and
would have concealed that the code changed after it was reviewed, which is exactly
what the gate exists to surface.

The loop converges because the gate watches `proposal.md`, `tasks.md` and
`features.json` but **not** `verification.md`: a non-gating finding can be
dispositioned without invalidating the attestation it responds to. A round that
ends with only non-gating Minors is the signal to archive, not to iterate again.

## Something the reviewer did that the author had not

It checked that the live `exit 0` was the *correct* answer rather than green
paint, by confirming #1419 carries a CodeRabbit review at `2026-09-02T02:09:11Z`
and a re-triage at `2026-09-03T02:33:33Z`, so `Evaluate` rightly returns
"triaged". Verifying a green result is right, not merely green, is the harder half.

## Promotions

- **Lesson:** `docs/lessons/lesson-260-the-path-with-no-seam-is-the-path-with-no-test.md`, landed in #1457.
- **ADR:** none — ADR-020 already governs the convergence; this establishes nothing new.
- **Pattern:** none yet. Revisit if "vendor CLI called directly, therefore untestable" turns up in a second project.

Closes #1454


mlorentedev:
## Review triage

**CI:** green on the PR head, including `spec-gate` and both Windows legs.

**Who actually reviewed: PR-Agent only.** CodeRabbit posted *"Review limit reached
— next included review available in 47 minutes"*. A notice is not a review, and
this is the third consecutive PR on which it has produced one (#1456, #1457,
#1461). Recorded rather than left to look like silence.

### PR-Agent (`github-actions[bot]`)

Effort 1/5, *no major issues detected*, *no security concerns identified*, *PR
contains tests*. **No code findings and nothing to apply.**

Its ticket-compliance section, however, contradicts itself in a way worth
recording: both blocks are headed **"🔶 Partially compliant"** while listing
**"(none)"** under non-compliant requirements and marking every requirement
compliant. The per-requirement rows are accurate; only the headline is wrong.

The second block also treats **#1457 — a pull request — as a ticket to satisfy**.
That is the seventh observed instance of ticket-scope inference from any issue
number appearing in a PR body (previously: #1440 treating #1433, a PR, the same
way; #1457 treating #1033, a cited reference, as a requirement; #1462 the same
headline defect). The code findings have been reliable throughout; this section
has not. Reviewer configuration, already carried as such — nothing to change here.

### Cross-session note

A peer session flagged this PR as untriaged rather than triaging it, on the
grounds that a disposition written by someone who did not read the review is
worse than an empty queue. That is the right call and worth stating.

It also passed on a finding from its own lane: `harness.LoadTriggers` walks up
from the cwd when the root it is given has no `triggers.json`, so a deployed
binary can pick up whichever repo the user happens to be standing in. **Checked
against this package: it does not have that defect.** Both call sites resolve the
registry at a fixed depth — `pr.go:85` joins `harness/review-attestation.json`
relative to the cwd, and `mem.go:199` joins it to an explicit cwd and treats a
failed `os.Stat` as "not applicable". Neither searches upward, so the queue either
finds the registry in the directory it was pointed at or says so.

**Nothing to apply. No blocking findings.** The substantive review of this change
is the pair of independent adversarial rounds recorded in
`specs/archive/CLI-071-triage-queue-transport/review.md`, not this table.

PR #1462, opened 2026-09-03
fix(harness): land what #1455 merged without, and archive the spec it closed

mlorentedev:
Completes HARNESS-110. Refs #1436 (already closed by #1455).

## Why this exists

**#1455 merged at 03:13:46Z while its reviewer findings were still being applied.** The independent
adversarial review had returned PASS with four Minor findings; three were being fixed when the merge
landed. Everything pushed after that went to a branch whose PR was already closed, so `main` carries
the suggester without them.

Measured against `origin/main` before writing this:

| | in main after #1455 |
|---|---|
| `roles.go`, `hookpayload.go`, `harness_suggest_hook.go` | present |
| `suggest-role` hook timeout | **NONE** |
| triggers pinned to the resolved root | **no** — still walks up from cwd |
| the three review coverage fixes | **absent** |
| `specs/HARNESS-110/` | **still ACTIVE**, for an issue #1455 closed |
| `specs/archive/HARNESS-110/` | absent |

That last row is the state the spec gate exists to prevent, sitting in `main`.

## What this lands

**1. A bounded timeout on the prompt hook.** It rides `emit_hooks` on `UserPromptSubmit`, which is
the interactive path: an unbounded run delays what the user typed, not a background tool call. AC8
budgets the work at 20ms, so 5s is 250x headroom and the point is only that the damage is bounded.
Asserted, so the next entry cannot omit it silently. (PR-Agent's finding on #1455.)

**2. The suggester resolves ONE root.** `runSuggestFromHook` called `harness.LoadTriggers`, which
walks up from the cwd when the given root has no `triggers.json`. The hook runs with cwd set to
whatever repository the user is in, so the trigger rules could come from one root while the personas
came from another — the two halves of a join disagreeing about which harness they belong to. It now
reads the file at the resolved root and parses it directly, matching `loadGatePersona`.

This was found by fixing finding 3 below, not by the review.

**3. Three tests that asserted less than they claimed**, from the independent review:

| Finding | Fix |
|---|---|
| REAL — the `persona_record_unreadable` row shared a fixture root with the unparseable-triggers row, so `LoadTriggers` failed first and the row was green without ever reaching its branch | Its own root with valid triggers, plus a `wantStderr` column so every row must prove which branch it reached |
| The `kind: invocable` filter was asserted by nothing — deleting it left the suite green | `TestResolveRolesExcludesNonInvocable`, mutation-checked: removing the filter turns it red |
| AC4 was verified by asserting the manifest contains a string — the form AC4 itself calls insufficient | `TestPromptHookReachesSettingsByConsequence` runs `LoadBindTargets → HookCommands → MergeHooks` and asserts the hook lands in the emitted settings |

The fourth finding (the latency test measures match+join only, excluding per-prompt disk work) is
noted rather than changed: the reviewer measured the real path at avg 7.13ms / max 9.14ms against a
20ms budget, and the load half is now bounded by the 5s timeout.

**4. The spec archive**, carrying the review verdict so it is auditable.

## Review provenance

Independent adversarial review: **PASS**, `nan/deepseek-v4-flash`, drawn from the pool, not the
implementer. Rubric Correctness B, Verification B, Scope A, Reliability A, Maintainability A,
Handoff-readiness B. No Blocker, no Major. Full text in `specs/archive/HARNESS-110/review.md`.

Disclosure: the first pool draw (`nan/qwen3.8-flash`) ran ~37 minutes and died on a provider-side
`400: "Input text data may contain inappropriate content"` with `willRetry: false`, settling without
writing a verdict — a content-filter false positive on repository text, not a review. Re-run
deliberately on the pool primary via `--reviewer`. Recording it because an aborted review and a clean
one are indistinguishable from outside unless someone says which happened.

## Verification

```
go build ./...                             OK
go test ./internal/harness ./internal/cmd  ok
go vet ./... + GOOS=windows go vet ./...   OK
golangci-lint run                          0 issues (v2.12.2, the versions.conf pin)
pre-commit                                 all hooks Passed
```


mlorentedev:
## Review triage

|---|---|---|
| **PR-Agent** (`github-actions`) | All 8 ACs listed compliant, **non-compliant: (none)**, **requires further human verification: (none)** | **Nothing to apply.** No finding was raised |
| **CodeRabbit** | *"Review rate limited"* | **No review ran.** Disclosed, not implied — a notice about a review is not a review. Second consecutive PR where its free tier produced nothing |
| **Independent adversarial** (`nan/deepseek-v4-flash`) | **PASS**, rubric B/B/A/A/A/B, no Blocker or Major | Ran against this content on the predecessor branch; its four Minor findings are what this PR carries. Verdict ships in `specs/archive/HARNESS-110/review.md` |

**One observation about the reviewer, not the code:** PR-Agent headed both sections
*"🔶 Partially compliant"* while listing zero non-compliant requirements and zero items needing human
verification — and it treated **#1455, a pull request, as a ticket to satisfy**. That is the
ticket-scope inference defect already measured 5/5 on another lane the same day, and it is tracked as
reviewer configuration. The per-requirement findings are accurate; the headline verdict is not.
Reading the headline instead of the rows would have manufactured work that does not exist.

## Checks

All green: `spec-gate` **pass** (it was the gate that flagged the unarchived spec this PR fixes),
`test (ubuntu-latest)`, `test (windows-latest)`, `integration`, `lint`, `lint-powershell`,
`attestation`, `review-attestation`, `goreleaser snapshot`, `GitGuardian`.

Local, reproduced this session:

```
go build ./...                             OK
go test ./internal/harness ./internal/cmd  ok
go vet ./... + GOOS=windows go vet ./...   OK
golangci-lint run                          0 issues (v2.12.2, the versions.conf pin)
```

Nothing outstanding. Ready to merge on owner authorization.

Issue #1460, opened 2026-09-03
CLI-072: dotf hooks install — retire the install-git-hooks twin pair, and the test twins with it


mlorentedev:
Measured 2026-09-03, before any code was written.

## Why this pair and not another

The `dotf` CLI migration drive measures progress by **setup-script LOC and twin-pair
count**. Two OPS tickets have closed well and LOC moved 4079 → 3992, but the pair
count has not moved at all: **six twin pairs at drive start, six today**
(`dotfiles-sync`, `install-dotf`, `install-git-hooks`, `obs-cli`, `utils`,
`vault-maintenance-weekly`). The queue as written was not pointed at the metric.

Of the six, `install-dotf` is bootstrap (ADR-020 C7, off the table) and `utils` is
the library both setups source, which shrinks by attrition and is never a target.
`install-git-hooks` is the cleanest remaining candidate: discrete behaviour, real
test coverage on both sides, no C7 constraint, and one caller per platform.

## The C7 objection, and why it does not hold

`scripts/install-git-hooks.ps1` opens by declaring itself exempt:

> Bootstrap stays shell (ADR-020 C7): the deployed `dotf` release carries no
> source tree, so setup-windows.ps1 places the dispatcher and `dotf doctor`
> verifies the wiring thereafter.

C7 reads, verbatim: *"**Irreducible shell bootstrap.** The step that provisions
**the tooling itself** stays shell (chicken-and-egg)."* Installing git hooks does
not provision the tooling — `dotf` already exists by then, so there is no
chicken-and-egg. C7 covers `install-dotf`; it does not reach this.

The stated reason — the release carries no source tree — **is already solved in
this repository**. `//go:embed` is used in six packages (`spec`, `initrepo`,
`vault`, `harness/model_map`, `harness/triggers`, `initrepo/scaffold`) precisely so
the binary is self-contained; `dotf spec init` writes templates on a machine with
no vault checked out. `git-hooks/` is **8 files** (`pre-commit`, `commit-msg`,
`pre-push`, `post-checkout`, `prepare-commit-msg`, `lib/{board-pickup,
chain-local-hook,memory-sink-guard}.sh`). Embedding them is the established
pattern here, not a new one.

This is a justification that outlived its mechanism (repo lesson 259): true when
written, false once embedding landed elsewhere.

## The actual finding: the code twins agree, the TEST twins do not

Both scripts implement the same behaviours and both carry the safety guards. The
verification does not match:

[...]

PR #1463, opened 2026-09-03
fix(harness): scan ids over REST and yield on ownership, not on issue number

mlorentedev:
## What

Fixes the two rules that governed `AREA-NNN` allocation, both of which were measured wrong today
while running #1448's AC5 classification, and records the failure that exposed them.

Documentation and skill prose only — no Go, no shell, no behaviour change.

## Why

**The scan was GraphQL.** `new-ticket/SKILL.md` computed the next free id with `gh issue list`,
which fails under a GitHub secondary rate limit. That is exactly the condition under which parallel
sessions are filing tickets, so the scan was unavailable precisely when it was load-bearing. Now
`gh api --paginate --jq`, one line per issue, PRs excluded, population printed.

**A throwaway allocator I wrote for the same pass proposed an id another session already held.**
`CLI-072` is #1460. The scan behind it put `TEST` at maximum `001` when the real maximum is `008`.
Cause: `--paginate` concatenates one JSON array per page and I split them with
`re.findall(r'\[.*?\]')`, which breaks on the first `]` inside an issue body — dropping pages
silently and returning a plausible number. Caught only because the plan was printed before it ran.
That is `docs/lessons/lesson-263`.

**The yield rule was wrong in 3 of 11 collisions.** It said the higher issue number renumbers.
`CLI-065`, `HARNESS-067` and `REFACTOR-011` are each owned by the *higher* number, because a spec
folder's frontmatter says so (`issue: mlorentedev/dotfiles#1363`). Applying the old rule would have
orphaned three active spec folders — manufacturing the stale-referent defect #1448 exists to catch.
The rule is now ownership-first: spec frontmatter, then inbound references, then the board `ID`
field, and ordering only as a tie-break.

## The lesson-261 collision, resolved by the new rule applied to itself

Two files claimed `lesson-261`: the merged-PR one (landed 22:15) and the secret-guards one
(22:25). Ordering says the later yields. **Ownership says the opposite and wins**: the later file
is cited from `AGENTS.md:65`, `docs/adr/adr-028:338` and `specs/SEC-001-secrets-run-guard/tasks.md`
AC8, while the earlier is cited by nothing but its own index row. So `261` stays where the
references point and the merged-PR lesson becomes `262`. `_index.md` also gains the row the
collision had swallowed — the secret-guards lesson was never indexed.

## Verification

```
pre-commit   Detect hardcoded secrets / Run dotfiles tests / bats @test names /
             Check referenced doc paths in instruction files / Validate message format   all Passed
```

The doc-path check is the one that matters here: it is what proves the renamed lesson and the new
`lesson-263-*` reference resolve, and that `AGENTS.md` and `adr-028` still point at a file that
exists.

## Not in scope

The detector itself (#1448 AC1-AC4) is deliberately not started — 174 issues opened against 103
closed in the last 14 days, so the meta-work cap applies and classification was the cheaper honest
half. The classification result, the fixtures it produced and the two findings above are recorded
on #1448.

Deploy note: `harness/skills/**` reaches `~/.claude/skills/` through `scripts/compile-harness.sh`,
so the change is inert for running agents until the next deploy.

Refs #1448

### DIFF
--- a/docs/lessons/_index.md
+++ b/docs/lessons/_index.md
@@ -278,6 +278,8 @@ tags: [lessons, index, dotfiles]
 | [258 - An argument passed through two independent quoting conventions is only as safe as the weaker one](lesson-258-an-argument-passed-through-two-independent-quoting-layers.md) | 2026-09-02 |  |
 | [259 - A justification outlives its mechanism as easily as a name outlives its contract, and the cleanup that taught us the first lesson left an instance of it](lesson-259-a-justification-outlives-its-mechanism.md) | 2026-09-02 |  |
 | [260 - The path with no seam is the path with no test, and it is the one that fails in production](lesson-260-the-path-with-no-seam-is-the-path-with-no-test.md) | 2026-09-02 |  |
-| [261 - A merged PR and a stuck PR are indistinguishable from the branch side](lesson-261-a-merged-pr-and-a-stuck-pr-look-identical-from-the-branch.md) | 2026-09-02 |  |
+| [261 - Never test secret guards against live credentials, and redact at the stream boundary](lesson-261-never-test-secret-guards-against-live-credentials-and-redact-at-the-stream-boundary.md) | 2026-09-02 |  |
+| [262 - A merged PR and a stuck PR are indistinguishable from the branch side](lesson-262-a-merged-pr-and-a-stuck-pr-look-identical-from-the-branch.md) | 2026-09-02 |  |
+| [263 - A population that cannot prove it is complete answers with a plausible number](lesson-263-a-population-that-cannot-prove-it-is-complete.md) | 2026-09-02 |  |
 
 - [2026-08-30 Intercepting Cobra Errors without Breaking SilenceErrors](2026-08-30-cobra-silence-errors-interception.md)
--- a/docs/lessons/lesson-262-a-merged-pr-and-a-stuck-pr-look-identical-from-the-branch.md
+++ b/docs/lessons/lesson-262-a-merged-pr-and-a-stuck-pr-look-identical-from-the-branch.md
@@ -1,5 +1,5 @@
 ---
-id: lesson-261
+id: lesson-262
 type: lesson
 status: active
 created: "2026-09-02"
--- a/docs/lessons/lesson-263-a-population-that-cannot-prove-it-is-complete.md
+++ b/docs/lessons/lesson-263-a-population-that-cannot-prove-it-is-complete.md
@@ -68,6 +68,29 @@ support it — produced by the tooling written to serve that ticket, roughly an 
 argued that a defect class surviving its own authors' knowledge cannot be fixed by writing it down
 again.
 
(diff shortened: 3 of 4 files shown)


mlorentedev:
## Review triage

|---|---|---|
| **PR-Agent** (`github-actions`) | Ticket-compliance blocks for #1448, #1460 and #1363. No security concerns, no major issues, effort 1/5 | Dispositioned below |
| **CodeRabbit** | *"Review skipped — this branch is under active development … CodeRabbit has automatically paused this review"* | **No review ran.** Disclosed, not implied |

### Findings

| # | Finding | Disposition |
|---|---|---|
| 1 | #1448 "Partially compliant" — *current duplicate IDs not classified in this PR*, AC1/AC2/AC4/AC6 not implemented | **Accurate about the diff, and no change.** The classification is real but is not diff-shaped: 11 issues renumbered, 2 closed on evidence, all recorded in a comment on #1448. AC1/AC2/AC4 are deliberately out of scope — 174 issues opened against 103 closed in 14 days, so the meta-work cap applies and this PR is the cheap half |
| 2 | #1460 and #1363 "Not compliant" — no `dotf hooks install`, no env-sweep implementation | **Declined.** Both are issues this PR merely *references*: #1363 is the issue a spec's frontmatter names when deciding a yield, #1460 is the id that proved my allocator wrong. Neither is a requirement this PR claims |

### Finding 2 is a reviewer defect, and it sharpens a known one

Eighth instance of PR-Agent inferring ticket scope wrongly. The previous reading was *"it treats a
pull request as a ticket"* (#1455 on #1462, #1433 on #1440, #1457 on #1461). This case refutes that
narrowing: #1460 and #1363 are **issues**, not PRs, and they were consumed the same way.

The bug is broader: **every referenced issue number is treated as a requirement source the PR must
satisfy.** The earlier three were referenced numbers whose being PRs was incidental. That matters
for the fix — "exclude PRs from ticket inference" would not have caught this one.

### A gate defect surfaced here, not caused by this PR

`review-attestation` reports **`[OK] attested — reviewed by: coderabbitai`** on this PR. The only
thing CodeRabbit posted is the auto-pause notice quoted above. **The gate accepted a notice about a
review as a review.**

That is precisely the distinction the pr-stewardship doctrine requires be made by content and never
by author — *"a review names files, lines, or claims; a notice talks about the review itself"* — and
the automation does it by author. A PR nothing reviewed can pass the gate that exists to prove it
was reviewed. Recorded as a measurement on #906 rather than filed as a new ticket.

### Verification

```
spec-gate  pass    lint  pass    lint-powershell  pass    integration  pass (1m19s)
test       pass (4m38s)          attestation  pass       changes  pass
test-windows  pending at the time of writing
```

Docs and skill prose only — no Go, no shell, no behaviour change. The load-bearing check is
pre-commit's *Check referenced doc paths in instruction files*, which is what proves the renamed
lesson resolves and that `AGENTS.md:65` and `adr-028:338` still point at a file that exists.

Issue #1429, opened 2026-09-02
TEST-008: install-dotf.bats reads the ambient PATH dotf, so it fails on the box that is changing the CLI and passes on one that never built it


mlorentedev:
## What

`tests/install-dotf.bats` resolves `dotf` from the **ambient PATH** instead of
from its own fixture, so its result depends on what happens to be installed on
the machine running it. Five tests fail on a box carrying a source build:

```
not ok 628 happy path: verifies sha256 and installs the binary
not ok 629 checksum mismatch aborts the install, leaving no binary
not ok 630 missing checksum entry aborts the install
not ok 632 converges over a running dotf: a live binary in dest is replaced, not refused
not ok 633 install leaves no staging artifact behind in dest
```

## Why it happens

`install_dotf` deliberately refuses to overwrite a source build:

```sh
if [ "$_dotf_current" = "dev" ]; then
    log_info "dotf is a source build (dev); leaving it in place ..."
    return 0
fi
```

That behaviour is correct and documented (#1305: the release lags the tree, and
replacing a dev build would make testing it certify nothing). But the tests read
that ambient binary, so on a developer box mid-change — exactly when the suite is
most needed — the install path short-circuits and `[ -x "$DEST/dotf" ]` fails.

## Why it matters

The failure is **backwards**: the suite is green on a machine that has never
built the CLI and red on the machine actively changing it. A test whose verdict
depends on the developer's own environment is not evidence either way, and the
five names above read as real install-path breakage when nothing is broken.

Same class as #1409 / BUG-771 — red on a clean tree locally, green in CI, from
fixture leakage rather than a regression.

## Acceptance criteria

- [ ] The tests resolve `dotf` from their own fixture, not from `PATH`.
- [ ] The suite gives the same verdict on a box with a `dev` build, with a
      released build, and with no `dotf` at all — asserted, not assumed.
- [ ] A red-direction test covers the `dev` short-circuit itself, so the
      behaviour that causes this stays deliberately verified rather than merely
      worked around.

Issue #695, opened 2026-07-07
BUG: README clone-into-~/.dotfiles layout kills setup mid-run and self-destructs the git-hooks dispatcher


mlorentedev:
**Problem.** README Quick Start clones INTO `~/.dotfiles`. With `CURRENT_DIR == DOTFILES_DIR`, the env-contract deploy `cp` (~`setup-linux.sh:1437`) fails "same file" and `set -euo pipefail` aborts setup there (exit 1) — env generate and the final doctor never run. Worse: `scripts/install-git-hooks.sh` `deploy_git_hooks` does `rm -rf dest` + unchecked `cp src/. dest` with src==dest — it EMPTIES the dispatcher, still logs `[SUCCESS]`, the second setup run refuses (`has no pre-commit dispatcher`), and doctor's fix text ("run dotfiles setup") is circular. GUARD-001 ends up inert. CI only exercises the other layout (`~/dotfiles-repo`, tests/Dockerfile.integration:48), so the documented path is the untested one. Related: README Requirements omit `curl` — without it setup dies even earlier (`setup-linux.sh:238`, unguarded `$(curl …)` under set -e).

**Repro:** process-audit E2 S1/S5 and E3 T8 (exact sequences in the report).

**Fix direction.** Decide the supported layout first (gates the fix): either make in-place safe (guard every same-file copy; `deploy_git_hooks` no-op when src==dest; add a CI variant for the README layout) or change the README to the copy layout and have setup refuse in-place with a clear message. README truth pass is #677.

Source: docs/adr/audit-010-process-workflows.md §4 P2 (CONFIRMED) + §6 Q1.

PR #1464, opened 2026-09-03
feat(hooks): dotf hooks install, retiring the install-git-hooks twin pair and its test twins

mlorentedev:
Spec: `specs/CLI-072-dotf-hooks-install/`

## Twin pairs 6 → 5 — the first time the drive's own metric has moved

The `dotf` migration drive measures itself in setup-script LOC and twin-pair
count. Two OPS tickets had closed well and LOC had moved 4079 → 3992, but the
pair count sat at six from the drive's first day. `install-git-hooks` is retired
here: `scripts/install-git-hooks.{sh,ps1}` and
`tests/install-git-hooks.{bats,Tests.ps1}` are gone, both setups repoint at
`dotf hooks install`, and `scripts/` falls **7921 → 7571**.

## The finding: the code twins agreed, the TEST twins did not

13 bats cases against 9 Pester ones for the same behaviours. Four were verified
on Linux only — and **the guards themselves existed on both sides**, so this was
a coverage gap and not a live defect. It still meant a destructive `rm -rf` ran
on Windows behind a guard nobody had ever exercised.

| Verified on Linux only | Now, from one source on both legs |
|---|---|
| `Test-GuardDispatcher` — an equivalent dispatcher elsewhere is ACTIVE; a `pre-commit` without the guard warns | `TestInstallWiresHooksPath` subtests 4–5 |
| `Test-SameHooksPath` — a trailing-slash variant counts as wired | `TestInstallWiresHooksPath` subtest 3 |
| the `#695` self-mirror guard | `TestInstallDoesNotDestroyItsOwnSource`, plus the first-install case bats could not reach |
| a missing source dispatcher directory | `TestInstallRefusesABadSource` |

The Pester-only drive-root refusal is kept too. Test drift shows no symptom until
the untested guard is the one that fails, which is why one Go suite replaces two
that had quietly diverged.

## Three design decisions, each settled by measurement before any code

**`//go:embed` was the first design and was measured out.** `git-hooks/` sits at
the repo root, **outside the Go module**, so embedding meant relocating a
directory the README documents as a landmark — and with it `.gitattributes`'s
`eol=lf` rule, the CI path filter, `architecture.md`, three audit ADRs, and three
unrelated bats suites that read hook files from the root. A relocation change
wearing a port's clothes. Keeping a `--source` directory also keeps three guards
*reachable* that embedding would have silently retired, `#695` among them.

**On Windows the step MOVED, it was not repointed.** `setup-windows.ps1`
installed hooks ~270 lines **before** `Install-Dotf`. Harmless while it sourced a
PowerShell twin; a silent failure the moment it invoked `dotf`, because the
surrounding `Write-Warn` would have turned "binary not found" into a skipped
memory-sink guard on every fresh box. `tests/guard-setup-hooks-order.bats` spans
the two call sites nothing spanned before — which is why the inversion survived.

**`git` sits behind a runner seam**, because `git config --global` has
machine-wide blast radius and a suite that drove the real binary could rewire the
developer's `~/.gitconfig`. One integration test drives real `git` against a
throwaway `GIT_CONFIG_GLOBAL`, which is what proves the fake speaks the real
dialect — the CLI-071 lesson about a fake that agreed only with itself.

## Two guards came out stronger than their originals

`checkDotfilesDir` refuses a root with `filepath.Dir(p) == p`, true of `/` and
`C:\` alike, where the shell needed a POSIX check *plus* a separate PowerShell one
— which is exactly why the drive-root case existed only in Pester. `sameDir` uses
`os.SameFile` and short-circuits when the destination does not exist: the
first-install case the bats test for `#695` could not reach, because it assumes
both paths exist.

## Verification

```
go build / vet / GOOS=windows vet / test / -race → clean    golangci-lint → 0 issues (v2.12.2 pin)
shellcheck setup-linux.sh → clean                          bash -n / zsh -n → OK
bats tests/*.bats → BATS_EXIT=1, 1520 ok, 7 not ok         features.json → 10/10, all non-vacuous
```

**The 7 bats failures are pre-existing and were verified, not assumed.** All seven
fail identically on a clean detached `origin/main` worktree: six in
`install-dotf.bats` (#1429, the `dev` build on PATH) and `BUG-771` in
`skills-pipeline.bats` (#1409). "Known" is not "measured on this tree".

**15 mutations, all caught** — including unregistering the command from `root.go`,
making the installer write where doctor does not look, and inverting the setup
ordering comparison.

**Two escaped the first battery, and both were defects in the tests rather than
the code.** Each asserted *that* something was refused rather than *which*
diagnosis came back. Removing the empty-dir guard still refused — via the root
branch — while telling a user with an unset `DOTFILES_DIR` that `"."` is a
filesystem root. A byte comparison in `samePath` fell through to the
equivalent-dispatcher branch, which **also** declines to write: identical
observable outcome, different message. Both tests now read the message, because
where two branches differ only in what the user is told, the message *is* the
behaviour.

**A false green I produced and caught:** `bats tests/*.bats | tail -15` reported
"exit code 0" — that is `tail`'s status, not the suite's, and the captured file
held only the last 15 lines. Re-run without the pipe, capturing `$?` directly.

## Doctor

The GUARD FAIL remedy now names ``run `dotf hooks install` `` instead of "run
dotfiles setup". That string is user-facing twice, since CLI-070 lifts FAIL
remedies verbatim into a Next-steps block — a stale remedy is *promoted*, not
buried. `checks_guard_agreement_test.go` asserts doctor and the installer agree on
the dispatcher path by **running both** against one temp deploy root; two
constants compared to each other would pass if both were wrong together.

## Reported rather than netted away

**Setup LOC went 3992 → 4025 (+33)**: the Windows block grew to record why it
moved, the Linux block to resolve `dotf` by path (the #1202 idiom — the process
may not carry `~/.local/bin` yet). Net repo surface is −317. The drive tracks two
metrics and here they moved in opposite directions; a net figure would have hidden
the direction.

## Prose sweep (lesson 259)

Grepped for the retired names across the repo, **comments included**. One real
breakage: `tests/setup-linux.bats:492-494` asserted setup *sources* the deleted
script. Past-tense references are kept deliberately — recording where behaviour
went is what the lesson asks for — and the ordering guard strips comment lines
before checking for invocations, for exactly that reason.

Refs #1460


mlorentedev:
## Review triage

**Who actually reviewed: PR-Agent only.** CodeRabbit posted *"Review limit reached
— next included review available in 42 minutes"*. That is a notice, not a review,
and it is now the **fourth consecutive** PR on which it has produced one (#1456,
#1457, #1461, #1464). Recorded rather than left to look like silence: this change
has not been read by CodeRabbit.

### Applied

| Finding | Disposition |
|---|---|
| **Potential file-mode issue** — `copyTree` writes `0o644` and relies on `markExecutable`; if that fails partway, hooks end up non-executable and git refuses to run them | **Applied** in `3e929b3`. The real defect was narrower and worse than described: `markExecutable` skipped an entrypoint on **any** stat error, so a permission problem or a broken mount was indistinguishable from "this dispatcher does not carry that hook type", and the deploy reported success. Only `os.IsNotExist` is skippable now. This is the spec's own class — an error path whose failure looked exactly like its success — so it is fixed rather than tracked. |

### Declined, with reasons

Both remaining items are ticket-compliance inferences against issues this PR
merely **references**:

| Finding | Disposition |
|---|---|
| **#695 "partially compliant"** — *in-place layout still causes setup failure (env-contract deploy `cp` fails "same file")*; *missing curl requirement in README* | **Declined.** #695 is cited because its self-mirror guard is ported here, and that half is confirmed compliant. The env-contract `cp` failure is a **different** defect in a different code path (`audit-010:165` documents it), and the README's curl prerequisite is not this PR's subject. Neither was claimed. |
| **#1429 "partially compliant"** — *test still reads from PATH, not fixture; no new test for dev short-circuit* | **Declined.** #1429 is cited as the **explanation** for six pre-existing bats failures, verified failing identically on a clean `origin/main` worktree. Citing a known-issue number to explain a measurement is not undertaking to fix it. |

**Ninth instance of the same reviewer defect**, and it now has a precise
statement: PR-Agent treats **every issue number appearing in a PR body or commit
message as a requirement source the PR must satisfy** — whether the referent is a
ticket (#695, #1429, #1363), a pull request (#1455, #1433, #1457), or a citation
explaining a measurement. "Exclude PRs from ticket inference" would not fix it;
the bug is referent resolution, not referent type. Reviewer configuration, carried
as such.

Worth stating plainly against that record: **the code findings have been reliable
throughout, and this one was real and is applied.** The ticket-compliance section
is the part that has never been right.

### CI

Green on the merged head except `test-windows` and `review`, in flight at the time
of writing; `3e929b3` restarts them. Earlier `cancelled` runs on superseded heads
are concurrency cancellations, not failures — checked rather than read either way.

The substantive review of this change is still owed: the independent adversarial
review runs against `main` after merge, and it, not this table, is what the spec
gate requires before archiving.

PR #1465, opened 2026-09-03
docs(shell): add PIPESTATUS to the prohibited patterns, in the silent group

mlorentedev:
## What

One row in the `Shell Compatibility (MANDATORY)` table of `.claude/CLAUDE.md`, plus the one-word
count fix the row makes necessary. Documentation only.

| Prohibited | Use instead |
|---|---|
| `${PIPESTATUS[0]}` to check a piped command | drop the pipe and use `$?`, or `set -o pipefail` when the pipe is needed |

## Why

zsh spells it `$pipestatus[1]` and indexes arrays from 1, so the bash form expands to **nothing**.
`echo "EXIT=${PIPESTATUS[0]}"` prints `EXIT=` — not an error, not a wrong number, an absence that
reads as inconclusive-but-fine.

Measured twice on 2026-09-02 in this repo, by two different sessions:

- wrapping `scripts/compile-harness.sh --deploy` to confirm it succeeded — the deploy had in fact
  succeeded, but the check proved nothing and I only found that out by grepping the deployed files;
- `bats tests/*.bats | tail -15` reporting `exit code 0`, which was **`tail`'s** status. The real
  run was `BATS_EXIT=1`.

The second is the dangerous shape: a verification that reads a pipeline's exit status reports the
*last* command's, so a failing `cmd` passes silently for as long as nobody looks.

## Placement, and the count that had to move with it

It goes in the table's **last group**, not the first. The table separates rows that break loudly
from those that "fail **silently** — they return an empty or single-element result instead of an
error, and empty reads as a finding". `${PIPESTATUS[0]}` under zsh returns empty; that is the
group's signature exactly, and the group's own standing advice — *re-run it in the other shell
before believing an empty result* — is precisely what catches it.

The paragraph closing that group read **"The last three rows fail silently"**. A fourth silent row
makes that sentence false, so it becomes four in the same edit. Adding the row without touching the
count would have created a claim whose referent moved underneath it — the defect class #1448
tracks, manufactured inside the change that documents its cousin.

## Why the table and not a lesson

`PIPESTATUS` already appears in two files under `docs/lessons/` and in neither `.claude/CLAUDE.md`
nor `AGENTS.md`. The knowledge existed and still did not reach the command being typed, because
lessons are read when someone goes looking and this table is read at the moment of writing shell.
That is the same argument as the `new-ticket` concurrency guard, which was written in prose on
2026-07-07 and was followed by 21 duplicate ids.

Verified `.claude/CLAUDE.md` is hand-maintained, not generated: `harness/manifest.json`'s
`agents.presence[0]` maps **`~/.claude/CLAUDE.md`** to source `ai/claude/CLAUDE.md`, which the
deploy log confirms (`[deploy] presence -> /home/manu/.claude/CLAUDE.md`). The repository's own
`.claude/CLAUDE.md` has no writer and no generated marker, so this edit survives the next deploy.
The table is not duplicated anywhere live — the only other matches are archived spec proposals.

## Verification

```
pre-commit  Detect hardcoded secrets / Run dotfiles tests / bats @test names /
            Check referenced doc paths in instruction files / Validate message format   all Passed
```

No shell, Go or PowerShell changed, so the two verification loops in the file itself are untouched.

Refs #1448


mlorentedev:
## Review triage

|---|---|---|
| **PR-Agent** (`github-actions`) | Ticket-compliance block for #1448: headline "🔶 Partially compliant", **zero** compliant requirements, AC1–AC6 all listed non-compliant. No security concerns, no major issues, effort 1/5 | **Declined**, with reasons below |
| **CodeRabbit** | *"Review limit reached — next included review available in 37 minutes"* | **No review ran.** Disclosed, not implied |

### Findings

| # | Finding | Disposition |
|---|---|---|
| 1 | #1448 AC1–AC6 "Not implemented; PR is documentation only" | **Declined.** Accurate and irrelevant. This PR adds one row to a table and carries `Refs #1448`, not `Closes` — it never claimed an AC. AC1–AC4 are deliberately unstarted (meta-work cap: 174 issues opened against 103 closed in 14 days); **AC5 is done** and recorded on #1448 — 11 issues renumbered, 2 closed on probed evidence, duplicated ids carrying open work 16 → 6 |

### The finding is two known reviewer defects stacked, and one is now proven

**Ninth instance of scope inference by referenced number.** `Refs #1448` was read as a claim to
satisfy every acceptance criterion on #1448. Same mechanism as #1463 (which consumed #1460 and
#1363, both merely mentioned) and as #1455/#1433/#1457 before it.

**The headline aggregation bug, in its cleanest form yet.** The block reads **"Partially
compliant"** above a list of **zero** compliant requirements and six non-compliant ones. The
inverse was observed on #1461 — "Partially compliant" above *every* requirement compliant and
"(none)" non-compliant. The headline is therefore decoupled from its own rows **in both
directions**, which settles it as an aggregation defect rather than judgement noise.

Both belong on the reviewer-config ticket, not here.

### On the attestation gate — this run is correct, and worth saying so

`review-attestation` reports `[OK] attested — reviewed by: github-actions (comment-shaped review:
"## PR Reviewer Guide")`. That is **accurate**: PR-Agent did post a review, and the gate credited
the reviewer that produced it rather than the one that did not.

Recorded deliberately, because the measurement filed on #906 an hour ago is the *opposite* case —
on #1463 the same gate reported `reviewed by: coderabbitai` when CodeRabbit's only output was an
auto-pause notice. The gate is not uniformly broken; it fails when the sole output is a notice and
it resolves the author anyway. Narrowing the defect matters as much as reporting it.

### Verification

```
all checks green — spec-gate, lint, lint-powershell, test, test-windows, integration,
                   changes, attestation, review-attestation, GitGuardian
```

Documentation only: one table row and the count in the sentence below it. No shell, Go or
PowerShell changed.

Issue #1467, opened 2026-09-04
CLI-073: the reviewer registry is resolved against the CWD, so triage-queue fails loudly and the session-start probe fails silently from any subdirectory


mlorentedev:
Both commands that bind Definition of Done §4 resolve the reviewer registry
against the process CWD, not against the repository root. One `os.Stat` decides
whether the triage mechanism runs at all, and from any subdirectory it decides
"no".

## Measured

`dotf` built from `main` @ `eaf8a91`, 2026-09-03.

```
$ cd ~/Projects/dotfiles && dotf pr triage-queue
[OK] no reviewer output is awaiting a disposition        # exit 0

$ cd ~/Projects/dotfiles/cli && dotf pr triage-queue
pr triage-queue: read reviewer registry: open harness/review-attestation.json: no such file or directory
                                                          # exit 1
```

`harness/review-attestation.json` is present and tracked in both cases. The
question is not "is there a registry" but "where did the process happen to be
standing".

## Root cause

One shared defect, two call sites, two opposite symptoms.

- `cli/internal/cmd/pr.go:85` — the `--registry` flag defaults to the *relative*
  `filepath.Join("harness", "review-attestation.json")`, so Go resolves it
  against the CWD.
- `cli/internal/cmd/mem.go:199` — `memTriageQueue` builds
  `filepath.Join(cwd, "harness", "review-attestation.json")` from the
  SessionStart payload's `cwd` (or `os.Getwd()`), with the same result.

`spec.RepoRoot(start)` already exists at `cli/internal/spec/spec.go:58` and
already handles the worktree case (`.git` as a file, not only a directory).
Neither call site uses it.

## Why this matters more than a path bug

**`pr triage-queue` fails in the direction the doctrine cannot distinguish.**
The rule is explicit that a non-zero exit is *"a queue you must read, never an
empty one"* — it exits non-zero both when work is pending and when the question
could not be answered, precisely so an unanswerable queue is never mistaken for
a clear one. That design is correct, and this bug abuses it: from a
subdirectory the command reports "unanswerable" forever, for a reason that has
nothing to do with the queue. A signal that is always red is a signal nobody
reads.

**`mem` fails in the direction its own comment forbids.** `memTriageQueue`
returns `nil` when the `os.Stat` misses, which disables the session-start probe
silently. The comment directly above it states the intended contract:

> A missing registry means "not applicable" and stays silent; a registry that is
> present and unanswerable is reported loudly.

[...]

Issue #1466, opened 2026-09-03
TOOL-020: PR-Agent's ticket-compliance section treats any referenced number as a requirement, and its headline ignores its own rows


mlorentedev:
## What

PR-Agent's **ticket compliance** section treats *every issue number appearing anywhere in a PR* as a
requirement source the PR must satisfy, and its headline verdict is computed independently of the
rows beneath it. Measured **nine times** between 2026-09-01 and 2026-09-02 by two sessions.

Its **code** findings are not in question — see the counter-evidence section. This is scoped to the
ticket-compliance block.

## Defect 1 — any referenced number becomes a requirement source

The referent does not have to be the ticket the PR closes. Three kinds observed:

| Kind | Instances | How the number got into the PR |
|---|---|---|
| **Pull requests** | #1455 (on #1462), #1433 (on #1440), #1457 (on #1461) | cited as the PR whose gap this one closes |
| **Unrelated issues** | #1460, #1363 (on #1463) | one names a spec's owner, one proved an allocator wrong |
| **Citations** | #695, #1429 (on #1464) | #695 as a ported guard's origin, #1429 to *explain* six pre-existing bats failures |

The third kind is the sharpest: **citing a known-issue number to explain a measurement was read as
undertaking to fix it.** On #1465 — a PR adding one row to a markdown table, carrying `Refs #1448`
and not `Closes` — every one of #1448's six acceptance criteria was reported non-compliant.

**Consequence for any fix:** "exclude pull requests from ticket inference" resolves three of nine.
The rule needs to be *closing keywords only* (`Closes`/`Fixes`/`Resolves`), never bare references.

## Defect 2 — the headline is decoupled from its own rows, in both directions

| PR | Rows | Headline |
|---|---|---|
| #1461 | every requirement compliant, "(none)" non-compliant | 🔶 **Partially compliant** |
| #1465 | **zero** compliant, six non-compliant | 🔶 **Partially compliant** |

Both directions from the same verdict string settles this as aggregation rather than judgement: the
headline does not read the rows at all. It is also the half a human actually reads, so a PR with a
perfect row set is reported as partial and one with no compliant requirement reads the same.

## Counter-evidence — the code review half is reliable

Recorded so this ticket cannot be misread as "turn PR-Agent off". Across the same nine PRs its code
findings were **correct every time**, including:

- #1455: the `suggest-role` hook carried no `timeout`. Real; applied.
[...]

PR #1468, opened 2026-09-04
chore(spec): archive CLI-072 after an independent review, applying the two findings that survived measurement

### LINKED ISSUE #1460: CLI-072: dotf hooks install — retire the install-git-hooks twin pair, and the test twins with it

mlorentedev:
Archives CLI-072 and closes #1460. The implementation itself merged as #1464
(`c375102`): `dotf hooks install` replaced `scripts/install-git-hooks.{sh,ps1}`
and both test twins, taking the twin-pair count from **6 to 5** — the first
movement in that metric since the CLI migration drive began.

## Independent adversarial review

**PASS.** `nan/mimo-v2.5`, drawn at random from `harness/reviewer-pool.json`,
not the implementer. Reviewed against `eaf8a91`. No Blocker, no Major.

| Dimension | Grade |
|---|---|
| Correctness | A |
| Verification | B |
| Scope | A |
| Reliability | A |
| Maintainability | A |
| Handoff-readiness | B |

Full verdict: `specs/archive/CLI-072-dotf-hooks-install/review.md`.

## Disposition of the five findings

**Applied (2).**

- *`copyTree` strips CR from every file, including a hypothetical binary.* The
  finding was right, and worse than stated: `isText`'s NUL-byte probe — the only
  thing standing between that strip and a corrupted file — **had no test at
  all**. Removing it left the suite green, because `TestInstallNormalisesCRLF`
  only asserts that text *loses* its CR. `TestInstallLeavesNonTextFilesByteVerbatim`
  now covers it, verified by mutation: with the guard removed the new case fails
  and the CRLF case still passes.
- *`Install` is not transactional.* Documented in its godoc. Deployment can
  succeed while the wiring fails, leaving the guards on disk and inert; a caller
  cannot see that from the signature, and one that swallows the error turns every
  guard off silently.

**Declined (3), each on a measurement rather than an argument.**

- *"Set `features.json` states to `pass` before archiving."* Declined on two
  independent grounds. First, doing it would have **invalidated the review that
  asked for it**: `features.json` is a contract file
  (`cli/internal/spec/review.go:24`) and the staleness gate compares contract
  files against `reviewed_sha` across both committed history *and* `git status
  --porcelain` — a check whose own comment names this exact shape ("get a passing
  review, rewrite the acceptance criteria in the working tree, archive"). Second,
  the premise is false: `archive.go` has no reference to `features` or `state` in
  309 lines, and this archive succeeding proves it. The field is vestigial
  repo-wide — across every archived spec it carries 506 `pending`, plus
  `verified` (42), `passing` (32), `verifying` (15), `done` (15), `passed` (12)
  and `implemented` (6), and **zero** `pass`. Seven unvalidated spellings with
  nothing consuming any of them is why a careful reviewer read it as a gate;
  filed separately, not fixed here.
- *"Assert the `C:\` drive-root case on Linux too."* The suggested assertion
  would fail. `path/filepath` compiles per-GOOS: measured on Linux,
  `filepath.Clean("C:\\")` is `C:\` and `filepath.Dir` of it is `.`, so
  `filepath.Dir(clean) == clean` is **false** and the root check does not fire.
  On Linux `C:\` is a relative single-element filename, not a drive root. The
  `windowsOnly` skip is correct, not a coverage gap.
- *A broken git binary read as "`core.hooksPath` unset".* Raised as a Question,
  explicitly out of scope, no action requested.

## Verification

Run in this branch, at the pinned toolchain:

```
go build ./...                 BUILD_OK
go vet ./...                   VET_OK
GOOS=windows go vet ./...      WINDOWS_VET_OK
go test ./...                  exit 0, 23 packages ok, no FAIL lines
golangci-lint run              exit 0, 0 issues (v2.12.2, matches versions.conf)
```

Windows vet included deliberately: the Windows CI leg compiles the same tree and
fails the whole package on a build error a Linux-only loop cannot see.


mlorentedev:
## Review triage

Who actually reviewed: **PR-Agent** (`github-actions`, `## PR Reviewer Guide`,
the counting marker). **CodeRabbit posted nothing at all on this PR** — not a
review, not a limit notice. Disclosed rather than glossed: one reviewer produced
substantive output, and the registry treats CodeRabbit as advisory, so this is
attested, not unreviewed.

Separately, this PR *carries* the independent adversarial review that gates the
archive — `nan/mimo-v2.5`, PASS, disposition of all five of its findings in the
PR body and in `specs/archive/CLI-072-dotf-hooks-install/review.md`. The table
below is the CI/reviewer triage, not that.

| Source | Finding | Disposition | Reason |
|---|---|---|---|
| PR-Agent | No major issues; no security concerns; effort 1/5; tests present | **Noted** | No code finding to act on. Its code findings have been reliable here; the absence of one is informative. |
| PR-Agent | "Ticket compliance: **#1464** — fully compliant" | **Declined** | #1464 is a **pull request**, not a ticket. Tenth instance of the referent-resolution bug tracked as #1466: it treats every issue number in a body or commit message as a requirement source. The #1460 half of the same analysis is correct and accepted. |
| CI | `test-windows` — `cancelled` at 30m06s (step 7, `setup-windows.ps1` end-to-end) | **Re-run, not overridden** | Conclusion was `cancelled` (a ceiling), not `failed`. Norm for this job is 9–10m (#1464: 9m03s, #1462: 10m24s). This PR's production diff is one godoc comment plus one Go test, in a package `setup-windows.ps1` does not execute, so it has no mechanism to treble that step. Re-running to distinguish flake from regression rather than merging past a required check. |
| CI | Remaining 11 checks | **Green** | `lint`, `lint-powershell`, `test`, `test (ubuntu-latest)`, `spec-gate`, `attestation`, `changes`, `integration`, `add-to-project`, GitGuardian, `review`. |

### One measurement worth recording, because it nearly misled me

`test-windows` **skips on `main`** under the CI path filter — it was `skipped`,
not passed, on both post-merge runs for #1464 and #1465. So "main is green" is
not evidence that this job passes on main's tree, and the check that actually
exercises `setup-windows.ps1` only ever runs on a PR. I started to reason from
the green main runs and would have concluded the wrong thing; the real
comparison is against the PR runs of #1464 and #1462, which is where the 9–10m
norm above comes from.

### Filed, not fixed here

- **#1467** (CLI-073) — the reviewer registry is resolved against the CWD, so
  `dotf pr triage-queue` exits 1 from any subdirectory and, worse,
  `memTriageQueue`'s stat misses and the session-start probe silently does not
  run.
- **#1469** (CLI-074) — a `dev`-stamped `dotf` is a terminal state with no drift
  check; the binary on this machine could not serve the hook config deployed
  alongside it, and every prompt failed with `unknown flag: --from-hook` until a
  human read the error.
- **#1470** (GUARD-010) — `features.json`'s `state` is written by five
  vocabularies and read by nothing; it is what the adversarial reviewer mistook
  for the archive gate.

mlorentedev:
## Review triage — addendum: merging with `test-windows` red

Recording this before the merge, not after. **`test-windows` is a required
check and it is RED on this PR.** It was not overridden quietly and it did not
go green; it is red for a reason that has nothing to do with this change, and
the evidence for that is below.

### What the failure actually is

`test-windows` was `cancelled` — not `failed` — twice, at 30m06s and 30m07s,
both times in step 7 (`Run setup-windows.ps1 end-to-end`). The job's own log
gives the cause: `setup-windows.ps1` reconciles the pi package manifest with one
serial `npm install` per package, and on tonight's Windows runner each takes
between 54 seconds and 7 minutes.

| Package | Duration |
|---|---|
| `npm:pi-effort@0.0.8` | 54s |
| `npm:pi-web-access@0.24.2` | 93s |
| `npm:@ayulab/pi-rewind@0.4.6` | **421s** |
| `npm:@narumitw/pi-plan-mode@0.53.0` | 134s |
| `npm:@narumitw/pi-goal@0.53.1` | 200s |
| `npm:pi-subagents@0.56.0` | 153s |
| `npm:pi-memory@0.4.2` | 251s |
| `npm:pi-cc-extensions@0.8.67` | cancelled at the ceiling |

Everything before that step completes in about 2m30s. The remaining 27 minutes
are npm. Tracked as **#1472**.

### Why this PR is not the cause — three independent grounds

1. **Mechanical.** `test-windows` never runs `go test`. Its steps are: build
   `dotf`, run `setup-windows.ps1`, doctor gate, Pester suites, bats subset.
   `go build` does not compile test files, so the new
   `TestInstallLeavesNonTextFilesByteVerbatim` cannot execute on this job at
   all. The only production change in this PR is a **godoc comment**, which has
   no runtime behaviour.
2. **Independent reproduction.** #1471 — a different branch, a different author,
   a different diff, the same `eaf8a91` base — hit the same wall in the same
   step and also completed as `cancelled`. Neither PR touches
   `ai/pi/packages.json`.
3. **Unchanged input.** `ai/pi/packages.json` has not been modified since #1226.
   The same step passed at **9m03s** on #1464 and **10m24s** on #1462 with the
   identical package list. The variable is npm registry latency, which is not a
   property of this repository.

A required check that fails identically with and without this change has stopped
measuring this change.

### One thing that nearly produced the wrong diagnosis

`test-windows` is **`skipped`** on `main` under the CI path filter — skipped,
not passed, on the post-merge runs for both #1464 and #1465. So "main is green"
was never evidence about this job, and the only place it genuinely runs is a PR.
I started to reason from those green main runs and would have concluded this PR
broke it. The real comparison is against the *PR* runs of #1464 and #1462, which
is where the 9-10m norm above comes from.

### Disposition

Merged with `test-windows` red, deliberately and with the owner's decision, on
the three grounds above. Not re-run a third time: two cancellations at the
ceiling plus a named cause means another attempt buys thirty minutes and no
information. #1472 carries the fix (cache the installs keyed on the manifest
hash, with a timeout bump as a companion) and is the blocker for every PR
touching `cli/` until it lands.

mlorentedev:
## Review triage — round 2 (new reviewer output since the addendum)

Triggered by output that arrived **after** my 02:14:31Z addendum and that the
queue did not surface: CodeRabbit posted at 03:00:26Z and edited in place to
03:11:46Z, so `created_at` never moved past my triage (TOOL-019, #1422).

| Source | Finding | Disposition | Reason |
|---|---|---|---|
| CodeRabbit | **"Reviews paused"** — auto-pause after an influx of reviewed commits, against `ff389ff...9d8e6b8`, 8 files selected | **Disclosed, not dispositioned** | A notice about the review, not a review. No findings were produced, so there is nothing to apply. Recorded because a paused review leaves this PR less reviewed than a green attestation implies — this is the configured-auto-pause shape, distinct from the quota wall (#906). |
| PR-Agent | Round-1 findings | **Already dispositioned** | See the triage comment of 01:26:53Z; unchanged since. |
| CI | `test-windows` | **Red, expected, tracked** | Failed at **45m5s** — it picked up #1475's raised ceiling and exhausted that too. Not this PR's diff: its only production change is a godoc comment. Root cause CI-001 (#1472); the fix is #1482, which takes the pi reconcile off the PR path. |

**Reviewer count: one.** PR-Agent reviewed; CodeRabbit did not. Stated plainly
rather than left behind a green `review-attestation`.

Issue #1469, opened 2026-09-04
CLI-074: a dev-stamped dotf is a terminal state with no drift check, so the binary silently cannot serve the harness config deployed alongside it


mlorentedev:
`install_dotf` refuses to overwrite a binary stamped `dev`. That refusal is
correct for a developer mid-edit and has no expiry, no drift check and no
diagnostic, so a `dev` build made once from a feature branch stays on `PATH`
indefinitely while the deployed harness config moves past it. Nothing compares
the two. The result is a binary that silently cannot serve the config it is
wired into.

## Measured, 2026-09-03

`~/.local/bin/dotf` on msi was a `dev` source build dated 2026-09-02 18:07,
built from `feat/gate-decision-record`. `main` had moved four merges past it.
Two live failures, neither of which named the cause:

**1. Every single prompt failed.** `compile-harness.sh --deploy` had written the
`UserPromptSubmit` hook from #1455:

```
"command": "/home/manu/.local/bin/dotf harness suggest --from-hook"
```

The binary predates that flag, so Claude Code reported on every prompt:

```
UserPromptSubmit hook error
Failed with non-blocking status code: Error: unknown flag: --from-hook
```

The persona suggester was therefore off for the entire life of that binary,
failing open exactly as designed — which is why nobody noticed until a human
read the error text.

**2. `dotf hooks` did not exist.** #1464 repointed both setup scripts at
`dotf hooks install`. The next `setup-linux.sh` run on this machine would have
failed at the hooks step, on a box whose `dotf` had been current when the step
was written.

Both cleared by rebuilding from `main` and installing over the top. Verified by
consequence: `dotf hooks --help` exits 0, and
`echo '{"prompt":"x"}' | dotf harness suggest --from-hook` prints
`[suggest] prompt arrived as "prompt"` and exits 0.

## Why this is a defect and not just a stale machine

The refusal is a *policy* with no *feedback path*. Three properties combine:

- `dev` is a terminal state. Nothing ever un-stamps it, so the machine most
  likely to be developing the CLI is the machine guaranteed to run the oldest
  one.
- The harness config is deployed declaratively and the binary imperatively.
  `harness/manifest.json` is the SSOT for what gets wired; nothing is the SSOT
  for what the wired binary can answer.
- The failure surfaces in the harness, not in `dotf`. The operator sees a Claude
  Code hook error naming a flag, with no hint that a stale binary is the cause
  or that `dotf doctor` would have known.

[...]

PR #1471, opened 2026-09-04
fix(harness): resolve a named dispatch through the dispatch that created it

### LINKED ISSUE #1434: HARNESS-109: a named subagent dispatch sends its NAME as agent_type, so the gate resolves no persona and enforcement silently turns off

mlorentedev:
Naming a subagent turned its own gate off.

`agent_type` carries the caller-supplied **name** when a dispatch is named, so the persona lookup missed, `loadGatePersona` failed open by its own contract, and enforcement was silently off for every call that subagent made. Under the current `warn` canary that costs a warning nobody reads. Under `enforce: block` it is a one-word opt-out, which is why every promotion on the orchestrator's roadmap sits behind this ticket.

## Scale — this is the normal case, not an edge case

Of the **274** decision records carrying an `agent_type`, **271 resolved to no persona**. Only 3 of those were the probe #1434 was filed from; the other 268 are ordinary working dispatches (`kubelab-harness` 103, `web080-adversarial` 76, `kubelab-reality` 75, `kubelab-1e` 14). The only two that ever resolved name `reviewer`, dispatched with no name.

Provenance checked rather than assumed, after a reviewer challenge: the journal's schema is uniform (no record predates #1435), and the two resolving records are *interleaved* with the unresolved ones rather than from a later binary — one population, not a fix landing. The honest limit runs the other way: every `agent_type` record falls in one 3.5-hour window on 2026-09-02. Large in calls, small in sessions, and the proposal says so.

## The ticket's first candidate fix is refuted, not deferred

#1434 proposed *"fall back to a second payload field if one carries the true type"*, flagged as needing measurement. Read out of the Claude Code 2.1.260 executable's own payload builder, the complete base field set is:

```
session_id, transcript_path, cwd, scratchpad_dir, prompt_id,
permission_mode, agent_id, agent_type, effort
```

`agent_type` is the only type-carrying field, and `agent_id` is `a<name>-<hash>` — the name again. **There is nothing to fall back to.** That direction could only ever have produced a fallback that silently never fires.

## What this does instead

The gate already runs on the **parent's** `Agent` call, whose arguments carry both the name and the true type — and `PreToolUse` is synchronous, so that hook completes before the child exists *by construction*, not by winning a race. A per-session map in the state dir turns the child's `agent_type` back into a persona.

The roster check is the other half. `general-purpose`, `Explore` and `Plan` have no persona **by design**; they were being reported as `role-unresolved` with `ENFORCEMENT IS OFF`, which is what made 271 records read as faults and left the genuine faults unfindable among them. An unnamed dispatch is recorded under its own type, so presence in the map means *"the gate witnessed this and knows what it really is"* — a witnessed non-persona is a quiet `no-role`, and only a dispatch the gate never saw stays loud. That is why no hardcoded list of built-in agents appears anywhere.

`no-role` is reused rather than adding a ninth `Outcome`: the vocabulary is closed and pinned by tests, the fact is genuinely the same one, and `role_requested` keeps the raw name.

## Verified live

One owner-authorised named dispatch, on the real hook path. #1434's own table, with the row this PR adds:

| dispatch | payload `agent_type` | `role_resolved` | outcome |
|---|---|---|---|
| `subagent_type: reviewer`, no name | `reviewer` | `reviewer` | `warn` — 4 skills |
| `subagent_type: reviewer`, `name: gate-probe-agent-type` *(before)* | `gate-probe-agent-type` | *(none)* | **`role-unresolved`** |
| `subagent_type: reviewer`, `name: harness109-probe` *(after)* | `harness109-probe` | **`reviewer`** | **`warn` — the same 4 skills** |

The shared `~/.local/bin/dotf` was backed up and **restored byte-identically** afterwards (`cmp`-confirmed), so the box is not left running a `dev` build of an unmerged branch — the trap #1469 describes. `compile-harness.sh --deploy` was deliberately **not** run, so no session's live settings file was rewritten mid-flight; the `timeout` fields go live with the next ordinary deploy.

## Also: the gate's hook was unbounded

#1455's review gave `harness suggest` `timeout: 5` because an unbounded hook on the interactive path delays what the user typed. `harness gate` runs on *every tool call* on that same path and carried none. The new guard asserts the **class** rather than the instance — and immediately found a second one nobody was looking for: agy's `BeforeTool` gate. Both now carry `timeout: 5`.

Measured in the same executable: a timed-out hook returns `blocked: false` unless `timeoutFailsClosed`, and that flag is set only for a call served to a cloud session. Locally a bound converts a stall into a fast fail-open — the gate's own contract — and cannot cause a block. Recorded as lesson 264, because it is not derivable from the documentation and the next person to bound a hook will need it.

## Verification

- `go build`, `go vet`, **`GOOS=windows go vet`** — all clean.
- `go test ./...` — all packages ok, no regressions.
- `golangci-lint run` at the pinned **2.12.2** from `versions.conf` — **0 issues**.
- All seven `features.json` commands **executed**, not written and trusted. That caught a real defect in one of them: an unanchored `--- PASS: Test` count matched indented subtests too, so the entry expected 4 and got 18.

## Out of scope, deliberately

Declaring `enforce` on the six bare personas and promoting `warn → block` — that reads the decision journal for evidence, and this PR only makes the journal worth reading. Blocking on an unresolved role stays forbidden (#1434's own "do not close by", and lesson 219). The residual `role-unresolved` population — a `Workflow` step, an agent resumed from an earlier session, an in-process teammate — is named in the proposal, not built for. A `dotf doctor` check for the unresolved rate is worth filing once this has shrunk the set to that residual.

Closes #1434


mlorentedev:
## Review triage

**Who actually reviewed**: PR-Agent (`github-actions`) and **CodeRabbit — a real review, 5 actionable comments**, not a limit notice. Disclosed because CodeRabbit posted a *"review limit reached"* notice on 4 of the last 5 PRs and one on #1475 tonight, so a green `review-attestation` has not meant "two reviewers looked" recently. Here it did.

Separately, this change carried **two rounds of independent adversarial review** from `harness/reviewer-pool.json`, neither by the implementer, each on a different model from the random draw: round 1 `nan/glm5.3-flash` → **FAIL**, round 2 `agy/gemini-3.1-pro-high` → **PASS-WITH-GAPS**. Both artifacts are committed (`fc9d5eb` preserves round 1, which the re-review would otherwise have overwritten).

### CodeRabbit

| # | Finding | Disposition | Why |
|---|---|---|---|
| 1 | `harness_gate.go`: apply the dispatch mapping **before** the persona lookup, so `name: reviewer` + `subagent_type: builder` resolves to builder | **Already applied** (`e096583`) | Independently the same defect round 1's reviewer found and reproduced. CodeRabbit reviewed the pre-fix commit. Verified against current code: `lookup := role; if !fromFlag { if trueType, witnessed := dispatched[role]; witnessed { lookup = trueType } }`, with `TestADispatchNamedAfterAnotherPersonaResolvesToItsTrueType` covering it. Two reviewers converging on it independently is the strongest signal on this PR. |
| 2 | `bind_test.go`: assert the emitted timeout **equals** `c.Timeout` and is > 0, not that the key exists | **Applied** (`a05aeab`) | Correct and I had it wrong. `strings.Contains(raw, "\"timeout\"")` passes on any value, on a value disagreeing with the manifest, and on the word appearing anywhere in the document. Same class as the `grep -c` that counted indented subtests. Now decodes and compares, and fails if the hook never reached the document at all. Mutation-checked. |
| 3 | `dispatch.go`: serialize the read-modify-write with a **cross-process lock**; temp+rename alone still loses a concurrent update | **Ticketed, not applied** | Real and correctly diagnosed — two dispatches in one turn can lose one mapping. Consequence is bounded and fails **open**: the losing subagent resolves no persona, which is the pre-change behaviour, never a block. Round 2's reviewer classified the same area Minor/THEORETICAL. The right fix is not a lock but switching to `O_APPEND` JSONL with latest-wins folding — the pattern `decision.go` already uses in this package — and that is an on-disk format change that deserves its own review rather than being slipped in after two passing rounds. Filed as a follow-up. |
| 4 | `manifest.json`: make the PreToolUse timeout **fail closed** — deny the tool action, or remove the 5s timeout | **Declined, with evidence** | This inverts the gate's entire documented contract. `dotf harness gate` fails open by design (lesson 219; #1434's own *"Do not close by"* forbids blocking on an unresolved role, because it trades a silent bypass for a session that cannot run). It is also not achievable as described: measured in the Claude Code 2.1.260 executable, a timed-out hook returns `blocked: false` unless `timeoutFailsClosed`, which is set **only** for a call served to a cloud session. Locally the timeout converts a stall into a fast fail-open, which is what the gate promises. Removing the 5s instead would restore the defect this PR fixes. Reasoning recorded in lesson 264 so it is not re-litigated. |
| 5 | `bind_test.go`: 53-line test exceeds the 40-line guideline | **Applied** (`a05aeab`) | Taken — the extraction is where the value comparison belonged anyway, so it landed with #2. |

### PR-Agent

No major issues, no security concerns, tests present; ticket compliance on #1434 read as fully compliant. **One known false positive, disclosed**: it treated **#1435 — a merged PR, not a ticket** — as a second requirement source. That is #1466 (TOOL-020), now its **10th** observed instance. No action on this PR.

### CI

All green except `test-windows`, which is **red for a reason that is not this diff**: #1472 (CI-001) — the job spends 27 of its 30 minutes on serial `npm install`s and cancelled two unrelated PRs at the ceiling on a night nothing changed. Fix is open as **#1475**, which should merge **before** this PR. Merging this one past a red required check would need disclosing; the intent is to wait for #1475 instead.

### Spec

`specs/HARNESS-109/` archived in this PR (`077cc40`), which is what the archive-on-merge gate requires and what it correctly refused the first push for.

mlorentedev:
## Review triage — re-triage at `a05aeab`

The earlier triage (02:40:33Z) is superseded: **PR-Agent updated its review in place at 02:43:01Z**, 2.5 minutes after it. `created_at` never moved, so `dotf pr triage-queue` reported this PR clear. It was not.

|---|---|---|---|
| 1 | PR-Agent | **Unrelated changes** — `ai/pi/{models,settings}.json` and `ai/pi/README.md` are unrelated to #1434/#1435/#1469; mixing them into a bug fix raises review complexity and regression risk | **Accepted as accurate. Declined as a change.** The finding is correct — this is scope creep. It is *owner-directed* scope creep: the owner hand-edited those files mid-session and asked for them in this PR. Splitting them now costs a 20–45 minute `test-windows` roll on a PR that is already green, against a queue where #1468 just exhausted 45 minutes. Recorded rather than silently accepted. |
| 2 | PR-Agent | #1434 — compliant, no non-compliant requirements; AC6's live probe needs a real dispatch and cannot be verified statically | **Accepted, no change.** The probe was run with explicit owner authorisation: a named dispatch of `reviewer` produced `role_resolved: reviewer`, where #1434's own table recorded `role-unresolved`. Evidence is in `specs/archive/HARNESS-109/verification.md`. |
| 3 | PR-Agent | #1435 — "does not change the block path", "does not fix `TestMergeAgainstTheRealDeployedSettings`" | **Declined.** Neither is in this PR's scope. The block path is explicitly out of scope while `#1434` forbids blocking on an unresolved role, and that test is #1429 — fixture leakage, green in CI, local-only. |
| 4 | PR-Agent | #1469 — the hook timeout is added but `dotf doctor` validation of the emitted entry is not implemented | **Accepted, ticketed not fixed.** Correct: this PR adds `timeout: 5` and an assertion that the emitted value matches the manifest. The doctor-side validation is #1469's own scope and stays there. |
| 5 | PR-Agent | Effort 4/5 · PR contains tests · no security concerns | **Noted.** |

### Disclosure: the head commit is unreviewed by the reviewer the attestation names

`review-attestation` reports `[OK] attested — reviewed by: coderabbitai`. CodeRabbit's only formal review is `COMMENTED` at 01:49:47Z against **`74ca6b5c`**. Head is `a05aeab3` — `ahead_by 8`, 16 files.

The unreviewed fraction is not arbitrary code. It is precisely what was written *in response to* the two adversarial rounds:

| file | at review | now | added after |
|---|---|---|---|
| `dispatch.go` | 149 | 244 | **39%** |
| `harness_dispatch_test.go` | 285 | 412 | 31% |
| `dispatch_test.go` | 172 | 267 | 36% |

**Not blocking**, and the reason is not "close enough": the independent adversarial pool covered this PR across **two rounds with a different model each time**, neither the implementer (`nan/glm5.3-flash` → FAIL, then `agy/gemini-3.1-pro-high` → PASS-WITH-GAPS), and the CodeRabbit review that *did* run found the boundedness defect `a05aeab` fixes. But the fix for a reviewer's finding going back unreviewed by that reviewer is disclosed here and belongs in the merge note, not assumed away.

Tracked on #906. The defect is **staleness only** — `review-attestation` classifies notice-vs-review correctly (proven on #1480, where it skipped CodeRabbit's rate-limit notice and named PR-Agent). It just never compares the attested review against head.

PR #1475, opened 2026-09-04
ci: cache npm for the Windows setup run, which spent 27 of its 30 minutes there

### LINKED ISSUE #1472: CI-001: test-windows spends 27 of its 30 minutes on serial npm installs, so a required check now blocks every PR touching cli/
`test-windows` reconciles the pi package manifest by running `npm install` once
per package, serially, inside the job. On the Windows runner each install
currently takes between 54 seconds and 7 minutes, so the step alone consumes
~25 of the job's 30-minute budget and the job is cancelled before it reaches
the doctor gate, the Pester suites or the bats subset.

This is a **required status check on `main`**, so while it is in this state no
PR touching `cli/` can merge.

## Measured, 2026-09-04

Two independent PRs, different branches, different authors, both off
`eaf8a91`, both cancelled at the 30-minute ceiling in step 7
(`Run setup-windows.ps1 end-to-end`):

- #1468, run `33822719040` — cancelled at 30m06s, re-run cancelled at 30m07s.
- #1471, run `33826326013` — same step, past 22 minutes at the time of writing.

The log from #1468's second attempt gives the breakdown. Timestamps are the
start of each package install:

| Package | Duration |
|---|---|
| `npm:pi-effort@0.0.8` | 54s |
| `npm:pi-web-access@0.24.2` | 93s |
| `npm:@ayulab/pi-rewind@0.4.6` | **421s** |
| `npm:@narumitw/pi-plan-mode@0.53.0` | 134s |
| `npm:@narumitw/pi-goal@0.53.1` | 200s |
| `npm:pi-subagents@0.56.0` | 153s |
| `npm:pi-memory@0.4.2` | 251s |
| `npm:pi-cc-extensions@0.8.67` | cancelled at the ceiling |

Everything before this step is fast: the whole run reaches "Installing pi
packages" by 01:31:53, about 2m30s in. The remaining 27 minutes are npm.

## This is a margin defect, not a regression

`ai/pi/packages.json` has not changed. The manifest landed in #1226 and its last
touch was unrelated. Nothing in either PR's diff goes near it — #1468's
production change is a single godoc comment, and `test-windows` never runs
`go test` at all (its steps are: build `dotf`, run setup, doctor gate, Pester,
bats subset), so a Go test file cannot influence it.

`test-windows` passed on #1464 at **9m03s** and on #1462 at **10m24s**. The
same step, the same package list. The difference is npm registry latency, which
is not a property of the repository. The job has therefore always been one slow
registry night away from failing, and this is that night.

Worth stating plainly because it shaped the diagnosis: **`test-windows` is
`skipped` on `main`** under the path filter — it was skipped on both post-merge
runs for #1464 and #1465. So "main is green" is not evidence this job passes,
and the only place it genuinely runs is a PR. A latent failure here is invisible
until it blocks someone.

## Options

1. **Cache the pi package installs.** The manifest is version-pinned, so the
   install set is a pure function of `ai/pi/packages.json`. An `actions/cache`
   keyed on that file's hash turns a 25-minute step into a restore on every run
   where the manifest has not changed. Highest value, and it makes the job
   faster on Linux too.
2. **Install the packages concurrently** rather than serially. Cheaper to write,
   but it trades a deterministic slow step for a nondeterministic one, and npm
   concurrency on Windows has its own failure modes.
3. **Skip pi package reconciliation in CI**, gated on an env var, and assert the
   manifest is well-formed instead of installing from it. This tests less; the
   installs are exercising real npm behaviour that has broken before.
4. **Raise the job timeout.** Not recommended alone — it buys headroom without
   addressing that a 30-minute job spends 27 minutes in a package manager, and
   the next slow night moves the ceiling again.

Recommend option 1, with option 4 as a small companion so a cache miss on a slow
night still completes.

## Immediate consequence

#1468 and #1471 are both blocked on a required check that neither can turn
green. Whatever is done here, merging either PR before this is fixed means
merging past a red required check, which should be disclosed on the PR rather
than done quietly.

mlorentedev:
`test-windows` reconciles the pi package manifest with one `pi install` per package, **serially**, and each shells out to npm. On 2026-09-04 those eight installs took 54s to 7 minutes apiece and consumed **27 of the job's 30 minutes**, cancelling two unrelated PRs (#1468, #1471) at the ceiling.

Neither diff went near `ai/pi/packages.json`. The variable is npm registry latency, which is not a property of this repository — the job passed at 9m03s on #1464 and 10m24s on #1462 with the identical package list, so it had always been one slow night away from failing. It is a **required check**, so while it is in this state nothing touching `cli/` can merge.

## Measured before building, because the fix rested on an unchecked premise

The premise is that `pi install` uses the shared npm cache at all — if pi ran its own store or passed `--cache`, a warm npm cache would buy nothing and the job would still die. Read out of pi's own `dist/core/package-manager.js`:

```js
async installNpm(source, scope, temporary) {
    const installRoot = this.getNpmInstallRoot(scope, temporary);
    this.ensureNpmProject(installRoot);
    await this.runNpmCommand(this.getNpmInstallArgs([source.spec], installRoot));
}
// getNpmInstallArgs -> ["install", <spec>, "--prefix", installRoot, "--legacy-peer-deps"]
// getNpmCommand    -> { command: "npm", args: [] }   (unless a settings override exists)
```

Plain `npm`, no `--cache` anywhere; `--prefix` moves the install *destination*, not the cache. The shared cache is the right layer.

## The obvious fix does not work, and that is worth recording

Caching **pi's own** npm directory keyed on the manifest hash was the first proposal. It would have cached the artifacts and still paid the 27 minutes: `setup-windows.ps1:1305-1333` skips a package only when it is listed in the **live** `~/.pi/agent/settings.json`, which is seeded fresh on every runner, so `$piPresent` is empty and all eight reinstall no matter what is on disk. Caching npm's cache leaves the reconcile loop honestly running eight installs and makes each one a local unpack.

## What this does

- Asks npm for its cache directory rather than hardcoding one. A hardcoded path that stops being right caches nothing *while still reporting success* — and a green job would hide it.
- `actions/cache` keyed on `hashFiles('ai/pi/packages.json')`. Every entry in that manifest carries a pinned version (`tests/pi-config.bats` refuses an unpinned one), so the same manifest always resolves to the same tarballs. A manifest edit misses and pays full price once, which is correct — that run installs something genuinely new.
- `npm_config_prefer_offline` on the setup step. This is what makes the cache *decide* anything: a warm cache alone still lets npm round-trip the registry for metadata on every package. Set on the CI step and **not** in `setup-windows.ps1` — this is a property of a throwaway runner, not of a developer's machine, and ADR-020 C7 keeps the bootstrap scripts out of it.
- `timeout-minutes` 30 → 45 as the ticket's recommended **companion**, not the fix: it stops a cold cache on a slow night cancelling before the doctor gate, Pester and bats run, and a cancelled run tells you nothing about any of them.

## Not claimed

That every package becomes fast. `@ayulab/pi-rewind` took 421s against 54s for `pi-effort`, and package installs do **not** pass `--ignore-scripts` — only pi's self-update path does — so part of that spread may be a postinstall the cache cannot touch. **The per-package timings in the next run are the check**, and they are the difference between "the job went green" and "the job went green because npm was fast again". That verification is owed on this PR before it is called done.

Also rejected, per the ticket: seeding `settings.json` with the package list so `$piPresent` is populated. It would make the job fast immediately and turn the reconcile loop into a no-op that installs nothing and asserts nothing — a pass that means less than the red it replaced.

## SDD skip rationale

The functional change is 12 lines of workflow YAML; the gate counts 59 because the reasoning above is written into the file as comments, which is this repository's house style and not something to strip to get under a threshold. There is no public contract, no new dependency, and no multi-PR sequence — it is a CI-runner margin defect with a measured cause, a filed ticket (#1472) and a single-file fix. A `specs/` folder here would restate the ticket and the commit message, which is the "duplicated true rule" shape ADR work has already been bitten by.

Closes #1472

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -249,7 +249,21 @@ jobs:
 
   test-windows:
     runs-on: windows-latest
-    if: github.event_name == 'pull_request'
+    # RUNS ON main TOO, and that is half of CI-001's fix rather than a cost.
+    #
+    # A GitHub Actions cache is readable from the branch that wrote it, that
+    # branch's base, and the DEFAULT branch — never from a sibling PR. So a
+    # cache populated only on PR branches warms nothing for the next PR, and the
+    # step below would have helped re-runs of one PR and nobody else. Populating
+    # it on main is what makes every PR start warm.
+    #
+    # It also closes the reason this defect stayed invisible. The job was
+    # `skipped` on every push to main — measured, 6 of the last 6 — so "main is
+    # green" was never evidence about it, and the only place it genuinely ran
+    # was a PR. A required check that accumulates no evidence on the default
+    # branch is a latent failure with nowhere to show up until it blocks
+    # someone, which is exactly how this arrived. `integration`, the other heavy
+    # job, already runs on both for the same reason.
     # CI-001 (#1472): 45 rather than 30 as the COMPANION to the npm cache below,
     # not as the fix. The cache removes the cost on a hit; this stops a cold
     # cache on a slow registry night from cancelling the job before it reaches
@@ -264,7 +278,7 @@ jobs:
       - uses: actions/checkout@v7
 
       - name: Install age (deterministic release + registry PATH — BUG-025)
-        if: needs.changes.outputs.code == 'true'
+        if: github.event_name == 'push' || needs.changes.outputs.code == 'true'
         shell: pwsh
         run: |
           $ver = ((Get-Content versions.conf | Select-String '^AGE_VERSION=').Line -split '=', 2)[1]
@@ -284,7 +298,7 @@ jobs:
           & (Join-Path $ageDir 'age-keygen.exe') --version
 
       - name: Sandbox secrets + minimal vault
-        if: needs.changes.outputs.code == 'true'
+        if: github.event_name == 'push' || needs.changes.outputs.code == 'true'
         shell: pwsh
         run: |
           New-Item -ItemType Directory -Force "$env:USERPROFILE\.config\age" | Out-Null
@@ -308,13 +322,13 @@ jobs:
       # tree; the integration container builds from source for the same
       # reason) and make it reachable by effect before setup runs.
       - uses: actions/setup-go@v7
-        if: needs.changes.outputs.code == 'true'
+        if: github.event_name == 'push' || needs.changes.outputs.code == 'true'
         with:
           go-version-file: cli/go.mod
           cache-dependency-path: cli/go.sum
 
       - name: Build dotf from this PR and put ~/.local/bin on PATH (TEST-003)
-        if: needs.changes.outputs.code == 'true'
+        if: github.event_name == 'push' || needs.changes.outputs.code == 'true'
         shell: pwsh
         run: |
           $bin = "$env:USERPROFILE\.local\bin"
@@ -353,7 +367,7 @@ jobs:
       # being right caches nothing while still reporting success — the failure
       # this repository keeps cataloguing, and one a green job would hide.
       - name: Resolve npm's cache directory
-        if: needs.changes.outputs.code == 'true'
+        if: github.event_name == 'push' || needs.changes.outputs.code == 'true'
         id: npmcache
         shell: pwsh
         run: |
@@ -364,7 +378,7 @@ jobs:
           Write-Host "npm cache: $dir"
 
       - uses: actions/cache@v4
-        if: needs.changes.outputs.code == 'true'
+        if: github.event_name == 'push' || needs.changes.outputs.code == 'true'
         with:
           path: ${{ steps.npmcache.outputs.dir }}
           key: npm-pi-${{ runner.os }}-${{ hashFiles('ai/pi/packages.json') }}
@@ -384,7 +398,7 @@ jobs:
       # property of a throwaway runner, not of a developer's machine, and the
       # bootstrap scripts stay out of it (ADR-020 C7).
       - name: Run setup-windows.ps1 end-to-end (PS 5.1 -> BUG-005 re-exec)
-        if: needs.changes.outputs.code == 'true'
+        if: github.event_name == 'push' || needs.changes.outputs.code == 'true'
         shell: powershell
         env:
           DOTFILES_REPO_DIR: ${{ github.workspace }}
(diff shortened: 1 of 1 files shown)


mlorentedev:
## Review triage

**Who actually reviewed**: PR-Agent (`github-actions`) only. **CodeRabbit did NOT review this PR** — it posted a *"Review limit reached / You've used all free OSS reviews for now"* notice. That is a notice about a review, not a review, and this PR is proceeding with one reviewer. Disclosed rather than implied, because the green `review-attestation` cannot tell the two apart (#906).

### PR-Agent

No blocking findings. Two items were raised as *"requires further human verification"* and **both are correct — they are the same gap this PR's own description names, and neither is answerable before merge**:

| # | Item | Disposition |
|---|---|---|
| 1 | "Verify the cache actually warms on a main push and that subsequent PR runs hit it (requires observing a real CI cycle after merge)" | **Accepted as an owed verification, not resolved.** This is precisely the point: `test-windows` was `skipped` on 6 of the last 6 main runs, so nothing about this can be observed until it merges and runs there once. Checking it is a post-merge obligation on this PR, not a claim it already makes. |
| 2 | "Confirm the 45-minute timeout is sufficient when the cache misses on a genuinely new manifest" | **Accepted, unresolved, and measurable now.** This PR's own `test-windows` run IS a cold-cache run — the key never existed — so its duration is the answer to exactly this question. |

**One known false positive, disclosed**: PR-Agent read **#1468 — a pull request, not a ticket** — as a second requirement source and evaluated this PR against it. Tenth observed instance; tracked as #1466 (TOOL-020). No action here.

### What this PR does NOT claim

Stated in the description and repeated here because it is the thing most likely to be read as proven:

- **This PR's own run is not evidence the fix works.** The cache is cold; this run populates it. A green `test-windows` here means the 30 → 45 timeout stopped the cancellation, not that the cache did anything.
- **It is not claimed that every package becomes fast.** `@ayulab/pi-rewind` took 421s against 54s for `pi-effort`, and package installs do not pass `--ignore-scripts` (only pi's self-update path does), so part of that spread may be a postinstall the cache cannot touch. The per-package timings on the first warm run are the check.

### Self-reported defect, found after opening

`6c85af7` alone would not have helped any PR but this one. A GitHub Actions cache is readable from the writing branch, its base, and the **default** branch — never from a sibling PR — and `test-windows` was `if: pull_request`, so the cache would only ever have been written where no other PR could read it. Fixed in `e28d282` by running the job on `main` too. Recorded here rather than silently amended, because the corrected commit is the one that makes the first one do anything.

### Merge order

This PR should merge **first**: while `test-windows` is red, nothing touching `cli/` can merge, which currently blocks #1468 and #1471. Then #1468, then #1471, then #1419 (release) last.

Issue #1478, opened 2026-09-04
CI-002: every heavy job gates on one path filter, so a Go-only PR pays 15 minutes reinstalling pi packages it never touched


mlorentedev:
## Why

Every job in `ci.yml` that costs real time gates on the **same** path filter, so any
change to almost anything runs all of them.

There are exactly two filters in the whole workflow. `powershell`, and `code` — and
`code` matches `ai/**`, `cli/**`, `scripts/**`, `powershell/**`, `windows/**`,
`tests/**`, `.github/**`, `harness/**`, `sensitive/**`, the dotfiles themselves,
`versions.conf`, `packages.json`, both setup scripts and more. `lint`, `test`,
`test-windows` and `integration` all gate on that one output.

That is not a defect in HARNESS-041 (#552), which introduced it: its question was
*"does this diff need CI at all?"*, and one bucket answers that correctly. This ticket
asks the next question — *"which parts of CI does this diff need?"* — which the single
bucket cannot answer.

### Measured

- **#1471** is a pure Go change under `cli/`. It touches no shell script, no
  PowerShell, nothing under `ai/`. It ran `test-windows` for **23m2s** and the Docker
  `integration` build. Both were guaranteed no-ops for that diff.
- **`test-windows` costs 22–23 minutes**, of which the pi package reconcile is
  **883s of 1348s (~66%)** — nine serial `pi install` calls, one of which
  (`npm:@ayulab/pi-rewind@0.4.6`) takes **422s on its own**.
- The reconcile's result is a function of `ai/pi/packages.json` and the reconcile code
  in the setup twins. **Nothing else can change it** — every entry is version-pinned
  and `tests/pi-config.bats` refuses an unpinned one.

So the repository pays ~15 minutes per PR to reinstall nine packages whose manifest
the PR did not touch, because the filter cannot distinguish "you edited Go" from
"you edited the pi manifest".

## What

1. **Granular filters.** Split `code` into filters that name what actually invalidates
   each job (`go`, `shell`, `pi`, `harness`, `ci`, …) and gate each job on the filters
   it genuinely depends on. Keep every required job *reporting* — the
   required-check-safe shape HARNESS-041 established (jobs run, heavy steps skip) is
   the constraint, not something to revisit.

2. **`DOTFILES_SKIP_PI_PACKAGES`, honoured by both setup twins.** Skips the pi package
   reconcile, logging the reason — never silently. CI sets it on `test-windows` for
[...]

PR #1480, opened 2026-09-04
ci: drop the npm cache from test-windows — no measured benefit, real cost

mlorentedev:
> **Correction, 03:55Z.** An earlier version of this PR claimed the cache made the job 2.6x slower. **That was wrong** and is retracted below. The cache is removed for buying nothing, not for causing harm — different justifications, different urgency. How it was wrong is recorded because it is the same error this PR's own ticket keeps cataloguing.

Removes the npm cache added in #1475. #1475 merged on the cold run's green; the runs since refute the hypothesis it was built on.

## What is measured

Three runs of the same workflow, same manifest hash `2890f0f4`:

| run | restore | package phase |
|---|---|---|
| #1475 PR, cold | miss on the key **and** the `npm-pi-Windows-` prefix | **883s** |
| **main, COLD** | miss on the key **and** the prefix | **2201s** |
| #1475 PR, warm | confirmed hit, 325 MB in 8s | **2269s** |

**Two cold runs differ by 2.5x.** The warm run sits alongside the slower cold one, not beyond it. So the cache's effect is **not separable from run-to-run variance**, and the earlier "2.6x slower with a confirmed hit" was n=1 against n=1 across a spread nobody had measured.

What the data does support: a confirmed 325 MB hit produced **no benefit**, so **the time is not on the download leg**.

## Why remove it then

- No measured benefit, on the only measurement that could show one.
- It restores and saves ~310 MB every run, and holds 325 MB of the repo's Actions quota per manifest hash — evicting caches that do work.
- An inert step reads as *"npm caching is handled here"*, which stops the next person looking. That is the expensive part.

## The actual finding

Per-package on **main's cold** run:

| package | s | | package | s |
|---|---|---|---|---|
| `pi-effort` | **421.4** | | `pi-subagents` | 132.2 |
| `pi-web-access` | 63.5 | | `pi-memory` | **421.5** |
| `@ayulab/pi-rewind` | 122.4 | | `pi-cc-extensions` | **422.5** |
| `@narumitw/pi-plan-mode` | 53.5 | | `pi-add-dir` | 158.2 |
| `@narumitw/pi-goal` | 405.7 | | | |

Every observation of that band, across every run:

```
pi-rewind 421s | pi-rewind 422s | pi-plan-mode 422s
pi-effort 421.4s | pi-memory 421.5s | pi-cc-extensions 422.5s
```

**Six observations at 421 ±1s across five different packages.** That is a fixed ~7-minute timeout landing on whichever install happens to hit it, and **how many hit it per run** is what moves the phase between 883s and 2201s — one on the first cold run, four on main's.

There is no slow package. `@ayulab/pi-rewind` was believed to be a deterministic 7-minute outlier on two samples that agreed; its other three samples are 122s, 269s and 422s. `"27 of 30 minutes"`, `"22–23 minutes"` and `"43 minutes"` are the same job with a different number of timeouts in it.

420s is suspiciously round. Identifying it needs the `pi install` redirect gone — both setup twins discard its stdout **and** stderr on the package loop, which is why six identical 421s and nothing to explain them is all anyone has.

## What changes

18 production lines, all of them the cache: the resolve step, `actions/cache@v4`, `npm_config_prefer_offline`. The rest of the diff is the workflow comment, rewritten to record the above.

Untouched from #1475, and never dependent on the cache being right:

- `test-windows` still runs on pushes to `main` — it was `skipped` on 6 of the last 6, so "main is green" was never evidence about it.
- `timeout-minutes: 45` stays and stays temporary. It converts a cancelled job into a slow one; it is a loan against a cost nobody has diagnosed.

## Not closed by this

**#1472 (CI-001) stays open.** Taking the reconcile off the PR path is **#1478**, whose task one is now removing the redirect — that is the difference between guessing at the timeout and reading its error text.

## SDD skip rationale

Removal of a CI-only step with no measured benefit. 18 production lines, all in `.github/workflows/ci.yml`; the remaining diff is the comment recording the measurement. No behaviour outside the workflow changes.

Refs #1472, #1478

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -275,29 +275,35 @@ jobs:
 
   test-windows:
     runs-on: windows-latest
-    # RUNS ON main TOO, and that is half of CI-001's fix rather than a cost.
+    # RUNS ON main TOO, because a required check that accumulates no evidence on
+    # the default branch is a latent failure with nowhere to show up until it
+    # blocks someone.
     #
-    # A GitHub Actions cache is readable from the branch that wrote it, that
-    # branch's base, and the DEFAULT branch — never from a sibling PR. So a
-    # cache populated only on PR branches warms nothing for the next PR, and the
-    # step below would have helped re-runs of one PR and nobody else. Populating
-    # it on main is what makes every PR start warm.
+    # The job was `skipped` on every push to main — measured, 6 of the last 6 —
+    # so "main is green" was never evidence about it, and the only place it
+    # genuinely ran was a PR. That is exactly how CI-001 arrived: a job nobody
+    # could see degrading until it cancelled two unrelated PRs at the ceiling.
+    # `integration`, the other heavy job, already runs on both for this reason.
     #
-    # It also closes the reason this defect stayed invisible. The job was
-    # `skipped` on every push to main — measured, 6 of the last 6 — so "main is
-    # green" was never evidence about it, and the only place it genuinely ran
-    # was a PR. A required check that accumulates no evidence on the default
-    # branch is a latent failure with nowhere to show up until it blocks
-    # someone, which is exactly how this arrived. `integration`, the other heavy
-    # job, already runs on both for the same reason.
-    # CI-001 (#1472): 45 rather than 30 as the COMPANION to the npm cache below,
-    # not as the fix. The cache removes the cost on a hit; this stops a cold
-    # cache on a slow registry night from cancelling the job before it reaches
-    # the doctor gate, the Pester suites or the bats subset — which is what
-    # happened to two unrelated PRs, and a cancelled run tells you nothing about
-    # either. Raising it alone would have been option 4 on the ticket, and was
-    # rejected there: a job that spends 27 of 30 minutes in a package manager
-    # does not get better by being given 45.
+    # CI-001 (#1472): 45 rather than 30 — and READ WHAT THIS DOES NOT DO.
+    #
+    # It fixes nothing. It converts a cancelled job into a slow one. The work
+    # takes exactly as long as it did; the ceiling merely stopped catching it.
+    # What the raise buys is the ability to OBSERVE the problem instead of
+    # having it reported as a failure of whichever PR paid for it — and a
+    # cancelled run tells you nothing about that PR.
+    #
+    # The ticket rejected raising the ceiling as "option 4" and that rejection
+    # stands: CI-001 does not close here. Measured on this branch, the pi
+    # package reconcile is ~66% of the setup step and ~57% of the job, and its
+    # cost tracks npm registry latency, which is not a property of this
+    # repository. Two runs of the SAME commit spent 1175s and >1900s in that
+    # step. The second crossed 30 minutes — not for a defect, for a slow night.
+    #
+    # So this is a margin on an undiagnosed and unbounded cost, which is a loan,
+    # not headroom: the next slow night eats 45 the way this one ate 30. It is
+    # deliberately temporary and it is not a substitute for taking the reconcile
+    # off the blocking path. Lower it again when that lands.
     timeout-minutes: 45
     needs: [changes]
     steps:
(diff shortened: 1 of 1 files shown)


mlorentedev:
## Fourth cold sample

Pulled from #1419's `test-windows` (job `100895294846`) — cold, confirmed by `Cache not found for input keys: npm-pi-Windows-2890f0f4…, npm-pi-Windows-` and it created the cache at 03:27:50.

**Package phase: 1331s.** So the cold spread is now measured rather than asserted:

| run | cache | phase |
|---|---|---|
| #1475 PR | cold | **883s** |
| #1419 | cold | **1331s** |
| main | cold | **2201s** |
| #1475 PR re-run | **warm hit** | 2269s |

Three cold runs of the same nine pinned packages span **883s–2201s (2.5x)**. The one warm run is 2269s — just above the cold maximum, and nowhere near separable from that spread at n=1. The retraction in the PR body stands and is now better supported than when I wrote it.

## Seventh observation of the 421s band

`@narumitw/pi-plan-mode` — **422.0s** in this run.

```
pi-rewind 421s | pi-rewind 422s | pi-plan-mode 422s | pi-plan-mode 422.0s
pi-effort 421.4s | pi-memory 421.5s | pi-cc-extensions 422.5s
```

**Seven observations at 421 ±1s across five different packages.** Package-independence is no longer an inference from three points.

## One thing I am not claiming

Earlier I suggested the hit-count alone drives the phase. It does not, and the fourth sample is what shows it: this run had **one** hit and still took 1331s against the 883s run's **one** hit, and the warm run had one hit and took 2269s. The baseline cost of the non-timeout installs also moves between runs (`pi-cc-extensions` 37s / 227s / 252s / 422s across the four).

So: the ~421s timeout is real, fixed, and belongs to the operation rather than any package — and it is **one** contributor to the spread, not the explanation of it. Reading its error text needs the `pi install` redirect removed (#1478, task one).

mlorentedev:
## Review triage

Reviewed at `3ab9af2` (head). Two reviewers produced output; **one of them did not review.**

|---|---|---|---|
| 1 | PR-Agent | **#1475 "partially compliant"** — the cache/prefer-offline/resolve steps are reverted | **Declined.** That is the PR's purpose, and PR-Agent says so itself ("intentionally reverts… based on measured evidence"). Nothing to apply. |
| 2 | PR-Agent | **#1472 "partially compliant"** — the core problem is not addressed; the job still pays full cost | **Accepted, no change.** Correct, and already stated in the PR body under *Not closed by this*: **#1472 stays open**, the reconcile comes off the PR path under #1478. Agreement, not a gap. |
| 3 | PR-Agent | **#1478 "partially compliant"** — no granular filters, no `DOTFILES_SKIP_PI_PACKAGES`, timeout still 45 | **Declined.** #1478 is a separate ticket with its own spec; implementing it here would be scope creep. PR-Agent itself calls the sequencing "right". |
| 4 | PR-Agent | Estimated effort 1/5 · no security concerns · no major issues · **"No relevant tests"** | **Accepted, no change.** Correct: the diff is a workflow step removal plus its comment. There is no unit test for "a cache step is absent"; the evidence is the four measured runs in the body and the comment. |
| 5 | CodeRabbit | **`> [!WARNING] Review limit reached — you've used all free OSS reviews for now.`** | **No review ran.** Disclosed, not applied — there is nothing to apply. See below. |

### Disclosure: one reviewer, not two

CodeRabbit's check reports `pass` with the text **"Review rate limited"**. That is a *notice about the review*, not a review: it names no file, no line and no claim. So this PR has been reviewed by **PR-Agent only**.

`review-attestation` handled this **correctly** and it is worth recording, because it is the counter-example to the failure on #1471: it attested `reviewed by: github-actions (comment-shaped review: "## PR Reviewer Guide")` — it skipped CodeRabbit's notice and named the reviewer that actually produced a review. Notice-vs-review classification by content works. What it does not check is whether the attested review covers the *head commit*, which is the separate defect on #906.

### Known reviewer defect, 12th instance

PR-Agent's compliance section treats **#1475 — a pull request — as a ticket** to satisfy. Tracked; not a finding against this PR.

### Not triaged here

Nothing. The four measured runs, the retraction of the "cache is harmful" claim and the seven `421 ±1s` observations are in the PR body and the comment above, all produced this session.

mlorentedev:
## Rebased onto `91e3834` (#1482 merged) — conflict resolved, and this PR is now the owed measurement

#1482 landed the CI-002 reconcile skip, which touches the same `test-windows` block this PR edits. Rebased and resolved by hand; force-pushed as `b314de8`.

**Resolution: both changes kept, neither weakened.**

| | |
|---|---|
| removed | the resolve step, `actions/cache@v4`, `npm_config_prefer_offline` — 18 production lines, all of them the cache |
| kept from `main` | `DOTFILES_SKIP_PI_PACKAGES`, the `pi` filter, the never-skip-on-push guard |

`npm_config_prefer_offline` went with the cache and the comment now says why: **it existed only to make that cache decide anything, and against no cache it is a flag with nothing to prefer.** Leaving it would have been a setting whose entire rationale had just been deleted.

The merged comment also corrects two numbers that were stale by the time #1482 landed — **four** runs, not three (883s / 1331s / 2201s / 2269s), and **seven** observations of the 421 ±1s band, not six (`@narumitw/pi-plan-mode` at 422.0s from #1419's cold run).

## This PR is now the measurement CI-002 owes

Its diff is **only** `.github/workflows/ci.yml`. That matches the `code` filter and **not** the `pi` filter — so the skip applies to this PR's own `test-windows` run.

Prediction, recorded before the run finishes rather than after: **~8 minutes**, against 22–28 minutes for the same job before #1482 (this PR's previous run: 37m36s; #1482's own: 28m4s; #1471's: 23m2s; #1468's: cancelled at 45m).

If it does not come in around 8 minutes, the skip is not doing what its own tests say it does, and that matters more than this PR does.

Verified locally after the rebase: YAML parses, `tests/pi-packages.bats` 23/23.

mlorentedev:
## Prediction resolved: 7m24s, and the verdict is the absence of lines

I predicted ~8 minutes on this PR before the run finished. Result: **7m24s**, against **37m36s** for the same PR on the same single file before `91e3834`.

But the number is the weakest of the three signals, so the pass is on the other two:

```
06:08:56Z   DOTFILES_SKIP_PI_PACKAGES: 1
06:13:05Z   [WARNING] DOTFILES_SKIP_PI_PACKAGES set - pi package reconcile
            skipped (nothing installed, nothing verified)
```

and **zero** `Installing pi package` lines. The variable reached the script, the guard fired, the block did not run. A fast total with neither string present would have meant the setup script never reached the block — a different bug wearing the same green.

Full acceptance write-up on #1478, including the caveat that this PR is *skip + cache removal* rather than a clean control for the skip alone.

## State

Rebased onto `91e3834`, conflict resolved by hand keeping both changes, `MERGEABLE`, all checks green. Triaged. Reviewer count on this PR remains **one** — PR-Agent; CodeRabbit's only output here was its rate-limit notice.

PR #1474, opened 2026-09-04
docs(lessons): record the correct-measurement-wrong-question shape, and land an orphaned regen

mlorentedev:
Docs-only. Touches no `cli/` and no shell, so it should not trip #1472's `test-windows` wall — the path filter reduced #1465's run of that job to 12s.

**Lesson 265** records three near-misses from one session as a single failure. The genus: *the verification succeeds*, and returns a plausible value that answers a question nobody asked.

| Measured | Believed to mean | Actually meant |
|---|---|---|
| `tail` exited 0 | the suite passed | `tail` ran |
| `test (windows-latest)` green | Windows CI is healthy | one 1m25s Go job passed |
| a tracked file changed unbidden | a tool is writing it | a person is writing it |

The first happened hours after #1465 added that exact row to the prohibited-patterns table, which is the point: a rule that is read is not a rule that is followed. The third cost two agents half an hour of correct measurements and mutually refuted mechanisms, because neither tested the framing — only the mechanisms inside it.

Cross-links lesson 260 as the same genus at a different altitude.

**Also lands an orphan.** `harness/skills/handoff/SKILL.md` was sitting modified in the main checkout with its vault source (`00_meta/skills/handoff/SKILL.md`) already committed and no PR to carry the regeneration. Content is the deploy-path wording going agent-agnostic; `generated_sha` moves with it.

**Possible trivial conflict:** `docs/lessons/_index.md` gains one row here and one in #1471 (lesson 264). Whichever merges second resolves one line.

PR #1482, opened 2026-09-04
ci: take the pi package reconcile off the PR path

mlorentedev:
**PR 1 of CI-002.** Takes the pi package reconcile off the pull-request path.

## Why now

`test-windows` is a required check whose cost is **unbounded**, not merely high:

| PR | `test-windows` |
|---|---|
| **#1468** | **fail — 45m5s, cancelled** (02:58:19 → 03:43:24) |
| #1480 | pending |
| #1474 | pending, behind |
| #1471 | pass 23m2s, behind |

Four PRs open, none mergeable. The 30-minute ceiling cancelled two unrelated PRs; it was raised to 45 and #1468 exhausted 45 as well. **Raising the ceiling cannot win against this** — it only postpones the loss.

The cost is the reconcile: nine serial `pi install` calls, measured across four runs of the same nine **version-pinned** packages.

```
883s   1331s   2201s   2269s
```

#1471 — a Go-only change under `cli/` — paid 23m2s of it for a manifest it never touched.

So this is not a slow job being made faster. It is an **unbounded** one being taken off the blocking path.

## What changes

**`DOTFILES_SKIP_PI_PACKAGES`**, honoured by both twins as the **first** branch of the guard chain — before the `pi`/`npm`/`jq` probes, because the runner sets it precisely so the expensive block never starts. It logs what was **not verified**, not merely that it skipped: a line reading *"skipped"* is indistinguishable from *"fine"* to someone scanning a green log.

**A `pi` filter** in `changes` covering everything that can change what the reconcile does — the manifest and both twins, listed **whole**. Nothing else can change it: every entry is pinned and `tests/pi-config.bats` refuses an unpinned one, so the same manifest always resolves to the same install set. The twins are whole rather than line-scoped because a filter cannot see line ranges, and a guard that silently under-matches is worse than one that runs too often.

**One `env:` line** on `test-windows`, set when the event is `pull_request` **and** the filter is false.

It is an environment variable and not a config key or a `packages.json` entry: a durable switch would let a real machine end up permanently unconverged, which is the opposite of what setup is for.

## The design's weak seam, stated rather than assumed

**After this, the reconcile runs only on pushes to `main`** — so `main` becomes the only place a reconcile regression can surface, and the guard's value rests entirely on **a red `main` being noticed**. That sentence is in the workflow comment, not just here.

A scheduled run was considered and **rejected for exactly this reason**: it reproduces the property that made CI-001 invisible, where the evidence exists and nobody encounters it. `test-windows` was `skipped` on 6 of the last 6 main runs before #1475, which is how a required check degraded with nobody able to see it.

This is also a **real coverage reduction, not a relocation**. TEST-003 exists so a PR proves setup works end to end on Windows. Detection moves from pre-merge to the next push to `main`. Recorded as a decision with its latency named.

## Tests: mutated, not merely run

Five new assertions in `tests/pi-packages.bats` (17 → 22). Each was shown to **fail against a mutated tree**:

| Mutation | Caught by |
|---|---|
| the skip fires on `push` too | AC4 |
| the `pi` filter loses `setup-windows.ps1` | AC5 |
| the Linux guard moves below the `$PI_BIN` probe | AC2 |
| the Windows guard moves below the `Get-Command pi` probe | AC2 |
| the loud message becomes a bare "skipped" | AC3 |

**That pass found two defects in my own tests**, both green-for-the-wrong-reason and neither visible from a passing run:

1. **`grep -n X f \| grep -v '^ *#'` does not filter comments.** `grep -n` prefixes a line number, so every line starts with a digit and `^ *#` never matches. The ordering assertion was anchoring on the explanatory comment *above* the block — which precedes everything — so it passed wherever the guard actually sat.
2. **`[ ! -x "$PI_BIN" ]` also appears ~150 lines earlier**, in pi's own install block, so a file-wide search compared against the wrong branch and the corrected assertion then failed on the *clean* tree.

Both fixed by slicing the reconcile block with `awk` before comparing.

## Verified in this session

```
bash -n / zsh -n setup-linux.sh    OK          shellcheck   15 pre-existing infos, 0 new
yaml.safe_load(ci.yml)             OK          bats tests/*.bats   1541 tests, exit 0
bats tests/pi-packages.bats        22/22       features.json f1..f5  5/5, each executed
```

`pwsh` is absent on this machine, so the PowerShell twin's syntax is covered by CI's `lint-powershell`.

## Not verified, and owed rather than claimed

- **The Linux guard end to end.** `integration`'s container has no npm, so `setup-linux.sh`'s reconcile block has **never executed in CI at all** — it logs `npm not found` and stops. Pre-existing gap, recorded in the proposal, not introduced or fixed here. The Linux guard is verified structurally and by mutation only.
- **The effect on job duration.** Measured by the first `pull_request` run after this lands.

## Not closed by this

**#1472 (CI-001) stays open.** Why an install costs ~421s is unchanged and unknown: seven observations at **421 ±1s across five different packages** say it is a fixed timeout landing on whichever install hits it, and how many hit it per run is what moves the phase between 883s and 2201s. Reading its error text needs the `pi install` redirect removed from both twins — task one of the follow-up.

The wider filter audit is **PR 2** and stays separate: it re-gates four jobs at once on a workflow whose only end-to-end verification is the job we are trying not to run.

Refs #1478, #1472

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,19 +66,27 @@ jobs:
               - '**/*.psm1'
               - '.PSScriptAnalyzerSettings.psd1'
               - '.github/workflows/ci.yml'
-            # What can change what the pi package reconcile DOES: the manifest
-            # it reads, and the code in either twin that reads it. Nothing else
-            # can, because every entry is version-pinned and
-            # tests/pi-config.bats refuses an unpinned one -- so the same
-            # manifest always resolves to the same install set.
+            # What can change what the pi package reconcile DOES. This set is
+            # MAINTAINED BY HAND and is not proven closed — an earlier version
+            # of this comment claimed "nothing else can", and that claim is
+            # exactly what stopped `versions.conf` being noticed. Under-matching
+            # here is silent: the PR still runs test-windows, it just runs it
+            # with the reconcile skipped, and it goes green.
             #
-            # Both setup scripts are listed WHOLE rather than by the reconcile
-            # block, because a filter cannot see line ranges and a guard that
-            # silently under-matches is worse than one that runs too often.
+            # - ai/pi/**        the manifest and the deployed configs
+            # - setup-*.{sh,ps1} the code that reads them, listed WHOLE because
+            #                   a filter cannot see line ranges
+            # - versions.conf   PI_VERSION selects the pi binary that PERFORMS
+            #                   the reconcile (setup-linux.sh:722,
+            #                   setup-windows.ps1:1148). A version bump is the
+            #                   PR class that most needs nine real installs run
+            #                   by the NEW pi, and it was the one class that
+            #                   would have skipped them.
             pi:
               - 'ai/pi/**'
               - 'setup-linux.sh'
               - 'setup-windows.ps1'
+              - 'versions.conf'
               - 'tests/pi-*.bats'
 
   lint:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,10 @@ jobs:
     outputs:
       code: ${{ steps.filter.outputs.code }}
       powershell: ${{ steps.filter.outputs.powershell }}
+      # CI-002 (#1478): the first filter that is narrower than "is this code at
+      # all". HARNESS-041 asked whether a diff needs CI; this asks which PART of
+      # CI a diff needs, which one bucket cannot answer.
+      pi: ${{ steps.filter.outputs.pi }}
     steps:
       - uses: actions/checkout@v7
       - uses: dorny/paths-filter@v4
@@ -62,6 +66,20 @@ jobs:
               - '**/*.psm1'
               - '.PSScriptAnalyzerSettings.psd1'
               - '.github/workflows/ci.yml'
+            # What can change what the pi package reconcile DOES: the manifest
+            # it reads, and the code in either twin that reads it. Nothing else
+            # can, because every entry is version-pinned and
+            # tests/pi-config.bats refuses an unpinned one -- so the same
+            # manifest always resolves to the same install set.
+            #
+            # Both setup scripts are listed WHOLE rather than by the reconcile
+            # block, because a filter cannot see line ranges and a guard that
+            # silently under-matches is worse than one that runs too often.
+            pi:
+              - 'ai/pi/**'
+              - 'setup-linux.sh'
+              - 'setup-windows.ps1'
+              - 'tests/pi-*.bats'
 
   lint:
     runs-on: ubuntu-latest
@@ -397,12 +415,27 @@ jobs:
       # set here, on the CI step, and NOT in setup-windows.ps1: this is a
       # property of a throwaway runner, not of a developer's machine, and the
       # bootstrap scripts stay out of it (ADR-020 C7).
+      #
+      # CI-002 (#1478): on a PR whose diff cannot change what the pi reconcile
+      # does, skip it. That block is 883-2200s of this job — four runs of the
+      # same nine pinned packages, measured 2026-09-04 — and a Go-only PR was
+      # paying all of it. `pi` is the filter above; anything it matches runs the
+      # reconcile as before.
+      #
+      # PUSHES TO main NEVER SKIP, and after this the reconcile runs ONLY there.
+      # That is deliberate and it is the whole guard: coverage does not move to
+      # a schedule nobody reads. But it does mean main is now the only place a
+      # reconcile regression can surface, so the guard's value rests entirely on
+      # a red main being noticed. Stated out loud because the alternative is
+      # leaving it as an assumption, and an unread signal is how CI-001 stayed
+      # invisible for as long as it did.
       - name: Run setup-windows.ps1 end-to-end (PS 5.1 -> BUG-005 re-exec)
         if: github.event_name == 'push' || needs.changes.outputs.code == 'true'
         shell: powershell
         env:
           DOTFILES_REPO_DIR: ${{ github.workspace }}
           npm_config_prefer_offline: "true"
+          DOTFILES_SKIP_PI_PACKAGES: ${{ (github.event_name == 'pull_request' && needs.changes.outputs.pi != 'true') && '1' || '' }}
         run: .\setup-windows.ps1
 
       # The gate. Setup itself stays non-fatal on doctor (degrade-not-break, C9,
--- a/setup-linux.sh
+++ b/setup-linux.sh
@@ -850,9 +850,26 @@ fi
 # version check above already reaches for $PI_BIN for the same reason.
 #
 # Idempotent by set difference, so a second run installs nothing and says so.
+#
+# DOTFILES_SKIP_PI_PACKAGES exists for one caller: a throwaway CI runner whose
+# diff cannot change what this block does (CI-002, #1478). The reconcile is
+# 883-2200s of a 20-45 minute job — measured across four runs of the same nine
+# pinned packages — and it reinstalls the same manifest whether or not the PR
+# went near it. Set it and the block is skipped LOUDLY; the coverage does not
+# move to a schedule nobody reads, it stays on pushes to the default branch
+# where a failure is somewhere people already look.
+#
+# It is deliberately not a config key or a flag in packages.json. A durable
+# switch would let a real machine end up permanently unconverged, which is the
+# opposite of what this script is for. An environment variable is scoped to the
+# one process that sets it and dies with it.
 PI_PACKAGES_SRC="$CURRENT_DIR/ai/pi/packages.json"
 if [ -f "$PI_PACKAGES_SRC" ]; then
-    if [ ! -x "$PI_BIN" ]; then
+    if [ -n "${DOTFILES_SKIP_PI_PACKAGES:-}" ]; then
+        # Loud, and it says what was NOT verified rather than only what was
+        # skipped: a line reading "skipped" is indistinguishable from "fine".
+        log_warning "DOTFILES_SKIP_PI_PACKAGES set — pi package reconcile skipped (nothing installed, nothing verified)"
+    elif [ ! -x "$PI_BIN" ]; then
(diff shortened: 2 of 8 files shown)


mlorentedev:
## Review — independent read, no stake in this diff

Reviewed as a reviewer rather than a co-author: I did not write any of this and
have no branch depending on it. One **Major**, the rest is confirmation.

The mechanism is right, the ordering contract is right, and the two mutation
defects you found in your own tests were real defects — the `grep -n` one in
particular would have made the ordering assertion permanently green. Slicing the
block with `awk` before searching is the correct fix and the `[ ! -x "$PI_BIN" ]`
collision 150 lines up is a genuine trap.

---

### Major — `versions.conf` is missing from the `pi` filter, and the test encodes the omission

The filter:

```yaml
pi:
  - 'ai/pi/**'
  - 'setup-linux.sh'
  - 'setup-windows.ps1'
  - 'tests/pi-*.bats'
```

Its comment claims:

> What can change what the pi package reconcile DOES: the manifest it reads, and
> the code in either twin that reads it. **Nothing else can** […]

That is false. **`versions.conf:16` is `PI_VERSION=0.84.4`**, and both twins read
it to decide which `pi` binary performs the reconcile:

    setup-windows.ps1:1148   if ($line -match '^\s*PI_VERSION\s*=\s*(.+?)\s*$')
    setup-windows.ps1:1157   $piPkg = "@earendil-works/pi-coding-agent@$piVersion"
    setup-windows.ps1:1172   & npm install -g --ignore-scripts "...@$piVersion"
    setup-linux.sh:722       PI_PKG="@earendil-works/pi-coding-agent${PI_VERSION:+@$PI_VERSION}"

`pi install` is the thing being exercised. A different `pi` can resolve, unpack
or record packages differently — that is the entire reason the version is pinned
in the first place.

**The failure is exactly the one your filter comment warns about.** `versions.conf`
IS in the `code` filter, so a `PI_VERSION` bump still runs `test-windows` — and
now runs it with the reconcile **skipped**. The one PR class where you most want
nine real `pi install` calls executed is the one class that silently stops
executing them, and it goes green.

Worse, the test cannot catch it:

```bash
@test "pi packages: the CI pi filter covers the manifest and BOTH twins" {
    # Everything that can change what the reconcile does. A filter that misses
    # one of these skips the reconcile on the exact PR that needed it […]
    echo "$filter" | grep -q "ai/pi/\*\*"
    echo "$filter" | grep -q "setup-linux.sh"
    echo "$filter" | grep -q "setup-windows.ps1"
}
```

The comment states the property; the assertions enumerate today's three entries.
So the test asserts the **implementation**, not the property, and is green by
construction against any omission — including this one. It is the same shape as
the ordering test you already fixed: an assertion that cannot fail for the reason
it exists.

**Fix is two lines.** Add `- 'versions.conf'` to the filter, and
`echo "$filter" | grep -q "versions.conf"` to the test.

I would also weaken the filter comment from "Nothing else can" to something that
does not assert completeness. A comment claiming a closed set invites exactly the
reading that stopped this being noticed — and having now checked, I can only say
I found one omission, not that there are none left.

---

### Confirmations, no action needed

- **The env expression is correct at both ends.** When false it yields `''`;
  bash `[ -n "${VAR:-}" ]` and PowerShell `if ($env:VAR)` both treat an empty
  string as unset, so an empty value never skips. `needs: [changes]` is present
  on `test-windows`, so `needs.changes.outputs.pi` resolves.
- **Pushes genuinely never skip** — the expression short-circuits on
  `github.event_name == 'pull_request'` before the filter is consulted.
- **`PI_BIN` is hardcoded** at `setup-linux.sh:712` (`$HOME/.local/bin/pi`), so
  it introduces no further external input. I checked for a second gap of the
  same class and did not find one.
- **`tests/pi-*.bats` in the filter is over-broad but errs safe** — a test change
  cannot alter runtime behaviour, so it only causes the reconcile to run when it
  need not. Correct direction for a guard.
- **A PR editing only the `pi` filter skips its own reconcile**, so a filter
  typo is not caught by the PR that introduces it. Inherent to path filters,
  worth knowing, not worth solving here.

### On the Linux twin having no end-to-end coverage anywhere

You asked me to decide rather than deciding alone, so: **acceptable, and not this
PR's debt.** `integration`'s container has no npm and the Linux reconcile has
therefore never executed in CI — that predates this change, this change does not
worsen it, and fixing it here would mean adding npm to the integration image
inside a PR about skipping work in CI.

But it should not stay a paragraph in a proposal that archives. It deserves its
own ticket, because right now the Linux half of a system that installs
third-party code with full system access is verified by `grep` alone. Filing it
is cheap; discovering it from an incident is not.

mlorentedev:
## Review triage

|---|---|---|---|
| 1 | **dotfiles-d5** (peer session, adversarial read) | **Major — `versions.conf` missing from the `pi` filter.** `versions.conf:16` is `PI_VERSION=0.84.4`, read by both twins to choose the pi binary that *performs* the reconcile (`setup-linux.sh:722`, `setup-windows.ps1:1148`). A version bump still runs `test-windows` — `versions.conf` is in `code` — and runs it with the reconcile skipped. The one PR class that most needs nine real installs by the **new** pi was the class that would silently stop executing them. | **Accepted. Fixed in `46af234`.** Verified independently before applying. |
| 2 | dotfiles-d5 | The filter test **asserts the implementation** — it greps the entries someone already wrote, so it is green by construction against any omission, including this one. | **Accepted, and fixed beyond the report.** Added a **derived** assertion: ask the repository where `PI_VERSION` is declared and require that file to be in the filter. Survives the pin moving or the file being renamed. |
| 3 | dotfiles-d5 | Drop *"Nothing else can"* from the filter comment — it is the sentence that made the omission invisible. | **Accepted.** The comment now states the set is hand-maintained and **not proven closed**. |
| 4 | dotfiles-d5 | `tests/pi-*.bats` in the filter is over-broad. | **Accepted, no change.** It errs safe: it can only cause the reconcile to run when it need not. |
| 5 | dotfiles-d5 | The Linux twin has no end-to-end coverage — acceptable and not this PR's debt, but it should not archive as a paragraph in a proposal. | **Accepted, ticket owed.** Filing it under CI-002's ownership. The Linux half of a system that installs third-party code with full system access is verified by grep alone. |
| 6 | PR-Agent | #1478 — all six acceptance criteria compliant, **no** non-compliant requirements; effort 2/5; contains tests; no security concerns; no major issues. | **Noted.** Its one "requires human verification" — the end-to-end effect on job duration — is already recorded in `verification.md` as owed, measured by the first `pull_request` run after this lands. Not claimed here. |

### Two defects the fix exposed, in this PR's own tests

Both are the class this repository keeps cataloguing, and neither was visible from a passing run.

**Third instance in this file:** the `awk` slicing the filter block terminated on *"the next 12-space key"*, which never fires there. The slice ran **past** the block and swept up later lines mentioning the same paths, so deleting an entry still matched. An earlier "strip the comments" fix had caught only one of the things being over-captured — treating the symptom. It now stops at the first non-entry, anchored on the **list** rather than on what follows it.

**And a trap worth stating plainly:** the first mutation run reported all four filter entries `NOT CAUGHT`. That was my harness, not the tests — `setup-windows.ps1` appears twice in `ci.yml` (the `code` filter and the `pi` filter) and I was deleting the first. **"The mutation was not caught" and "the mutation was not applied" look identical from the outside.** I nearly rewrote working assertions because of it. Re-run targeting inside the `pi` block, all four entries now fail when dropped.

### Who reviewed

**dotfiles-d5** (independent peer session, produced the only Major) and **PR-Agent**. CodeRabbit posted `> [!NOTE] Currently processing new changes in this PR` — a notice, not a review; if it produces one, this triage gets an addendum rather than being assumed complete.

mlorentedev:
## Review triage — addendum at `46af234` → `e0b0ad1`

PR-Agent re-ran against `46af234` (the versions.conf fix) and **edited its comment in place at 04:22:47Z**, 2m25s after the triage above at 04:20:22Z. `created_at` stayed at 04:14:03, so `dotf pr triage-queue` reported **exit 0**. It was wrong — live instance of #1422, recorded there, on the PR that is the critical path.

|---|---|---|---|
| 7 | PR-Agent | **#1478 — fully compliant**, all seven acceptance criteria listed individually, no non-compliant requirements. Effort 3/5, contains tests, no security concerns. | **Noted.** |
| 8 | PR-Agent | Focus area *"Shell sourcing pattern"* at `setup-linux.sh:862-863` — then states it itself: *"doc-only concern — no defect… the guard logic is correct and zsh-compatible. No prohibited shell patterns introduced."* | **Declined, self-cancelling.** The lines are a comment. Verified independently anyway: `[ -n "${DOTFILES_SKIP_PI_PACKAGES:-}" ]` parses under `bash -n` and `zsh -n`, and the `:-` default is required precisely because an unset variable under `set -u` would abort. |
| 9 | PR-Agent | Focus area on `features.json` — *"states are all `pending` with empty evidence, which is expected at this stage… No issues found."* | **Verdict noted; the EVIDENCE it quoted was a real defect.** See below. |

### The finding was in what it quoted, not in what it said

PR-Agent quoted `f2` back verbatim:

```
"verification": "... ~/.local/bin/bats tests/pi-packages.bats 2>&1 | grep -c '^ok ' | grep -qx 22"
```

The versions.conf fix added the **23rd** test. So that command **fails** — and it is the command `spec archive` reads as evidence. The review said "no issues found" and simultaneously printed the broken assertion.

A hard-coded count is brittle in the one direction a spec is guaranteed to move: adding tests. Worse, it fails for a reason unrelated to the behaviour it claims to verify — the signal would read *"AC2–AC5 are broken"* when what happened is that the suite grew.

**Fixed in `e0b0ad1`.** `f2` now asserts the suite passes **and** that each named guarantee is present by name, proven in both directions rather than assumed:

| mutation | result |
|---|---|
| add a 24th test | f2 still passes — no longer brittle |
| rename a guarantee out of the file | f2 fails — not vacuous either |

All five verification commands re-executed: **5/5 pass**.

### Reviewer count

**PR-Agent** and **dotfiles-d5** (peer session). CodeRabbit has posted only `> [!NOTE] Currently processing new changes` — a notice, not a review. Two reviewers, and the one that found the Major was the peer, not either bot.

PR #1485, opened 2026-09-04
docs(lessons): 267 — a mutation harness must prove the mutation landed

mlorentedev:
Docs-only. One lesson file plus its index row.

## What it records

Mutation-testing four CI path-filter assertions on #1482 reported **all four `NOT CAUGHT`**. The obvious reading was that the assertions were vacuous and needed rewriting.

They were fine. **The harness was broken.** `setup-windows.ps1` appears twice in `ci.yml` — the `code` filter and the new `pi` filter — and the mutation did `replace(entry, '', 1)`, deleting the **first** occurrence. It edited a different block, left the one under test untouched, and the assertions correctly passed against unmutated input.

> **"The mutation was not caught" and "the mutation was not applied" are indistinguishable from outside the harness.**

Both produce a green suite and the words `NOT CAUGHT`. One means the test is worthless; the other means the test is fine and the harness lied. They demand opposite responses, and the wrong one — rewriting working assertions — costs more than the bug would have.

`assert anchor in source` is **not** the guard against it: it proves the string exists *somewhere*, which is exactly the property that was true and useless. The anchor must be located the way the code under test locates it — by block, by line range, by structure.

## Why it is worth a lesson rather than a commit message

Fifth member of **one family** found in a single session, every one *a green result that measured the wrong thing*, none visible from a passing run:

| where | what it measured instead |
|---|---|
| `features.json` counting `--- PASS: Test` | 18, including indented subtests, where 4 was expected |
| `grep -n X f \| grep -v '^ *#'` "filters comments" | `grep -n` prefixes a line number, so no line starts with `#` — nothing filtered |
| an `awk` slice terminating on "the next 12-space key" | never fires; the slice ran past the block and matched unrelated lines |
| `features.json` asserting `grep -c '^ok ' \| grep -qx 22` | broke when the suite grew to 23, for a reason unrelated to the behaviour |
| **this one** | the mutation deleted a line in a different block |

The recurring shape: a check whose **target** is specified by a pattern matching more, or less, or elsewhere than intended — while the **assertion** on top of it is perfectly correct. The verification is never the assertion alone; it is the assertion plus a proof it was pointed at the right thing.

## Numbering

267. 264 is #1471, 265–266 are #1474 — deconflicted with the peer session, no `_index.md` collision with either.

Refs #1478, #1422


mlorentedev:
## Review triage

|---|---|---|---|
| 1 | PR-Agent | **#1482 "partially compliant"** — *"does not implement the actual CI changes (env var, path filter)"*, *"does not add the mutation tests (they exist in #1482)"*, *"does not record the decision about latency"* | **Declined — the referent is wrong.** #1482 is a **pull request**, not a ticket, and it is **already merged** (`91e3834`). This PR does not "fail to implement" #1482's changes; #1482 implemented them. A docs-only PR being marked non-compliant for not containing the code it documents is the defect, not a finding. |
| 2 | PR-Agent | **#1471 "partially compliant"** — *"this PR is unrelated to #1471's requirements… does not address any #1471 requirements"* | **Declined, same defect.** #1471 is also a pull request. The analysis states its own premise: the PRs are unrelated. |
| 3 | PR-Agent | *"Requires human verification: lesson content accuracy, and that the index row is properly placed"* | **Accepted, verified.** Index row sits between 263 and the undated tail, matching the file's existing ordering; 267 is deconflicted against 264 (#1471) and 265–266 (#1474), so no `_index.md` collision with the peer's branch. Content: every claim in the lesson is a measurement from this session's transcript, and the five family members it tabulates each have a commit or PR comment behind them. |
| 4 | CodeRabbit | *(no output on this PR)* | **Nothing to dispose of.** |

### Known reviewer defect — 13th instance

PR-Agent read **two pull requests as tickets** on one PR. Tracked on **#1466**.

Worth adding a bound the ticket does not yet have: on **#1482** it got the referent **right**, correctly analysing #1478 and returning a full-compliance verdict. So the defect is *not* "every referenced number becomes a requirement" — it behaves correctly when the PR body's `Refs` points at a real issue, and misfires when the body mentions PR numbers. That narrows the fix.

### Reviewer count

**One.** PR-Agent only; CodeRabbit posted nothing here. Recorded rather than implied.

### Note on this PR's cheapness

`test-windows` completed in **11s** — the `code` filter excludes `docs/**`, so the heavy path never started. That is HARNESS-041 working as designed, and it is the contrast case for CI-002's PR 2: the coarse filter already handles *"no code at all"*; what it cannot answer is *"which part of the code"*.

Issue #1487, opened 2026-09-04
CI-004: a third push to main silently discards the second pending CI run, so a merge can get no CI at all


mlorentedev:
## What happened

Three pushes to `main` on 2026-09-04. The middle one produced **no jobs at all**:

| run | created | window | jobs run | conclusion |
|---|---|---|---|---|
| `33845866062` | 06:47:51 | 06:47:51 → 07:33:17 | all 6 | cancelled (ceiling) |
| `33846609699` | 06:58:13 | 06:58:13 → 07:06:07 | **none** | cancelled |
| `33847200968` | 07:06:05 | 07:06:05 → 08:18:37 | all 6 | cancelled (ceiling) |

`33846609699` was cancelled 7m54s after creation without a single job starting, and the
run that cancelled it is the one created 8 seconds *before* its own cancellation.

## Why

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

On `push` this evaluates to `cancel-in-progress: false`, which is deliberate and correct —
a post-merge run should not be killed by the next merge. But GitHub's concurrency model
holds **at most one pending run per group**: when a third run enters a group that already
has one running and one pending, the *pending* one is cancelled. `cancel-in-progress:
false` protects the run that is executing; nothing protects the one that is waiting.

So the guarantee people read off that config — "every push to main gets a run" — does not
hold. It holds only for pushes spaced further apart than a run takes.

That spacing is currently ~45 minutes (#1486), which makes this trivially reachable: two
PRs merged within the same hour and one of them silently gets no CI at all. It is the
same shape as the defect CI-002 was trying to close — the signal is not red, it is
absent — and it defeats the same guard, which is why it is worth a ticket rather than a
comment.

## Notes toward a fix

- The group key is per-ref, so main's pushes all share one group. Keying pushes by SHA
  (`${{ github.workflow }}-${{ github.ref }}-${{ github.event_name == 'push' && github.sha || '' }}`)
  gives every push its own group and no push is ever queued behind another. It also
  removes serialisation on main, which is the property that made the queue exist —
  weigh runner cost against it.
- Whatever the fix, it needs an assertion. There is no test today that reads the
  concurrency block at all, and this is a config whose failure mode is a run that does
  not exist, so nothing downstream notices.

## References

- #1486 — the 45-minute main runs that make the queue deep enough for this to bite.
[...]

Issue #1486, opened 2026-09-04
CI-003: the pi reconcile is unobservable and its cost varies 24x on identical input, so main's Windows verification is a coin flip against the ceiling


mlorentedev:
> **Corrected 2026-09-05.** The first version of this body said main's `test-windows`
> "does not complete, so it verifies nothing anywhere". That was true of the three runs
> it cited and **false of the current head** — a peer session pointed at two later runs,
> and checking them directly confirms it. The correction is kept in place rather than
> edited away, because what replaced it is a stronger claim, not a retreat: the numbers
> below are the same evidence read without the sampling error.

## Why this is not a reopen of #1472

#1472 was closed at 02:57:57Z on 2026-09-04, before #1480 and #1482 landed. Its
diagnosis — *"the difference is npm registry latency, which is not a property of the
repository"* — has since been refuted by measurement, and its options list is half
resolved (the cache was tried and removed in #1480; the CI skip shipped in #1482).
Rewriting it wholesale would destroy the record of what was believed when.

## What is broken

**The reconcile's cost is unbounded, with a 24x spread on byte-identical input**, and it
sits inside a job with a 45-minute ceiling. So whether main verifies Windows setup at all
is a coin flip on registry behaviour, not a property of the code.

Five pushes to main on 2026-09-04, same nine version-pinned packages every time:

| run | head | package phase | setup step | outcome |
|---|---|---|---|---|
| `33845866062` | — | (killed mid-loop) | 45m06s | **cancelled at ceiling** |
| `33846609699` | — | — | — | **cancelled before any job started** (#1487) |
| `33847200968` | — | ≥2430s | 44m19s | **cancelled at ceiling** |
| `33930483404` | `9c2758a` | **101.8s** | 6m24s | success |
| `33930629978` | `7b00e84` | — | 6m57s | success |

Two of the three that ran to a verdict produced **no Windows verification at all**:
`Post-setup doctor gate`, `Run Pester suites` and `Run PowerShell bats subset` are all
`skipped`, and the run is reported `cancelled` — neither a pass nor a failure, so nothing
goes red about it.

**101.8s against ≥2430s is the finding.** Same commit lineage, same pins, same runner
image. A cost that varies by 24x cannot be planned around, and a ceiling is a coin toss
against it rather than a limit on it.

## Measured: the 421s constant belongs to the operation, not to any package

Per-package durations, derived from the runner timestamps of consecutive
[...]

Issue #1484, opened 2026-09-04
TEST: setup-linux.sh's pi package reconcile has never run in CI — the integration container has no npm


mlorentedev:
## What is not covered

`setup-linux.sh`'s pi package reconcile has **never executed in CI**.

The `integration` job builds a container and runs `setup-linux.sh` end to end. That container has **no npm**, so the block guards out immediately:

```
[WARNING] npm not found — skipping pi install (install Node.js, then re-run setup)
[WARNING] pi not installed — skipping pi package reconcile (re-run setup after pi installs)
```

Measured on job `100895294663` (run 33830064842), and it is why `integration` completes in ~1m29s while `test-windows` — which *does* run it — takes 22–43 minutes.

So the Linux half is verified by **grep alone**: `tests/pi-packages.bats` asserts the block's *shape* (it calls `$PI_BIN`, it uses `grep -Fxq`, it guards on npm, it honours `DOTFILES_SKIP_PI_PACKAGES`). Nothing has ever run it.

## Why it is worth a ticket rather than a paragraph

This block installs **nine third-party npm packages** into an agent. `ai/pi/packages.json`'s own header states the stakes:

> *"Pi packages run with full system access — extensions execute arbitrary code and skills can instruct the model to run executables."*

That is inside an agent holding `NAN_API_KEY` with write access to these repositories. The Windows twin's copy of that logic is exercised on every run; the Linux twin's is exercised on nobody's.

Raised in review on **#1482** by the peer session, whose judgement was: acceptable, and *not* #1482's debt — it predates the change, the change does not worsen it, and adding npm to the integration image inside a PR about *skipping* work would be scope creep. But it should not archive as a paragraph in a spec.

## Options

1. **Add Node/npm to the integration image.** Honest coverage, and it makes `integration` pay some of the cost `test-windows` pays — which is the cost CI-002 is removing from the PR path, so it needs care not to reintroduce the same problem on Linux.
2. **A narrow harness test** that drives the reconcile block against a stub `pi` and a fixture manifest, asserting the set difference and idempotence, without installing anything real. Cheaper, faster, and tests the logic rather than npm.
3. Accept and document. Current state, and it is what this ticket exists to stop being the default by omission.

Option 2 looks right: the untested thing is the *reconcile logic*, not npm's ability to install packages.

## Notes

- Pre-existing. Not introduced by #1482 or CI-002.
[...]

PR #1488, opened 2026-09-05
fix(setup): log what a pi install cost and what it printed, in both twins

mlorentedev:
Refs #1486 — **part 1 of 2, so this does not close it.** #1486 §What.1 is the
observability; §What.2 is the bound, and the whole argument of this PR is that the bound
cannot be chosen until §What.1 has produced data. The spec stays `implementing` and
archives with part 2.

## The finding

The pi package reconcile costs **101.8s or ≥2430s** for the same nine version-pinned
packages, on the same runner image. **24x on identical input.** Two pushes to main:

| run | package phase | setup step | outcome |
|---|---:|---:|---|
| `33847200968` | **≥2430s** | 44m19s | **cancelled at the 45-minute ceiling** |
| `33930483404` | **101.8s** | 6m24s | success |

The cancelled one produced **no Windows verification at all** — `Post-setup doctor gate`,
`Run Pester suites` and `Run PowerShell bats subset` are all `skipped` — and `cancelled`
is neither a pass nor a failure, so nothing went red. The fast one reports
`pi packages: 9 installed, 0 already present, 0 failed`, which is what makes 101.8s a
*complete* reconcile rather than one that skipped work.

Per package:

| package | cancelled run | successful run |
|---|---:|---:|
| `npm:pi-effort@0.0.8` | **421.4s** | 1.4s |
| `npm:pi-web-access@0.24.2` | 195.2s | 42.0s |
| `npm:@ayulab/pi-rewind@0.4.6` | **421.7s** | **1.8s** |
| `npm:@narumitw/pi-plan-mode@0.53.0` | 309.6s | 13.2s |
| `npm:@narumitw/pi-goal@0.53.1` | **421.5s** | 2.3s |
| `npm:pi-subagents@0.56.0` | 35.2s | 12.7s |
| `npm:pi-memory@0.4.2` | 42.2s | 1.7s |
| `npm:pi-cc-extensions@0.8.67` | 345.0s | 25.2s |
| `npm:pi-add-dir@1.3.1` | ≥238.4s (killed) | 1.6s |

`@ayulab/pi-rewind` — the package #1472 singled out as the slow one — ran **421.7s and
1.8s**. `pi-memory` ran **421.5s, 42.2s and 1.7s** on one pin. Ten observations in a
421 ±1s band across seven packages: a fixed timeout on the install *operation*, landing
on whichever installs hit it. How many hit it per run is the whole variance.

## Why this PR does not fix it

Because it cannot yet, and that is the point. Both twins discarded the install's stdout
**and** stderr:

```powershell
& pi install $piPkgName 2>$null | Out-Null              # setup-windows.ps1:1345
```
```sh
"$PI_BIN" install "$pi_pkg" >/dev/null 2>&1             # setup-linux.sh:910
```

Between two consecutive `Installing pi package` lines the log holds **nothing**. So a
421s install that **succeeded slowly** and one that **failed after retries** are
indistinguishable — and they demand opposite bounds. A bound that kills a slow success
turns working setup into broken setup; a bound on an install that was failing anyway is
free. Durations alone cannot settle it either: one package spans 1.4s–421.7s.

Neither twin logged elapsed time, so every number above was reconstructed from Actions'
own line timestamps — which means the same diagnosis is simply unavailable on a real
machine.

## What changed

Both twins, together:

- **Elapsed seconds logged on every outcome** — fast success, slow success, failure. This
  is the line that answers *"did the 421s end in success?"*, which no amount of captured
  output answers on its own.
- **The install's output captured, not discarded**, and emitted **fenced** when the
  install fails or crosses a verbosity threshold. Fenced because "the install printed
  nothing at all" is itself a finding, and a bare dump makes it indistinguishable from
  "nothing was captured" — which is the bug being replaced.
- **`run 'pi install X' to see why` is gone.** It asked a human to reproduce by hand,
  later, a failure the machine held in a variable at the time. On the Windows runner that
  reproduction is not available at all.
- **`$ErrorActionPreference` pinned and restored in a `finally`** around the PowerShell
  call. `2>&1` on a *native* command turns stderr into `ErrorRecord`s, and under `Stop`
  those terminate — so an install that writes a warning and exits 0 would abort setup.
  The redirect being replaced never met that hazard because it threw stderr away, which
  means **the hazard arrives with the fix**.
- **`$piRc` defaulted to failure before the guarded call.** It was assigned only *inside*
  the `try`. A throw would leave it unassigned — and reading an unassigned variable under
  `Set-StrictMode -Version Latest` is itself a terminating error, thrown outside the `try`
  meant to contain it. On iteration 2+ it would not even be unassigned: it would still
  hold the *previous* package's result, so a broken install would be counted as whatever
  the last one did. An unknown outcome is a failure, never a success.

## Deliberately not done

- **No bound.** #1486 §What.2 orders it: chosen from the data this produces.
- **No ceiling change.** #1475's 30 → 45 was an explicitly temporary loan that came due
  twice in one morning. Raising it again is the same loan.
- **No `ci.yml` change at all.** `setup-*.{sh,ps1}` are already in the `pi` filter, so
  **this PR's own `test-windows` runs the full reconcile against the new code** — it is
  the behavioural test *and* the run that produces the data. If it is cancelled at the
  ceiling, the captured output up to the kill is still in the log; a cancelled run keeps
  its log, which is the property that made this measurement possible in the first place.

## Verification

```
tests/*.bats                 1549/1549 pass, exit 0 (clean tree)
tests/pi-packages.bats       31/31 (23 -> 31)
bash -n / zsh -n             OK
shellcheck --severity=error  OK
setup-windows.ps1            ASCII-only (.gitattributes:36 marks it eol=crlf;
                             PSScriptAnalyzer fails CI on non-ASCII without a BOM)
git diff --check             OK
```

**Twelve mutations, twelve caught** — `python3 specs/CI-003/mutate-assertions.py`, exit 0. Each
prints the mutated line before and after, at its line number **inside the reconcile
block**, before the suite runs:

```
== AC1  neither twin discards  [setup-linux.sh]
    @ setup-linux.sh:926
      -if pi_pkg_out=$("$PI_BIN" install "$pi_pkg" 2>&1); then
      +if pi_pkg_out=$("$PI_BIN" install "$pi_pkg" >/dev/null 2>&1); then
      1 test(s) selected -> CAUGHT
```

That shape is lesson 267's requirement, and it is why the harness is committed rather
than described: the previous one did a repo-wide `replace(x, '', 1)`, deleted an
identically-spelled line from a *different* block, and reported four sound assertions as
vacuous. This one reports **HARNESS FAULT** — never `NOT CAUGHT` — for an identical file,
a missing target, or a `-f` filter selecting zero tests, so the two are no longer spelled
the same.

The same reasoning ran *before* any assertion was written: block anchors were checked for
uniqueness first. `pi install` appears in pi's **own** install block ~130 lines earlier in
`setup-windows.ps1`, carrying its own `run: npm install -g …` advice, so a file-wide
assertion would have compared against the wrong code and passed.

**Both directions on the two checks that print nothing when they pass** — a green command
that has never been shown capable of being red is not evidence:

- shell syntax + shellcheck + ASCII-only: appended one `é` to `setup-windows.ps1` →
  **exit 1**.
- the harness-fault path: replaced one mutation target with a string that does not exist
  → `HARNESS FAULT: target not found in setup-linux.sh at/after line 866`, **exit 1**.

## Corrections made while building this

- **#1486's original body said main "does not complete, so it verifies nothing anywhere".
  That was false of the current head** — true of three morning runs, and a peer session
  pointed at two later ones that succeeded. Verified directly, then the ticket's title and
  body were corrected. Three cancellations read as a state is the same sampling error that
  made #1472 call `pi-rewind` the slow package.
- **One assertion's first draft was wrong.** `the slow threshold gates VERBOSITY only`
  searched for `break|continue|exit|return` unanchored and matched the literal `exit`
  inside the fence's own `(exit $pi_pkg_rc, …)` label — a false *finding*, which is the
  same disease as a false pass. Anchored now, with the reason kept in the test.
- **And again, on the `$piRc` ordering assertion.** It first anchored on `try {`, but the
  reconcile block opens a `try` ~40 lines earlier to parse the manifest, so it compared
  against *that* one and reported a correctly-placed line as misplaced. Re-anchored on
  `$ErrorActionPreference = 'Continue'`, unique in the block. Three occurrences of one
  shape in one file is why the block slices and the anchor-uniqueness check are
  load-bearing rather than ceremony.

## Not verified here

- **`setup-linux.sh`'s reconcile has still never executed in CI** (#1484): the
  `integration` container has no npm, so the block logs `npm not found` and skips. The
  Linux half is verified structurally and by mutation, not end to end. Unchanged by this
  PR, stated rather than implied.
- Also filed from this measurement: **#1487** — a third push to main discards the second
  *pending* run (`cancel-in-progress: false` protects the running run, not the queued
  one), so two merges close together can leave one commit with no CI at all.

### DIFF
--- a/setup-linux.sh
+++ b/setup-linux.sh
@@ -900,6 +900,14 @@ if [ -f "$PI_PACKAGES_SRC" ]; then
             pi_pkg_added=0
             pi_pkg_failed=0
             pi_pkg_present=0
+            # A VERBOSITY threshold, not a bound. Nothing is killed for crossing
+            # it; it only decides whose captured output is worth printing. Being
+            # wrong therefore costs log lines, never a broken install — which is
+            # why it is a literal here and not a knob. Sized from the measured
+            # distribution on the Windows runner (#1486): normal installs land
+            # at 35-345s and the anomaly at 421 ±1s, so there is no clean gap to
+            # sit in, and this deliberately errs toward printing too much.
+            PI_INSTALL_SLOW_SECONDS=120
             while IFS= read -r pi_pkg; do
                 [ -n "$pi_pkg" ] || continue
                 if printf '%s\n' "$PI_PKG_PRESENT" | grep -Fxq "$pi_pkg"; then
@@ -907,11 +915,40 @@ if [ -f "$PI_PACKAGES_SRC" ]; then
                     continue
                 fi
                 log_info "Installing pi package $pi_pkg ..."
-                if "$PI_BIN" install "$pi_pkg" >/dev/null 2>&1; then
-                    pi_pkg_added=$((pi_pkg_added + 1))
+                # Both the elapsed time and the output used to be discarded here
+                # (`>/dev/null 2>&1`), and #1486 is what that cost: nine installs
+                # whose durations could only be reconstructed from GitHub's own
+                # line timestamps, and whose 421s outliers cannot be told apart
+                # from a slow success even in hindsight. The machine had the
+                # diagnostic in its hands and threw it away, then told the reader
+                # to reproduce it by hand.
+                pi_pkg_started=$(date +%s)
+                if pi_pkg_out=$("$PI_BIN" install "$pi_pkg" 2>&1); then
+                    pi_pkg_rc=0
                 else
-                    log_warning "pi install $pi_pkg failed — run \"$PI_BIN install $pi_pkg\" to see why"
+                    pi_pkg_rc=$?
+                fi
+                pi_pkg_elapsed=$(( $(date +%s) - pi_pkg_started ))
+
+                if [ "$pi_pkg_rc" -eq 0 ]; then
+                    pi_pkg_added=$((pi_pkg_added + 1))
+                    if [ "$pi_pkg_elapsed" -ge "$PI_INSTALL_SLOW_SECONDS" ]; then
+                        log_warning "pi install $pi_pkg took ${pi_pkg_elapsed}s — over the ${PI_INSTALL_SLOW_SECONDS}s diagnostic threshold, output follows"
+                    else
+                        log_info "pi package $pi_pkg installed in ${pi_pkg_elapsed}s"
+                    fi
+                else
                     pi_pkg_failed=$((pi_pkg_failed + 1))
+                    log_warning "pi install $pi_pkg failed after ${pi_pkg_elapsed}s (exit $pi_pkg_rc) — output follows"
+                fi
+
+                # Fenced, so that EMPTY output is itself legible. "The install
+                # printed nothing at all" is a finding; a bare dump makes it
+                # indistinguishable from "we never captured anything".
+                if [ "$pi_pkg_rc" -ne 0 ] || [ "$pi_pkg_elapsed" -ge "$PI_INSTALL_SLOW_SECONDS" ]; then
+                    printf '%s\n' "--- pi install $pi_pkg (exit $pi_pkg_rc, ${pi_pkg_elapsed}s) ---"
+                    printf '%s\n' "$pi_pkg_out"
+                    printf '%s\n' "--- end pi install $pi_pkg ---"
                 fi
             done <<EOF
 $PI_PKG_WANTED
--- a/setup-windows.ps1
+++ b/setup-windows.ps1
@@ -1336,18 +1336,63 @@ if (Test-Path -LiteralPath $piPackagesSrc -PathType Leaf) {
             $piAdded = 0
             $piFailed = 0
             $piAlready = 0
+            # A VERBOSITY threshold, not a bound. Nothing is killed for crossing
+            # it; it only decides whose captured output is worth printing. Being
+            # wrong therefore costs log lines, never a broken install - which is
+            # why it is a literal here and not a knob. Sized from the measured
+            # distribution on this runner (#1486): normal installs land at
+            # 35-345s and the anomaly at 421 +/-1s, so there is no clean gap to
+            # sit in, and this deliberately errs toward printing too much.
+            $piSlowSeconds = 120
             foreach ($piPkgName in $piWanted) {
                 if ($piPresent -contains $piPkgName) {
                     $piAlready++
                     continue
                 }
                 Write-Info "Installing pi package $piPkgName ..."
-                & pi install $piPkgName 2>$null | Out-Null
-                if ($LASTEXITCODE -eq 0) {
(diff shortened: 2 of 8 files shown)


mlorentedev:
## Review triage

**One reviewer produced substantive output on this PR, not two.** Stated rather than implied, because a green rollup and two bot comments look like two reviews.

|---|---|---|
| **CodeRabbit** | *"review paused by coderabbit.ai"* | **Not a review — a notice.** No files, lines or claims. Nothing to dispose of, and this PR should not be read as having had CodeRabbit's opinion. |
| **PR-Agent** | `No major issues detected`, no security concerns, contains tests, effort 3/5. Ticket compliance on #1486: every requirement compliant or explicitly deferred, zero non-compliant. | **Accepted.** No code findings to apply. |
| **PR-Agent, ticket scope** | Also analysed **#1472 and #1475** as if they were tickets this PR must satisfy. Both are closed, and #1475 is a *PR*. | **Declined.** Same inference defect already filed on #1466; this is another instance, not a new one. Its own output says "N/A — not the target ticket", so it cost nothing here. |
| **CI** | 10/10 green including `test-windows` (`33932370090`, job success, 7m59s). | Accepted. |

## The run's own verdict, which is not "green"

`test-windows` executed the full reconcile against the new code — the behavioural test this PR was written to get. What it proves, and what it does not:

```
pi-effort         2s     pi-web-access    44s     pi-rewind         2s
pi-plan-mode      9s     pi-goal           2s     pi-subagents     10s
pi-memory         2s     pi-cc-extensions 12s     pi-add-dir        1s
[SUCCESS] pi packages: 9 installed, 0 already present, 0 failed
```

- **AC3 is proven end to end.** Nine of nine installs logged their elapsed time on a real Windows runner. Before this PR that number was zero, and every duration in #1486 had to be reconstructed from GitHub's line timestamps.
- **AC6 was exercised.** All nine went through the pinned `$ErrorActionPreference` and the `finally`; setup did not abort. The hazard the fix introduces did not fire.
- **The fenced-output path was NOT exercised.** Nothing crossed the 120s threshold and nothing failed, so no capture block printed. Not claimed as verified.
- **The data #1486 §What.2 needs has NOT arrived.** 84s total, zero installs in the 421 ±1s band. Which is what "the constant lands at random" predicts, and it is why part 2 waits rather than guessing a bound from a fast night.

So the merge case here is "the instrument works and costs nothing", not "the problem is understood".

Issue #1269, opened 2026-08-27
CLI-050: cut over knowledge-crystallize.{sh,ps1} callers to dotf vault crystallize


mlorentedev:
Slice of CLI-023 (#492), scoped down after #492 turned out to be all-or-nothing (9 scripts,
depends on unstarted CLI-022). This ticket flips only the pair CLI-021 (#490) already fully
ported with golden byte-parity: `knowledge-crystallize.{sh,ps1}` -> `dotf vault crystallize`.

## Scope

- Repoint `scripts/vault-maintenance-weekly.sh:22` (and `.ps1:25`) to call
  `dotf vault crystallize --all` instead of the shell twin.
- Remove `setup-linux.sh:117` chmod +x and the `setup-windows.ps1:1593-1598` deploy block for
  the twin.
- Delete `scripts/knowledge-crystallize.{sh,ps1}`.
- Repoint docs: project `.claude/CLAUDE.md`, `docs/runbooks/guide-knowledge-distillation.md`,
  `scripts/vault.sh:25` comment, `scripts/utils.ps1:113` comment.
- Locate and repoint the SessionStart hook text that tells agents to run
  `./scripts/knowledge-crystallize.sh` (source not yet located, per CLI-021's tasks.md flip
  checklist).
- Retire or gate `tests/golden/crystallize/{capture.sh,ORACLE}` deliberately — deleting the
  oracle changes what the parity suite means; do not let it go red then get silently patched.
- Vault-side files (`00_meta/skills/crystallize/SKILL.md`, two pattern files,
  `CURRENT-STATE.md`) — separate commit direct to vault `master`, sequenced after this PR
  merges (vault commits bypass PR review per this repo's standing convention).

## Out of scope

- `vault-health.sh`, `vault-maintenance-weekly.{sh,ps1}` themselves, `obs-cli.{sh,ps1}`,
  `check-md-escapes.sh`, and the spec-gate cluster (`check-spec-gate.sh`,
  `check-backlog-integrity.sh`, `check-backlog-merged.sh`) — those stay in #492, blocked on
  CLI-022 (#491, spec gate port, unstarted) and CLI-021 increments 2-3 (`vault health`,
  `vault maintain`, unstarted).

## Verification

- Before deleting: confirm the *installed* `dotf` binary (not just the repo tree) has
  `vault crystallize` (two-tier deploy verification).
- `git diff --stat` after: only the callers/docs listed above plus the two deleted scripts.
[...]

Issue #492, opened 2026-06-21
CLI-023: vault + spec-gate cutover — repoint callers, delete the cluster


mlorentedev:
AUDIT-007 **Phase B / PR7**. Risk: high. Parent: `docs/adr/audit-007-cli-convergence-state.md`.

## Scope
Repoint CI / pre-commit / the `vault.sh` dispatcher / cron / aliases to the Go nouns, then delete the 7-script vault+spec-gate cluster + 2 .ps1.

## Deletes
- vault cluster: knowledge-crystallize.{sh,ps1}, vault-maintenance-weekly.{sh,ps1}, vault-health.sh, obs-cli.{sh,ps1}, check-md-escapes.sh
- spec-gate: check-spec-gate.sh, check-backlog-integrity.sh, check-backlog-merged.sh

## Depends on
- CLI-021 (vault noun built) + CLI-022 (spec gate built).

## Acceptance
- [ ] All callers repointed; cluster deleted; guard-grep clean; CI green on both OSes.

Issue #490, opened 2026-06-21
CLI-021: dotf vault — build crystallize/maintain/health (resolve noun collision)


mlorentedev:
AUDIT-007 **Phase B / PR5**. Risk: medium. Build+test only, no deletes. Parent: `docs/adr/audit-007-cli-convergence-state.md`.

## Scope
Resolve the `dotf vault` noun collision: today `dotf vault` is an entry-scaffolder (project/work), DISJOINT from knowledge crystallize/health/maintain. Build `vault crystallize` / `vault maintain` / `vault health` beside the shell; flip nothing yet.

## Related (cross-ref, different surface)
- #400 (unify dotf init project templates with vault SSOT), #461 (re-vendor vault templates / drift) — template surface, not the knowledge noun.

## Acceptance
- [ ] `dotf vault {crystallize,maintain,health}` implemented + table-driven tests; no twin deleted in this PR.

mlorentedev:
**State: spec written, implementation not started.** Unassigning so the board stops showing this as In Progress — nobody is working it.

The spec is on branch `feat/CLI-021-dotf-vault-knowledge` (pushed, no PR): `specs/CLI-021-dotf-vault-build-knowledge/`. It records three things worth not rediscovering:

- The port goes **under the existing `vault` noun**, not top-level — a top-level `dotf crystallize` would recreate the exact collision this ticket exists to resolve.
- **Build-beside, no deletes**, per the parent ADR (`audit-007`, row 5). The cutover is PR7 / CLI-023.
- The **full flip inventory** for CLI-023: 4 executable callers (`vault-maintenance-weekly.{sh,ps1}`, `setup-linux.sh:117`, `setup-windows.ps1:1593-1598`), 3 doc references, 4 vault files (incl. `crystallize/SKILL.md:55`, which already flags the pending convergence in its own text), and the SessionStart hook string. **None of these should be repointed until the flip** — the shell stays canonical through this ticket.

Two things block or shape the first increment:

1. **Open question:** does `vault health` mean the shell's local checks, or the Hive `vault_health` MCP tool? Same word, two surfaces. Blocks increment 2, not increment 1.
2. **The `.sh`/`.ps1` pair is already divergent** — found by #850 / BUG-060, which had to fix the same defect twice. So "port the twin" is ambiguous where they disagree; Linux is the reference, per the CLI-024 precedent.

[...]

PR #1489, opened 2026-09-05
feat(cli): dotf vault maintain, composing crystallize and health in-process

mlorentedev:
Increment 3 of CLI-021 — the last of the three subcommands that resolve the `dotf vault` noun collision. Built beside the twins: no caller repointed, no script deleted, the crontab (`setup-linux.sh:1605`) and Task Scheduler (`setup-windows.ps1:2185`) entries untouched. That cutover is CLI-023 (#492).

## The ticket's premise was partly false, and measuring it was the first work

The spec described three unbuilt increments. Measured against `origin/main` before writing anything:

| Increment | Real state |
|---|---|
| 1 · `vault crystallize` | done **and cut over** — the twins were deleted by CLI-050 (#1269) |
| 2 · `vault health` | **built** — `health.go` (528 LOC), registered, with a 16/16 parity suite |
| 3 · `vault maintain` | genuinely absent |

A comment at `cli/internal/cmd/vault.go:60` asserted health and maintain were "still unbuilt" while `newVaultHealthCmd()` was registered forty lines below it. Reading that comment instead of probing the target would have meant reimplementing increment 2. Lessons 256/257, again. The comment is corrected here.

## Composition is in-process, not `exec dotf`

The shell twin had no choice and pays for it: cron starts with a minimal PATH that excludes `~/.local/bin`, so its bare `dotf` call silently no-ops under `|| true` every Sunday. It hardens PATH to compensate and carries a regression guard (`tests/vault-maintenance-weekly.bats:147`) for the day someone removes that.

There is no PATH to harden when there is no subprocess. A bats case asserts this structurally by running the binary under `PATH=/usr/bin:/bin`.

This is not "improving while translating" — the defect removed is an artifact of subprocess composition, not a behaviour of the maintenance run.

## A finding is not a failure

The one real design decision, and the reasoning is in `RunMaintain`'s doc comment because it will outlive this PR.

`RunHealth`'s 0/1/2 answers *what did the report find*. Maintain's status answers *did the run do its job*. Propagating health's code would make the status depend on whether a desktop GUI happened to be running — nondeterministic from cron's view, and code 2 (GUI unreachable) would mail the owner a failed-job notice **every Sunday the laptop was shut**. False alarms on a weekly channel train the owner to ignore it, taking the desktop notification — the signal this command was built around — down with it.

So the finding is routed rather than dropped: report body to the log, count to the notification urgency, verdict to stdout for whoever ran it by hand. The only error is being unable to write the log.

`maintainExitCode` was **deleted** rather than made to `return 0` — a function ignoring both its parameters is a trap for the next reader who "fixes" it by wiring it to `os.Exit`. The `int` left `RunMaintain`'s signature with it.

This is deliberately *not* the fork `vault_health.go:23-33` took. There the oracle had a **gap** (nothing called it standalone); here the oracle made a **decision**, and it made it as a cron job.

## Linux is the reference

The `.ps1` twin runs no health step at all despite its own header claiming it does — already recorded in `divergences.md` §Increment 2, not re-decided here. `maintain` runs both steps on every OS, per the ADR-020 precedent CLI-024 set.

## Testing

**No golden corpus, deliberately.** The twin is a 52-line wrapper around two subcommands whose byte-parity is already proven by `knowledge-crystallize-go-parity.bats` and `vault-health-go-parity.bats`. A third fixture scheme would measure the same thing again. What is left is the wrapper's own behaviour.

- **22 table tests** (`maintain_test.go`) — issue regex, notification threshold, both OS log paths, composition against injected steps. No real vault, no real log location, no desktop bus.
- **4 end-to-end bats cases** against the built binary, alongside the 13 shell cases in the same file.
- **Mutation-verified, with the harness proving each mutation LANDED before believing the red** (lesson 267). Five mutations, five kills: notification threshold `>0`→`>1`; the code-1 verdict line silenced; the crystallize error swallowed; `LOCALAPPDATA` ignored; and health's exit code propagated — which reddens both the unit test and the bats case, so the decision above is guarded at both layers.

Two test defects were found by review, not by a red suite, and the second commit fixes them. One assertion ended in `|| true` — the shape flagged as Major on #1482 — and removing it turned the test red, revealing the assertion was *also* wrong (`grep -qE` is case-sensitive; the log lines are `[WARNING]`). The `|| true` had been hiding a defect in the test itself. The other case fired a real desktop notification on every local run.

`healthOptions()` is **extracted** from `vault_health.go` rather than copied — the `$VAULT_DIR`-then-ADR-025 cascade is a contract, and two copies drift. `DefaultLogFile` is split over `logFileFor(goos, home, localAppData)` so both OS branches are tested from either host; reading `runtime.GOOS` inline leaves the branch you are not running on covered only by the compiler.

## Spec corrections in the same PR

The next session inherits the ticket, not the transcript.

- The flip checklist's four "callers" rows all named `knowledge-crystallize.{sh,ps1}`, deleted by CLI-050. Replaced with the six that actually exist — including **`cli/internal/mem/session_start.go:135`, where the Go CLI shells out to the very script it is replacing**. That is the row that makes deleting `vault-health.sh` a Go change, not a shell change.
- The acceptance criterion said "`git diff` touches only `cli/` and `specs/`", which increments 1 and 2 both broke honestly (a twin port needs its tests). Corrected to the negative `verification.md` has actually been enforcing: nothing under `scripts/`, `setup-*.{sh,ps1}`, or the vault.
- Found while re-verifying, **out of scope and unfixed**: `harness/skills/crystallize/SKILL.md:70` tells agents to run `knowledge-crystallize.sh`, which CLI-050 deleted, naming the working command only as a parenthetical.

## Verification

`go build ./...`, `go vet ./...`, `GOOS=windows go vet ./...`, `go test ./... -count=1`, and `golangci-lint` at the `versions.conf` pin (v2.12.2) — all clean. Full bats suite 1546/1546.

Diff scope confirmed: nothing under `scripts/`, `setup-*.{sh,ps1}`, or the vault.

Not run on an actual Windows box. The `%LOCALAPPDATA%` path is unit-tested from Linux; the PowerShell toast in `NotifyDesktop` is unexercised anywhere and is fire-and-forget by design, so a failure there is silent by construction. Stated as a gap, not covered.

Refs #490

### DIFF
--- a/cli/internal/cmd/vault.go
+++ b/cli/internal/cmd/vault.go
@@ -45,6 +45,7 @@ Subcommands:
 	cmd.AddCommand(newVaultProjectCmd())
 	cmd.AddCommand(newVaultCrystallizeCmd())
 	cmd.AddCommand(newVaultHealthCmd())
+	cmd.AddCommand(newVaultMaintainCmd())
 	cmd.AddCommand(newSearchCmd())
 	return cmd
 }
@@ -59,11 +60,20 @@ Subcommands:
 //
 // Built beside the twins in CLI-021, then cut over in CLI-050 (#1269): the
 // shell/PowerShell pair is deleted, every caller repoints here, and this is
-// the sole implementation. `vault health` and `vault maintain` (increments 2
-// and 3 of #490) are still unbuilt, and the rest of the cluster
-// (vault-health.sh, vault-maintenance-weekly.{sh,ps1}, obs-cli, the spec-gate
-// scripts) is still shell — that full cutover is CLI-023 (#492), blocked on
-// those plus CLI-022.
+// the sole implementation.
+//
+// All three of #490's increments are now built — crystallize (this file),
+// health (vault_health.go) and maintain (vault_maintain.go). Only crystallize
+// is CUT OVER; the other two are built beside twins that are still the
+// canonical invocation (scripts/vault-health.sh and
+// scripts/vault-maintenance-weekly.{sh,ps1}, the latter still wired to cron at
+// setup-linux.sh:1605 and to Task Scheduler at setup-windows.ps1:2185). That
+// cutover, plus the rest of the cluster (obs-cli, the spec-gate scripts), is
+// CLI-023 (#492), blocked on CLI-022.
+//
+// This comment previously said health and maintain were "still unbuilt" while
+// health was registered forty lines below it. Lessons 256/257: probe the
+// target, never trust a block's own comment about itself.
 func newVaultCrystallizeCmd() *cobra.Command {
 	var all bool
 
--- a/cli/internal/cmd/vault_health.go
+++ b/cli/internal/cmd/vault_health.go
@@ -60,31 +60,7 @@ missing from PATH), 2 the GUI is unreachable.`,
 		SilenceUsage:  true,
 		SilenceErrors: true,
 		RunE: func(c *cobra.Command, _ []string) error {
-			name := vaultName
-			if name == "" {
-				name = os.Getenv("VAULT_NAME")
-			}
-			if name == "" {
-				name = "knowledge"
-			}
-
-			dir := os.Getenv("VAULT_DIR")
-			if dir == "" {
-				dir = env.ResolvePath("VAULT_PATH")
-			}
-			if dir == "" {
-				if home, herr := os.UserHomeDir(); herr == nil {
-					dir = home + "/Projects/knowledge"
-				}
-			}
-
-			code, err := vault.RunHealth(c.OutOrStdout(), vault.HealthOptions{
-				VaultDir:   dir,
-				VaultName:  name,
-				Verbose:    verbose,
-				ScriptsDir: memScriptsDir(),
-				BashPath:   mem.ResolveBash(),
-			})
+			code, err := vault.RunHealth(c.OutOrStdout(), healthOptions(vaultName, verbose))
 			if err != nil {
 				return err
 			}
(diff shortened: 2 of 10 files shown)


mlorentedev:
## Review triage

|---|---|---|
| **CodeRabbit** | *"Review limit reached — next included review available in 52 minutes"* | **No review ran.** This is a notice about the review, not a review. Recorded as a disclosure, not dispositioned — there is nothing to dispose of. |
| **PR-Agent** (`## PR Reviewer Guide`) | *"No major issues detected"*, *"No security concerns identified"*, *"PR contains tests"*, effort 3/5 | **Accepted.** No code findings to apply. |
| **PR-Agent — ticket compliance** | Three sections, all "Partially compliant", all with `(none)` compliant | **Declined, with reason below.** |
| **CI** | Every check green except `test-windows`, still running | Open — will be re-read before merge. |

### Why the ticket-compliance section is declined

This is **TOOL-020 (#1466)** again, and this instance is a sharper case than the ones already on that ticket. It read three numbers out of the PR body and treated each as a requirement this PR must satisfy:

- **#492** is CLI-023 — the *future* cutover PR, which this PR's body names four times as explicitly out of scope.
- **#1269** is CLI-050, a **merged PR**, not a ticket.
- **#1482** is also a **merged PR**, cited only as context for an unrelated CI measurement.

It then listed as "non-compliant requirements" the exact things this PR's acceptance criteria require it **not** to do:

> - No callers repointed
> - No scripts deleted
> - No setup script changes
> - No workflow changes

Those are the spec's stated non-goals, and `verification.md` calls the negative *"the proof that matters most here"*. The tool inverted a stated non-goal into an unmet requirement — which is a stronger failure than "treated a PR number as a ticket", because the PR body states the constraint in prose and the analysis contradicts it.

Its own headline (*"No major issues detected"*) also disagrees with its rows, which is the second half of #1466's description.

Measurement added to #1466.

### Disclosure

**One reviewer produced substantive output, not two.** CodeRabbit's free-tier quota was exhausted, so its comment is a notice. This PR has been reviewed by PR-Agent alone.

PR #1276, opened 2026-08-27
feat(cli): cut over knowledge-crystallize.{sh,ps1} to dotf vault crystallize

mlorentedev:
## Summary
- Deletes `scripts/knowledge-crystallize.{sh,ps1}` — CLI-021 (#490) already built `dotf vault crystallize` beside them with byte-identical golden parity; this is the cutover.
- Repoints every caller: the weekly maintenance script, `setup-linux.sh`/`setup-windows.ps1` deploy steps, the SessionStart hook's staleness message, and the docs/comments that told a human to run the shell script.
- Sliced out of CLI-023 (#492), which bundles this with unstarted work (spec-gate, `vault-health`, `vault-maintain`) into an all-or-nothing 9-script deletion — too large a unit to land now. #492 stays open for the rest.

## Why not the full #492
CLI-022 (#491, spec-gate port) hasn't started, and CLI-021 increments 2-3 (`vault health`, `vault maintain`) aren't built. Crystallize is the one piece with full golden-parity coverage already proven, so it ships now on its own ticket (CLI-050 / #1269) rather than waiting on unrelated work.

## Test plan
- [x] `cd cli && go build ./... && go vet ./... && go test ./...` — all packages ok
- [x] `GOOS=windows go vet ./...` — clean
- [x] `golangci-lint run` (pinned 2.12.2) — 0 issues
- [x] `shellcheck --severity=error setup-linux.sh scripts/*.sh` (CI's actual invocation) — exit 0
- [x] `bats tests/*.bats` — 1462 tests, 0 failures
- [x] Verified no coverage gap before deleting the shell-invoking bats files: `TestIsYAMLBlockScalar` + `TestProcessProjectRefusesYAMLAndLeavesFileIntact` in `cli/internal/vault/crystallize_test.go` already cover what `knowledge-crystallize-yaml-guard.bats` asserted against the shell
- [x] Fixed two bats files the deletion broke (a false-passing filename-grep test in `setup-windows.bats`, and `vault-maintenance-weekly.bats`'s sandbox stub that still shelled out to a deleted script name) rather than leaving them red or silently deleting them

Full details in `specs/CLI-050-knowledge-crystallize-cutover/`.

Closes #1269

PR #1490, opened 2026-09-05
fix(harness): the crystallize skill told agents to run a script deleted in #1276

mlorentedev:
Two lines, and the second is why the first survived.

## The defect

`harness/skills/crystallize/SKILL.md:70`, step 7 of a skill agents follow in order:

```
- Run: `knowledge-crystallize.sh` (or `dotf vault crystallize`)
```

`scripts/knowledge-crystallize.sh` has not existed since **#1276** cut it over to the
CLI. The instruction names the dead one first and demotes the working one to a
parenthetical, so an agent reading top to bottom runs the script that is gone.

Found by a peer session while reviewing something else; verified here before acting on it.

## It is one line, not a class — measured before assuming

Ten bare filenames in `harness/skills/` resolve to nothing on disk. That number is
misleading, and checking which of them this repository *ever tracked* is what separates
them:

```
knowledge-crystallize.sh       tracked-in-history: 2   <- added, then deleted
deploy.sh / entrypoint.sh / head.sh / helm.sh
hermes-startup.sh / script.sh / validate.sh
vault-pull-daemon.sh / vault-validate.sh        tracked-in-history: 0
```

Nine of ten are generic example names in prose or files belonging to the Hermes agent
repo. So **one real, nine noise** — and that ratio is not just a filter, it is a
measurement of the exact false-positive rate that made `check-doc-paths.sh` stop
resolving bare filenames:

> Bare filenames are NOT checked. Resolving them by basename was tried and reverted: it
> flagged vault patterns, `machine.json` and `review.md` on AGENTS.md — files that
> legitimately live elsewhere. The cost is a real blind spot (a backticked `missing.md`
> passes), accepted deliberately.

**That decision stands, and this PR does not widen the guard.** The cost it named was
accepted knowingly; this is the cost being paid once, with evidence, not a reason to
reopen it. A basename-resolving guard would have flagged all ten.

## The second edit

`check-doc-paths.sh`'s auto-discovery excluded `harness/` with this reason:

```
#   harness/ — generated artifacts from vault; checked by compile-harness.sh --check
```

`--check` covers **enforced-region drift and clean rendering**. It does not resolve paths
named inside skill prose — and the same script's own header already warns
*"do not read a passing --check as type validation"*. So the exclusion promised a
coverage nobody had, and that is what stops the next reader looking. Corrected to state
what is actually covered, and which blind spot really owns this case.

This is the shape lesson 256 records: reading the block's own comment is what fails;
probe the target. The comment was probed, not believed.

## Verification

```
bash -n / zsh -n scripts/check-doc-paths.sh   OK
shellcheck --severity=error                   OK
./scripts/check-doc-paths.sh                  exit 0 (guard still passes)
./scripts/compile-harness.sh --check          exit 0
tests/check-doc-paths.bats                    19/19 pass
pre-commit (5 hooks)                          all passed
```

No test added, deliberately. The change is a corrected instruction and a corrected
comment; the guard that would assert the first is the one measured above as
9-of-10 false positives, and asserting a comment's prose is not a thing to build.

### DIFF
--- a/harness/skills/crystallize/SKILL.md
+++ b/harness/skills/crystallize/SKILL.md
@@ -67,7 +67,7 @@ tags: [tag1, tag2]
 - Update `## Last Crystallized: YYYY-MM-DD` to today
 
 ### Step 7 — Stamp
-- Run: `knowledge-crystallize.sh` (or `dotf vault crystallize`)
+- Run: `dotf vault crystallize` (add `--all` to stamp every project)
 - This updates currentDate and Last Crystallized automatically
 
 ### Step 8 — Report
--- a/scripts/check-doc-paths.sh
+++ b/scripts/check-doc-paths.sh
@@ -68,7 +68,15 @@ if [ "$#" -eq 0 ]; then
     # Auto-discover instruction files when invoked without arguments (BUG-088, #1021).
     # Governed files: all active agent instructions and READMEs.
     # Exclusions for stated reasons:
-    #   harness/ — generated artifacts from vault; checked by compile-harness.sh --check
+    #   harness/ — rendered from the vault; compile-harness.sh --check covers it,
+    #              but ONLY for enforced-region drift and clean rendering. It does
+    #              not resolve paths named in skill prose, and #1486's session found
+    #              harness/skills/crystallize/SKILL.md telling agents to run
+    #              knowledge-crystallize.sh, deleted by #1276. Recorded because the
+    #              earlier wording here read as "harness/ is covered", full stop,
+    #              which is how that line survived. Not closed by widening this
+    #              guard: the reference is a BARE FILENAME, so the blind spot
+    #              documented above owns it, not this exclusion.
     #   specs/   — per-feature historical proposals and archived logs, not standing instructions
     #   docs/    — historical decision records/lessons mentioning retired scripts by design
     if command -v git >/dev/null 2>&1 && { [ -d "$REPO_ROOT/.git" ] || [ -f "$REPO_ROOT/.git" ]; }; then


mlorentedev:
## Review triage

**Neither reviewer produced a code finding on this PR, and one of them did not review it at all.** Saying so explicitly, because a green rollup plus two bot comments reads like two reviews.

|---|---|---|
| **CodeRabbit** | *"Review limit reached — next included review available in 48 minutes… You've used all free OSS reviews for now."* | **Not a review — a notice.** It names the two files it *would* have read and then does not read them. This PR should not be recorded as having had CodeRabbit's opinion. |
| **PR-Agent — code** | `No major issues detected`, no security concerns, effort 1/5. No findings. | **Accepted.** Nothing to apply. |
| **PR-Agent — "PR contains tests"** | False. | **Declined, and worth naming.** This PR adds no test, deliberately and explicitly ("No test added, deliberately", with the reason). The flag appears to be inferred from the diff touching a path near tests rather than from a test existing. |
| **PR-Agent — ticket compliance** | Read **#1276** as a ticket this PR must satisfy, and listed as *non-compliant*: "Delete `scripts/knowledge-crystallize.{sh,ps1}`", "Repoint all callers", "test plan verifications not evidenced". | **Declined.** #1276 is a **merged PR**, not a ticket, and it is the PR that *already did* those things. Its deletion of those scripts is the premise of this change — the whole defect is that a skill kept naming one of them afterwards. So the analysis asks this PR to perform work whose completion is the reason it exists, and to delete files that do not exist. |

## On that last row

This is the fifth or sixth instance tonight of PR-Agent inferring a ticket from a *PR* number and then inverting the relationship. The peer session filed it on **#1466**; this is another instance rather than a new defect, so it goes here and not into a second ticket — we already duplicated one issue this evening and it is not free.

It is the sharpest instance so far, though, and worth recording on #1466 for that reason: the earlier cases misread scope, while this one asserts that a PR failed to do the thing its own referenced predecessor did. A reviewer that treats "already done upstream" as "not done here" will mark every follow-up fix non-compliant.

## CI

11/11 green, including `test-windows` and `integration`. Verified via the REST `check-runs` endpoint — `gh pr checks` is GraphQL and was refusing at the time, while `gh api rate_limit` simultaneously reported 5000/5000 remaining because it cannot see the secondary limit.

PR #1492, opened 2026-09-05
docs(lessons): 268 — a refused question and a negative answer print the same thing

mlorentedev:
Seven instances in one evening, across two sessions, through five different mechanisms. Every one was a query whose **failure** and whose **negative answer** produced identical output — and in every case the negative answer was the thing being looked for, so silence read as a finding.

| # | Query | Printed | Actually meant |
|---|---|---|---|
| 1 | `gh pr checks \| grep -iE 'test-windows\|fail\|pending'` | nothing | the command had exited with `GraphQL: API rate limit already exceeded`; the grep was searching an error message |
| 2 | a poll loop on the peer session | `(no test-windows yet)` | the API declined to answer |
| 3 | `gh project item-list --limit 200` | `#490 not found on the board` | the page was capped |
| 4 | the same at `--limit 400` | found `#490` | **kubelab's** #490 — the board is cross-repo and a number identifies nothing alone |
| 5 | `gh api rate_limit` | `core 5000/5000, graphql 5000/5000` | GraphQL's **secondary** limit was exhausted, and that endpoint does not report it |
| 6 | `echo "LINT_EXIT=${PIPESTATUS[0]:-done}"` | `LINT_EXIT=done` | zsh spells it `$pipestatus[1]`; the bash form expands to nothing |
| 7 | `grep -qE … \|\| true` | green | an assertion that could not fail |

**Instance 5 is the sharpest.** The diligent move — *check whether you are rate limited before believing you are* — returned a confident, precise, wrong answer. The endpoint was working perfectly and answering truthfully about a budget that was not the constraint. A health check aimed at the wrong subsystem is worse than none, because it converts "I do not know" into "I verified".

**Instance 7 is the one that bit hardest here.** Removing the `|| true` turned the test **red**, and the red was correct: `grep -qE` is case-sensitive and the log lines are `[WARNING]`. The suppressor was hiding a defect in the *assertion*, not in the code under test.

**Instance 6 is a row in this repo's own `.claude/CLAUDE.md` prohibited-patterns table**, tripped by an agent that had read the table. That is the argument for the lesson existing at all: knowing the class by name does not prevent the instance.

**An eighth arrived while opening this PR.** `gh pr create` failed with `GraphQL: API rate limit already exceeded` one minute after the file was written; this PR was opened over REST instead. Recorded here rather than added to the lesson, which already has enough instances to make its point.

## Why it is not covered by 267

Lesson 267 is a mutation that silently failed to **apply**, producing a passing test. This is a query that silently failed to **run**, producing an empty result. Both are absence read as information, from opposite directions. 265 and 266 are the measurement-shaped members of the same family.

The through-line: **the reassuring reading is always the one that requires no further work**, and nothing in the tooling will ever prompt for the other one.

## Practical rules it lands on

- Separate the exit status from the content — a pipeline that greps a command's output has already discarded the bit that distinguishes the two cases.
- Make the absent case *say* absent: `… || echo "(none)"`.
- A result of exactly `--limit N` is the ceiling, not data.
- Ask the specific object, not the containing set — "which project is this issue in" is one call and cannot truncate.
- Never `|| true` an assertion.
- In a shared-quota environment, do not poll: five sessions share 5000/hr, and a 30-second watcher makes *everyone's* checks unreadable rather than slow.

Docs-only. Branched off `7b00e84` so the `_index.md` row lands after 267's rather than conflicting with it.


mlorentedev:
## Review triage

|---|---|---|
| **CodeRabbit** | *"Review limit reached"* | **No review ran.** Notice, not a review. Disclosed. |
| **PR-Agent** | `## PR Reviewer Guide`; ticket compliance against **#490** | **Declined** — see below. |
| **CI** | green | — |

### Why the ticket analysis is declined

This is a **docs-only** PR adding one lesson file and one index row. It carries no ticket reference, closes nothing, and touches nothing under `cli/`, `scripts/` or `specs/`. PR-Agent nonetheless evaluated it against **#490** — CLI-021, a spec archived by a different PR tonight — and reported `(none)` compliant.

**#1466** (TOOL-020) again, and the third instance today. What makes this one useful: on **#1496**, one PR away and the same hour, the same tool read #490 *correctly*, listing three accurate compliant requirements with evidence attributed across #882, #1489 and #1269. So the section is not uniformly broken — it is right when the PR genuinely implements the ticket and wrong when it has to decide whether a referenced number is a requirement at all. Recorded on #1466, because a conditional defect can be scoped; an unconditional one can only be switched off.

### Content note

The lesson gained an eighth instance after the initial review (`3934cdb`): `gh api .../check-runs` returns entries from **every** run on a SHA including superseded ones, so `cancelled` from a replaced run is spelled identically to `cancelled` from a timeout. Found by the parallel session on its own PR while this document was being written — which is the point the lesson makes, arriving unprompted.

Issue #1493, opened 2026-09-05
HARNESS: compile-harness.sh --refresh silently REVERTS repo-side skill fixes when the vault is behind, and nothing distinguishes that from a forward regen


mlorentedev:
## What happened

On a clean checkout at `52035f1`, `./scripts/compile-harness.sh --refresh` modified six files. One was the record I was migrating. Four were legitimate forward regeneration (the vault holds the auto-merge exception paragraphs, the repo did not). **One was a reversion**: it rewrote `harness/skills/new-ticket/SKILL.md` back to an older, worse version, deleting work that had landed in the repo.

What `--refresh` removes from that skill:

- The `gh api --paginate` REST recipe for scanning issue ids, **and the warning that explains it**: *"Never hand-split `--paginate` output… dropping pages silently and returning a plausible-but-low maximum"*, citing `docs/lessons/lesson-263-*` and the 2026-09-02 incident where a scan proposed `CLI-072`, an id another session already held.
- The four-step **yield rule** for id collisions (spec folder → inbound references → board `ID` field → higher `#number`), with its measurement: *"Ordering was the whole rule until 2026-09-02, and it was wrong in 3 of 11 collisions resolved that day."*
- The `wc -l < /tmp/nnn-scan.txt` population check — *"a short scan is a wrong answer"*.

What it puts back in their place: `gh issue list --repo … --limit 800`, and a one-line yield rule that is the exact ordering rule measured wrong in 3 of 11 cases.

`gh issue list` is **GraphQL**. It is the command that was refusing all evening under GitHub's secondary rate limit while REST answered fine — which is precisely the failure the removed comment exists to warn about, and precisely when parallel sessions are filing tickets and the scan matters most.

## Why this is not #836

**#836 is phantom drift**: `--refresh` faithfully pulls a vault that is mid-write, `doctor` reports FAIL, and re-running minutes later is clean. Transient, and nothing is lost.

This is the opposite. It is **persistent, one-way, and it destroys content**. The vault is behind the repo for this record and will stay behind until someone pushes the repo's version back, so *every* `--refresh` on *every* checkout re-applies the reversion. Two sessions ran `--refresh` tonight.

## Why nobody notices

`--refresh` treats the vault as the SSOT unconditionally, which is correct as a rule and silent as an implementation: there is no signal that distinguishes

- "the vault is newer, this is a forward regen" from
- "the repo is newer, this is a revert",

[...]

PR #1494, opened 2026-09-05
feat(harness): builder becomes the second enforceable persona, gating two of nine

mlorentedev:
Refs #1497. Second persona migrated to the enforceable form; five to go.

> **Reference corrected after review.** This originally said `Refs #1420`, and PR-Agent was right to push on it: #1420 is *"no capability maps to the skill-invocation tool"*, which this PR does not advance. It was referenced because it was the nearest open issue — and the ticket this PR *does* advance did not exist. Filed as **#1497**. `Refs`, not `Closes`: five personas remain.

## Why this is the step that matters

The orchestrator's plumbing is already live on this machine — `PreToolUse` runs `dotf harness gate --harness claude`, `UserPromptSubmit` runs `dotf harness suggest --from-hook`. What was missing is not machinery, it is **data**:

| persona | form | `enforce:` declarations |
|---|---|---|
| **reviewer** | mapping | 4 |
| **builder** | *inline* | **0** |
| architect, curator, planner, shipper, hermes-nan | inline | 0 |

One of seven could bind anything. The rest resolve and the gate acts on nothing — so the `[persona]` line on every prompt was a *suggestion*, not an obligation. `builder`'s own prose meanwhile said its skills were "**enforced** by hook", which was false for all nine of them.

## What changed

`test-driven-development` and `test` carry `enforce: warn`. **The other seven stay bare strings** — declared, reported by `UnmigratedSkills`, ungated.

That split is the design, not a compromise. `debug-hardware`, `mcp-builder` and `async-python-patterns` are irrelevant to nearly all work in this repository, and a gate that names them on every call teaches the reader to scroll past `[gate]` lines. The severity is worth exactly as much as the attention it still commands. The two that remain are the two that hold whatever the task is: a build-phase change without a test that would have failed without it is not finished, and TDD's rule is *before* the implementation.

The mixed list is a supported parse path, not a formatting accident:

- `- golang-pro` → `EnforceUnset`, the gate refuses to act on it;
- `- id: golang-pro` with no `enforce:` → **hard parse error** (`severity is required in the mapping form`).

That asymmetry is what makes a partial migration expressible without inventing a third state, and it is why the loader applies no default: defaulting to `warn` would make every unmigrated persona *"silently inert while every check reported it as wired"*, and defaulting to `block` would turn every declared skill into a hard gate overnight.

## Verified behaviourally, not structurally

Against a `dotf` built from `main` (the local binary is a `dev` build and would have measured a different branch):

```
$ gate --role builder            # nothing consumed
[gate] warn: test not consumed
[gate] warn: test-driven-development not consumed

$ gate ... Skill{test-driven-development}   # consume one
$ gate --role builder
[gate] warn: test not consumed
```

Exactly the two gated skills, **none of the seven ungated ones**, and the warning drops as the skill is consumed. Note the gate reads the **deploy dir** by default, so this used `--repo-root`; the change is live on a real machine only after a deploy.

## Tests

Two gaps, both real:

1. **The mixed form had no coverage.** The existing cases are pure-mapping and pure-flat. The shape this whole design rests on was untested.
2. **Nothing asserted that a migrated persona stays migrated** — and that is not hypothetical. `compile-harness.sh --refresh` regenerates these records from the vault, so a checkout whose vault half is stale rewrites a migrated record back to inline and the gate goes quiet with nothing red. Filed as **#1493**, found while doing exactly this. Asserted by name rather than by count: each migration is a decision, and a count makes the test a chore on every edit.

I did **not** add a redundant test — my first draft duplicated `TestPersonaRefusesToReadSkillsAsEmpty/mapping_without_enforce`, and I deleted it rather than leaving two assertions of one property.

## One existing test had to move, and the reason generalises

`TestEveryGateDecisionLeavesADurableRecord` pinned the **live roster**, using `builder` as its example of "a persona whose skills carry no severity". So it went red the moment the migration it was implicitly measuring succeeded — a failure with nothing to do with durable records.

Repointed at a fixture (`unmigratedRepoRoot`), mirroring the `blockingRepoRoot` helper already beside it. The neighbouring `TestRealPersonaRecordsLoadAndReportTheirMigrationState` refuses to pin the migration count for the same reason; this is that refusal one layer down. **A test keyed to migration progress is an incentive to leave a persona unmigrated so the test keeps passing.**

## Not in this PR

- **The vault definition** is committed separately to the knowledge repo's `master` (`08a3063e`) — that is the SSOT; `harness/agents/builder/AGENT.md` here is the regenerated record. Staged by explicit path, with a concurrent session's untracked file in that tree left untouched.
- **`--refresh` also produced five unrelated files**, four forward regen and one reversion. All five reverted out; the reversion is #1493 and it is already sitting uncommitted in the main checkout.
- **No severity is `block`.** #1434 stands: a *named* subagent dispatch sends its name as `agent_type`, so its role never resolves and enforcement silently turns off. Nothing is promoted until a real dispatch is observed resolving a persona, with two dispatches of one role carrying different `agent_id`s.

## Verification

```
go build ./... / go vet ./...        OK
GOOS=windows go vet ./...            OK
go test ./... -count=1               all packages ok
golangci-lint run                    0 issues (2.12.2, matches versions.conf pin)
tests/*.bats                         1554/1554 pass, exit 0
compile-harness.sh --check           OK: no harness drift
```


mlorentedev:
## Review triage

|---|---|---|
| **CodeRabbit** | *"Review limit reached — next included review available in 34 minutes."* | **Not a review — a notice.** No files, no lines, no claims. This PR has not had CodeRabbit's opinion and should not be read as having had it. |
| **PR-Agent — code** | Effort 2/5, no security concerns, **"Recommended focus areas for review" empty**. | **Accepted.** No code findings. |
| **PR-Agent — ticket compliance on #1420** | *"all acceptance criteria are unmet by this PR"*, listing five unmet requirements. | **Applied, not declined** — see below. |
| **CI — `integration`** | `failure` | **Infrastructure, not this diff.** Re-run requested. |

## PR-Agent was right, and I fixed the cause rather than the symptom

Its five "non-compliant requirements" against #1420 are all accurate: this PR adds no skill-invocation capability, does not touch `resolve-capabilities`, and proves no persona can *invoke* a skill. It is a `Refs`, not a `Closes`, so nothing was claimed — but the reference was still wrong, and the reason is worth stating rather than waving away:

**#1420 was referenced because it was the nearest open issue, and the ticket this PR actually advances did not exist.** Migrating five remaining personas to the enforceable form is the standing orchestrator goal's next step and had no ticket at all — board drift, not a labelling slip. Filed as **#1497**, and this PR now refs that.

So the disposition is *applied*: PR-Agent surfaced a real defect by way of a misframed one. Its own framing ("the target ticket's ACs are unmet") is the same PR-vs-ticket inference recorded on #1466; here it happened to point at something true.

## `integration` failed on an Ubuntu mirror, not on this diff

```
Connection failed [IP: 91.189.91.82 80]
E: Failed to fetch .../binary-amd64/Packages.gz
   File has unexpected size (742980 != 742981). Mirror sync in progress?
ERROR: process "/bin/sh -c apt-get update && apt-get install -y ..." exit code: 100
```

The Docker build never completed, so **no test ran** — this is not a red test, it is a job that did not get to start one. Re-run requested.

Worth noting as a gap rather than a fix here: the `test` job already wraps `apt-get update` in a three-attempt retry with explicit `Acquire::Retries` and timeouts, added for exactly this failure mode. `tests/Dockerfile.integration` has no equivalent, so the same transient takes the whole job down. Filing that separately rather than widening this PR.

## Everything else green

`changes`, `lint`, `lint-powershell`, `test`, `test-windows` — all success. `test-windows` passed **without** the pi reconcile, correctly: this diff touches `harness/**` and `cli/**` but nothing in the `pi` filter, which is CI-002 behaving as designed.

Local, before pushing: `go build` / `go vet` / `GOOS=windows go vet` clean, `go test ./... -count=1` all packages ok, `golangci-lint` 0 issues at the pinned 2.12.2, `tests/*.bats` 1554/1554, `compile-harness.sh --check` no drift.

And the round-trip, re-verified after the vault was fixed in #1493: `--refresh` reproduces `harness/agents/builder/AGENT.md` **byte-identically** — the record does not appear in `git status` at all, so this PR is not carrying a stale regen.

mlorentedev:
## Review triage — second pass, posted after the merge

Disclosure first: this PR merged at 02:52:43Z with reviewer output that my triage did not cover. Both bots re-ran on the `update-branch` head `09e53c5` and **edited their existing comments in place** — CodeRabbit at 02:44:37Z, PR-Agent at 02:45:38Z — while the triage above is timestamped 02:15:09Z.

`dotf pr triage-queue` returned **exit 0** throughout. It compares `created_at`, and an edit in place does not move it. **This is the fourth recorded occurrence of that exact failure**, and it is the reason this comment exists rather than nothing: the queue said clear, and the queue was wrong. Nothing here changes the merge — the findings are dispositioned below and none of them is a defect in the diff — but "green queue" was not evidence that anything had been read.

| source | output on `09e53c5` | disposition |
|---|---|---|
| **CodeRabbit** | *"Review limit reached — next included review available in 56 minutes."* | **Still not a review.** Second notice on this PR. It names the three files it *would* have read and the base it would have diffed, which reads like work product and is not one. **This PR merged without CodeRabbit's opinion, both times of asking.** |
| **PR-Agent — effort/security** | Effort 2/5, no security concerns, PR contains tests. | Accepted. |
| **PR-Agent — ticket compliance (#1497)** | "Partially compliant": four ACs unmet. | **Declined, with the reason below.** |
| **PR-Agent — focus area: "Documentation Accuracy"** | Claims `builder` lacks the `skill` capability, so "the gate can warn but the persona cannot act on the warning". | **Declined — the claim is false.** Refuted below. |

### The focus area is factually wrong, and it is worth showing why

PR-Agent's only code-adjacent finding asserts that no persona has a capability permitting skill invocation. Three independent artifacts say otherwise:

1. `harness/agents/builder/AGENT.md:13` — `capabilities: [read, search, edit, shell, skill]`.
2. The deployed `~/.claude/agents/builder.md` frontmatter — `tools: Read, Glob, Grep, Edit, Write, Bash, Skill`.
3. The live dispatch surface in this session lists `builder` with `Skill` among its tools.

So a `builder` dispatch can invoke the skills the gate names. The documentation caveat PR-Agent asks for would document a constraint that does not exist, which is worse than silent — it would be the *record* claiming a limitation while the *system* has none, the mirror image of the defect this PR fixes.

### The unmet ACs belong to the epic, not to this PR

The four it lists — the other personas unmigrated, "all seven declare an explicit `enforce`", and a genuine dispatch shown producing `warn` — are #1497's acceptance criteria for the **whole migration**. #1497 states *"one persona per PR"* in its own What section, and PR-Agent's compliant list includes that line. So it is measuring a five-PR epic against PR two of five and reporting the remainder as non-compliance. Same PR-vs-ticket inference recorded on #1466 and #1440.

### One of them is real, though, and it is not closed by declining the framing

*"Verification that a genuine dispatch of `builder` produces `warn` in the decision record"* — that has **not** been observed, and the previous field test is the evidence that it matters rather than evidence against it. Five real dispatches were read back from the #1435 record; the `builder` one produced **`allow`, not `warn`**, because the *deployed* record was still flat while the checkout carried the migrated one. Role resolution and severity enforcement fail independently, and only the decision record tells them apart.

It could not be closed before this merge, for a reason worth stating rather than waving: the gate reads the **deploy dir**, which is owned by `setup-linux.sh` / `dotf deploy`. Hand-copying one record into it to manufacture the observation would be a second writer on a surface with one owner — and it would prove the gate reads a file I placed, not that a deploy produces the state. The honest sequence is deploy, then dispatch, then read the record.

**Carried to #1497 as a precondition on the remaining migrations**, not left here. It is also the standing blocker on any `enforce: block` promotion, which #1434 already holds for a different reason.

### CI

All 14 checks success on `09e53c5`, one skipped (`goreleaser release`, expected on a PR). `integration` passed on its re-run after the Ubuntu mirror failure filed as #1498 — infrastructure, no test had run.

Issue #1241, opened 2026-08-26
HARNESS-084: the capped doctrine payload has outgrown its hardest target twice, and has no budget policy


mlorentedev:
## Problem

`.gemini/GEMINI.md` receives the compact doctrine payload against a **hard platform limit**: Antigravity caps *each* rules file at 12000 characters, so an overflow is not rejected loudly — the excess is simply not read.

It has now been breached twice, and both times the only thing that noticed was a bats assertion:

| When | Size | Cause | Rescue |
|---|---|---|---|
| 2026-08-22 | 12114 | a doctrine id was added to `doctrine.inject` | `render_region_compact` — drop the provenance blockquote (737 chars) |
| 2026-08-25 | 12045 | the roster went from one persona to seven | drop autonomous entries, compact a restatement (135 chars) |

Current headroom after the second rescue: **90 characters.**

## Why a third rescue is the wrong answer

Both fixes were correct in themselves — each dropped meta-text and kept every rule byte-for-byte, which is the right principle. But they were *reactive*, and each one bought less headroom than the last while the payload's growth drivers stayed unbounded:

- **`doctrine.inject`** grows whenever a standing rule is added. There are nine ids today and nothing caps their total.
- **The persona block** grows with the roster. Six invocable personas today; the roster's own "Open / provisional" section already contemplates an `sre`/observability role and a `platform` split. **The next persona breaches the cap again** — a role with a skill list the size of `builder`'s is 150 characters on its own, and there are 90 left.

So the next addition to either driver silently fails a platform limit, and the failure surfaces as a red bats assertion on somebody's unrelated PR — which is exactly how it surfaced this time.

## What a fix looks like

The question is not "how do we save characters once more" but "what is this payload's budget, and who pays when it is exceeded". Options, cheapest first:

1. **Report the margin, don't just assert it.** The bats test currently fails at the cliff edge. Having `dotf doctor` (or the test) print remaining headroom per target turns a cliff into a gauge — you find out at 200 characters left, not at −45.
2. **Give the persona block its own sub-budget** and degrade deterministically when it is exceeded: full skill lists while they fit, role names plus a pointer once they do not. A declared degradation is honest; discovering the limit through a red test is not.
[...]

PR #1495, opened 2026-09-05
chore(harness): land the pending regen, and stop --refresh reverting the new-ticket skill

mlorentedev:
Lands a harness regen that has been sitting **uncommitted in the primary checkout for twenty hours** (all six files carry mtime `2026-09-04 00:03:29`, one `--refresh` run), and fixes the reason one of those six was a *reversion* rather than a regen.

## What is here

**Two commits.**

`718e748` — the four genuinely-forward files: `AGENTS.md`, `ai/claude/CLAUDE.md`, `ai/orca/ORCA.md`, `harness/enforced/no-auto-merge.md`. Purely additive: the two paragraphs the vault's `pattern-git-workflow` already holds (the narrow machine-generated/machine-verified auto-merge exception with its four conditions, and the rule that a branch prefix may be matched to **forfeit** something but never to **grant** it), plus the generated-region sha bump `d96bda9e8359af7f → c069c43e9923fab1`.

`80839ac` — `harness/skills/new-ticket/SKILL.md`'s `generated_sha`, after the underlying vault record was restored.

## Why the sixth file needed a vault fix, not a commit

`compile-harness.sh --refresh` treats the vault as SSOT **unconditionally**, which is correct as a policy and silent as an implementation: "the vault is newer, this is a forward regen" and "the repo is newer, this is a revert" both arrive as `M harness/…`. In that run, four were the first and one was the second.

The vault's `00_meta/skills/new-ticket/SKILL.md` was **three weeks stale** (mtime 17 Aug). So every `--refresh` on every checkout was rewriting the repo backward, deleting:

- the `gh api --paginate` REST recipe and its warning,
- the *"never hand-split `--paginate` output"* paragraph citing lesson 263 and the 2026-09-02 incident where a silently-short scan proposed `CLI-072` — an id another session already held — putting `TEST` at `001` when the real maximum was `008`,
- the `wc -l` population check,
- the four-step yield rule and *"renumbering is a title-only edit"*,

and restoring in their place:

```bash
gh issue list --repo "$REPO" --state all --limit 800 --json title | python3 -c "
```

**Two independent defects in that one line.** `gh issue list` is GraphQL, which has been refusing all evening under a secondary rate limit across five concurrent sessions on this account — the exact conditions the deleted comment names (*"when parallel sessions are filing tickets and the scan matters most"*). And `--limit 800` is a ceiling that reads as a population, which is lesson 268 (#1492) instances 3 and 4: tonight a board query at `--limit 200` reported an issue absent, and the same query at `--limit 400` returned exactly 400 and matched **another repository's** issue of the same number.

So the reversion would have reintroduced tonight's blocker *and* the defect class the deleted paragraph exists to prevent, inside the skill that warns about it.

**Fixed at the source instead**: the vault record is restored at `knowledge@613e0b21`, verified purely additive bar five lines each superseded by the better version of the same instruction. `--refresh` then produces the correct content again, and this PR carries only the leftover sha.

## Verification

- **Additivity checked, not assumed.** The uncommitted tree sits on a base twenty-odd commits behind `main`, so copying it forward could have reverted whatever landed in between. Against `origin/main` the only removed line in each of the four is the sha marker it replaces, and `harness/enforced/no-auto-merge.md` removes nothing at all.
- **Round-tripped.** `--refresh` after the vault fix leaves `new-ticket` content-identical; the diff was one `generated_sha` line, which is what `80839ac` lands.
- `compile-harness.sh --check` → `no harness drift`. `harness-generated-sha`, `agents-md`, `compile-harness` bats all pass.

## Deliberately excluded

- **`harness/skills/handoff/SKILL.md`** — part of the same refresh, but its change had already landed on `main` via another PR, so the copy produced no diff.
- **`harness/agents/builder/AGENT.md`** — regenerates from the vault right now because a parallel session migrated that persona to the enforceable `skills:` mapping form. In flight and theirs to land.

Refs #1493


mlorentedev:
## Review triage

|---|---|---|
| **PR-Agent** | `## PR Reviewer Guide` | Read; no code findings requiring action on this diff. |
| **CI — `test`** | **FAILURE**, `skills-pipeline.bats:212` | **Real, mine, and NOT fixed here.** Diagnosis below. |
| **CI — everything else** | green or running | Open. |

### The failure

```
.gemini/GEMINI.md is 13632 chars, at or over its 12000 cap
```

Caused by `718e748`, which added **1677 chars** to `harness/enforced/no-auto-merge.md`. Confirmed by running `skills-pipeline.bats` at both revisions: case 18 **passes at `52035f1`** and **fails at `80839ac`**.

**But 13632 − 1677 = 11955.** Before this PR the hardest target sat **45 characters** under its cap. This diff did not create a size problem; it was the first thing to touch doctrine after the margin reached 0.4%, and it added no new policy — only paragraphs the vault had already been serving for hours.

That is **HARNESS-084 (#1241)**, "the capped doctrine payload has outgrown its hardest target twice, and has no budget policy", third instance, with the margin measured for the first time. Posted there.

### Why this PR is left red

The three available moves are all decisions that belong to #1241, not to a regen PR:

- **Raise the cap** — the number presumably encodes a real Gemini limit; changing it because a test went red is picking the number that makes red go away.
- **Trim the vault paragraphs** — they are the auto-merge doctrine's actual content, and the existence of a *compact* payload implies the compiler should shorten, not the source.
- **Compact per target** — the right fix, and out of scope here.

The alternative — dropping `harness/enforced/no-auto-merge.md` from this PR and landing the other three files — would put the repo's `AGENTS.md` and `CLAUDE.md` in a state where they carry the two paragraphs and the enforced source does not, which is drift manufactured to make CI green. Not doing that.

**What this PR still fixes stands on its own and is independent of the cap**: the vault record behind `harness/skills/new-ticket/SKILL.md` is restored (`knowledge@613e0b21`), so `compile-harness.sh --refresh` no longer rewrites that skill backward on every checkout — deleting the REST scan recipe and restoring `gh issue list --limit 800` in its place (#1493). That half is already landed in the vault and is not blocked by this PR's CI.

Handing over with this red and diagnosed rather than merged or forced.

mlorentedev:
## Status: still red, now with a concrete route — and a redundant hunk

**#1241 was decided (budget the payload per id, do not raise the cap).** That turns this PR from "parked pending a policy" into "parked pending a specific edit", so recording the route rather than leaving it looking blocked.

**The overflow is entirely this PR's own doctrine addition.** Measured on this branch just now:

```
payload = 13632   cap = 12000   over_by = 1632
```

and the diff adds exactly two paragraphs to `harness/enforced/no-auto-merge.md` (~1677 chars). Nothing else here contributes: the other four files are the regen and the stamp.

So the route out is **not** trimming 1632 characters of unrelated meta-text out of an already-compact payload. It is deciding whether the auto-merge *exception* — the four-condition carve-out and the branch-name-matching rule — belongs in the **compact** doctrine payload at all, or only in the full `AGENTS.md`/`CLAUDE.md` render. The compact payload's established rule is *drop meta-text, never the rules*; a 1677-character elaboration of when a rule does not apply is the case that rule was written for and the case it does not obviously settle. That is a real decision, not a mechanical trim.

**Not doing it in this session** — it is queued behind a `dotf secrets run` TTY defect the owner raised, which breaks `pi` outright.

**Also: the `harness/skills/new-ticket/SKILL.md` hunk here is now redundant.** #1503 lands the identical stamp change (`cd89b515ade938f7` → `1d1c81043049e6c9`) and is green. I opened #1503 without checking that this PR already carried it. This hunk will drop out on the next rebase; noting it so it is not mistaken for a conflict.

mlorentedev:
Follow-up: this PR's marker change landed in the **generated** record only.

`harness/enforced/no-auto-merge.md` gained `<!-- full-only:begin/end -->`, but `00_meta/patterns/pattern-git-workflow.md` — the vault source it is regenerated from — did not. So the next `compile-harness.sh --refresh` strips them again. Reproduced today on the first refresh after this merged, on clean `d03033e`.

Same class this PR was fixing for the new-ticket skill, one file over: a generated record edited without its source. Second occurrence.

Not cosmetic — those markers keep the auto-merge **exception** (1686 chars, four qualifying conditions) out of the compact doctrine payload that agy and codex receive instead of the constitution. No live exposure: the deployed `GEMINI.md` came from `d03033e`, which has the markers.

Fixed at source in `knowledge` `0c835e6d`; `--refresh` is idempotent again. The missing guard is **#1525**.

PR #1501, opened 2026-09-05
docs(spec): CLI-063 targets dotf deploy, after three of the ticket premises measured false

mlorentedev:
Spec only. No production code, no twin touched, nothing repointed.

Refs #1339.

## What changed about the ticket

#1339 asked for `dotf agent claude sync`. Before writing code I checked its premises against the
tree, and three of them do not hold:

| #1339's premise | Measured |
|---|---|
| `dotf agent` is the right home ("already exists, so it's a subcommand, not a new noun") | **False.** ADR-032:122-124 fixes `dotf agent` as the *run* side — *"one is idempotent and offline, the other spawns processes and consumes quota"* — and states that the **noun** is the part it fixes. Config sync is idempotent and offline. |
| "`dotf deploy` has no merge kind" | **False.** `strategy: merge` shipped with AI-039 (#1322) and is in production on `copilot-settings`, `copilot-config`, `agy-settings`. |
| This is new surface | **False.** `ai/deploy.json`'s own `$comment`: *"the MCP registration blocks are deliberately absent — they are slices 2 and 3."* |

`ai/deploy.json` already routes `pi`, `orca-keybindings`, `copilot-settings`, `copilot-config`,
`copilot-mcp` and `agy-settings`. **Claude's configs are the only ones left in the shell twins.** So
a separate `dotf agent claude sync` would not have been a new path — it would have been a *second*
path for the one agent that hasn't migrated, which is ADR-020's twin shape rebuilt rather than
collapsed.

Owner decision 2026-09-05: extend `dotf deploy`. Spec renamed accordingly.

## The scope got smaller, and better defined

`mergeInto` (`cli/internal/deploy/deploy.go:433`) writes the source's **top-level** keys into the
destination. Claude's six policies split against that:

| Key | Policy | Covered today? |
|---|---|---|
| `attribution`, `autoCompactEnabled`, `precomputeCompactionEnabled` | template wins | **yes** |
| `env`, `enabledPlugins` | **nested** object merge | no — top-level replace loses the box's keys |
| `permissions.allow` | **array union, deduped** | no — top-level replace drops the box's entries |

So the "largest and subtlest" increment is not a 100-line jq port. It is three per-key behaviours
the manifest cannot yet express — and they generalise beyond Claude.

## The oracle question is settled, not deferred

Golden characterization capture does **not** transfer from CLI-021. Four capabilities run at four
points of a setup, their effect is on `$HOME` and on `claude` itself rather than stdout, and two of
them shell out. `tasks.md` names a per-capability oracle instead (policy table tests, a plugin drift
guard, a threshold unit test, a fake `claude` on PATH). No golden file is to be added.

## Risk carried forward explicitly

`hooks` must not gain a second writer. HARNESS-045 AC1 moved it out of `merge_claude_settings`
because the jq assignment **replaced the whole SessionStart array and deleted a live third-party
group**. `tasks.md` puts the negative assertion in increment 1 — before the code that could violate
it exists.

## Verification

Spec-only change; no code paths touched.

```
pre-commit: Detect hardcoded secrets .......... Passed
            Run dotfiles tests ................ Passed
            Check bats @test names ............ Passed
            Check referenced doc paths ........ Passed
            Validate message format ........... Passed
```

### DIFF
--- a/specs/CLI-063-dotf-deploy-claude/features.json
+++ b/specs/CLI-063-dotf-deploy-claude/features.json
@@ -1,0 +1,58 @@
+[
+  {
+    "id": "CLI-063-dotf-deploy-claude-f1",
+    "behavior": "Claude's capabilities are declared in ai/deploy.json and reachable through `dotf deploy`, not through a new command noun",
+    "verification": "cd cli && go test ./internal/deploy/... -run TestClaudeEntriesDeclared",
+    "state": "pending",
+    "evidence": ""
+  },
+  {
+    "id": "CLI-063-dotf-deploy-claude-f2",
+    "behavior": "ai/claude/plugins.json is the sole plugin list; the test fails if either twin's hardcoded ids drift from it",
+    "verification": "bats tests/claude-plugins-drift.bats",
+    "state": "pending",
+    "evidence": ""
+  },
+  {
+    "id": "CLI-063-dotf-deploy-claude-f3",
+    "behavior": "env and enabledPlugins merge nested (a box-only key survives) and permissions.allow unions deduped; each case fails under top-level replace",
+    "verification": "cd cli && go test ./internal/deploy/... -run TestPerKeyMergePolicy",
+    "state": "pending",
+    "evidence": ""
+  },
+  {
+    "id": "CLI-063-dotf-deploy-claude-f4",
+    "behavior": "A template key absent from the declared policy errors loudly instead of being silently skipped (the outputStyle defect)",
+    "verification": "cd cli && go test ./internal/deploy/... -run TestUnknownTemplateKeyIsAnError",
+    "state": "pending",
+    "evidence": ""
+  },
+  {
+    "id": "CLI-063-dotf-deploy-claude-f5",
+    "behavior": "ai/deploy.json version is bumped and a decoder that predates the per-key field refuses the file rather than deploying every entry as plain merge",
+    "verification": "cd cli && go test ./internal/deploy/... -run TestManifestVersionRefusal",
+    "state": "pending",
+    "evidence": ""
+  },
+  {
+    "id": "CLI-063-dotf-deploy-claude-f6",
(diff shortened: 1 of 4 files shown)


mlorentedev:
## Review triage

**Who actually reviewed:** `pr-agent` (full review, re-run on `5f48e82`). **CodeRabbit did NOT review** — it posted a rate-limit notice, not findings. A notice that no review ran leaves that reviewer's pass unperformed, and this PR is proceeding on pr-agent's review alone. Disclosed rather than implied by a green tick.

**pr-agent: no major issues detected.** Its ticket-compliance items, dispositioned:

| Finding | Disposition |
|---|---|
| "Create a new command `dotf agent claude sync`" — non-compliant | **Declined, deliberately.** This is the change. ADR-032:122-124 fixes `dotf agent` as the run side; the PR body carries the three falsified premises. The reviewer is correctly reporting that the spec diverges from the ticket — that divergence is the deliverable, and #1339 has been retitled and commented so the ticket no longer advertises the command we are not building. |
| "Delete the duplicated shell blocks / remove the pinning bats tests" — non-compliant | **Declined, in scope terms.** Explicitly Out of scope in `proposal.md`: port, prove parity, then delete (the #1450 precedent). The pinning tests are the only oracle for the shell behaviour and are the last thing to go, never a step alongside. |
| "Requires further human verification: the spec redefines the command home… requires re-evaluation against ADR-032" | **Correct, and already done.** That re-evaluation is what produced the change; the owner made the call on 2026-09-05. This is the one item where the reviewer identified the right question rather than a defect. |
| #1322 partial compliance | **Not this PR's ticket.** #1322 is referenced as the origin of `strategy: merge`, not as work this PR closes. |

**Checks:** `CI`, `pr-agent`, `review-attestation`, `spec-gate`, `Add to bitácora` — all SUCCESS. `MERGEABLE CLEAN`.

**Nothing applied**, because nothing was a defect. Spec-only change; no production code paths touched.

Issue #1502, opened 2026-09-05
HARNESS: compile-harness.sh --check returns rc=0 on a tree --refresh would rewrite, so it never detected either reversion


mlorentedev:
## Why

`compile-harness.sh --check` is the repository's drift guard for generated harness artifacts. It is run in CI, it is what agents cite when they say the tree matches the vault, and its own name states the proposition. **It does not check that proposition.**

Measured 2026-09-05 on a clean tree at `ca81157`, using `harness/skills/crystallize/SKILL.md`:

| state | md5 | contains `knowledge-crystallize.sh` (deleted in #1276) | `--check` |
|---|---|---|---|
| A — as committed, #1490's fix in place | `1bb56a86` | no | `[check] OK -> record crystallize`, **rc=0** |
| B — after `--refresh` reverted it | `d84f1c33` | **yes** | `[check] OK -> record crystallize`, **rc=0** |

Both states pass. `--check` returns rc=0 on a tree that `--refresh` — the same script, one flag apart — would itself rewrite.

This is not confined to skill records. `--check` prints `[check] OK -> AGENTS.md` on a tree where `--refresh` adds two doctrine paragraphs to that file. On the same run, `--refresh` moved **seven** files while `--check` reported no drift on any of them.

## The distinction that makes this its own ticket

#1493 is the incident: `--refresh` reverts repo-ahead fixes, it has now done so twice (`new-ticket`, `crystallize`), and both times the root cause was a generated file fixed without its source. That is a workflow hazard with a known remedy — fix the SSOT.

**This is a broken instrument.** The remedy is different, the urgency is different, and they will be fixed at different speeds if they share a ticket. More importantly, the instrument is why the incidents go unnoticed: both reversions were caught by a human reading `git status`, never by the guard whose job that is. `--check` passing on a PR was never evidence that its tree matched the vault, and at least two agents (myself and a parallel session) had been reading it as exactly that.

Same shape as lesson 268 — *"no drift"* and *"I did not check for drift"* print identically.

## What

`--check` must fail whenever `--refresh` would change a file. Whatever it compares today is not the rendered output; if it were, states A and B could not both pass.

The cheapest honest implementation is to define `--check` *in terms of* `--refresh` rather than reimplementing the comparison: render to a temporary tree, diff against the working tree, exit non-zero with the file list on any difference. Two renderers that are supposed to agree is the drift-generating shape this repository has been bitten by repeatedly; one renderer and a diff cannot disagree with itself.

[...]

Issue #1499, opened 2026-09-05
HARNESS: the role-join guard measures rule coverage, so five skills owned by no persona sit undetected


mlorentedev:
## Why

`TestRoleJoinDrift` (HARNESS-108, AC3) is the only guard that spans `harness/triggers.json` and the persona roster. It watches the join collapsing, and it does that well. It cannot see a skill that no persona owns, because it counts **rules that resolve**, not **skills that are owned** — and a rule resolves as soon as *one* of its skills reaches a persona.

So an orphaned skill inside an otherwise-resolving rule is invisible: the rule still resolves, the count still clears the floor, the guard stays green, and the skill contributes nothing to role resolution for the rest of its life.

Measured 2026-09-05 against `harness/agents/**` and `harness/triggers.json` (18 rules, 7 personas, 39 skills on disk):

**Five skills exist in `harness/skills/` that no persona declares:**

| skill | named by a trigger rule? |
|---|---|
| `pr-review-triage` | **yes** — `git-workflow` |
| `project-maturation` | **yes** — `security-and-quality-audit` |
| `catchup` | no |
| `dispatching-parallel-agents` | no |
| `research-prompt` | no |

The two marked *yes* are the live half: they sit in a rule's `skills` array, so the join reads them, finds no owner, and drops them. Both rules still resolve through a sibling skill (`git-workflow` via `using-git-worktrees` → shipper; `security-and-quality-audit` via `audit` → reviewer), which is exactly why nothing is red.

`pr-review-triage` is the sharpest case and is being fixed in #1497, where it goes to `shipper` — whose mandate already says *"An open PR is not finished work... dispositioned"* in prose while declaring no skill for it. **That fix is a fix for one row, not for the class.** The guard that let it sit there is unchanged, and `project-maturation` is still orphaned behind the same blind spot.

**Two rules additionally carry no `skills` key at all** — `shell-standards` and `powershell-ascii-only`. That one is deliberate and already documented (`roles_test.go`: *"two of the 18 rules are pattern-only and have no owner"*), which is why the floor is 16 of 18 and not 18. Recording it here so a future reader does not "fix" it.

## What

A guard on **skill ownership**, beside the existing one on rule coverage. Two distinct assertions, because they fail for different reasons:

1. **Every skill named in `triggers.json` is declared by at least one persona.** This is the one that would have caught `pr-review-triage`. It is a closed set on both sides and cheap to assert.
[...]

Issue #1497, opened 2026-09-05
HARNESS: five of seven personas still declare skills in the flat form, so the gate resolves them and enforces nothing


mlorentedev:
## Why

The orchestrator's plumbing is live. `PreToolUse` runs `dotf harness gate --harness claude`, `UserPromptSubmit` runs `dotf harness suggest --from-hook`, and the decision record (#1435) makes every outcome observable. What is missing is **data, not machinery**: five of seven personas declare their skills in the legacy flat list, which the loader reads as `EnforceUnset`, so the gate resolves them and acts on nothing.

Measured 2026-09-05:

| persona | form | gated skills |
|---|---|---|
| `reviewer` | mapping | 4 (canary, #1412) |
| `builder` | mixed | 2 (#1494) |
| `architect` | flat | **0** |
| `curator` | flat | **0** |
| `planner` | flat | **0** |
| `shipper` | flat | **0** |
| `hermes-nan` | flat | **0** |

`TestRealPersonaRecordsLoadAndReportTheirMigrationState` logs **29 skills still in the flat form**.

The effect is that the `[persona]` line on every prompt is a *suggestion*, not an obligation, for five of seven roles — and each of those personas' own prose says its skills are "enforced by hook", which is false until it migrates. `builder`'s said exactly that for all nine of its skills.

## What

Migrate the remaining five, one persona per PR, each at `enforce: warn`.

**The shape is settled by #1494 and should not be re-litigated per persona:**

- **Gate the unconditional skills only; leave the situational ones as bare strings.** The list accepts a mix — a string entry yields `EnforceUnset`, a mapping entry missing `enforce` is a hard parse error — so a partial migration is expressible without inventing a third state. `builder` gates `test-driven-development` and `test`, and leaves `debug-hardware`, `mcp-builder`, `async-python-patterns` and three others declared-and-ungated.
- **Why not gate everything:** a gate that names irrelevant skills on every call teaches the reader to scroll past `[gate]` lines. The severity is worth what the attention it commands is worth, and `UnmigratedSkills` keeps the ungated ones visible rather than silent.
- **The owner picks the severities**, as they did for the `reviewer` canary and for `builder`.
- **The vault is the SSOT.** Edit `$VAULT_PATH/00_meta/agents/definitions/<persona>/AGENT.md`, commit to `master`, then `compile-harness.sh --refresh` regenerates `harness/agents/<persona>/AGENT.md`. Editing the record in the repo is clobbered.

[...]

PR #1504, opened 2026-09-05
feat(harness): shipper becomes the third enforceable persona, and adopts an orphaned skill

mlorentedev:
Third of the five personas #1497 covers. `reviewer` was the canary (#1412), `builder` was the second (#1494), this is `shipper`.

## What changed

`shipper` declared its four skills in the flat list, which `LoadPersona` reads as `EnforceUnset`, so `dotf harness gate` resolved the persona and acted on nothing. Its own "Forced skills" section said the skills were *"enforced by hook, not left to memory"* — false for all four, and false since the record was written. Same defect `builder`'s prose carried for all nine.

| skill | before | after | why |
|---|---|---|---|
| `using-git-worktrees` | flat | **`enforce: warn`** | Isolation applies to every change without exception |
| `docker` | flat | **`enforce: warn`** | This repo does ship a container — `tests/Dockerfile.integration`, built by `integration`, is where delivery actually breaks here |
| `pr-review-triage` | **absent** | **`enforce: warn`** | See below |
| `helm` | flat | bare string | Nothing here is delivered by Helm |
| `terraform` | flat | bare string | Nothing here is delivered by Terraform |

The prose now states that the last two are ungated **because of this repository**, not as a property of the skills. A persona travels between repositories — in one that ships charts or state files those two *are* the ship phase — and a severity was chosen against one. That distinction was missing from `builder`'s rewrite and is worth having in the record rather than in a PR description.

## `pr-review-triage` was owned by nobody

Found while auditing the join, not while writing this. The skill exists in `harness/skills/`, is named by the `git-workflow` rule in `harness/triggers.json:91`, and **was declared by no persona at all** — while this record's mandate has required *"an open PR is not finished work... dispositioned"* since before the skill was written. So the join read it, found no owner, and dropped it.

`TestRoleJoinDrift` stayed green through all of that, because it counts **rules that resolve**, not **skills that are owned** — and `git-workflow` resolves through `using-git-worktrees` regardless. That guard gap is filed as **#1499**, with `project-maturation` still sitting behind it. This PR closes the one row, not the class.

The new assertion is deliberately on the skill **alone**:

```go
if got := ResolveRoles(Suggestion{Skills: []string{"pr-review-triage"}}, personas); ...
```

Asserting the rule would have passed before this change too. Only the single-skill form distinguishes "the rule resolves" from "this skill has an owner".

## Also: the crystallize stamp, which is my own loose end

`--refresh` reverted `harness/skills/crystallize/SKILL.md`, putting back `knowledge-crystallize.sh` — the script deleted in #1276 and removed from that record by #1490 six hours earlier. Root cause was mine: **#1490 fixed the generated file and left the vault source stale**, so `--refresh` was propagating a stale SSOT *correctly*. Fixed at source in knowledge `57b6b1ae`; the `generated_sha` here is what makes the round-trip clean.

Second instance of that shape in two days after `new-ticket`. Fixing a generated file without its source is not a fix, it is a countdown.

## Two findings filed rather than folded in

- **#1502** — `compile-harness.sh --check` returns **rc=0 on both** the correct and the reverted `crystallize`, and passes `AGENTS.md` by name while `--refresh` adds two paragraphs to it. A guard returning OK on a tree it would itself rewrite is why neither reversion was ever caught by anything but a human reading `git status`.
- **#1499** — the ownership blind spot above.

One `--refresh` on clean `ca81157` moved **seven** files in three directions: four forward (auto-merge doctrine paragraphs, owned by another branch), one backward, one stamp-only, one mine. Every line of output reads `[refresh] skill record: ...` regardless. The five that are not mine are reverted here; direction was checked per file, not assumed.

## Verification

Run in this session:

- `go build ./...`, `go vet ./...`, `GOOS=windows go vet ./...`, `go test ./... -count=1` — all clean
- `golangci-lint run` — **0 issues** at the pinned 2.12.2 (`versions.conf`), version confirmed against the pin
- `~/.local/bin/bats tests/*.bats` — **1554/1554**, exit 0
- `compile-harness.sh --check` — no harness drift
- `shellcheck scripts/compile-harness.sh` — clean

**Both new assertions proven in the failing direction**, each mutation confirmed to have landed before the verdict:

| mutation | result |
|---|---|
| shipper reverted to the flat form | `TestMigratedPersonasStillGateSomething`: *"shipper has been migrated but now gates nothing"* |
| `pr-review-triage` removed from shipper | `TestResolveRoles`: *"should resolve to [shipper], got []"* |

**Payload cap, predicted before the work and confirmed after:** 11974 of 12000, 26 to spare. The migration itself costs **0 characters** — confirmed on three trees all landing on 11956 — because the presence block renders ids only and severity is `gate`'s input, not the payload's. The single new id costs 18. Numbers and the budget consequence are on #1241.

## Not claimed

No genuine `shipper` dispatch has been observed producing `warn`. That requires the record to be deployed first, and the deploy dir is written by `setup-linux.sh` / `dotf deploy` — hand-placing a file there would prove the gate reads a file I put there, not that a deploy produces the state. It is also still blocked in general by **#1434**: a *named* dispatch sends its name as `agent_type`, so its role never resolves and the gate silently turns off. Carried as a precondition on the remaining migrations and on any `block` promotion.

Refs #1497

### DIFF
--- a/cli/internal/harness/persona_test.go
+++ b/cli/internal/harness/persona_test.go
@@ -249,7 +249,7 @@ func TestMigratedPersonasStillGateSomething(t *testing.T) {
 		byName[p.Name] = p
 	}
 
-	for _, name := range []string{"reviewer", "builder"} {
+	for _, name := range []string{"reviewer", "builder", "shipper"} {
 		p, ok := byName[name]
 		if !ok {
 			t.Errorf("%s is missing from the roster", name)
--- a/cli/internal/harness/roles_test.go
+++ b/cli/internal/harness/roles_test.go
@@ -42,6 +42,16 @@ func TestResolveRoles(t *testing.T) {
 		t.Errorf("test-driven-development should resolve to [builder], got %v", got)
 	}
 
+	// `pr-review-triage` is named by the `git-workflow` rule and, until shipper
+	// adopted it, was declared by no persona at all — so the join read it, found
+	// no owner and dropped it. The rule still resolved through a sibling skill,
+	// which is exactly why nothing was red. Asserting the skill ALONE is what
+	// distinguishes "the rule resolves" from "this skill has an owner"; only the
+	// second is what changed. See #1499 for the guard that cannot tell them apart.
+	if got := ResolveRoles(Suggestion{Skills: []string{"pr-review-triage"}}, personas); len(got) != 1 || got[0] != "shipper" {
+		t.Errorf("pr-review-triage should resolve to [shipper], got %v", got)
+	}
+
 	// Sorted output is the determinism contract: same input, same bytes.
 	multi := ResolveRoles(Suggestion{Skills: []string{"terraform", "audit", "read-all-adrs"}}, personas)
 	for i := 1; i < len(multi); i++ {
--- a/harness/agents/shipper/AGENT.md
+++ b/harness/agents/shipper/AGENT.md
@@ -1,7 +1,7 @@
 ---
 generated: true
 generated_from: 00_meta/agents/definitions/shipper/AGENT.md
-generated_sha: ead9e03f1abf72e1
+generated_sha: 86c1ecfacc81c6fe
 id: agent-shipper
 type: agent
 status: active
@@ -11,7 +11,15 @@ description: Ship-phase persona. Invoke to get a verified change into the world 
 kind: invocable
 model: mid
 capabilities: [read, search, edit, shell, skill]
-skills: [using-git-worktrees, docker, helm, terraform]
+skills:
+  - id: using-git-worktrees
+    enforce: warn
+  - id: docker
+    enforce: warn
+  - id: pr-review-triage
+    enforce: warn
+  - helm
+  - terraform
 owner: manu
 ---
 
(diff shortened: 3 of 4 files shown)


mlorentedev:
## Review triage

|---|---|---|
| **CodeRabbit** | *"Review limit reached — next included review available in 36 minutes."* | **Not a review — a notice.** Third consecutive notice across #1494 and this PR. **This change has not had CodeRabbit's opinion and must not be read as having had it.** |
| **PR-Agent — ticket compliance (#1497)** | Partially compliant, **"Non-compliant requirements: None"**, five compliant items listed. | Accepted. |
| **PR-Agent — verify the `generated_sha` matches the vault** | Requires human verification. | **Applied. Verified below — both stamps correct.** |
| **PR-Agent — confirm no unrelated files were modified** | Requires human verification. | **Applied. Verified below — exactly four files.** |
| **PR-Agent — focus area: "Unrelated change" (crystallize stamp)** | Says the crystallize `generated_sha` is unrelated to the shipper migration and creates noise. | **Declined, with the reason below.** Factually accurate; the conclusion does not follow. |
| **CI** | 16 of 16 checks, 0 failures. | Accepted. |

### The two verification requests, answered by measurement

PR-Agent asked for both rather than asserting either, which is the right shape for a claim it cannot check. Both confirmed:

```
harness/agents/shipper/AGENT.md       generated_sha: 86c1ecfacc81c6fe
sha256 $VAULT/00_meta/agents/definitions/shipper/AGENT.md  -> 86c1ecfacc81c6fe   ✓

harness/skills/crystallize/SKILL.md   generated_sha: 1b585b8def0a52a1
sha256 $VAULT/00_meta/skills/crystallize/SKILL.md          -> 1b585b8def0a52a1   ✓
```

Files changed against `origin/main` — exactly the four the description names, no fifth:

```
cli/internal/harness/persona_test.go
cli/internal/harness/roles_test.go
harness/agents/shipper/AGENT.md
harness/skills/crystallize/SKILL.md
```

That matters more than it looks, because the `--refresh` that produced this PR moved **seven** files. The other three were reverted after checking the direction of each; two were another branch's doctrine additions and one was a stamp belonging to a parallel session's PR (#1503).

### Why the crystallize stamp stays

PR-Agent is right that it is a different subject from the shipper migration, and right that its immediate cause was `compile-harness.sh --refresh`. I am keeping it, and the reason is causal rather than convenient:

**That stamp is the completion of the fix this PR's own `--refresh` forced.** The refresh reverted `crystallize/SKILL.md` to reference `knowledge-crystallize.sh` — a script deleted in #1276 and removed from that record by #1490 six hours earlier. The root cause was mine: #1490 fixed the generated file and left the vault source stale. I fixed the source (knowledge `57b6b1ae`), and after a source fix the content matches while the stamp still does not, because it was computed from the old source.

Dropping the stamp from this PR does not reduce noise, it relocates it: the next person to run `--refresh` sees `crystallize` dirty, investigates, and rediscovers a closed issue. Carrying it is what makes the round-trip idempotent for everyone else.

On the merge-conflict half of the finding: the parallel session's #1503 carries the **`new-ticket`** stamp, a different file. There is no overlap, and this was coordinated rather than assumed — the two stamps were found together and split by owner deliberately, so neither of us reaches into the other's tree.

Filed rather than folded in: **#1502** (`--check` returns rc=0 on both the correct and the reverted file, which is why neither reversion was ever caught) and **#1499** (the join guard counts rules that resolve, not skills that are owned). Lesson in **#1505**.

### Verification, run in this session

`go build` · `go vet` · `GOOS=windows go vet` · `go test ./... -count=1` — all clean. `golangci-lint run` — **0 issues** at the pinned 2.12.2, version checked against `versions.conf` rather than assumed. `bats tests/*.bats` — **1554/1554**. `compile-harness.sh --check` — no drift. `shellcheck` — clean.

**Both new assertions proven in the failing direction**, each mutation confirmed to have landed before any verdict was read:

| mutation | caught by |
|---|---|
| shipper reverted to the flat form | `TestMigratedPersonasStillGateSomething` — *"shipper has been migrated but now gates nothing"* |
| `pr-review-triage` removed from shipper | `TestResolveRoles` — *"should resolve to [shipper], got []"* |

Presence payload **11974 of 12000**, predicted before the change was written and confirmed after.

PR #1505, opened 2026-09-05
docs(lessons): 269 — a generated file fixed without its source is a countdown

mlorentedev:
Documents the class behind two reversions in two days, and the measurement explaining why neither was caught by anything but a human reading `git status`.

## The class

`harness/skills/**` and `harness/agents/**` are generated from the vault, with `generated_from` and `generated_sha` in their frontmatter. Twice this week a defect was fixed by editing the **generated** record and merging it, and `compile-harness.sh --refresh` undid the fix the next time anyone ran it — **correctly**, because the vault source was still stale.

- **`new-ticket`** — the fix sat uncommitted in the main checkout for ~20h before anyone noticed. Nothing reported it.
- **`crystallize`** — #1490 removed a reference to `knowledge-crystallize.sh`, a script deleted in #1276. Six hours later a `--refresh` put the deleted script's name straight back. Root cause was mine: I fixed the render and left `00_meta/skills/crystallize/SKILL.md:67` alone. Fixed at source in knowledge `57b6b1ae`; the stamp is carried in #1504.

## The measurement, which is what makes it a guard bug

| state | `crystallize/SKILL.md` md5 | contains the deleted script | `--check` |
|---|---|---|---|
| as committed, #1490's fix in place | `1bb56a86` | no | `OK -> record crystallize`, **rc=0** |
| after `--refresh` reverted it | `d84f1c33` | **yes** | `OK -> record crystallize`, **rc=0** |

`--check` returns rc=0 on a tree that `--refresh` would itself rewrite, and prints an identical line either way. On the same run it passed `AGENTS.md` by name while `--refresh` added two doctrine paragraphs to it.

So neither reversion was missed through carelessness — the instrument that exists to catch them was answering a different question, in the affirmative. That is **#1502**, filed separately from the reversion incident (**#1493**): an incident and a broken instrument get fixed at different speeds and should not share a ticket.

## Two details worth having written down

**Blast radius.** One `--refresh` on clean `ca81157` moved seven files in three directions — four forward (doctrine owned by another branch), one backward, one stamp-only, one intended. Every line of output reads `[refresh] skill record: …` regardless, so direction has to be checked per file with `git diff`.

**Content-identical ≠ clean after `--refresh`.** After a source fix the record's content matches but `generated_sha` still differs, computed from the old source. Both fixes this week achieved the first and neither the second until the stamp was carried. The second is what stops the next person seeing noise and learning to ignore it.

## Not claimed

This is documentation, and the lesson says so about itself: this is the **third** time the class has been written down (lessons 009 and 092 are the earlier two) and it recurred anyway. Documentation is not a mechanism — #1502 is. The lesson is worth having so the next occurrence is recognised in seconds rather than investigated, not because writing it prevents anything.

## Verification

- `scripts/check-doc-paths.sh` — rc=0
- Every lesson cross-referenced in the body (009, 092, 214, 220, 267, 268) confirmed to exist on disk rather than assumed
- `_index.md` row added alongside the file
- Pre-commit hooks green: secrets, dotfiles tests, bats `@test` names, referenced doc paths, message format

Docs-only; no production LOC.

### DIFF
--- a/docs/lessons/_index.md
+++ b/docs/lessons/_index.md
@@ -286,5 +286,6 @@ tags: [lessons, index, dotfiles]
 | [268 - A refused question and a negative answer print the same thing](lesson-268-a-refused-question-and-a-negative-answer-print-the-same-thing.md) | 2026-09-04 |  |
 | [265 - A correct measurement answering the wrong question, three times in one session](lesson-265-a-correct-measurement-answering-the-wrong-question.md) | 2026-09-03 |  |
 | [266 - Two measurements agreeing is not reproducibility](lesson-266-two-measurements-agreeing-is-not-reproducibility.md) | 2026-09-04 |  |
+| [269 - A generated file fixed without its source is a countdown, and the drift guard says OK either way](lesson-269-a-generated-file-fixed-without-its-source-is-a-countdown.md) | 2026-09-05 |  |
 
 - [2026-08-30 Intercepting Cobra Errors without Breaking SilenceErrors](2026-08-30-cobra-silence-errors-interception.md)
--- a/docs/lessons/lesson-269-a-generated-file-fixed-without-its-source-is-a-countdown.md
+++ b/docs/lessons/lesson-269-a-generated-file-fixed-without-its-source-is-a-countdown.md
@@ -1,0 +1,138 @@
+# Lesson 269 — A generated file fixed without its source is a countdown, and the drift guard says OK either way
+
+**Date:** 2026-09-05
+**Context:** Two reversions in two days (`new-ticket`, `crystallize`), found by two different sessions, neither found by the guard whose job it is
+
+## What happened
+
+`harness/skills/**` and `harness/agents/**` are **generated** from the knowledge vault. Each record
+carries its provenance in frontmatter:
+
+```yaml
+generated: true
+generated_from: 00_meta/skills/crystallize/SKILL.md
+generated_sha: d6d6cf0c738eab7d
+```
+
+`scripts/compile-harness.sh --refresh` re-renders every one of them from the vault. Twice in two
+days, someone fixed a defect by editing the **generated** file, merged it, and had the fix silently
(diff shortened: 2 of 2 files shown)


mlorentedev:
## Review triage

|---|---|---|
| **CodeRabbit** | *"Review limit reached — next included review available in 29 minutes."* | **Not a review — a notice.** Fourth consecutive notice tonight across #1494, #1504 and this PR. **This change has not had CodeRabbit's opinion.** |
| **PR-Agent** | Effort 1/5, **no major issues detected**, no security concerns, *"No relevant tests"*. | **Accepted, including the tests observation** — see below. |
| **CI** | 11 of 11 checks, 0 failures, 0 pending. | Accepted. |

### On "No relevant tests"

Correct and intended. This is a documentation-only change: one lesson file and one `_index.md` row, no production LOC. There is nothing here for a test to assert that `check-doc-paths.sh` does not already cover, and it does — rc=0.

What was verified instead, since a prose file's failure mode is a dangling reference rather than a broken build: **every lesson cross-referenced in the body (009, 092, 214, 220, 267, 268) was confirmed to exist on disk** rather than assumed. That check is the point of the exercise here — lesson 268, cited in this very file, is about exactly this: a query whose negative answer and whose failure print the same thing. Writing a lesson that cites a lesson that does not exist would have been the failure documenting itself.

### The disclosure this PR's own subject demands

Two of tonight's four PRs merged or will merge without CodeRabbit having reviewed them, and this one is documentation *about* guards that report success without having checked anything. So it is worth being exact rather than ironic about it: **a notice that no review ran leaves the change unreviewed.** PR-Agent did review it and found nothing major; CodeRabbit did not review it at all. Proceeding is allowed, proceeding silently is not.

### Scope note

The lesson deliberately claims no preventive power for itself. It is the **third** documentation of this class — lessons 009 and 092 are the earlier two — and it recurred anyway, which is stated in the body. The mechanism is **#1502**, not this file.

PR #1459, opened 2026-09-03
feat(secrets): guard secrets run against environment introspection commands

mlorentedev:
## Summary

Adds a deterministic guard in `dotf secrets run` to refuse commands and shell snippets whose purpose is to dump the process environment or inspect secrets to standard output, upholding ADR-028 doctrine.

## Changes

- Implement `assertSafeChildCommand` in `cli/internal/cmd/secrets.go` refusing direct commands (`env`, `printenv`, `export`) and shell wrappers (`sh -c`, `bash -lc`, `bash -i -c`, etc.) containing `env`, `printenv`, `export`, `set`, or `declare`.
- Normalize tokens by stripping quoting/escaping delimiters to prevent regex boundary evasion.
- Add 20-case table-driven unit test suite `TestAssertSafeChildCommand` in `cli/internal/cmd/secrets_test.go`.
- Update `ai/claude/settings.json` permissions deny list with environment introspection commands.
- Register `openrouter` provider in `ai/pi/models.json` using `${OPENROUTER_API_KEY}`.
- Scaffold and populate `specs/SEC-001-secrets-run-guard/`.

Refs #1458

PR #1508, opened 2026-09-05
fix(secrets): give an interactive child a pty, so secrets run stops breaking TUIs

### LINKED ISSUE #1506: SEC-002: dotf secrets run gives every child a pipe, so any interactive TUI exits silently
## Symptom

`pi`, wrapped in `.zshrc:92` / `.bashrc:114` as

```sh
pi() { dotf secrets run --only NAN_API_KEY,OPENROUTER_API_KEY -- pi "$@"; }
```

returns to the prompt in ~2s with **exit code 0 and zero bytes** on stdout and stderr. Reproduces in every directory; not repo-specific.

## Root cause (diagnosed at the fd level by the owner; confirmed by reading the code, do not re-diagnose)

`cli/internal/cmd/secrets.go:504-506` wraps the child's streams in a redacting writer, and `runChild` (`:618`) assigns them straight onto `exec.Cmd`:

```go
stdout := newRedactWriter(cmd.OutOrStdout(), injected)   // an io.Writer, NOT an *os.File
c.Stdin, c.Stdout, c.Stderr = stdin, stdout, stderr
```

`exec.Cmd` passes a file descriptor through **only** when the writer's dynamic type is `*os.File`; for anything else it allocates an `os.Pipe`. So the child's fd 1 and fd 2 are pipes:

```
DIRECT:    in=0  out=0  err=0
VIA DOTF:  in=0  out=1  err=1        fd1 -> pipe:[...], fd2 -> pipe:[...]
```

`pi` sees stdout is not a TTY, declines to start its TUI, and exits 0 silently.

Introduced by `a720b9d` (#1459); shipped in the binary built from `eaf8a91`, installed 2026-09-03. `runChild`'s own comment encodes the assumption that broke — *"a wrapper that owns no terminal"*.

Already ruled out by the owner: pi's version, extensions (`--no-extensions`), Orca extensions (`ORCA_PANE_KEY=`), `trust.json`, and secret injection itself. `pi -p '...'` works because headless needs no TTY — **which is why the reviewer pool never surfaced this**.

## A second defect the pty alone does not fix

`redactWriter.Write` (`secrets.go:750-756`) holds back a **fixed** `maxSecretLen - 1` bytes on *every* write, unconditionally:

```go
if len(data) >= r.maxSecretLen && r.maxSecretLen > 1 {
    split := len(data) - (r.maxSecretLen - 1)
    r.tail = data[split:]          // withheld until the NEXT write
```

With a ~64-character API key that is **63 bytes of every frame withheld until the next write arrives**. Correct for a batch pipe that ends in `Flush()`; wrong for an interactive stream, where the tail of each frame — typically the cursor positioning — must reach the terminal now. Giving the child a pty without fixing this yields a TUI that renders late and corrupt rather than not at all.

The hold-back must be narrowed to *bytes that are actually a proper prefix of some secret*, not a fixed window.

## Extent — checked, and it is not systemic

All 15 sites in `cli/internal` that assign a child's `Stdout`:

| Site | Shape | Affected |
|---|---|---|
| `cmd/secrets.go:618` | `redactWriter` | **yes — this bug** |
| `cmd/spec.go:92` | `io.MultiWriter(os.Stdout, f)` | same mechanism, but the child is headless by design (the reviewer). **Latent, not active** |
| `tools/install.go:111` | `in.Out`, defaults to `os.Stdout` | no — dynamic type is `*os.File`, fd passes through |
| `cmd/update.go:104`, `cmd/spec.go:57` | `os.Stdout` | no |
| 9 x `secrets/*`, `agent/process.go`, `doctor/system.go` | `&bytes.Buffer` capture-by-design | no |

Exposed wrappers: `pi` (broken) and `opencode` (renders 14383 bytes through the pipe today, so it tolerates a non-TTY) — in **both** `.zshrc` and `.bashrc`.

## The fix

When the parent's stdout is a terminal, give the child a **pty** instead of a pipe: allocate it, pump `pty -> redactWriter -> os.Stdout` and `os.Stdin -> pty`, put the parent tty in raw mode (restored on exit *and* on signal), forward SIGWINCH, propagate the child's exit code.

**Do not** take the cheap alternative of passing `os.Stdout`/`os.Stderr` through unredacted when they are terminals. That drops the #1459 guarantee exactly where the session transcript is captured, which is the case the guarantee exists for.

Constraints:

- Redaction still applies to everything, including across write boundaries (see the second defect above).
- A pty merges stdout and stderr onto one stream. Decide deliberately and record it.
- Cross-OS: pty path behind a build tag, current pipe behaviour as the Windows fallback. `golang.org/x/term` and `x/sys` are already deps; `creack/pty` would be new and needs a pin (or an `x/sys` implementation instead — a real choice for the proposal).
- Keep `assertSafeChildCommand` (#1459) on both paths.
- Do not regress headless `pi -p`: the reviewer pool runs it across parallel worktrees.
- `~/.zshrc` is deployed from `.dotfiles/.zshrc` — never hand-edit the deployed copy.

## Acceptance criteria

1. A committed test asserts the child gets a **pty** when the parent has one and a **pipe** when it does not.
2. A committed test asserts an injected secret routed through the pty is redacted before reaching the parent's stdout, **including when split across writes**.
3. Evidence in the PR, produced in-session and from a binary rebuilt from the branch (the installed `dotf` is `vcs.modified=true`, built from a dirty tree):
   - `dotf secrets run --only NAN_API_KEY,OPENROUTER_API_KEY -- pi` launches the TUI, stays alive, accepts typing.
   - `dotf secrets run --only NAN_API_KEY -- pi -p 'reply with OK'` still prints OK.
   - `opencode` through its wrapper still works (the unaffected canary).
4. Reported as exit code and byte count, never by pasting the captured stream.

## Repro harness

`script(1)` and a bare `pty.fork()` both mislead — `pty.fork()` gives a 0x0 winsize and the TUI refuses to draw. Set the window size after forking:

```python
fcntl.ioctl(fd, termios.TIOCSWINSZ, struct.pack('HHHH', 40, 120, 0, 0))
```

Then compare `pi` against `command pi`: broken exits 0 in ~2s with 0 bytes; working stays alive and renders ~40KB.

## Known stopgap — do NOT ship

`dotf secrets run ... -- script -qfec pi /dev/null` gives pi a pty and the TUI renders. Not the fix: `AGENTS.md:63` and dropping the `/dev/null` makes `script(1)` write a `typescript` file holding everything the child printed — a secrets sink.

## Process

SDD applies (public contract, >50 LOC, cross-OS, new I/O path on the secrets boundary).

mlorentedev:
Closes #1506. Spec: `specs/SEC-002-secrets-run-pty/`.

## The defect

`dotf secrets run` is **the only sanctioned way to hand a secret to a process** — it is what the
doctrine points every wrapper at instead of exporting a credential into the ambient shell. Since
#1459 it silently broke every interactive child: `pi` returned in ~2s with **exit 0 and zero bytes**.

So the supported path failed and the *unsupported* one — ambient `$NAN_API_KEY` — worked. A security
control that breaks the tool it protects gets routed around, restoring exactly the exposure ADR-028
exists to prevent.

**Root cause** (diagnosed at the fd level in the ticket, confirmed by reading): `exec.Cmd` passes a
descriptor straight through **only** when the writer's dynamic type is `*os.File`. A `redactWriter`
never is, so every child got an `os.Pipe`:

```
DIRECT:    in=0  out=0  err=0
VIA DOTF:  in=0  out=1  err=1        fd1 -> pipe:[...], fd2 -> pipe:[...]
```

`runChild`'s own comment encoded the assumption that broke — *"a wrapper that owns no terminal."*

## The pty was the easy half

Found while reading, before writing any code: `redactWriter.Write` withheld a **fixed**
`maxSecretLen - 1` bytes on *every* write, regardless of content.

```go
split := len(data) - (r.maxSecretLen - 1)
r.tail = data[split:]          // invisible until the NEXT write
```

With a 64-byte key that is **63 bytes of every frame held dark** until more output happens to
arrive. Unobservable against a pipe drained at `Flush()`; fatal against a terminal, where the tail
of a frame is its cursor positioning. **A pty alone would have produced a TUI that renders late and
misplaced instead of not at all.**

The hold-back is now prefix-aware — withhold only a trailing run that is a *proper prefix* of some
secret, which is zero in the common case. That is the correct rule for the pipe path too, so this
removes a fixed-latency window that was always wrong and merely invisible in batch use.

## What is deliberately NOT done

- **Redaction is never dropped.** Passing `os.Stdout` through untouched when it is a terminal would
  make the TUI perfect and void the #1459 guarantee precisely where a session transcript is
  captured — the case the guarantee exists for. `assertSafeChildCommand` runs on **both** paths,
  with a test pinning it to the new one.
- **Windows keeps the pipe path.** ConPTY is a different API that the Linux CI leg cannot prove, and
  a half-working pseudo-console fails the same silent way as the bug being fixed. Stated as a known
  limitation in `secrets_child_windows.go`, not discovered later.
- **`cmd/spec.go:92`** carries the identical `io.MultiWriter` shape, but its child is the reviewer,
  headless by design — so the defect is **latent, not active**. Fixing it changes how every
  adversarial review's transcript is captured. Different blast radius, out of scope, recorded.

## Extent — checked, and not systemic

All 15 sites in `cli/internal` that assign a child's `Stdout`: **1 active** (this), **1 latent**
(`spec.go:92`), and 13 correct — `tools/install.go` defaults `in.Out` to `os.Stdout` whose dynamic
type *is* `*os.File`, and the rest are `&bytes.Buffer` capture-by-design.

## Evidence

From a binary **rebuilt from this branch** — the installed `dotf` is `vcs.modified=true`, built from
a dirty tree. Reported as exit status and byte counts, never the captured stream (~40KB of escape
sequences, and the channel `secrets-never-in-output` governs).

Harness: `pty.fork()` with `TIOCSWINSZ` set to 40x120 **after** the fork — a bare `pty.fork()`
leaves the window at 0x0 and the TUI refuses to draw, which is a false negative.

| Case | Old binary | New binary |
|---|---|---|
| `secrets run -- pi` under a real pty | `bytes=0`, dead in 1.5s | **`bytes=41904`, alive** |
| `pi` directly (reference) | — | `bytes=50969`, alive |
| `secrets run -- pi -p 'reply with exactly: OK'` | exit 124 @ 90s | **exit 0, `OK`, 118s** |
| `secrets run -- opencode --version` (canary) | — | exit 0 |

**`pi -p` looked like a regression and was not.** At a 90s budget it timed out — but so did the
*old* binary, identically. At 240s the new one returns `OK` in 118s. Only the controlled comparison
established that; the new binary alone would have supported either conclusion.

### Mutation proof (lesson 267)

The tests were proven in the failing direction, and the mutation was proven to have **landed** — the
patch asserts its anchor before applying, so a silent no-op cannot pass as a result:

```
mutation applied            <- holdBack reverted to the old fixed-window rule
--- FAIL: TestRedactWriter_ReleasesFrameWithNoSecretPrefixImmediately
--- FAIL: TestRedactWriter_HoldsBackATrailingSecretPrefix
mutated_rc=1 / restored_rc=0
```

### Checks

```
go build ./...                 clean
go vet ./...                   clean
GOOS=windows go vet ./...      clean
go test ./...                  all packages pass (4 pre-existing runChild tests unchanged)
golangci-lint run (2.12.2)     0 issues
```

## New dependency

`github.com/creack/pty v1.1.24`, behind a build tag. `x/term` and `x/sys` were already deps. Chosen
over hand-rolled `/dev/ptmx` ioctls because those differ across Linux and the BSDs, and a
per-platform path that only ever runs on the author's box is the same class of defect as the one
being fixed.


mlorentedev:
## Review triage

Recorded after the merge, so the disposition exists on the PR rather than only in the spec folder.

**Independent adversarial review, two rounds, reviewer `nan/deepseek-v4-flash`** (random draw from
`harness/reviewer-pool.json`, not the implementer):

| Round | SHA | Verdict |
|---|---|---|
| 1 | `a636844` | **FAIL** — one REAL Major |
| 2 | `f7fee84` | **PASS** — no Blocker, no Major, 10 Minor all THEORETICAL |

**The Major was correct and worth the round trip.** AC1 asked for the `isTerminal` seam to be
exercised in both directions; the seam existed and nothing touched it. Both path tests called
`runChildPTY`/`runChild` directly, bypassing the call-site condition — so the single line the whole
fix rests on was unverified, and inverting it would have restored the bug with every test still
green. Fixed by extracting `wantsInteractiveChild()` and driving the seam true and false, proven in
the failing direction.

Five of six round-1 findings applied; full dispositions in
`specs/archive/SEC-002-secrets-run-pty/verification.md`. Two were declined and both are stated
rather than silently dropped:

- The grandchild-holds-the-pty-slave hang — recorded as a known limitation, because a deadline would
  truncate a legitimately long-running TUI, which is the normal case here.
- `features.json` `state: "pass"` — the template reserves the terminal state for the harness after
  it runs the verification command. Evidence populated; `state` left `pending`.

**Both rounds' `review.md` are committed unedited** in the archived spec folder, so the provenance of
the FAIL→PASS flip is readable rather than asserted.

`pr-agent` and `CI`/`cli`/`spec-gate`/`review-attestation` all SUCCESS on the merged head.

PR #1512, opened 2026-09-05
docs(lessons): renumber the secrets-wrapper lesson to 270, resolving a collision

mlorentedev:
Rename plus one index row plus the H1. No content changes.

Two lessons numbered **269** reached `main` about ninety minutes apart, from two sessions working in
parallel:

```
ddad532  21:20  269 - a generated file fixed without its source is a countdown   (#1505)
46dc9d8  22:42  269 - a security wrapper that breaks the tool it protects        (#1508)
```

Mine is the later one, so mine renumbers to **270**. Nothing references the old filename, so the
change is contained.

**The part that actually misleads is the index**, which carried two rows both labelled 269 — a
reader following either link lands on a file whose number does not identify it.

## Cause, since the fix does not prevent a repeat

The number is allocated by reading `docs/lessons/`, and two sessions that read the directory before
either has written get the same answer. The check is a directory listing at authoring time and it
cannot see a lesson still sitting on someone else's branch. Neither session did anything wrong and
neither could have noticed.

Not fixing that here — a real fix is either a CI assertion that lesson numbers are unique in the
merged tree (cheap, catches it at the PR rather than after) or dropping sequential numbering
entirely. Worth a ticket; this PR just unbreaks the index.

### DIFF
--- a/docs/lessons/_index.md
+++ b/docs/lessons/_index.md
@@ -284,9 +284,9 @@ tags: [lessons, index, dotfiles]
 | [264 - When the documentation does not name a field, read the executable — twice it decided a design](lesson-264-a-hooks-timeout-semantics-are-not-in-the-documentation.md) | 2026-09-03 |  |
 | [267 - A mutation harness must prove the mutation landed](lesson-267-a-mutation-harness-must-prove-the-mutation-landed.md) | 2026-09-04 |  |
 | [268 - A refused question and a negative answer print the same thing](lesson-268-a-refused-question-and-a-negative-answer-print-the-same-thing.md) | 2026-09-04 |  |
-| [269 - A security wrapper that breaks the tool it protects gets routed around](lesson-269-a-security-wrapper-that-breaks-the-tool-it-protects-gets-routed-around.md) | 2026-09-05 |  |
 | [265 - A correct measurement answering the wrong question, three times in one session](lesson-265-a-correct-measurement-answering-the-wrong-question.md) | 2026-09-03 |  |
 | [266 - Two measurements agreeing is not reproducibility](lesson-266-two-measurements-agreeing-is-not-reproducibility.md) | 2026-09-04 |  |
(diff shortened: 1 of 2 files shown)


mlorentedev:
## Review triage

**Who reviewed:** `pr-agent` (full review, 00:55). **CodeRabbit did not review** — no findings posted. Proceeding on pr-agent alone, disclosed.

**Verdict on this PR's own ticket (#1508): "Non-compliant requirements: None."** All three claims verified compliant — the rename, the H1, and the index row.

| Finding | Disposition |
|---|---|
| **#1505 "Partially compliant — this PR does not create the lesson document itself"** | **Declined; the reviewer inferred the wrong ticket.** #1505 is another session's *pull request*, not a ticket this PR closes. It appears here only because this PR's body quotes its merge SHA to establish which of the two lesson-269s landed first. This PR deliberately creates no document — it renames one. (Known pattern: pr-agent's ticket-scope inference has now been wrong on the majority of PRs where it fires.) |
| "Verify lesson cross-references exist on disk (requires filesystem check)" | **Done, not declared.** Every `lesson-*.md` link in `_index.md` resolved against disk: **0 broken.** |
| "Run `scripts/check-doc-paths.sh` (requires local execution)" | **Done.** Exit 0. |
| "Verify pre-commit hooks green (requires local execution)" | **Done.** All five passed at commit time, including *Check referenced doc paths in instruction files*. |

**Checks:** `MERGEABLE CLEAN`, nothing outside SUCCESS/SKIPPED.

**Scope note, deliberate:** this PR does **not** touch `lesson-256`, which is duplicated too — confirmed independently here:

```
$ git ls-tree --name-only origin/main docs/lessons/ | grep -oE "lesson-[0-9]+" | sort | uniq -d
lesson-256
lesson-269
```

So the 269 pair is the **second** occurrence, and the 256 pair has been in `main` since 2026-09-01 with nobody noticing. That is the argument for a guard rather than a habit: the detected case is the exception. Filed by the peer session as **#1514** (CI uniqueness assertion, with the AC that it must be proven against real `main`, which currently has two such pairs — a check that only passes on a clean tree would have passed all week). Jointly found.

**Triaged late.** I reported this queue clear in my handoff before pr-agent had posted; the peer caught it. The queue was right and I was early.

PR #1509, opened 2026-09-05
feat(harness): architect becomes the fourth enforceable persona, and a typo in any skill id now fails

mlorentedev:
## Summary

- Migrates `architect` to the enforceable mapping form — the fourth of the six invocable personas #1497 covers, and the first that gates its whole list rather than a subset.
- Adds `TestEveryDeclaredSkillHasARecord`, closing a gap found by mutation: a typo in a skill id passed `--check`, `go test` and bats alike.
- Removes a false claim the record had carried since it was written (`"enforced by hook, not left to memory"`), and corrects a `block` precondition that named a since-closed issue.

## SDD checklist

- [x] Issue exists on the bitácora GitHub Project — part of **#1497** (epic; this PR migrates one persona and does not close it)
- [x] Diff conforms to the ~300 LOC atomic cap — well under; one persona record, one test file
- [x] Spec folder — covered by the epic's own scope rather than a new `specs/<id>/`; `spec-gate` passes because the production diff is below its threshold
- [x] `proposal.md` Why / What / Acceptance criteria — carried by #1497, which states the migration's acceptance criteria for all six personas
- [x] `tasks.md` in TDD order — the guard test was written and proven failing (by mutation) before the migration it protects
- [x] `verification.md` — evidence is inline under **Verification** below, produced in-session

## SDD skip rationale

Not applicable — no `skip-sdd` label. This PR is one step of the #1497 epic, not an exemption from it. One persona per PR is a measured decision, not a convenience: `harness/agents/**` is **not** in spec-gate's `_excluded()`, so persona prose counts as production LOC and batching three migrations would cross the 50-line threshold.

## Test plan

- [x] `bats tests/*.bats` — **1554/1554**, exit 0
- [x] `shellcheck` — not applicable; no `.sh` changed in this PR
- [x] `cd cli && go test ./...` — passes, `-count=1`
- [x] `./scripts/compile-harness.sh --check` — no drift
- [x] Manual verification: with the deploy dir stale, the gate is silent; run against `--repo-root` at this branch it emits three `[gate] warn` lines naming `read-all-adrs`, `architecture-session` and `pattern-loader`. **The deployed record is not this one** — see the caveat under Verification.

---

Fourth of the six invocable personas #1497 covers, after `reviewer` (#1412), `builder` (#1494) and `shipper` (#1504).

## architect gates all three, and it is the only persona so far where that is right

[...]

PR #1513, opened 2026-09-05
fix(harness): the doctrine cap was asserted in characters while the payload was 47 bytes over it

mlorentedev:
The doctrine cap is asserted in **characters**. The payload is **47 bytes over it**, and the guard cannot see that because it measures the other unit.

## The measurement

```
wc -m  (characters)  11974      <- what tests/skills-pipeline.bats:211 asserts
wc -c  (bytes)       12047      <- 47 OVER the 12000 cap
```

The 73-byte gap is 33 em-dashes, four section signs, one accented vowel and an ellipsis — all prose, all multi-byte UTF-8 that `wc -m` counts as one character and a byte-counting consumer counts as three.

`manifest.json:145-155` declares `char_cap: 12000`, citing *"caps EACH rules file at 12000 characters"*. **Nothing here has verified that the platform counts that way**, and this issue already established what happens if it does not: Antigravity drops the overflow **silently** at the destination. No red check, no error. The bats assertion would keep passing at 11974 while the tail of the file never arrives.

## The fix does not decide the unit — it removes the question

Rather than settle what Antigravity counts (which needs an experiment against the live consumer, not a reading of its docs), `deploy_doctrine` now normalises typographic punctuation to ASCII **in a capped payload**. Bytes and characters then track each other, so the guard and the platform cannot disagree however the platform counts.

```
before             11974 chars   12047 bytes   OVER
after              11976 chars   11985 bytes   OK, both under
em-dash -> ' - '   12042 chars   12042 bytes   OVER  (measured and rejected)
```

The spaced and doubled variants were measured too: both *add* characters and land back over the cap. A single hyphen is the only fold that helps in both units.

**Only punctuation that does not alter the lexicon.** Dashes, curly quotes and the ellipsis are typographic — folding them is uglier and means exactly the same thing. Accents and section signs are deliberately **not** folded: `bitácora` → `bitacora` changes a word, `§4` → `S4` changes a reference. Five such characters survive, which is why the invariant is stated as *"under the cap in both units"* rather than *"ASCII"*, and why the deploy prints a note when any survive.

The substitutions are written as **hex escapes rather than literals**: a curly quote in shell source is SC1112, which fails `shellcheck` and therefore CI lint. I hit that on the first attempt — 3 new findings, `shellcheck` rc 1 — and it is why the escapes are there rather than the obvious characters.

## The guard now measures both units

That is the part that keeps this from recurring. **Proven in the failing direction**: with the normalisation disabled — which is exactly today's `main` — the test goes red with

```
not ok 18 HARNESS-056: the compact doctrine payload carries it and stays under its cap
# .gemini/GEMINI.md is 12047 BYTES, at or over its 12000 cap (chars: 11974)
```

The mutation was confirmed to have landed before the verdict was read.

## This corrects a claim I made on #1241 tonight

I reported that a persona migration costs **exactly 0** characters and concluded the remaining migrations were free of cap risk. The character half is right and independently confirmed on four trees. But it is only true in the unit that may not bind:

| what grows | characters | bytes |
|---|---|---|
| a new skill id | 18 | 18 |
| **a persona's prose** | **0** | **2 per em-dash** |

Tonight's four persona migrations added roughly ten paragraphs of vault prose and moved the character count by **zero** — the roster renders skill ids only. They moved the byte count. So the per-id budget decided on #1241 has been tracking the measure that is not growing, and the decision was made on an incomplete picture. Full numbers are on that issue.

## Coordination

Checked with the parallel session before writing any of this, because #1495 is parked on the same cap. It does not collide: #1495 is 1632 characters over, not 47, and this recovers ~24 — a different problem sharing a number. It also does not pre-empt that issue's decision (*budget, do not raise the cap*): the cap value is untouched, and the stated reason for not raising it was precisely that overflow truncates silently, which this removes a path to.

The lexicon caveat above is theirs, adopted: writing the boundary down now saves the argument about folding accents in six months.

## Verification

- `bash -n` and `zsh -n` — both clean
- `shellcheck scripts/compile-harness.sh` — **no new findings**; rc 1 and its SC2016 infos are identical on `main`, and SC1112 count is **0**
- `bats tests/*.bats` — **1554/1554**, exit 0
- `--deploy` into a throwaway `HOME` — 11976 chars / 11985 bytes, both under 12000, idempotent on a second run
- The script is `#!/usr/bin/env bash`; running it under `zsh` fails on `main` too, so that is a pre-existing limitation and not a regression here

Refs #1241

## Spec

`specs/HARNESS-111/`, gated on #1241. Six acceptance criteria, all executed in the implementing session, with the two that matter proven in the failing direction rather than observed green.

The spec-gate refused this push at 67 production LOC without one — 26 lines of code and 54 of comment, and the gate counts both. Rather than strip the comments to get under the threshold, which is the explanation the next reader needs, or reach for `skip-sdd` on my own judgement, the spec is the sanctioned path and #1241 already gives it an owner.

### DIFF
--- a/scripts/compile-harness.sh
+++ b/scripts/compile-harness.sh
@@ -1070,7 +1070,7 @@ build_agent_presence() {
 has_doctrine() { jq -e '.doctrine.deploy' "$MANIFEST" >/dev/null 2>&1; }
 
 deploy_doctrine() {
-    local ag_recdir="$1" agent file cap shadow file_abs payload sha begin tmp chars gen_chars ids
+    local ag_recdir="$1" agent file cap shadow file_abs payload sha begin tmp chars bytes gen_chars gen_bytes ids
     mapfile -t ids < <(jq -r '.doctrine.inject[]' "$MANIFEST")
     while IFS=$'\t' read -r agent file cap shadow; do
         [[ -n "$agent" ]] || continue
@@ -1082,6 +1082,52 @@ deploy_doctrine() {
             printf '\n'
             build_agent_presence "$ag_recdir" "$agent"
         } > "$payload"
+
+        # A CAPPED payload is normalised to ASCII punctuation, because a cap is
+        # only enforceable when its unit is unambiguous — and ours was not.
+        #
+        # Measured 2026-09-05: .gemini/GEMINI.md was 11974 CHARACTERS and 12047
+        # BYTES against a 12000 cap. The bats assertion reads `wc -m`, so it
+        # passed; a byte-counting consumer would have been 47 over and, per the
+        # platform's documented behaviour, would have dropped the overflow
+        # SILENTLY. The 73-byte gap was 33 em-dashes, four section signs, an
+        # accented vowel and an ellipsis.
+        #
+        # The gap also hid a second thing: persona prose costs ZERO characters
+        # (the roster renders skill ids only) and two bytes per em-dash. So the
+        # per-id budget on #1241 was tracking the measure that was not growing.
+        #
+        # This does not decide which unit the platform counts. It removes the
+        # question: an ASCII payload has bytes == chars by construction, so the
+        # guard and the platform cannot disagree however the platform counts.
+        #
+        # ONLY punctuation that does not alter the lexicon. Em/en dashes, curly
+        # quotes and the ellipsis are typographic; folding them is uglier and
+        # means exactly the same thing. Accents and section signs are NOT folded:
+        # `bitácora` -> `bitacora` changes a word, and `§4` -> `S4` changes a
+        # reference. Five such characters survive, which is why the invariant
+        # below is stated as "under the cap in both units" rather than "ASCII".
+        # The codepoints are written as \u escapes rather than as literals: a
+        # curly quote pasted into shell source is SC1112 ("this is a unicode
+        # quote"), which fails `shellcheck` and therefore CI lint. Writing them
+        # as escapes also says WHICH character is meant, where a literal in a
+        # monospace diff does not.
+        if [[ "$cap" != 0 ]]; then
+            tmp="$(mktemp)"
+            # Hex escapes, not literals: a curly quote or dash pasted into shell
+            # source is SC1112 ("this is a unicode quote"), which fails
+            # `shellcheck` and therefore CI lint. The escapes also name the exact
+            # codepoint, which a literal in a monospace diff does not.
+            LC_ALL=C sed \
+                -e 's/\xe2\x80\x94/-/g'    `# U+2014 em dash` \
+                -e 's/\xe2\x80\x93/-/g'    `# U+2013 en dash` \
+                -e "s/\xe2\x80\x99/'/g"    `# U+2019 right single quote` \
+                -e "s/\xe2\x80\x98/'/g"    `# U+2018 left single quote` \
+                -e 's/\xe2\x80\x9c/"/g'    `# U+201C left double quote` \
+                -e 's/\xe2\x80\x9d/"/g'    `# U+201D right double quote` \
+                -e 's/\xe2\x80\xa6/.../g'  `# U+2026 ellipsis` \
+                "$payload" > "$tmp" && mv "$tmp" "$payload"
+        fi
 
         # A shadow file wins over ours at read time, so a silent deploy here
         # would look successful while changing nothing the agent ever reads.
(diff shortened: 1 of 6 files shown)


mlorentedev:
## Review triage

|---|---|---|
| **CodeRabbit** | *"Review limit reached."* | **Not a review — a notice.** Sixth consecutive notice tonight (#1494, #1504, #1505, #1509, this). **This change has not had CodeRabbit's opinion.** |
| **PR-Agent — #1241, item 1** | Adding a persona or doctrine id now has a defined outcome when the budget is exceeded. | Accepted. |
| **PR-Agent — #1241, "headroom reported before it is exhausted"** | The deploy warns *after* overflow; no proactive margin gauge. | **Real, and declined for THIS PR with a reason** — see below. |
| **PR-Agent — #1241, "whether codex should receive skill lists"** | Not addressed. | **Correctly out of scope**, and PR-Agent says so itself. |
| **PR-Agent — #1495 "Not compliant"** | No files from #1495's scope appear in this diff. | **Declined — #1495 is a PR, not a ticket.** |
| **CI** | 11 of 11, 0 failures. | Accepted. |

### The headroom gauge is a good idea and belongs to #1241, not here

PR-Agent is right that the current warning fires *at* the boundary rather than before it, and that a margin report would have surfaced this defect months earlier instead of tonight. The numbers are already computed — `chars`, `bytes`, `gen_chars`, `gen_bytes` — so emitting *"N of 12000, M remaining"* on every deploy is a few lines.

Declined here for a scope reason rather than a cost one: **a headroom gauge is a budget instrument, and the budget policy is #1241's open decision.** What the threshold should be (200 characters? a percentage? per-target?) is exactly the question that issue is deciding, and picking one in this PR would answer it by implementation while the owner is still choosing. This PR fixes the *unit* the budget is counted in, which is a precondition for any threshold being meaningful — a gauge reading "26 remaining" in characters while 47 bytes over would have been a confidently wrong instrument.

Carried to #1241 rather than dropped.

### #1495 is a pull request, not a ticket

PR-Agent evaluated this diff against "#1495's scope" and found it non-compliant. #1495 is **the parallel session's open PR**, which this PR mentions only because it is parked on the same cap and I coordinated before writing. Nothing here was ever supposed to satisfy it, and its files are deliberately untouched — reverting another branch's doctrine regen out of my `--refresh` was one of the things this session did on purpose.

Same PR-vs-ticket inference already recorded on **#1466**, and the third occurrence tonight after #1494 and #1440. Its code findings continue to be accurate; its ticket inference continues not to be.

### What this PR actually establishes, restated because the compliance table buries it

`.gemini/GEMINI.md` was **11974 characters and 12047 bytes** against a 12000 cap, and the guard asserted the unit that passed. After the fold: **11976 / 11985**, both under.

The assertion is proven in the failing direction — with the normalisation disabled, which is today's `main`:

```
not ok 18 HARNESS-056: the compact doctrine payload carries it and stays under its cap
# .gemini/GEMINI.md is 12047 BYTES, at or over its 12000 cap (chars: 11974)
```

The mutation was confirmed present before the verdict was read, per lesson 267.

### Not claimed

**Which unit Antigravity counts is still unverified.** This change makes the answer stop mattering for the payload; it does not establish it. The experiment that would — a marked sentinel at the end of the payload, deployed, then asked for — is recorded on #1241 and is deliberately not part of HARNESS-111.

### Verification

`bash -n` / `zsh -n` clean · `shellcheck` **no new findings vs main**, SC1112 count **0** (the literal-character first attempt measured 3, which is why the escapes are hex) · `bats tests/*.bats` **1554/1554** · `--deploy` idempotent on a second run into the same `HOME`.

Spec `specs/HARNESS-111/`, six acceptance criteria, all executed in the implementing session.

mlorentedev:
## Review triage

Second disposition on this PR. The first (07:04Z) covered the review as it stood then; `update-branch` onto `46dc9d8` at ~17:30Z made both reviewers re-run, and **both edited their existing comments in place**, so `dotf pr triage-queue` still reported this PR as clear. It was not. Measurement filed on #1422 — fourth occurrence.

|---|---|---|
| pr-agent (re-run, `Review updated until commit beacea20`) | No major issues detected; no security concerns; tests present; review effort 2/5 | **Nothing to apply.** Re-read against the new head, not carried over from the first run. |
| pr-agent — ticket compliance, #1241 | "Non-compliant: adding a persona or doctrine id has a defined outcome when the budget is exceeded" and "whether `codex` should receive skill lists it has no skills for is decided" | **Declined — out of scope by design, stated in the spec.** `specs/HARNESS-111/proposal.md` §What: *"Out of scope: the cap value, the per-id budget policy (#1241's own decision), and any vault doctrine prose."* This PR fixes the unit the cap is measured in; it does not resolve #1241 and does not claim to. |
| pr-agent — ticket compliance, #1495 | "Non-compliant: land the pending harness regen / stop `--refresh` reverting the new-ticket skill" | **Declined — #1495 is a pull request, not a ticket.** Second instance of the same inference error today (the first treated #1433 as a ticket on #1440, the third treats #1494 and #1412 as tickets on #1509). Carried to the peer session that owns reviewer configuration. |
| pr-agent — headroom gauge | Remaining headroom per capped target is reported before exhaustion | **Already satisfied by this PR**, and the broader gauge is carried on #1241. |
| CodeRabbit | *"Review limit reached — next included review available in 59 minutes. You've used all free OSS reviews for now."* | **Not a review. No CodeRabbit review ran on this PR.** Recorded as a disclosure, not a pass — a notice that talks about the review rather than about the code leaves the PR unreviewed by that reviewer. |

**Disclosure:** the only substantive review on this PR is pr-agent's. CodeRabbit produced a capacity notice on both runs. Proceeding is allowed; proceeding silently is not, which is why this row exists.

**CI:** 10/10 green on `beacea20`, the post-`update-branch` head — verified against that SHA rather than against the superseded runs REST also returns for the previous head.

Issue #1516, opened 2026-09-05
BUG-093: Gate f is silently off on Windows — the worktree reaper deletes on 'I could not find out'


mlorentedev:
## What is wrong

`isHostProcessInside` (Gate f — "is a live process using this worktree?") read `/proc` with **no build-tag split**, and returned `false` when `os.ReadDir("/proc")` failed:

```go
entries, err := os.ReadDir("/proc")
if err != nil {
    return false   // "no process is inside"
}
```

On Windows `/proc` never exists, so the gate answered *"nobody is in there"* on every run. `isCandidateForReap` and `reapSingleWorktree` **delete the worktree on that `false`**, in a reaper whose stated premise (#1500) is fail-closed.

Same for any non-Linux platform, and for a Linux box where `/proc` is unreadable.

## How it was found

CodeRabbit flagged it on #1515 as a **macOS** exposure. This repo does not run macOS. It runs **Windows**, where `test (windows-latest)` is a required status check and the maintainer has a working box.

The tell is inside the same package: `lock_unix.go` and `lock_windows.go` both exist, so the file lock crossed the platform boundary deliberately and the process gate did not. Nothing caught the difference because the Linux CI leg exercises the only path that works.

## Why the review missed it

The refusing branch is executable **only on a platform this CI does not have**, so reading the code was the only available check. That is the same shape as the separator bug fixed in the same PR, and as lesson 268: the broken result and the correct one print identically.

## Fix

Split by build tag:

- **`sweep_proc_linux.go`** keeps the `/proc` walk and answers `true` on *every* failure path. "I could not find out" and "nobody is in there" must not produce the same answer when one of them deletes uncommitted work.
- **`sweep_proc_other.go`** answers `true` unconditionally until a real implementation exists (`ToolHelp32Snapshot` + per-process cwd on Windows, `libproc` on Darwin).
- `SweepReport.ProcessDiscovery` reports the inertness, and the command says so **before** the counts: an inert sweep prints `reaped 0` exactly like a machine with nothing to clean up.
- Gate f is reached through a package-level variable so the refusing branch is testable on any platform; `sweep_proc_other_test.go` asserts the platform answer itself on the Windows leg.

## Accepted consequence

`dotf worktree sweep` is **inert on Windows** until someone implements process enumeration there. That is deliberate: an inert sweep costs disk, a wrong one costs uncommitted work.

## Known gap

Mutation testing (anchors asserted before applying):

```
[...]

Issue #1500, opened 2026-09-05
CLI-075: dotf worktree lifecycle and fail-closed garbage collection


mlorentedev:
## Problem

Developer workstations accumulate dozens of stale, abandoned, or merged Git worktrees across parallel agent sessions (Claude Code, Antigravity, OpenCode, Pi). Today, 17 worktrees linger across ~/Projects/ because manual cleanup is hindered by legitimate fear of data loss (uncommitted changes, untracked scratch files, active agent processes).

Automating cleanup naively introduces critical failure modes:
1. False-positive deletion of freshly initialized branches (empty diff against base).
2. Deleting active workspaces of containerized or PID-namespaced agents invisible to host /proc.
3. Dumping plaintext secrets (.env) and bloated build caches (node_modules/, target/) into quarantine archives, violating ADR-028.
4. Concurrent sweep races between parallel agent sessions.

## Decision

Implement a fail-closed, lease-based worktree lifecycle management subsystem in Go under cli/internal/worktree/ exposed via dotf worktree:

1. Reconciliation-First Discovery: git worktree list --porcelain per-repo is the primary source of truth. Rejects submodules (gitdir pointing to modules/).
2. Lease/Heartbeat State Machine (.dotf-worktree.json): Worktrees declare intent, owner, and lease inside .dotf-worktree.json (ignored via .git/info/exclude).
3. Fail-Closed Reaper (dotf worktree sweep): Reaps only when positive proof exists (lease expired + git status clean + PR merged in GitHub + minimum age passed + flock acquired).
4. Git-Aware Quarantine: Never tar whole trees. Preserves uncommitted work via git diff patches and untracked files (excluding build caches and secrets).
5. Subcommands: dotf worktree list, dotf worktree add, dotf worktree sweep.

Issue #1517, opened 2026-09-05
CLI-076: Git-Aware Quarantine — requirement 4 of #1500 shipped unbuilt


mlorentedev:
## What is missing

Issue #1500 listed four requirements for `dotf worktree`. **#1515 shipped three.** Requirement 4 has no implementation:

> **4. Git-Aware Quarantine:** Never tar whole trees. Preserves uncommitted work via `git diff` patches and untracked files (excluding build caches and secrets).

`cli/internal/worktree/` contains no quarantine code. Verified by grep across the package after the merge: every `quarantine` hit in `cli/` is in `internal/doctor/checks_pi_extensions*`, which is unrelated.

## Why the descoping does not cover it

The CLI-075 proposal's **Out of scope** section reads:

> Full-directory tarball archives that duplicate `node_modules/`, `target/`, `.venv/` or risk leaking unencrypted `.env` secrets into trash directories (ADR-028 compliance).

That excludes the **tarball** form. Requirement 4 asks for precisely the *non-tarball* alternative — `git diff` patches plus selected untracked files. The two are opposite halves of the same sentence in the ticket, and only one of them was descoped.

The archived spec's independent review PASSed without noticing, and the archived `verification.md` marks every AC as passed.

## Why it matters

The whole premise of #1500 is that manual worktree cleanup is blocked by *"legitimate fear of data loss (uncommitted changes, untracked scratch files)"*. Requirement 4 is the part that removes the fear. Today:

- `dotf worktree done --force` removes a worktree with uncommitted changes and preserves nothing.
- `sweep` gates on `IsDirty`, so it refuses dirty trees — but a refusal is not preservation, and the ticket asked for preservation.

## Process note

**#1500 was closed by #1515's `Closes #1500`, with a quarter of it unbuilt.** This ticket is the honest remainder rather than a reopen, so the closed issue's history stays intact. Flagged during triage of #1515 by pr-agent, whose ticket-compliance finding was correct here — worth recording, because this repo's notes had it wrong 5 times out of 5 before this one.

## Acceptance criteria

- [ ] Dirty worktree teardown writes a `git diff` patch (tracked changes) rather than copying a tree.
- [ ] Untracked files are preserved by explicit selection, never by wholesale copy.
- [ ] Build caches (`node_modules/`, `target/`, `.venv/`) are excluded by construction, not by an allowlist that can be widened.
- [ ] `.env` and anything matching the secrets registry is **never** written to a quarantine location (ADR-028).
[...]

PR #1515, opened 2026-09-05
feat(cli): dotf worktree lifecycle and fail-closed garbage collection (#1500)

mlorentedev:
## Problem

Developer workstations accumulate dozens of stale, abandoned, or merged Git worktrees across parallel agent sessions (Claude Code, Antigravity, OpenCode, Pi). Today, 17 worktrees linger across `~/Projects/` because manual cleanup is hindered by legitimate fear of data loss (uncommitted changes, untracked scratch files, active agent processes).

Naive auto-cleanup introduces critical failure modes:
1. False-positive deletion of freshly initialized branches (empty diff against base).
2. Deleting active workspaces of containerized or PID-namespaced agents invisible to host `/proc`.
3. Destroying local scratchpads (`.env`, `notes*`, `*.tmp`).
4. Concurrent sweep races between parallel sessions.

## Solution

Introduces a fail-closed, lease-based worktree lifecycle management subsystem in Go under `cli/internal/worktree/` exposed via `dotf worktree`:

1. **Reconciliation-First Discovery (`dotf worktree list`)**: Discovers worktrees deterministically via `git worktree list --porcelain`. Rejects submodule checkouts (`modules/`). Evaluates worktree state machine (`ACTIVE`, `REAPABLE`, `DIRTY`, `UNMERGED`, `ORPHAN`). Supports `--all` for multi-repo scanning across `~/Projects` and `--json`.
2. **Standardized Creation (`dotf worktree add`)**: Enforces external sibling directory topology (`<repo>-wt-<slug>`). Initializes in-tree `.dotf-worktree.json` lease metadata and excludes it via `.git/info/exclude`.
3. **Fail-Closed Reaper (`dotf worktree sweep`)**: Protects against data loss by enforcing 6 sequential gates under a cross-process flock:
   - Gate A: Lease expired (`now > lease_expires_at`) with explicit opt-in (`reap_ok: true`).
   - Gate B: Branch merged into upstream base ref (`merge-base --is-ancestor` or GitHub PR merged). Caches PR lookups via `RealGitRunner` to prevent rate limiting.
   - Gate C: Zero uncommitted changes, untracked files, or non-disposable gitignored scratchpads (`.env`, `notes*`, `*.tmp`).
   - Gate D: Minimum age guard (>15m) to prevent 0-diff race on fresh branches.
   - Gate E: Not current working directory or host process cwd.
   - Gate F: Re-checked under file lock immediately prior to branch deletion. Full 40-char commit SHA logged to stderr for full recoverability.
[...]

PR #1518, opened 2026-09-05
fix(cli): make Gate f fail closed where it has no implementation

mlorentedev:
> **NOT READY — do not merge.** Round-2 adversarial review returned **FAIL** on seven REAL
> Majors. `Closes #1516` was deliberately REMOVED from this body: the issue is not fixed,
> and leaving the keyword would both close it falsely and keep the archive-on-merge gate
> demanding a spec archive that cannot happen without a passing review.

Refs #1516. Spec: `specs/BUG-093-gate-f-fail-closed/`.

Found while triaging **#1515** (now merged). CodeRabbit raised it there; this is the fix, split
out because #1515 merged before it was ready.

## The defect

`isHostProcessInside` is Gate f — the check that asks whether a live process is still using a
worktree before `sweep` deletes it. It read `/proc` with **no build-tag split**:

```go
entries, err := os.ReadDir("/proc")
if err != nil {
    return false   // "no process is inside"
}
```

On Windows `/proc` never exists, so the gate answered *"nobody is in there"* on **every run**.
`isCandidateForReap` and `reapSingleWorktree` delete on that `false`. A reaper whose entire
premise (#1500) is fail-closed had one gate that failed open, on a platform this repository
gates CI on and the maintainer actively uses.

**The reported platform was the wrong one.** CodeRabbit filed it as a macOS exposure; this repo
does not run macOS. Its blast radius was understated, not overstated.

The tell was inside the same package: `lock_unix.go` and `lock_windows.go` both exist, so the
file lock crossed the platform boundary deliberately and the process gate did not. Nothing
caught the difference because the Linux CI leg exercises the only path that works.

## Fix

- **`sweep_proc_linux.go`** keeps the `/proc` walk, and now answers `true` on *every* failure
  path. "I could not find out" and "nobody is in there" must not produce the same answer when
  one of them deletes uncommitted work.
- **`sweep_proc_other.go`** answers `true` unconditionally until a real implementation lands
  (`ToolHelp32Snapshot` + per-process cwd on Windows, `libproc` on Darwin).
- **`SweepReport.ProcessDiscovery`**, and the command says so **before** the counts:

```
note: no process-liveness check on this platform, so nothing can be reaped.
      Gate f refuses rather than guesses; remove one deliberately with `dotf worktree done`.
```

An inert sweep prints `reaped 0` exactly like a machine with nothing to clean up. Those are
different facts and the tool should not report them identically — which is this repo's
recurring failure shape, and lesson 268's.

## Two test files, and neither is redundant

The refusing branch is executable **only on a platform this CI does not have**, so reading the
code was the only check available — which is how the defect survived an adversarial review.

- Gate f is now reached through a package-level variable, so `sweep_proc_test.go` can drive
  **both** answers anywhere and assert the *caller* refuses. The allowing direction is there so
  the refusing one cannot pass vacuously by the gate refusing everything.
- `sweep_proc_other_test.go` (`//go:build !linux`) asserts the **platform answer itself**, and
  runs on the `test (windows-latest)` leg.

A seam test passes over a wrong implementation; a platform test cannot see whether the caller
honours the answer. Neither alone is sufficient.

## Accepted consequence

`dotf worktree sweep` is **inert on Windows** until process enumeration exists there. Deliberate:
an inert sweep costs disk, a wrong one costs uncommitted work. Same reasoning that kept the pipe
path on Windows in SEC-002 rather than shipping a half-working ConPTY.

## Verification

```
go build ./...                          clean
go test ./...                           all packages pass
GOOS=windows go vet ./...               clean
GOOS=darwin  go vet ./internal/worktree clean
golangci-lint run (2.12.2, the pin)     0 issues
```

All six `features.json` verification commands executed, all `rc=0`.

### Mutation, with the anchor asserted before applying

```
caller ignores Gate f                      mutated_rc=1  CAUGHT
linux impl fails open on unreadable /proc  mutated_rc=0  not caught locally
restored_rc=0
```

**The second gap is real and is not closed here.** Reaching it needs an unreadable `/proc`,
which the test cannot construct without injecting the directory read — a restructuring of a
function this change otherwise only moves. Recorded rather than presented as covered.

## Also filed

**#1517** — requirement 4 of #1500 (Git-Aware Quarantine) shipped unbuilt. #1515 carried
`Closes #1500`, so the issue closed with a quarter of it missing; the spec's *Out of scope*
excluded the *tarball* form, and the ticket asked for the `git diff` patch form, which is the
opposite half of the same sentence.


mlorentedev:
## Review triage

Three sources of reviewer output here, and they are not equivalent.

**1. Independent adversarial review (the one that mattered): round 1 FAIL.**
`agy/gemini-3.1-pro-high`, random draw from the pool, not the implementer. **1 Blocker, 2
Majors, all three REAL, all three applied** in `1de7910`. Full disposition table in
`specs/BUG-093-gate-f-fail-closed/verification.md`; the signed `review.md` is committed
unedited.

| Severity | Finding | Disposition |
|---|---|---|
| **Blocker** | `filepath.Abs` does not resolve symlinks; `/proc/<pid>/cwd` is already resolved by the kernel. A worktree reached through a symlinked path never matched, so Gate f answered "nobody is inside" with a shell sitting in it | **Applied.** `EvalSymlinks` on the target; `TestGateFSeesThroughASymlinkedWorktreePath` is the regression, and removing the resolution fails it |
| **Major** | `os.Readlink` errors return `false`, contradicting AC2's "every failure path answers true" | **Applied, and AC2 amended.** The reviewer was right that the code violated it *and* right that the obvious fix is wrong: EACCES on `/proc/1/cwd` for every non-root caller would make `sweep` permanently inert. A per-process read is now three-valued and `unreadable` is counted and surfaced |
| **Major** | `sweep_proc_linux.go` had zero coverage — only the seam was tested | **Applied.** `sweep_proc_linux_test.go`, six tests against the real `/proc` with live child processes |

**The AC2 amendment is declared, not slipped in.** Editing acceptance criteria after a review is
exactly what the archive gate exists to catch. It was done under a FAIL, not a PASS, it makes
the criterion weaker in letter and truer in substance, and `verification.md` carries the reason
in full for the round-2 reviewer to read alongside it.

**2. pr-agent** — ticket compliance for #1516: **no non-compliant requirements**, every stated
criterion listed back as met. Reviewed the pre-round-2 commit, so it did not see the Blocker
fix. **Noted.** Its ticket-scope inference again pulled in **#1515**, which is a pull request
and not a ticket — **declined**, the recurring error.

**3. CodeRabbit** — no review object and no inline comments on this PR. **Recorded rather than
assumed**: an unreviewed PR and a reviewed one look identical on a green board.

### Worth stating, because no reviewer found it

Changing Gate f's return type broke `sweep_proc_other_test.go` and **`go test ./...` on Linux
stayed green** — it never compiles a `//go:build !linux` file. That is this PR's own defect
class, reproduced by its author while fixing it. `GOOS=darwin go vet` caught it, which is why
AC7 names the cross-compile vet instead of trusting the Linux suite.

### Also filed from this work

**#1517** — Git-Aware Quarantine, requirement 4 of #1500, shipped unbuilt by #1515.

mlorentedev:
## Review triage

**Round 2: FAIL. Seven REAL Majors. This PR is now a draft and must not merge.**

`nan/qwen3.8-flash`, random draw, not the implementer, reviewing `dab7b6e`. It ran its own
mutations rather than reading — several findings below are "I deleted this line and the suite
stayed green", which is the only evidence that settles a coverage claim.

`Closes #1516` has been **removed from the body**. Leaving it would close the issue falsely and
would keep the archive-on-merge gate demanding a spec archive that cannot happen without a
passing review.

| # | Finding | Verdict |
|---|---|---|
| 1 | Gate f still fails open through a **mount namespace or bind mount** — `/proc/<pid>/cwd` resolves in the *process's* namespace, so a container mounting the worktree reports a different path entirely. The same door AC2c closed for symlinks, one step further out | **Accepted.** Real, unfixed |
| 2 | **`Uninspectable` is permanently maxed out.** `/proc/1/cwd` is EACCES on every Linux machine, so the count is never zero — it is not an anomaly signal, it is the normal state | **Accepted, and this is the serious one** — see below |
| 3 | `TestGateFMatchesASubdirectoryButNotASiblingWithASharedPrefix` **never asserts the "but not a sibling" half**; `otherRoot` is the same path as `target`, already asserted, so the block is a tautology whose failure branch calls `t.Skip` | **Accepted** |
| 4 | No test drives `reading.Uninspectable++`. Mutation to a no-op → **suite green, NOT CAUGHT** | **Accepted** |
| 5 | After `dab7b6e`, **no test drives Gate f's refusal through `SweepWithRunner` at all** — stubbing the seam to `false` in all five sweep tests removed the only integration-level coverage of the refusal→skip path | **Accepted.** My fix for the Windows red created this |
| 6 | `proposal.md` claims the drift risk is "**pinned by a test**". On Linux that test cannot fail: it skips when the constant is false, and its only assertion is unreachable | **Accepted.** A false claim in the spec |
| 7 | Round 2 **shipped a red required check on the platform this spec exists for** (`1de7910`, fixed at `dab7b6e`) | **Accepted.** Happened, and the reviewer is right to count it |

Four Minors also stand: the `Abs`/`EvalSymlinks` fail-closed branches are untested and
`verification.md` declares only one of the three gaps; AC6's ordering is asserted by `grep`, and
deleting the whole note block or moving it after the counts both leave the suite green; two
verification claims are stale; and **GUARD-010 #1470 is already open** for the `features.json`
state vocabulary I offered as a promotion candidate — filed, and I did not check.

### Why this is not a patch round

**Finding 2 invalidates the answer to round 1.** Round 1's Major was that an unreadable
`/proc/<pid>/cwd` silently reads as "not inside". The fix was to count those and report them.
The count is non-zero on every Linux machine, always — so it distinguishes nothing, and the
mitigation is vacuous in exactly the way this whole spec is about.

Findings 3, 4 and 6 are the same shape one level down, and all three are mine: **a test whose
name states a property it does not check**, a counter nothing exercises, and a spec claiming a
guarantee its test cannot provide. That is the thesis of this change, committed three times by
its author while writing it.

The honest position is that the observability design needs rethinking, not patching, so the
next session starts from the review rather than from a diff.

### What does stand

The **Blocker from round 1 is genuinely fixed and mutation-proven**: removing `EvalSymlinks`
fails `TestGateFSeesThroughASymlinkedWorktreePath`. Finding 1 extends that door rather than
reopening it. The build-tag split itself — the original defect, Gate f silently off on Windows —
is not disputed by either round.

`review.md` and its `review-request.json` are committed unedited at `eb9558b`, before any
response was written.

PR #1519, opened 2026-09-05
fix(docs): give every lesson its own number, and a guard so the next one cannot land

### LINKED ISSUE #1514: DOCS: lesson numbers collide between parallel sessions, and one collision has been in main unnoticed since 2026-09-01
## Why

Lesson numbers are assigned by listing `docs/lessons/` and taking the next free integer. Two sessions that list the directory before either has merged get the **same answer**, and nothing downstream rejects the collision.

Measured on `origin/main` 2026-09-05:

```
$ git ls-tree -r --name-only origin/main docs/lessons/ | grep -oE 'lesson-[0-9]+' | sort | uniq -d
lesson-256
lesson-269
```

```
docs/lessons/lesson-256-a-cleanup-blocks-own-description-of-what-it-deletes.md
docs/lessons/lesson-256-hooks-reload-personas-do-not-and-the-difference-was-assumed-away.md
docs/lessons/lesson-269-a-generated-file-fixed-without-its-source-is-a-countdown.md
docs/lessons/lesson-269-a-security-wrapper-that-breaks-the-tool-it-protects-gets-routed-around.md
```

**Two collisions, not one.** 269 happened tonight between two parallel sessions and was caught by one of them noticing; **256 has been sitting in `main` since 2026-09-01 and nobody noticed at all.** That is the part that makes this worth a guard rather than a habit: the detected case is the exception.

## What actually breaks

Not the files — they have distinct slugs and both are readable. **The index.** `docs/lessons/_index.md` ends up with two rows labelled the same number, so following either link lands on a file whose number does not identify it. Every cross-reference of the form "see lesson 269" becomes ambiguous, and this repository's lessons cite each other constantly — the two 269s are, by coincidence, the same family and would naturally cite one another.

## Why nobody could have caught it

The check available at write time is a listing of the working tree. It cannot see a lesson that exists only on another session's branch. Both authors did the correct thing and got the same number, which is a property of the numbering scheme rather than a mistake by either.

The same reasoning that made `MEMORY.md` handoffs move to `dotf mem handoff-write`: one shared mutable counter with N concurrent writers is a data-structure problem, and being more careful was never going to fix it.

## What

Either of these; they are alternatives, not steps.

**(a) Assert uniqueness on the merged tree.** A CI check that fails when two files under `docs/lessons/` share a `lesson-NNN` prefix. Cheap, catches it at the point where the collision actually becomes real, and would have caught 256 four days ago. Does not prevent the collision — it makes the second PR red, which is the right place for the cost to land.

**(b) Abandon sequential numbering.** Date-prefixed slugs (`2026-09-05-a-generated-file-fixed-without-its-source.md`) cannot collide between sessions, and `docs/lessons/` already contains one file in that form. The cost is every existing cross-reference by number, of which there are many.

Recommend **(a)**: it is a few lines, it is compatible with the existing corpus, and (b) can still follow later if numbering keeps hurting.

## Acceptance criteria

- [ ] A check fails when two files under `docs/lessons/` share a `lesson-NNN` prefix, naming both paths.
- [ ] Proven in the failing direction against the real `main`, which currently contains **two** such pairs — a check that only passes on a clean tree would have passed all week.
- [ ] The existing 256 collision is resolved (renumber the later one, fix its `_index.md` row and H1).
- [ ] `_index.md` rows and file numbers are asserted to agree, since the index is where the ambiguity actually bites.

## Notes

Found jointly by two parallel sessions; 269 was raised by the one that noticed, 256 by the other while measuring the first. Neither is "the" reporter. The 269 collision is already being resolved by renumbering the later file to 270 (#1512); this issue is the guard and the 256 residue, which that PR deliberately does not touch.

mlorentedev:
Closes #1514.

Three collisions reached `main` from parallel sessions. All three were found by a human noticing, never by a check:

| Number | What happened |
|---|---|
| `lesson-256` | two files, in `main` since **2026-09-01**, unnoticed for four days |
| `lesson-269` | two files, ninety minutes apart on 2026-09-05, fixed by #1512 |
| `lesson-268` | two files, landed by **#1515** on 2026-09-05 |

**The third one settles the argument.** It landed in a PR that was being triaged file by file at the time — by an agent reading the Go diff and twenty reviewer comments, hours after identifying the other two. "Be more careful" has now been tried and lost three times out of three.

## The fix

```
lesson-256-hooks-reload-personas-...  ->  lesson-271-...   (2026-09-02)
lesson-268-fail-closed-worktree-...   ->  lesson-272-...   (2026-09-05)
```

Later file of each pair renumbered, frontmatter `id` and heading updated, both added to `_index.md` — **neither was there**.

## Why the orphan check is not a second nicety

The duplicate set and the orphan set were **identical** on `main` at `4e2e2c4`: exactly two files, the same two, no others.

That is the shape of the event, not a coincidence. A session writes the lesson, takes a number that is already gone, and its index line either never arrives or loses the same race the number did. The orphan assertion catches the same mistake one step earlier — while it is still a missing line rather than a lost lesson, and a lesson absent from the index is invisible, since the index is the only surface anything reads.

## Why four assertions

The fourth exists because **fixing the collision broke a link**: `lesson-264` pointed at `[[lesson-256-hooks-reload-...]]`, and the rename left it dangling. A dangling wikilink is not a separate hazard — it is the predictable second-order effect of acting on the first three.

Writing it surfaced something worth keeping: **two link conventions are in use and both are legitimate.** `[[lesson-212]]` names a lesson by number; `[[lesson-268-full-slug]]` names the file. Asserting only the second reported correct state as broken — the same defect as reporting broken state as correct, with the sign flipped. Running against the real tree is what caught it; a fixture carrying one convention would have passed. That is also why there is no fixture: a check that only passes on a synthetic clean tree would have passed every one of the four days `lesson-256` sat duplicated.

## Verification

```
bats tests/guard-lesson-numbers-unique.bats   1..4, all ok
```

Mutation, anchor asserted before applying:

```
mutation applied  <- lesson-272 renamed back to the colliding 268
mutated_rc=1
    not ok 1 guard: no two lesson files share a number
    #   lesson-268-a-refused-question-and-a-negative-answer-print-the-same-thing.md
    #   lesson-268-fail-closed-worktree-garbage-collection-and-in-tree-leases.md
    not ok 2 guard: every lesson file is listed in _index.md
    not ok 3 guard: every _index.md lesson link points at a file that exists
restored_rc=0
```

All three fire on the real historical state, and the failure output names the colliding files rather than only reporting a count.

### DIFF
--- a/docs/lessons/_index.md
+++ b/docs/lessons/_index.md
@@ -288,5 +288,7 @@ tags: [lessons, index, dotfiles]
 | [266 - Two measurements agreeing is not reproducibility](lesson-266-two-measurements-agreeing-is-not-reproducibility.md) | 2026-09-04 |  |
 | [269 - A generated file fixed without its source is a countdown, and the drift guard says OK either way](lesson-269-a-generated-file-fixed-without-its-source-is-a-countdown.md) | 2026-09-05 |  |
 | [270 - A security wrapper that breaks the tool it protects gets routed around](lesson-270-a-security-wrapper-that-breaks-the-tool-it-protects-gets-routed-around.md) | 2026-09-05 |  |
+| [271 - Hooks are re-read, personas are not, and one measurement was generalised into both](lesson-271-hooks-reload-personas-do-not-and-the-difference-was-assumed-away.md) | 2026-09-02 |  |
+| [272 - Fail-closed worktree garbage collection requires in-tree leases and positive confirmation](lesson-272-fail-closed-worktree-garbage-collection-and-in-tree-leases.md) | 2026-09-05 |  |
 
 - [2026-08-30 Intercepting Cobra Errors without Breaking SilenceErrors](2026-08-30-cobra-silence-errors-interception.md)
--- a/docs/lessons/lesson-264-a-hooks-timeout-semantics-are-not-in-the-documentation.md
+++ b/docs/lessons/lesson-264-a-hooks-timeout-semantics-are-not-in-the-documentation.md
@@ -96,7 +96,7 @@ correct forever had the suite happened to hold four subtests.
 ## Related
 
 - `specs/HARNESS-109/proposal.md` — the refutation and the design it forced.
-- [[lesson-256-hooks-reload-personas-do-not-and-the-difference-was-assumed-away]] — the same class:
+- [[lesson-271-hooks-reload-personas-do-not-and-the-difference-was-assumed-away]] — the same class:
   a harness behaviour assumed rather than measured.
 - [[lesson-260-the-path-with-no-seam-is-the-path-with-no-test]] — why the timeout assertion had to
   run through `MergeHooks` rather than marshal the struct.
--- a/docs/lessons/lesson-271-hooks-reload-personas-do-not-and-the-difference-was-assumed-away.md
+++ b/docs/lessons/lesson-271-hooks-reload-personas-do-not-and-the-difference-was-assumed-away.md
@@ -1,11 +1,11 @@
 ---
-id: 256
+id: 271
 title: "Hooks are re-read and personas are not, and one measurement was generalised into both"
 date: "2026-09-02"
 tags: [lesson, harness, orchestrator, measurement]
 ---
 
-# 256 — Hooks are re-read, personas are not, and one measurement was generalised into both
+# 271 — Hooks are re-read, personas are not, and one measurement was generalised into both
 
 ## What happened
 
--- a/docs/lessons/lesson-272-fail-closed-worktree-garbage-collection-and-in-tree-leases.md
+++ b/docs/lessons/lesson-272-fail-closed-worktree-garbage-collection-and-in-tree-leases.md
@@ -1,5 +1,5 @@
 ---
-id: lesson-268
+id: lesson-272
 type: lesson
 status: active
 created: "2026-09-05"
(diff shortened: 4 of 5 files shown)


mlorentedev:
## Review triage

**Who actually reviewed:** `pr-agent` (github-actions) posted a PR Reviewer Guide.
**CodeRabbit did not review this PR** — no review object, no inline comments. Recorded because
a green board and an unreviewed PR look identical, and this repo has been caught by that three
times (#906).

| Finding | Disposition |
|---|---|
| #1514 ticket compliance: **no non-compliant requirements**; all four stated requirements met | **Noted.** The four it lists back are the four assertions |
| No code findings, no security concerns, effort 1/5 | **Noted** |
| Ticket-scope inference pulled in **#1512** and **#1515** as tickets this PR must satisfy | **Declined.** Neither is a ticket — #1512 and #1515 are pull requests. This is the recurring PR-Agent ticket-scope error, now 6 wrong out of 7 (it was right once, on #1515's quarantine gap → #1517) |

### Not found by either reviewer, and worth stating

`bats tests/*.bats` is not what proves this change. The proof is the **mutation**: restoring the
`lesson-268` collision fails three of the four assertions and names both colliding files. A
guard added to a tree it already passes on is indistinguishable from a guard that asserts
nothing, and neither reviewer checked that direction.

The fourth assertion also went in **wrong first**: it reported `[[lesson-212]]` as dangling
because two wikilink conventions are in use and it only knew one. Running against the real tree
caught it; a fixture would not have. That is a guard calling correct state broken — the same
defect as calling broken state correct, and it is in the file's comments so the next reader
does not re-derive it.

PR #1521, opened 2026-09-05
docs(lessons): 273 — the fix and the detector can be the same edit

mlorentedev:
Docs only. The session lesson from #1515 / #1518.

## The finding

`test (windows-latest)` failed on #1515 for **two reasons at once**: `t.TempDir()`'s Windows 8.3 short form (`RUNNER~1`) vs git's long form, and a separator convention — `ResolveMainRepoRoot` normalised through `filepath.Dir/Abs` while `ResolveWorktreeRoot` returned git's raw stdout, and git emits `/` on Windows too.

The obvious repair is a helper that resolves both spellings before comparing. It is correct for the 8.3 problem and **it makes the separator bug invisible**, because `filepath.Clean` on Windows rewrites `/` into `\`. With it in place, the fixed code and the unfixed code both pass.

**The cure would have removed the only thing that could detect the disease, in the same commit.** CodeRabbit independently suggested `os.SameFile`, which has the identical property — any identity comparison is blind to how the path was spelled. The separator had to be asserted separately, against the raw output, never through the comparison.

## Why it is worth a lesson rather than a comment

Five more instances landed in the same session, four of them mine, and **not one was found by reading**:

| What | What caught it |
|---|---|
| `render_region` as a pipe swallowed a `return 1` | three unrelated coverage tests |
| Two of three `! grep -q` asserted nothing — bash exempts `!` from `set -e` | this repo's `guard-bats-negation.bats` (#1034) |
| A broken `//go:build !linux` file; `go test ./...` on Linux never compiles it | `GOOS=darwin go vet` |
| A guard reporting correct state as broken — two wikilink conventions exist, it knew one | running against the real tree |
| A mitigation counter non-zero on **every** Linux machine, so it signals nothing | the round-2 adversarial reviewer |

The last was written as the answer to a reviewer's finding. **A permanently-lit warning light is a warning light that has been removed.**

## Verification

```
bats tests/guard-lesson-numbers-unique.bats   1..4, all ok
```

That guard landed in #1519 earlier today and validates this lesson: unique number, present in `_index.md`, and its `[[lesson-272]]` wikilink resolves.


mlorentedev:
## Review triage

**Who reviewed:** `pr-agent` (github-actions). **CodeRabbit did not** — no review object, no inline comments. Recorded, because an unreviewed PR and a reviewed one look identical on a green board.

| Finding | Disposition |
|---|---|
| **Incomplete reference** — `guard-bats-negation.bats` given as a bare basename, not locatable without guessing the directory | **Applied.** Now `tests/guard-bats-negation.bats` |
| Ticket compliance ×3: *"All implementation requirements are not addressed in this docs-only PR"* | **Declined.** It inferred three tickets from PR references and then reported a docs-only PR for not implementing them. The recurring ticket-scope error; this PR implements nothing by design |
| No security concerns | **Noted** |

### Verification after the fix

```
bats tests/guard-lesson-numbers-unique.bats   1..4, all ok
```

That guard landed in #1519 earlier today and validates this lesson against the real tree: unique number, present in `_index.md`, both `[[lesson-267]]`/`[[lesson-272]]` wikilinks resolve.

Issue #1523, opened 2026-09-05
BUG-094: Gate f compares paths as strings, so a process in another mount namespace reads as 'outside' and its worktree is deleted


mlorentedev:
`dotf worktree sweep`'s Gate f decides whether a live process is sitting in a worktree by
comparing `/proc/<pid>/cwd` against the worktree path **as strings**. `/proc/<pid>/cwd` is
resolved by the kernel in the *process's own mount namespace*, so a process that reaches the
worktree through a bind mount reports a path that does not exist on the host. The strings
differ, Gate f answers "nobody is inside", and the caller deletes the worktree.

BUG-093 closed the symlink instance of this class (`filepath.EvalSymlinks` on the target). This
is the same class arriving through a different door, and resolution does not close it.

## Reproduction (measured 2026-09-05)

```
docker run -d -u 1000:1000 -v /tmp/wtmount-repro/repo-wt-feat:/wt-mount-target \
           -w /wt-mount-target busybox sleep 60

host readlink /proc/2658284/cwd  -> /wt-mount-target
worktree path on host            -> /tmp/wtmount-repro/repo-wt-feat
dev/ino of /proc/<pid>/cwd       -> 50/992409
dev/ino of the worktree path     -> 50/992409     <- the same directory
string comparison                -> False         <- Gate f says "outside"
```

A second shape falls out of the same weakness: a worktree directory deleted and then recreated
makes `readlink` yield `".../repo-wt-feat (deleted)"`, which matches neither the equality nor the
prefix test. That one is not a data-loss path (the process holds the old inode), but it is the
same defect.

## Fix

Identity by `dev`+`ino` (`os.SameFile`) instead of by string, walking parents from
`/proc/<pid>/cwd/..` for the subdirectory case. `os.Stat("/proc/<pid>/cwd")` follows the magic
link and stats the real directory, so the comparison works across namespaces — the measurement
above shows it seeing what the string comparison does not.

Note the permission bound is unchanged: `os.Stat` on `/proc/<pid>/cwd` is gated by
`ptrace_may_access` exactly as `os.Readlink` is, so this widens *correctness*, not *reach*.

## Why it was declared rather than fixed in BUG-093

The code is ~20 lines; the **test** is the problem. Reproducing it needs a mount namespace.
`unshare -Urm` is refused on the development machine
(`kernel.apparmor_restrict_unprivileged_userns = 1`, Ubuntu's default since 24.04), and whether a
[...]

Issue #1470, opened 2026-09-04
GUARD-010: features.json state is a field nothing writes, validates or reads — seven spellings in the archive, and a reviewer just mistook it for the archive gate


mlorentedev:
Every `specs/*/features.json` carries a `state` field per feature. Nothing
writes it, nothing validates it, nothing reads it. It has accumulated seven
different spellings across the archive, and an adversarial reviewer just
mistook it for an archive gate and recommended an action that would have
invalidated its own review.

## Measured, 2026-09-03

Across every archived spec's `features.json`:

| `"state"` value | count |
|---|---|
| `pending` | 506 |
| `verified` | 42 |
| `passing` | 32 |
| `verifying` | 15 |
| `done` | 15 |
| `passed` | 12 |
| `implemented` | 6 |
| `pass` | 0 |

```
git grep -h '"state":' -- 'specs/archive/**/features.json' | tr -d ' ' | sort | uniq -c | sort -rn
```

Seven spellings, six of them synonyms for the same idea, and the modal value is
the scaffold default — meaning the field is left untouched in the overwhelming
majority of specs and hand-edited to an invented value in the rest.

`cli/internal/spec/archive.go` is 309 lines and contains no reference to
`features` or `state`. Nothing else in `cli/internal/spec/` reads it either.

## Why this is worth fixing rather than ignoring

It cost a real review cycle. On CLI-072 the reviewer (`nan/mimo-v2.5`) filed:

> `features.json` has all 11 entries at `state: "pending"` with empty
> `evidence`. [...] If `dotf spec archive` reads this state, it may reject the
> archive.

and recommended setting `state: "pass"` before archiving. Both halves are wrong
in an instructive way:

- `pass` is a spelling that has **never** been used in this repo.
- Acting on it would have invalidated the review, because `features.json` is a
  contract file (`cli/internal/spec/review.go:24`) and the staleness gate
  compares contract files against `reviewed_sha` over committed history *and*
  `git status --porcelain`.

The reviewer was being careful and reading the artifacts as written. A field
that looks like a state machine and is not one will keep producing this, and it
costs a reviewer round each time. It also erodes the artifact: an operator
reading `state: "pending"` on an archived, shipped, merged spec learns nothing
true.

## Options

Pick one; both are better than the status quo.

1. **Delete the field.** Cheapest honest fix. It records nothing that the
   spec's archived status and its `verification.md` do not already record
   better, and deletion cannot drift. Remove it from
[...]

PR #1526, opened 2026-09-05
fix(cli): make Gate f's claims falsifiable, and stop calling a constant a signal

mlorentedev:
Round 3 of the BUG-093 adversarial review returned **FAIL** on 6 REAL Majors and 4 REAL Minors.
Every one was real; none is argued down. This lands the dispositions.

The review did not find a broken behaviour. It found **claims no test could falsify** — including
the mitigation round 2 added as its answer to round 1.

Deliberately **no `Closes #1516`**: the spec is not archivable until a review passes, and the
archive PR is what closes the issue.

## The design blocker: a warning light wired to always-on

`UninspectableProcesses` was reported as a partial-scan warning. It is non-zero on **every** Linux
machine by construction — reading `/proc/<pid>/cwd` is gated by `ptrace_may_access`, so an
ordinary user can never read root's processes, another user's, or a same-uid process that set
`PR_SET_DUMPABLE=0`. Measured on this desktop: **364 of 571 processes out of reach, 20 of them the
caller's own**.

The obvious rescue — count only same-uid failures, so a healthy machine reads zero — was
**measured and rejected before designing on it**: 20 of 227 same-uid processes are EACCES
(`systemd --user`, `(sd-pam)`, `ssh-agent`, `chrome-sandbox`, `brave`). A uid-filtered counter is
still never zero.

So the count stays — deleting it would re-collapse the three-valued read round 1 required — but it
now states Gate f's **reach**, and nothing decides on it. The `> 0` gate on the message is kept
deliberately, with the reason in the code: `gateF` only scans worktrees that reached
`StateReapable`, so `0` also means "nothing was considered".

## Four properties that now fail when broken

| Was | Now |
|---|---|
| The sibling-prefix test's "sibling" was the target's own path, and its failure branch was a `t.Skip` | two tests, isolated roots; the naive-prefix mutation fails |
| Nothing drove `Uninspectable++` — it could pin at 0 forever | a real-`/proc` test, anchored on `/proc/1/cwd` being unreadable so it cannot go vacuous as root |
| Nothing drove the under-lock re-check in `reapSingleWorktree` | a seam answering `false` then `true`, with a call-count anchor |
| `processDiscoverySupported`'s Linux side was pinned by a test that skipped when the constant was false, then asserted a package-level `var` was non-nil | a direct assertion; flipping the constant now fails |

**Why the re-check survived two reviews with five tests pointing at it:** a seam with a *constant*
answer can never reach it. A constant `true` is refused by the first Gate f call; a constant
`false` never exercises the second. The seam has to answer `false` then `true` — which is exactly
the TOCTOU the re-check exists for. The shape of the double is part of the property, not plumbing.

`isCandidateForReap` is **deleted**: no production caller used it, so two tests were pinning a
refusal on a function that could not prevent a deletion.

## Declared, not fixed: the mount-namespace fail-open (#1523)

`/proc/<pid>/cwd` resolves in the process's *own* namespace, so a bind-mounted worktree compares
unequal and reads as outside. Reproduced independently rather than cited:

```
docker run -d -u 1000:1000 -v /tmp/wtmount-repro/repo-wt-feat:/wt-mount-target \
           -w /wt-mount-target busybox sleep 60
host readlink /proc/2658284/cwd -> /wt-mount-target
worktree path on host           -> /tmp/wtmount-repro/repo-wt-feat
dev/ino of both                 -> 50/992409   <- the same directory
string comparison               -> False       <- Gate f says "outside"
```

The fix is `os.SameFile` and it is ~20 lines. The **test** is the problem: it needs a mount
namespace, `unshare -Urm` is refused here (`kernel.apparmor_restrict_unprivileged_userns = 1`,
Ubuntu's default since 24.04), and whether `ubuntu-latest` permits one was not measurable from
this machine. A test that skips on the runner reports green while asserting nothing — the exact
shape of the vacuous test round 3 found in this same file. Filed as **#1523**, whose first task is
that measurement rather than a patch.

## Mutation results

The harness compares the file checksum after patching and reports **ANCHOR-MISS** rather than a
result if nothing changed, so a no-op cannot be read as "the tests caught it" (lesson 267).

```
=== mutations the round-3 review found SURVIVING ===
  CAUGHT   finding 3: prefix test loses its separator anchor
  CAUGHT   finding 4: Uninspectable++ becomes a no-op
  CAUGHT   finding 5: reapSingleWorktree drops the Gate f re-check
  CAUGHT   finding 6: Linux claims it has no process discovery
  CAUGHT   finding 7: unresolvable target fails OPEN
=== regressions from earlier rounds ===
  CAUGHT   round 1 Blocker: EvalSymlinks resolution removed
  CAUGHT   round 1: caller ignores Gate f entirely
=== declared-uncovered (expected to SURVIVE) ===
  SURVIVED AC2: filepath.Abs error fails open (unreachable by input)
```

## The Windows leg, actually executed

AC7 named `GOOS=windows go vet`, which **compiles** the non-Linux file and never runs it — which
is how round 2 shipped a red `test (windows-latest)` past a fully green local loop. AC7 now names
the Windows **test** leg, and a build-tag-flip proxy runs that implementation on Linux:
`rc=0`, the Windows leg would be green.

## Corrected, not quietly repaired

Three claims in `verification.md` were false at the tip: the pre-existing tests were **not**
untouched (`dab7b6e` edited five), the named subset is **7** PASS and not 3, and GUARD-010
**#1470 is already OPEN** for the `features.json` vocabulary.

## Verification

`go build ./...` · `go test ./...` · `go test -race ./internal/worktree/` ·
`GOOS=windows go vet ./...` · `GOOS=darwin go vet ./...` · `golangci-lint run` (2.12.2, the
`versions.conf` pin) — all `rc=0`, **0 issues**. All 8 `features.json` verifications re-executed,
all `rc=0`.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -20,11 +20,17 @@ All shell scripts MUST work in **both bash and zsh**. Before modifying any `.sh`
 | `set -- $var` / unquoted `$var` to split into fields | read line by line, or `${=var}` in zsh | **Fails silently.** zsh does not word-split unquoted parameters: you get one field containing everything, not N |
 | `. file` (no slash) to source from the cwd | `. ./file` | **Fails silently.** A slashless argument to `.` is searched on `$PATH` only; bash also falls back to the cwd, zsh does not. In a `$(...)` the result is an empty string, not an error |
 | `${PIPESTATUS[0]}` to check a piped command | drop the pipe and use `$?`, or `set -o pipefail` when the pipe is needed | **Fails silently.** zsh spells it `$pipestatus[1]` and arrays are 1-indexed, so the bash form expands to **nothing** — `EXIT=` rather than an error. A verification that reads a pipeline's exit status is exactly where this bites: `cmd \| tail` reports `tail`'s status, so a failing `cmd` passes |
+| `path=`, `status=`, `cdpath=`, `manpath=`, `fignore=` as ordinary variable names | any other name — `target`, `rc`, … | **Fails silently, and destructively.** In zsh these are *tied* to their uppercase counterparts: `path` IS `$PATH` as an array. Assigning a string to `path` wipes the command search path, so every later `cp`/`mv`/`grep` becomes "command not found" — and a script that interprets a failed command as "nothing matched" will report a clean result. Measured 2026-09-05: a mutation harness reported `ANCHOR-MISS` ("the pattern was not found") for all 8 mutations when the real cause was that `md5sum` no longer existed. `zsh -n` passes clean; only running it finds this |
 
-> The last four rows fail **silently**: they return an empty or single-element result instead of an
+> The last five rows fail **silently**: they return an empty or single-element result instead of an
 > error, and empty reads as a finding. Every row above them breaks loudly. Before believing an
 > empty result from a shell sweep, re-run it in the other shell — see `docs/lessons.md`,
 > *"a shell incompatibility that answers wrongly beats one that fails"*.
+>
+> **A script whose failure path reports "found nothing" needs a preflight.** The last row's damage
+> was not the wiped `PATH`, it was that the wipe arrived dressed as a result. Any script that
+> reports absence — no match, no diff, no mutation applied — should assert its tools exist before
+> it starts, so a broken environment exits loudly instead of answering the question wrongly.
 
 ## Verification Commands
 
--- a/cli/internal/cmd/worktree.go
+++ b/cli/internal/cmd/worktree.go
@@ -237,8 +237,9 @@ clean git status, confirmed merged PR, and minimum age.`,
 				// gateF only scans worktrees that reached StateReapable, so a
 				// run with nothing to consider also produces 0, and printing a
 				// reach for a scan that never happened would be its own lie.
-				c.Printf("Gate f reach: %d process(es) belong to root or another user and cannot be inspected;\n"+
-					"              one of theirs sitting in a worktree is invisible to this scan.\n",
+				c.Printf("Gate f reach: %d process(es) could not be inspected (other users, root, or\n"+
+					"              same-user processes that opt out); one of those sitting in a\n"+
+					"              worktree is invisible to this scan.\n",
 					report.UninspectableProcesses)
 			}
 
--- a/cli/internal/cmd/worktree.go
+++ b/cli/internal/cmd/worktree.go
@@ -225,10 +225,20 @@ clean git status, confirmed merged PR, and minimum age.`,
 				c.Printf("note: no process-liveness check on this platform, so nothing can be reaped.\n")
 				c.Printf("      Gate f refuses rather than guesses; remove one deliberately with `dotf worktree done`.\n")
 			} else if report.UninspectableProcesses > 0 {
-				// A partial scan, said out loud. These are almost always other
-				// users' processes, whose cwd only they and root may read, so
-				// the scan cannot rule out that one of them is inside.
-				c.Printf("note: %d process(es) could not be inspected, so this scan was partial.\n",
+				// Gate f's reach, not a warning. This number is non-zero on
+				// every Linux machine — reading /proc/<pid>/cwd is gated by
+				// ptrace_may_access, and a desktop runs hundreds of processes
+				// this user may not read (measured: 364 of 571, 20 of them the
+				// caller's own). It said "this scan was partial" before, which
+				// is a warning light wired to always-on, and a reader learns to
+				// skip those.
+				//
+				// Still gated on > 0, because 0 is ambiguous rather than good:
+				// gateF only scans worktrees that reached StateReapable, so a
+				// run with nothing to consider also produces 0, and printing a
+				// reach for a scan that never happened would be its own lie.
+				c.Printf("Gate f reach: %d process(es) belong to root or another user and cannot be inspected;\n"+
+					"              one of theirs sitting in a worktree is invisible to this scan.\n",
 					report.UninspectableProcesses)
 			}
 
--- a/cli/internal/worktree/sweep.go
+++ b/cli/internal/worktree/sweep.go
@@ -92,26 +92,42 @@ type SweepReport struct {
 	// and those are different facts.
 	ProcessDiscovery bool `json:"process_discovery"`
 	// UninspectableProcesses is the largest number of processes any single
-	// worktree's scan could not read the cwd of — almost always another user's,
-	// since /proc/<pid>/cwd is readable only by its owner and root. Non-zero
-	// means every reap decision in this run rested on a partial scan. Reported
-	// rather than acted on: refusing on it would make sweep inert on Linux,
-	// because /proc/1/cwd is unreadable to every non-root caller.
+	// worktree's scan could not read the cwd of.
+	//
+	// It states Gate f's REACH; it is not an anomaly signal, and an earlier
+	// version of this comment claimed it was. The number is non-zero on every
+	// Linux machine by construction: /proc/<pid>/cwd is gated by
+	// ptrace_may_access, so a scan run as an ordinary user can never read
+	// root's processes, another user's, or same-uid ones that set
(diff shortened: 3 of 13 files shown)


mlorentedev:
## Review triage

**Who actually reviewed:** PR-Agent. **CodeRabbit did not review this PR** — it posted
*"Review limit reached"*, which is a notice about the review, not a review. Disclosed rather than
counted, per the pr-stewardship rule: a notice that no review ran leaves that reviewer's pass
unmade.

Three code findings, all triaged. Applied in `86f4a7c`.

| # | Finding | Disposition |
|---|---|---|
| 1 | **Error Message Change** — "belong to root or another user" may be inaccurate; some uninspectable processes are same-uid with `PR_SET_DUMPABLE=0` | **Applied.** Correct, and contradicted by my own measurement in this very spec: 20 of 227 same-uid processes are EACCES here. The message named two classes where the data says three. Now: "other users, root, or same-user processes that opt out" |
| 2 | **Test Fragility** — `TestUninspectableProcessesAreCountedByTheRealScan` skips rather than fails when `/proc/1/cwd` is readable, so it may not run in some CI environments | **Applied, in the direction the finding points rather than the one it proposes.** The skip itself is right — as root the counter genuinely has nothing to count, so failing there would be a false alarm. What was wrong is that it was **invisible**: I went to the `test (ubuntu-latest)` log to prove the test had executed and could not, because `go test` without `-v` prints neither skips nor passes. A skip and a pass are the same colour in CI. The skip now announces `NOT EXECUTED`, names the condition and the euid. Verified locally that it does **not** skip here: `--- PASS`, not `--- SKIP` |
| 3 | **Verification Command Complexity** — the `grep -c '^--- PASS'` counts in `features.json` are fragile to test renames | **Declined, with the reason.** The fragility is real and it is the point: these fail **closed**. A renamed test makes `-run` match nothing, `go test` prints no `--- PASS` line, `grep -c` outputs `0` and exits **1**, so the verification fails rather than silently passing over a test that no longer exists. It has already earned its keep in this round — the count caught a stale claim of mine (f7 said 9, the command says 10) before the reviewer could. A rename that breaks it is a rename that should break it |

### Ticket-compliance analysis: read with care

PR-Agent measured this PR against **#1523** and **#1470** and reported both "partially compliant"
or "not compliant". Both are correct readings of the wrong question — neither is this PR's ticket:

- **#1523** was *filed by this PR* precisely to carry the mount-namespace fix, which is declared
  out of scope here with the reproduction and the reason (no CI-runnable test exists yet; its
  first task is that measurement). Reporting the PR as non-compliant with a ticket it opened for
  the work it deliberately did not do is the analysis working as designed and the conclusion not
  following.
- **#1470** (GUARD-010) is the `features.json` state-vocabulary guard, owned separately. This PR
  only corrected a stale claim that it had not been filed.

Its **#1516** reading is the relevant one, and its standalone pass flags
*"no evidence the Windows leg is tested via `sweep_proc_other_test.go` (not in diff)"*. Correct
that the file is not in the diff — it is unchanged and already on `main`. The evidence it asks for
is in the checks: **`test (windows-latest)` is green**, and `verification.md` records the
build-tag-flip proxy that executes that implementation on Linux, which is the gap `GOOS=windows
go vet` left open when round 2 shipped a red Windows leg.

### CI

All required checks green, including `test (windows-latest)` — the leg round 2 broke.

mlorentedev:
## Review triage

Addendum — PR-Agent re-reviewed after `86f4a7c` (the fixes for its own findings 1 and 2).

**Result: "No major issues detected."** No new code findings. The three from the previous pass
are dispositioned in the triage comment above and are not repeated here.

What it still reports as "non-compliant" is the scope of this PR, deliberately, and each is
stated rather than left implied:

| Reported | Why it is intended |
|---|---|
| "The PR does not close #1516; it is waiting for an adversarial review pass before archiving" | **Correct, and required.** A change that closes a spec gets an independent adversarial review before it archives, and `dotf spec archive` refuses without a passing one. The archive PR carries `Closes #1516`; this one must not |
| #1523: "no fix for dev+ino comparison is implemented" | #1523 was filed **by this PR** to carry that work. It is declared in *Out of scope* with the reproduction and the reason: no CI-runnable test exists yet, and its first task is that measurement |
| #1470: "no changes to the state field" | GUARD-010 is owned separately. This PR only corrected a stale claim that it had not been filed |

Its "requires further human verification" item — *"human review is required to pass the spec
archive gate"* — is exactly right, and is the next step: `dotf spec review BUG-093-gate-f-fail-closed`
against `main` once this merges, then an archive PR.

**Reviewer disclosure, unchanged:** PR-Agent reviewed. CodeRabbit posted *"Review limit reached"* —
a notice, not a review — so its pass was never made on this PR.

mlorentedev:
## Review triage

Third pass, after the mutation harness was committed. Both findings **REAL**; both applied in `6b4c19e`.

| # | Finding | Disposition |
|---|---|---|
| 1 | **Shell Compatibility** — `mutate.sh` may use bash-specific features and would fail under zsh, violating the mandatory both-shells rule | **Applied, and the finding understated it.** Measured rather than argued: under zsh the script did not merely "might fail", it reported **`ANCHOR-MISS` for all 8 mutations** — a clean-looking, completely wrong result. Root cause is not the shebang: the script used **`path`** as a local variable, and in zsh `path` is *tied* to `$PATH` as an array, so assigning a string to it wiped the command search path and every later `cp`/`md5sum`/`mv` became "command not found". Nothing was corrupted only because `cp` failed before any `.bak` was written — luck, not design. `local` does not help, and `zsh -n` passes clean. Renamed, plus a **preflight** that asserts its tools exist so a broken environment exits 2 loudly. Verified identical under both shells: 7 CAUGHT, 1 declared survivor, exit 0 |
| 2 | **Known Security Gap** — the mount-namespace fail-open is documented but not fixed; a REAL concern affecting data integrity | **Acknowledged; already dispositioned, and the severity assessment is endorsed.** It is right that this is a data-integrity gap, which is why it was reproduced with `docker` rather than accepted on the reviewer's word (`dev/ino` identical at `50/992409` while the string comparison says `False`), declared in *Out of scope*, and filed as **#1523** with the reproduction. The reason it is not fixed here is stated and is this spec's own lesson: the fix is ~20 lines, the **test** needs a mount namespace `unshare -Urm` will not grant under Ubuntu's apparmor default, and a test that skips on the runner reports green while asserting nothing. #1523's first task is that measurement, not a patch. Shipping the fix untested here would repeat the defect the spec exists to remove |

### What finding 1 turned up beyond this PR

Grepping the repo for the same class found **four more sites** — `scripts/check-spec-gate.sh:524`,
`scripts/compile-harness.sh:931`, `scripts/bitacora-rollout.sh:84,123`,
`tests/gitattributes-eol.bats:42,97`, and `tests/lib/refute.bash:54` (`local status=0`, which is a
**hard error** under zsh: `read-only variable`).

None is firing today, and that is the interesting part: two survive only because their function
bodies happen to contain no external command. Adding a single `grep` to `_excluded()` — in the spec
gate, a required check — would turn it into "command not found" for every call, in a script that
treats a failed command as "no match".

Measured and filed as **#1530** with the reproduction, plus a row in `.claude/CLAUDE.md`'s
prohibited-pattern table carrying the generalisation: **a script that reports absence needs a
preflight**, because for that shape a broken environment and a negative answer are the same output.
Not fixed here — it is a separate defect class in five files, none of them this PR's subject.

**Reviewer disclosure, unchanged:** PR-Agent reviewed all three passes. CodeRabbit posted
*"Review limit reached"* — a notice, not a review — so its pass was never made on this PR.

mlorentedev:
## Review triage

Final pass. PR-Agent re-reviewed after `6b4c19e` and reports **"No security concerns identified"** and **"No major issues detected"** — no new findings. Every finding from the three earlier passes is dispositioned in the comments above.

Queue closed. All checks green, including `test (windows-latest)`. `mergeStateStatus: CLEAN`.

**Reviewer disclosure:** PR-Agent reviewed, four passes. CodeRabbit never did — it posted *"Review limit reached"*, a notice about a review rather than a review, so this PR carries one reviewer and not two. Stated rather than left to be inferred from a green `review-attestation`.

PR #1524, opened 2026-09-05
fix(harness): shipper's block precondition named a closed issue and the wrong risk

mlorentedev:
## Summary

- **shipper's record named a closed issue as a live blocker.** It said a named dispatch sends its name as `agent_type`, "so its role never resolves and the gate silently turns off", and told the reader that stood "until that is fixed". #1434 closed 2026-09-04 via #1471 — ~19h *before* #1504 merged the sentence. The residual is narrower than the claim: a **witnessed** named dispatch is mapped back to its true type, and only an **unwitnessed** one still falls through to the roster.
- **Replaced with the precondition that does hold**, matching the correction #1509 already applied to architect: the gate reads the **deploy dir**, so a merged migration is not an enforced one — and the record written while it is inert is indistinguishable from the record written when everything passed (#1510).
- **Fixed at the source, not in the generated file.** `harness/agents/shipper/AGENT.md` carries `generated_sha`; the edit is in the vault (`knowledge` `eceee5da`) and this is the `--refresh` output. Editing the record directly is the failure mode that produced the side finding below.

### Evidence the correction is right, measured this session

The release deployed at `21:23:50Z`, and three genuine dispatches then landed in the gate ledger through the deployed `PreToolUse` hook — the one acceptance criterion no PR in this epic could assert:

```
21:32:21Z  agent_type=architect  agent_id=a51b7c05…  resolved=architect  outcome=warn
           warned=[architecture-session, pattern-loader, read-all-adrs]
21:33:45Z  agent_type=architect  agent_id=a918411d…  resolved=architect  outcome=warn
           warned=[architecture-session, pattern-loader, read-all-adrs]
21:33:47Z  agent_type=shipper    agent_id=a972839b…  resolved=shipper    outcome=warn
           warned=[docker, pr-review-triage, using-git-worktrees]
```

That is architect's "two dispatches of one role carrying different `agent_id`s" satisfied, plus shipper's first. A *named* dispatch resolving correctly was already evidenced by the 116 `adversarial-pr14 → reviewer` records the dispatch map produced — which is precisely why the removed sentence was false.

Driving v0.54.0's gate against each deployed record separately shows the contrast the epic is about — same binary, same deploy dir, same payload shape:

| persona | form | gate output |
|---|---|---|
| architect | mapping | 3 warns |
| shipper | mapping | 3 warns |
| builder | mapping (2 of 9) | 2 warns |
| curator | flat list | **silent** |
| planner | flat list | **silent** |

## SDD checklist

[...]

PR #1527, opened 2026-09-05
feat(harness): curator becomes the fifth enforceable persona, gating two of eight

mlorentedev:
## Summary

- **curator declared all eight skills in the flat list**, which the loader reads as `EnforceUnset`, so the gate resolved the persona and then acted on nothing. Its prose said they were *"enforced by hook, not left to memory"* — false for all eight, and false since the record was written.
- **`crystallize` and `handoff` now carry `enforce: warn`.** They are the two that hold on every invocation: this persona's trigger is *"after a substantial session"*, and those two **are** that phase — the promotion of session output into durable knowledge, and the checklist that makes a session's end a record rather than an ending.
- **The other six stay bare strings, and their own descriptions say why**: each names a trigger condition. `insights` is weekly maintenance or a pre-sprint check; `vault-doctor` runs when the vault reports structural failures; `genre-picker` applies when creating an artifact; `context-refresh` follows an ADR or a phase close; `place-knowledge` onboards a repository once; `dispose-proposals` is the weekly human gate. Gating all six would name five inapplicable obligations on every call and teach the reader to scroll past `[gate]` lines.

Fixed at the source (`knowledge` `331f2e91`), not in the generated record — `harness/agents/curator/AGENT.md` carries a `generated_sha` and a direct edit would be reverted by the next `--refresh`.

### Two of eight is the smallest gated fraction so far, and the prose says why

| persona | gated | of | rationale |
|---|---|---|---|
| architect | 3 | 3 | no skill in the list is situational |
| shipper | 3 | 5 | helm/terraform deliver nothing in this repo |
| curator | **2** | **8** | six of eight are condition-triggered by their own definition |
| builder | 2 | 9 | list mixes always-applicable with situational |

The record states this is a judgement about *these skills* rather than a weaker standard for this persona — otherwise the ratio reads as a target someone will later try to hit.

### Measured

- **Presence payload cost: 0 characters.** The declared id set is byte-identical before and after; severity is not rendered into the presence block. Same result #1504 measured for shipper.
- **Gate behaviour before this change**, driving v0.54.0 against the deployed record: curator was **silent** — `outcome: allow`, nothing warned. architect/shipper/builder each warned on their declared skills from the same binary and the same deploy dir. That contrast is the reason this PR exists.

## SDD checklist

- [x] Issue exists on the bitácora GitHub Project — `Refs #1497`
- [x] Diff conforms to ~300 LOC atomic cap — 29 insertions, 3 deletions, one generated record
- [ ] Spec folder `specs/<feature-id>/` is included in this PR (or `skip-sdd` label below)
- [ ] `proposal.md` has filled Why / What / Acceptance criteria
- [ ] `tasks.md` is in TDD order
- [ ] `verification.md` will be filled before merge (evidence + commit hashes)

## SDD skip rationale

One record migrated from the flat `skills:` list to the mapping form already specced, reviewed and shipped four times (#1412 reviewer, #1494 builder, #1504 shipper, #1509 architect). No logic, no contract change, no new dependency — the loader, the gate and the join all already handle both forms. Under the 50-LOC production threshold, which is why this is one persona per PR rather than curator and planner together.

## Test plan

- [x] `cd cli && go build ./... && go vet ./... && go test ./internal/harness/ ./internal/cmd/` — ok
- [x] `./scripts/compile-harness.sh --check` — `rc=0`, no harness drift
- [x] `--refresh` reproduces this record exactly from the vault source
- [x] Manual verification: declared id set compared before/after — identical, so the presence payload is unchanged

## Note for the reviewer

`--refresh` in this worktree also regenerates `harness/agents/shipper/AGENT.md`, because that record's vault source was corrected in `eceee5da`. That change belongs to **#1524** and was deliberately reverted out of this branch — this PR touches one file.

### DIFF
--- a/cli/internal/harness/persona_test.go
+++ b/cli/internal/harness/persona_test.go
@@ -249,7 +249,13 @@ func TestMigratedPersonasStillGateSomething(t *testing.T) {
 		byName[p.Name] = p
 	}
 
-	for _, name := range []string{"reviewer", "builder", "shipper", "architect"} {
+	for _, name := range []string{
+		"reviewer",
+		"builder",
+		"shipper",
+		"architect",
+		"curator",
+	} {
 		p, ok := byName[name]
 		if !ok {
 			t.Errorf("%s is missing from the roster", name)
--- a/harness/agents/curator/AGENT.md
+++ b/harness/agents/curator/AGENT.md
@@ -1,7 +1,7 @@
 ---
 generated: true
 generated_from: 00_meta/agents/definitions/curator/AGENT.md
-generated_sha: 283045d89bf80de8
+generated_sha: 89cb42c92f5098a8
 id: agent-curator
 type: agent
 status: active
@@ -11,7 +11,17 @@ description: Crystallize-phase persona. Invoke after a substantial session, or w
 kind: invocable
 model: top
 capabilities: [read, search, edit, shell, skill]
-skills: [vault-doctor, crystallize, insights, genre-picker, context-refresh, handoff, place-knowledge, dispose-proposals]
+skills:
+  - id: crystallize
+    enforce: warn
+  - id: handoff
+    enforce: warn
+  - vault-doctor
+  - insights
+  - genre-picker
+  - context-refresh
+  - place-knowledge
+  - dispose-proposals
 owner: manu
 ---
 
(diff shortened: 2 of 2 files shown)


mlorentedev:
## Review triage

Head `73c12d5`. Checks read as latest-run-per-name, because REST returns superseded runs for the same SHA and naive counting misreads them.

**Check status at time of writing: 13 green, `test-windows` still `in_progress`.** Not green-by-assumption — do not merge until it lands. Everything else, `integration` and `test (windows-latest)` included, has passed on this head.

### Reviewers who actually produced a review

| reviewer | output | disposition |
|---|---|---|
| pr-agent | effort 1/5, no security concerns, no major issues | **applied** (via the sibling finding), below |
| CodeRabbit | **no review ran** — "Review limit reached" | disclosed, not a review |

**CodeRabbit did not review this PR.** Its comment is the rate-limit notice — *"You've used all free OSS reviews for now"* — not findings. Recorded because a notice that no review ran leaves the PR unreviewed, and saying nothing would be a claim of review that never happened.

### Applied: the migration had no regression test

pr-agent raised this on the sibling PR (#1528) and it was **correct here too**, so it is fixed in both. `TestMigratedPersonasStillGateSomething` asserts the migrated set **by name**, precisely because the sibling test only *logs* the unmigrated count — so a record reverting to the flat form is invisible to it. Its own comment says each migration costs one line there. This PR added a migration and did not add the line, which means a stale vault half could have rewritten curator back to the flat form with nothing turning red.

Fixed in `73c12d5`, and **verified by mutation rather than by the suite going green**:

```
--- FAIL: TestMigratedPersonasStillGateSomething (0.00s)
    persona_test.go:271: curator has been migrated but now gates nothing — reverted to the flat form?
```

That is the test failing on a deliberately reverted record, then passing once restored. A guard never observed failing is not a guard.

The list is also reformatted to one name per line. The single-line form makes two concurrent persona PRs collide on the same line for no reason — which is exactly the situation #1527 and #1528 are in.

### Ignored: the ticket-compliance section

pr-agent inferred the wrong ticket again, treating **#1504 — a pull request** — as the issue to satisfy, and deriving "non-compliant requirements" from it (including a requirement to file #1502, which this PR never mentions). Known reviewer-configuration defect, now six consecutive occurrences. Its *code* observations are read on their merits and one of them is applied above; only the ticket inference is discarded.

### Prior red, classified rather than re-run away

`integration` failed once on the earlier head `50ba18b`, at *POLISH-005: second setup-linux.sh run exits 0 cleanly with zero config diff*. Before re-running I checked it against a mechanism instead of assuming a flake:

- both sibling PRs (#1524, #1528) — the same class of change on the same base — **passed** the identical test;
- `main` is green at `d03033e`;
- locally, two independent harness deploys with this record produce **byte-identical** trees, and a second deploy over an existing one changes nothing.

Nothing connects the failure to a one-record prose change, and it did not reproduce on the new head. Recorded as a flake in the idempotence test, not as a fix — no ticket exists for it, and one occurrence is not yet a pattern worth filing.

### Note for whoever merges

`main` carries `required_status_checks.strict: true`, so this PR and #1528 will each need `update-branch` after the other lands, and they touch the same test list — the conflict is one line and resolving it means keeping **both** names.

mlorentedev:
## Review triage — addendum, head `7c28092`

`update-branch` onto `43d4b0c` (#1524, now merged) moved the head, which reopened the review window. Re-read rather than assumed:

- **CodeRabbit: still no review.** Its comment now reads *"Next included review available in 33 minutes"* — and note it **edited that comment in place**, `created=21:44:23Z` / `updated=22:25:05Z`. `created_at` never moved, so `dotf pr triage-queue` cannot see the change. Fifth measured occurrence of #1422. Content, not metadata, is what classifies it: still a notice, still not a review.
- **pr-agent re-reviewed** and now reports **#1504 fully compliant**, effort 1/5, *PR contains tests*, no security concerns. Its remaining "non-compliant" items are that planner and hermes-nan are not migrated in this PR — by design: planner is #1528, and hermes-nan is **excluded, not pending** (#1507, `kind: autonomous`, zero gate decisions in ~8200 records). No action.

All checks green on this head at time of writing. **Nothing outstanding.**

Issue #1507, opened 2026-09-05
HARNESS: doctor counts an autonomous persona the gate cannot see, so its Pass state is unreachable without a meaningless migration


mlorentedev:
## Why

`dotf doctor`'s persona-skill check counts **every** record under `agents.record_dir` and does not filter by `kind`. `hermes-nan` is `kind: autonomous`; the other six personas are `kind: invocable`. So `hermes-nan`'s one skill (`agent-lifecycle`) is counted as unmigrated, and the check's `Pass` branch —

```go
case unmigrated == 0:
    rep.Pass(fmt.Sprintf("every persona skill declares an enforcement severity (%d)", total))
```

— is unreachable unless `hermes-nan` is migrated too.

**Gating it would enforce nothing.** Three independent facts, measured 2026-09-05:

1. **`kind: autonomous`.** Its own record says it: *"Unlike the invocable personas, you are not summoned for a phase of somebody's work — you run on a schedule, reconcile what you own, and record what you found."*
2. **It runs externally, on NaN.** `dotf harness gate` is a `PreToolUse` hook in this harness. A process on another machine never reaches it.
3. **It has never produced a gate decision.** Across 32 ledger files and ~8,200 recorded decisions in `~/.local/state/dotfiles/gate/`, `agent_type: hermes-nan` appears **zero** times. Not rare — absent.

So the counter as it stands asks for a severity on a persona the gate cannot see, and it will keep warning until someone supplies one.

## The failure mode this creates

A metric that can only go green through a meaningless change is a metric that will eventually produce that change. Whoever wants a clean `dotf doctor` will migrate `hermes-nan`, and the result is a persona declaring `enforce: warn` on a skill no gate will ever evaluate — **presence dressed as enforcement**, which is the exact phrase the loader's no-default-severity rule exists to prevent. The design deliberately refuses to default a severity so that "gated" always means someone chose it; this counter reintroduces the pressure to choose one for a persona where the choice is void.

It is also a live misreport today: the warning names `hermes-nan (1: agent-lifecycle)` alongside genuinely unmigrated personas, so the number overstates the remaining work and the list mixes two different things.

## What

Scope the check to `kind: invocable`, and make the exclusion visible rather than silent.

- Personas whose `kind` is not `invocable` are excluded from `total` and `unmigrated`.
- The check still **reports** them, as an informational line rather than a warning — an exclusion nobody can see is how the next reader concludes the check is broken.
[...]

PR #1528, opened 2026-09-05
feat(harness): planner becomes the sixth enforceable persona, gating one of six

mlorentedev:
## Summary

- **planner declared all six skills in the flat list**, which the loader reads as `EnforceUnset`, so the gate resolved the persona and then acted on nothing. Its prose said they were *"enforced by hook, not left to memory"* — false for all six. **This is the last record carrying that sentence.**
- **`spec` now carries `enforce: warn`**, and it is gated for a reason no other skill in this epic has: the obligation is *already enforced downstream by a machine*. `spec-gate` fails any PR over the production-LOC threshold that does not touch `specs/<id>/`. So skipping `spec` is not a stylistic omission — it is the one a CI job rejects **after the work is written**, which is the most expensive moment to discover it. `spec init` also refuses to scaffold without an open issue (ADR-018), making it the only one of the six with a structural gate of its own.
- **The other five each need something to already exist** before they apply: a backlog item (`enrich-us`), a spec or gathered requirements (`writing-plans`), a written plan file (`executing-plans`), a PRD (`prd-to-issues`), or a single ticket to file (`new-ticket`). Gating them would name five usually-absent preconditions on every call.

Fixed at the source (`knowledge` `c9e756c1`), not in the generated record.

### An observation recorded, deliberately not acted on

`executing-plans` runs a plan rather than shaping one, which sits oddly beside this record's own Boundaries — *"you do not write the implementation"*. It is left declared and ungated rather than quietly re-homed to builder: **moving a skill between personas is a decision about the work cycle, not a detail of this migration.** The record now says so in prose, so the next reader finds the question rather than re-deriving it. Flagging for the owner — happy to file it if it is worth a ticket.

### This completes the epic

| persona | gated | of | issue |
|---|---|---|---|
| reviewer | 4 | 4 | #1412 |
| builder | 2 | 9 | #1494 |
| shipper | 3 | 5 | #1504 (+ #1524 prose) |
| architect | 3 | 3 | #1509 |
| curator | 2 | 8 | #1527 |
| **planner** | **1** | **6** | **this** |

`hermes-nan` is excluded rather than pending (#1507): `kind: autonomous`, zero gate decisions in ~8200 records.

### Measured

- **Presence payload cost: 0 characters.** The declared id set is unchanged; severity is not rendered into the presence block.
- **Gate behaviour before this change**, driving v0.54.0 against the deployed record: planner was **silent** — `outcome: allow`, nothing warned — while architect, shipper and builder each warned from the same binary and the same deploy dir.

## SDD checklist

- [x] Issue exists on the bitácora GitHub Project — `Refs #1497`
- [x] Diff conforms to ~300 LOC atomic cap — 26 insertions, 3 deletions, one generated record
- [ ] Spec folder `specs/<feature-id>/` is included in this PR (or `skip-sdd` label below)
- [ ] `proposal.md` has filled Why / What / Acceptance criteria
- [ ] `tasks.md` is in TDD order
- [ ] `verification.md` will be filled before merge (evidence + commit hashes)

## SDD skip rationale

One record migrated from the flat `skills:` list to the mapping form already specced, reviewed and shipped five times (#1412, #1494, #1504, #1509, #1527). No logic, no contract change, no new dependency — the loader, the gate and the join already handle both forms. Under the 50-LOC production threshold, which is why this is one persona per PR rather than curator and planner together.

## Test plan

- [x] `cd cli && go build ./... && go vet ./... && go test ./internal/harness/ ./internal/cmd/` — ok
- [x] `./scripts/compile-harness.sh --check` — `rc=0`, no harness drift
- [x] `--refresh` reproduces this record exactly from the vault source
- [x] Manual verification: declared id set compared before/after — identical, so the presence payload is unchanged

## Note for the reviewer

`--refresh` in this worktree also regenerates the `shipper` and `curator` records, whose vault sources were corrected in `eceee5da` and `331f2e91`. Those belong to **#1524** and **#1527** and were deliberately reverted out of this branch — this PR touches one file.

### DIFF
--- a/cli/internal/harness/persona_test.go
+++ b/cli/internal/harness/persona_test.go
@@ -249,7 +249,13 @@ func TestMigratedPersonasStillGateSomething(t *testing.T) {
 		byName[p.Name] = p
 	}
 
-	for _, name := range []string{"reviewer", "builder", "shipper", "architect"} {
+	for _, name := range []string{
+		"reviewer",
+		"builder",
+		"shipper",
+		"architect",
+		"planner",
+	} {
 		p, ok := byName[name]
 		if !ok {
 			t.Errorf("%s is missing from the roster", name)
--- a/harness/agents/planner/AGENT.md
+++ b/harness/agents/planner/AGENT.md
@@ -1,7 +1,7 @@
 ---
 generated: true
 generated_from: 00_meta/agents/definitions/planner/AGENT.md
-generated_sha: d47c36b771742723
+generated_sha: 8965f5db19a175a7
 id: agent-planner
 type: agent
 status: active
@@ -11,7 +11,14 @@ description: Plan-phase persona. Invoke once a decision exists and the work need
 kind: invocable
 model: mid
 capabilities: [read, search, edit, shell, skill]
-skills: [spec, enrich-us, writing-plans, executing-plans, prd-to-issues, new-ticket]
+skills:
+  - id: spec
+    enforce: warn
+  - enrich-us
+  - writing-plans
+  - executing-plans
+  - prd-to-issues
+  - new-ticket
 owner: manu
 ---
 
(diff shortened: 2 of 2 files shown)


mlorentedev:
## Review triage

Head `fb3e26a`. Checks read as latest-run-per-name, because REST returns superseded runs for the same SHA.

**Check status at time of writing: 13 green, `test-windows` still `in_progress`.** Not green-by-assumption — do not merge until it lands. Everything else, `integration` and `test (windows-latest)` included, has passed on this head.

### Reviewers who actually produced a review

| reviewer | output | disposition |
|---|---|---|
| pr-agent | 1 focus area, effort 2/5, no security concerns | **applied** |
| CodeRabbit | **no review ran** — "Review limit reached" | disclosed, not a review |

**CodeRabbit did not review this PR.** Its comment is the rate-limit notice — *"You've used all free OSS reviews for now"* — not findings. Recorded because a notice that no review ran leaves the PR unreviewed, and leaving it unrecorded is indistinguishable from nobody having looked.

### pr-agent — "No Go tests added" → **applied, and it was right**

> Every prior persona migration in this series added Go test assertions verifying the persona now gates at least one skill and stays migrated. This diff contains only the generated `AGENT.md` record. […] a stale vault could revert it to inline form and nothing would turn red.

Correct, and the mechanism it names is not hypothetical: `compile-harness.sh --refresh` regenerates these records from the vault, so a checkout whose vault half is stale rewrites a migrated record back to the flat form and the gate goes quiet with nothing red. `TestMigratedPersonasStillGateSomething` exists for exactly that and asserts **by name** — its own comment says each migration costs one line there. This PR added a migration and omitted the line.

Fixed in `fb3e26a`, and **verified by mutation rather than by the suite going green**: reverting a migrated record to the flat form makes the test fail by name, then pass once restored. A guard never observed failing is not a guard.

The same gap was present on the sibling PR #1527 and is fixed there too. The list is now one name per line, so the next migration adds a line instead of editing one.

**This is the finding of the session from a reviewer** — the two `AGENT.md` diffs were reviewed by a human eye for prose accuracy and the missing test line was invisible in that reading, because the test lives in a file neither PR touched.

### Ignored: the ticket-compliance section

pr-agent inferred the wrong ticket again, treating pull-request numbers as issues to satisfy. Known reviewer-configuration defect, six consecutive occurrences. The code observation above was read on its merits and applied; only the ticket inference is discarded.

### Note for whoever merges

`main` carries `required_status_checks.strict: true`, so this PR and #1527 will each need `update-branch` after the other lands, and both add a name to the same list — the conflict is one line and resolving it means keeping **both** `curator` and `planner`.

mlorentedev:
## Review triage — addendum, head `b208e01`

`update-branch` onto `43d4b0c` (#1524, now merged) moved the head, which reopened the review window. Re-read rather than assumed:

- **CodeRabbit: still no review.** Its comment now reads *"Next included review available in 33 minutes"* — and it **edited that comment in place**, `created=21:46:20Z` / `updated=22:25:02Z`. `created_at` never moved, so `dotf pr triage-queue` cannot see it. Fifth measured occurrence of #1422. Still a notice, still not a review.
- **pr-agent re-reviewed**: effort 1/5, *PR contains tests*, no security concerns, one focus area.

### pr-agent — "Incorrect Count" → **declined**

> The test now includes five personas (reviewer, builder, shipper, architect, planner), but the PR title states this is the "sixth" enforceable persona.

Declined: this is an artifact of reviewing one PR of a pair in isolation. `curator` is the fifth and is added to the same list by the sibling PR **#1527**, deliberately kept separate because `harness/agents/**` is not in spec-gate's `_excluded()`, so two persona migrations in one PR would cross the 50-LOC production threshold. The list reads six once both land.

The reviewer could not have known that from this diff alone — which is also why both PRs' triage says the merge conflict on this exact hunk **must be resolved by keeping both names**.

All checks green on this head at time of writing. **Nothing outstanding.**

mlorentedev:
## Conflict resolved — head `91485ff`

#1527 merged (`5101938`), which produced the conflict both triage comments predicted: a single hunk in `TestMigratedPersonasStillGateSomething`, `curator` from main against `planner` from this branch.

**Resolved by keeping both**, in roster order:

```go
for _, name := range []string{
	"reviewer",
	"builder",
	"shipper",
	"architect",
	"curator",
	"planner",
} {
```

**Verified by mutation rather than by the merge compiling** — reverting *both* records to the flat form makes the guard name both, so neither name was silently dropped in the resolution:

```
--- FAIL: TestMigratedPersonasStillGateSomething
    persona_test.go:272: curator has been migrated but now gates nothing — reverted to the flat form?
    persona_test.go:272: planner has been migrated but now gates nothing — reverted to the flat form?
```

Records restored, `go build`/`vet`/`test ./internal/harness/ ./internal/cmd/` green.

**Net diff against `main` is unchanged by the merge** — the check that the resolution took nothing else with it:

```
 cli/internal/harness/persona_test.go |  1 +
 harness/agents/planner/AGENT.md      | 29 +++++++++++++++++++++++---
```

One wrinkle worth recording: the earlier `update-branch` was run through the REST API, so the remote branch had a merge commit the local worktree never fetched. Merging `origin/main` locally therefore produced a non-fast-forward push. Reconciled by merging the remote branch in and re-checking the net diff — **not** by force-pushing over it, which would have discarded the API's merge commit.

PR #1531, opened 2026-09-06
fix(cli): run Gate f last, immediately before the removal

mlorentedev:
Round 4 of the BUG-093 adversarial review — `agy/gemini-3.1-pro-high`, random draw, **a different
provider family** from round 3's `nan/qwen3.8-flash`, and not the implementer. Verdict **FAIL** on
two findings, down from ten.

Reviewed `5101938`; the launch pin and `reviewed_sha` agree, so the provenance problem round 3
flagged is closed.

No `Closes #1516` — the spec archives only on a PASS, and this verdict is FAIL.

## Finding 1 (Major) — applied

The TOCTOU re-check sat **before** `isDirty` and `isMerged`, both of which shell out to git. That
latency lived inside the window between "nobody is in there" and the directory being deleted, so a
shell that `cd`s in while git runs would be missed.

Gate f now runs immediately before `executeWorktreeReap`. It is also the **cheaper** order — Gate f
walks all of `/proc`, so running it after the cheap refusals means fewer walks, not more.

`TestGateFIsTheLastCheckBeforeRemoval` records the actual call sequence through `SweepWithRunner`
and asserts the check immediately preceding the removal *is* Gate f, with an anchor that fails if
the fixture never reaped (which would make the assertion vacuous).

## Finding 2 (Minor) — refuted, and the cause was mine

The review reported that a `filepath.Abs` failure "fails OPEN (`Inside: false`)". It does not:

```
$ grep -n "Inside: false" cli/internal/worktree/sweep_proc_linux.go
(no output)
```

The code returns `GateFReading{Inside: true}` — fail *closed*. The string exists nowhere under
`cli/`.

**Where it does exist is my own harness.** Round 3 committed `mutate.sh` so that `features.json` f8
would be a real check instead of a grep for a word in a markdown file. That put *deliberately
broken code* — the mutation payloads — into the reviewer's reading surface, and a payload was cited
as the implementation.

The harness defended itself against being **run** wrongly and not against being **read** wrongly.
It now opens with a banner naming the argument order (`<real code> <broken code>`), stating that
the 5th argument is sabotage, and naming this exact misreading.

## The harness caught a defect in the harness

The first version of the round-4 mutation *added* the front check while leaving the rear one in
place. It reported **SURVIVED** — and that was correct: the last check before the removal was still
Gate f, so the window had not widened. A duplicate is not a move. The mutation now replaces the
whole function body.

```
  CAUGHT   round 4: Gate f moves back ahead of the git shell-outs
  caught: 8   declared survivors: 1   regressions/anchor-miss: 0   exit 0
```

## Verification

`go build ./...` · `go test ./...` · `go test -race ./internal/worktree/` ·
`GOOS=windows go vet ./...` · `GOOS=darwin go vet ./...` · `golangci-lint run` (2.12.2, the
`versions.conf` pin) — all `rc=0`, **0 issues**. `bash specs/BUG-093-gate-f-fail-closed/mutate.sh`
— **exit 0**.

The reviewer's own `review.md`, `review-request.json` and transcript are committed unedited.

### DIFF
--- a/cli/internal/worktree/sweep.go
+++ b/cli/internal/worktree/sweep.go
@@ -198,9 +198,20 @@ func executeWorktreeReap(repoRoot string, info Info, runner SweepRunner) bool {
 }
 
 func reapSingleWorktree(repoRoot string, info Info, runner SweepRunner, now time.Time, absCwd string) bool {
-	// TOCTOU checks under lock: host process/cwd guard, clean status, merge status, lease expiration
+	// TOCTOU checks under lock: clean status, merge status, lease expiration,
+	// and LAST the host-process guard.
+	//
+	// The order is the point, not a style choice. isDirty and isMerged shell out
+	// to git, so a Gate f check placed before them leaves that latency inside
+	// the window between "nobody is in there" and the removal -- a shell that
+	// cds in while git is running would be missed. Gate f goes immediately
+	// before executeWorktreeReap so the window is as narrow as this code can
+	// make it. TestGateFIsTheLastCheckBeforeRemoval pins the order.
+	//
+	// It is also the cheaper order: Gate f walks all of /proc, so running it
+	// after the cheap refusals means fewer walks, not more.
 	absWT, err := filepath.Abs(info.Path)
-	if err != nil || absWT == absCwd || hostProcessInside(absWT).Inside {
+	if err != nil || absWT == absCwd {
 		return false
 	}
 	if isDirty(info.Path, runner) {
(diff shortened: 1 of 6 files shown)


mlorentedev:
## Review triage

**Nothing to apply.** PR-Agent reviewed and reports **"No major issues detected"** and **"No
security concerns identified"**. Recorded rather than left unwritten: a PR with no findings and no
triage comment is indistinguishable from one nobody looked at, and it sits in
`dotf pr triage-queue` forever.

**Reviewer disclosure:** PR-Agent reviewed. **CodeRabbit did not** — it posted *"Review limit
reached"*, which is a notice about a review rather than a review. This PR therefore carries one
reviewer, not two.

All checks green, including `test (windows-latest)`.

Issue #1533, opened 2026-09-06
HARNESS-111: the adversarial review gate reviews an empty diff on main, and blocks archive on its own THEORETICAL findings


mlorentedev:
The adversarial-review gate has two architectural defects. They were found together on
BUG-093, which took **four rounds and four FAILs** against a repo history of 30 PASS and only
4 specs ever needing more than one round.

## Root 1 — the review has no base, so on `main` it reviews nothing

`harness/skills/adversarial-review/SKILL.md:136` specifies the scope as
`git diff <base>...HEAD` against the merge base. **The launcher records no base.**
`review-request.json` carries only:

```json
{ "reviewed_sha": "...", "reviewer": "...", "requested_at": "...", "review_digest_before": "..." }
```

There is no base concept anywhere in `cli/internal/spec/review.go`. The reviewer must guess it,
guesses `main`, and **a review launched on `main` after the work merged therefore has an empty
diff**.

That is not hypothetical. Measured on BUG-093:

| Round | Launched on | Diff | Evidence verbs in its own findings |
|---|---|---|---|
| 3 | branch `dab7b6e` | real | "Mutation: … rc=0, NOT CAUGHT", "Measured through the production path", "Reproduced with `docker run`" |
| 4 | `main`, post-merge | **empty** | "Code read", "Code read" — nothing executed |

Round 4 declared `git diff main...HEAD` as its source and that diff did not exist. It produced one
THEORETICAL finding and one that was **factually wrong** (it cited `Inside: false`, a string that
appears nowhere in `cli/` — it had read a mutation payload out of the spec folder's committed
harness as if it were the implementation).

**The documented workflow guarantees this.** The doctrine requires the review before archiving,
archiving follows the merge, so the review naturally lands post-merge — where its primary input is
gone.

Note the fix is *not* "review the branch pre-merge": that reopens the provenance problem that
`dab7b6e` caused in round 2, where a commit landed after the review launched. The base must be
pinned; the head is free to be `main`.

## Root 2 — the verdict discards the reviewer's own Reality classification

The skill classifies every finding REAL / THEORETICAL / SPECULATIVE, then the aggregation ignores
the axis: **any Major ⇒ FAIL**. On BUG-093 round 4 the reviewer marked its own Major
**THEORETICAL** and it still blocked the archive, costing a full merge-and-re-review cycle for a
race-window narrowing with no demonstrated exploit.

## Acceptance criteria

[...]

PR #1535, opened 2026-09-06
feat(spec): give the adversarial review a base to diff against, and let the verdict use its own reality axis

mlorentedev:
Refs #1533 — deliberately **not** `Closes`.

The archive-on-merge gate refuses a PR that closes a spec's issue without archiving the spec in
the same PR, and the spec cannot archive without a passing adversarial review. So the issue is
closed by the archive PR, not this one. (My first push carried `Closes` and the gate caught it —
working as designed.)

BUG-093 took **four rounds and four FAILs**, against a repository history of 30 PASS,
7 PASS-WITH-GAPS, 1 FAIL, and only **4 specs ever needing more than one round**. A 4× outlier is
not a fact about that spec. It is two defects in the gate.

## Root 1 — the review had no base, so on `main` it reviewed nothing

The skill defines the scope as `git diff <base>...HEAD`. **The launcher recorded no base** —
`review-request.json` had `reviewed_sha`, `reviewer`, `requested_at`, `review_digest_before`, and
no base concept existed anywhere in `cli/internal/spec/`. So the reviewer guessed, guessed `main`,
and a review launched *on* `main` after the work merged diffed nothing.

Measured on BUG-093's own rounds:

| Round | Launched on | Diff | Evidence verbs in its own findings |
|---|---|---|---|
| 3 | branch `dab7b6e` | real | "Mutation: … rc=0, NOT CAUGHT" · "Measured through the production path" · "Reproduced with `docker run`" |
| 4 | `main`, post-merge | **empty** | "Code read" · "Code read" — nothing executed |

Round 4 declared `git diff main...HEAD` as its source and that diff did not exist. One of its two
findings was **factually wrong**: it cited `Inside: false`, a string that appears nowhere in
`cli/`, having read a mutation payload out of the spec folder's committed harness as if it were
the implementation.

**The documented workflow guarantees this** — the doctrine requires the review before archiving,
archiving follows the merge, so the review lands where its primary input is gone.

### The fix, and why not the PR's base branch

`ResolveReviewBase` returns the **parent of the commit that first added the spec folder**.

The PR route (`issue:` → PRs closing it → oldest PR's base) needs the network, `gh` auth and
human-editable metadata, and answers differently depending on which of a spec's several PRs you
ask. This one is local, offline, deterministic — and correct in both cases that matter:

- **on the work branch** — the adding commit is on the branch, so the parent is where it left `main`;
- **on `main` after a squash** — the adding commit *is* the squash commit, so the parent is `main`
  immediately before the work landed.

That second case is the whole point: it is the one the old code got wrong by having no answer.

The base is recorded as `base_sha`, stated in the prompt, and **a launch that cannot be scoped is
refused** rather than allowed to degrade into a folder-read wearing a verdict.

Note this is deliberately *not* a delta-since-last-round. The base is fixed and the head moves — a
per-round delta would stop round N seeing a fix round 1 forced, so it could not judge whether a
late change interacts with an early one.

## Root 2 — the skill contradicted itself

Stated twice, disagreeing:

> *Reality section:* "weight each finding by `severity × reality` … a **SPECULATIVE** finding
> cannot, by itself, move the verdict below PASS"
>
> *Verdict list:* "**FAIL** — at least one blocker or major"

A reviewer following the second returns FAIL on a Major it has itself labelled THEORETICAL. That is
round 4 exactly — a race-window narrowing with no demonstrated exploit, costing a full
merge-and-re-review cycle.

Not a missing policy: a contradiction inside one document, resolved in favour of the rule already
stated more precisely, with the reason recorded inline so the next editor does not "simplify" it
back.

**Blockers are exempt from the softening, at any reality tag.** Round 1's Blocker on BUG-093 was a
real data-loss bug (`filepath.Abs` does not resolve symlinks, so a symlinked worktree failed open
with a process inside). That is what the gate is for and it is untouched.

## Verification

All 6 `features.json` verifications executed, all `rc=0`.

- **Real git repositories, not fixtures**, for the resolver — including the post-squash case, whose
  test **asserts the anchor first**: `git diff main...HEAD` must be empty in the fixture or it
  `t.Fatal`s, so it cannot pass on a repo where the old behaviour also worked.
- **Both refusal paths plus the positive direction**, so the refusals cannot pass vacuously by the
  launcher refusing everything.
- **Guard mutation-proven**: reintroducing the exact old verdict line turns 3 of 6 assertions red.

```
go build ./...              rc=0     go test ./...            rc=0
GOOS=windows go vet ./...   rc=0     GOOS=darwin go vet ./...  rc=0
golangci-lint run ./...     0 issues (2.12.2, the versions.conf pin)
bats guard-review-verdict-honours-reality.bats   6/6 ok
```

## One thing this does not take effect without

**The reviewer reads the DEPLOYED skill**, `~/.pi/agent/skills/adversarial-review/SKILL.md`, which
`compile-harness.sh` writes. Measured after the edit: `diff -q` → **DIFFERS**. Root 2 is inert
until a deploy runs. Stated here and in `verification.md` rather than discovered when the next
review returns FAIL on a THEORETICAL finding and somebody concludes the fix did not work.

Root 1 has no such gap — it is compiled into `dotf`.

### DIFF
--- a/cli/internal/cmd/spec.go
+++ b/cli/internal/cmd/spec.go
@@ -63,6 +63,15 @@ var runCommand = func(dir string, argv []string) error {
 // rand.IntN in production, a fixed index in tests so a launch is reproducible there.
 var reviewerDraw = rand.IntN
 
+// resolveReviewBase and headSHAOf are seams for the review scope (HARNESS-111).
+// The command-level fixtures build a fake .git directory rather than a real
+// repository, so the production resolvers answer "" there and every launch test
+// would hit the new refusal. The resolvers themselves are tested against real
+// git histories in internal/spec — a seam proves the caller honours the answer,
+// and only a real repository proves the answer is right.
+var resolveReviewBase = spec.ResolveReviewBase
+var headSHAOf = spec.HeadSHA
+
 // runForeground runs the reviewer in this terminal, streaming its output to both
 // the screen and the transcript.
 //
@@ -176,7 +185,24 @@ is the only record of how.`,
 			}
 
 			skill := spec.ReviewerSkillPath(chosen.Runner)
-			prompt := spec.ReviewPrompt(id, repoRoot, chosen.ID, chosen.Runner, skill)
+			// The base the reviewer will diff against. Refusing when it cannot be
+			// resolved, or when it IS the head, is the point: a review with an
+			// empty diff does not fail, it quietly becomes a reading of the spec
+			// folder and reports findings with no execution behind them. That is
+			// worse than no review, because it arrives wearing a verdict.
+			baseSHA := resolveReviewBase(repoRoot, specDir)
+			headSHA := headSHAOf(repoRoot)
+			if baseSHA == "" {
+				return fmt.Errorf("cannot resolve a review base for specs/%s: no commit adds that folder.\n"+
+					"The reviewer diffs base...HEAD, so without a base it would browse the folder and\n"+
+					"call that a review. Commit the spec first", id)
+			}
+			if baseSHA == headSHA {
+				return fmt.Errorf("the review base and HEAD are the same commit (%s).\n"+
+					"There is nothing to review: the spec folder was added by HEAD itself", baseSHA[:min(12, len(baseSHA))])
+			}
+
+			prompt := spec.ReviewPrompt(id, repoRoot, chosen.ID, chosen.Runner, skill, baseSHA)
 			argv, err := spec.ReviewerCommand(chosen, prompt, timeout, repoRoot)
 			if err != nil {
 				return err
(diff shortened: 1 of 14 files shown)


mlorentedev:
## Review triage

**Nothing to apply.** PR-Agent reviewed and reports **"No major issues detected"** and **"No
security concerns identified"**.

One finding did arrive, from CI rather than a model, and it is worth recording because it was the
gate under change catching its author: the first push carried **`Closes #1533`**, and `spec-gate`
refused with *"SDD archive-on-merge violation: this PR closes an issue whose spec is still
active"*. Correct — a spec cannot archive without a passing adversarial review, so the issue is
closed by the archive PR, not this one. The body now reads `Refs #1533` and says why.

**Reviewer disclosure:** PR-Agent reviewed. **CodeRabbit did not** — *"Review limit reached"* is a
notice, not a review.

All checks green, including `spec-gate` and `test (windows-latest)`.

PR #1541, opened 2026-09-06
chore(spec): archive BUG-093, Gate f fail-closed

### LINKED ISSUE #1516: BUG-093: Gate f is silently off on Windows — the worktree reaper deletes on 'I could not find out'

mlorentedev:
Archives BUG-093 on its round-5 review, and records why the reviewed contract is archived unchanged.

## The review

Round 5 returned **PASS-WITH-GAPS** — the first pass after four consecutive FAILs — from `nan/qwen3.8-flash`, drawn at random and not the implementer. Its Sources line is the evidence that the two roots fixed in #1535 worked: it reviewed `git diff 5a81e8a1...HEAD (the launcher's base_sha)` and ran **12 mutations**, where round 4 reviewed "Code read" and ran none.

Its single Major is the bind-mount fail-open, declared out of scope in the proposal and owned by **#1523**. The reviewer honoured the declaration instead of blocking on it, which is the amended verdict rule working on its first use.

## Why the contract files are unchanged

The review's "Recommended next steps (before archive)" asked for edits to `proposal.md` and `features.json`. Those are the files the archive gate measures staleness against, so applying them invalidated the review that requested them:

```
Error: review.md is stale: features.json, proposal.md changed after reviewed_sha 5a68096
```

The gate is right and the recommendation was wrong. `PASS-WITH-GAPS` means the gaps are tracked, not fixed; and editing the reviewed files makes the review's own findings refer to text that no longer exists — a worse inconsistency than the stale number it would correct. Both files are restored byte-for-byte to `5a68096`, so the spec archives exactly as reviewed.

Each of the seven recommendations is dispositioned in `verification.md` (excluded from the staleness check for precisely this purpose) as applied, ticketed, or declined with a reason:

- **Applied, outside the contract set:** `flip-proxy.sh` (the four-file build-tag flip, self-verifying), the `/proc` fail-open declared as a survivor in `mutate.sh`, and the order test shipped in #1531.
- **Verified:** `test-windows: success` at `36c27fb15a6f`, `195cc1887b9a` and `6f9b0d95729e`.
- **Ticketed:** the tmpfs measurement carried into #1523.
- **Declined with a reason:** three evidence/prose refreshes, since the review already records the true values.

The residual is stated in the table rather than hidden: f3, f5 and f8 name pre-round-5 commands.

## Follow-up

The template defect behind this is real and not fixed here — `harness/skills/adversarial-review/SKILL.md:282` tells every reviewer to head its list *"(before archive)"*, and `:153` says a Major requires a *"spec update ... before archive"*. Both instruct exactly the edit the gate refuses. That lands on #1533 with the other review-gate roots.

Closes #1516


mlorentedev:
## Review triage

|---|---|---|
| PR-Agent (`github-actions`) | **No major issues detected**, no security concerns, tests present | Nothing to apply. |
| PR-Agent — ticket compliance | Analysed #1535, #1531, #1523 as "tickets" | **Noted, not a finding.** #1535 and #1531 are pull requests, not tickets; the reviewer's ticket-scope inference misreads PR references again. The substance it reports is accurate. |
| CodeRabbit | *"Review limit reached — next included review available in 23 minutes"* | **Not a review.** Disclosed rather than counted: a notice about the review is not a review, and this PR carries one substantive reviewer, not two. |
| CI | 12 checks green, `test-windows` included | — |

**Independent adversarial review**: round 5, `nan/qwen3.8-flash`, drawn at random and not the implementer. Verdict **PASS-WITH-GAPS**; its single Major is the bind-mount fail-open, declared out of scope and owned by #1523. Archived at `reviewed_sha` `5a68096` with the contract files byte-identical to the reviewed state; the seven recommendations are dispositioned in `verification.md`.

PR #1543, opened 2026-09-06
fix(harness): stop the review skill asking for the edits the archive gate refuses

mlorentedev:
Third root under HARNESS-111. Roots 1 and 2 shipped in #1535; this is the one that surfaced immediately afterwards, on the first spec to archive under the fixed launcher.

## The defect

Two lines of the review skill told every reviewer to demand contract-file edits before a spec archives:

- `SKILL.md:153` — *"**Major**: likely bug or significant gap; fix or spec update required **before archive**."*
- `SKILL.md:282` — the heading itself: *"### Recommended next steps **(before archive)**"*

The archive gate measures a review's staleness against `proposal.md`, `tasks.md` and `features.json`. So a reviewer returning **PASS-WITH-GAPS** and heading its list "before archive" has asked for exactly the edits that invalidate the verdict it just issued. Applying them gives:

```
Error: review.md is stale: features.json, proposal.md changed after reviewed_sha 5a68096
```

Measured on BUG-093 (#1516) round 5: four of seven recommendations targeted the contract set, on a spec that had already spent four rounds.

## The fix, and what was rejected

**The gate is right; the template was wrong.** A verdict is about a state — the review's own findings quote text that must still be on disk when the spec archives, so editing the reviewed files makes the review internally inconsistent, a worse defect than the stale evidence line it would correct. And `PASS-WITH-GAPS` already means the gaps are *tracked*, not fixed.

What was missing is a routing rule, now stated in the skill: the **contract set** is those three files; under **FAIL** a contract edit is the point and a re-review follows; under a **passing verdict** the set is closed and recommendations are dispositioned in `verification.md`, which the staleness check already excludes for exactly this purpose.

**Rejected, and recorded in `tasks.md`:** making the staleness check field-granular so `evidence`/`verification` edits stop invalidating a review. `verification` is the command the reviewer *ran*, so exempting it opens the "get a pass, rewrite in the working tree, archive" bypass the comment at `review.go:187` already names — and `proposal.md` prose has no field structure to key on. The coarseness is the check working.

## Also

The refusal named three exits — re-review, waive, force — all of which discard or bypass a verdict that is usually perfectly good, so an operator reading only those reaches for an override with a passing review sitting beside it. It now names the review-preserving exit first.

`TestStaleRefusalOffersTheExitThatKeepsTheReview` asserts the exit is present, says where dispositions go, and is offered *before* each override. Mutation-proven: reverting `review.go` to the old message fails it.

Both copies of the skill are updated — the repo record under `harness/skills/` here, and the vault source at `00_meta/skills/`, pushed to `master` and verified by content on `origin/master` (0 defective forms, 4 routing-rule lines).

## Verification

`go build`, `go vet`, `go test ./...`, `GOOS=windows go vet`, `GOOS=darwin go vet` all rc=0; `golangci-lint run` 0 issues on the pinned 2.12.2.

Refs #1533


mlorentedev:
## Review triage

|---|---|---|---|
| 1 | PR-Agent | **Vault Source Not Updated** — the diff changes only `harness/skills/`, so the two copies will diverge | **Declined — false as a fact, but right about the limitation.** The vault is a *separate repository* (`00_meta/skills/` lives in the knowledge vault, not here), so its change cannot appear in this diff no matter what. It was committed and pushed, and verified on `origin/master` **by content**, not by assumption: `0` occurrences of the two defective forms, `4` of the routing rule — the same counts as the copy in this PR. See below for the part of the finding that is real. |
| 2 | PR-Agent | **Test Fragility** — the guard greps the literal `"restore the contract files"`, so a semantics-preserving reword would fail it. LOW, "not a blocking issue" | **Declined, deliberately.** The assertion has to anchor on something a reader would actually see, and the property under guard *is* textual: the operator learns the exit exists by reading it. Extracting the fragments into shared constants would make the test assert a constant against itself — vacuous, which is the failure mode lesson 274 was written about on this same spec. A reword that keeps the meaning should update the test; that edit is cheap and visible, and it is the correct cost of the guard. |
| 3 | CodeRabbit | *"Review limit reached"* on the companion PR #1541 | **Not a review.** Disclosed, not counted. |

### The real half of finding 1, kept rather than dismissed

Nothing in CI verifies that the two copies of a skill agree, and CI cannot easily reach the vault — it is a different repository and is not checked out. Today the only thing asserting parity is the spec's own `HARNESS-111-f7`, a manual verification command. That is a genuine gap the reviewer surfaced by accident, and it is worth a ticket on its own rather than a hurried guard in this PR; the routing-rule change here is what #1533 asked for, and widening it to skill-parity CI would be a different concern.

**CI**: `lint`, `lint-powershell`, `spec-gate`, `attestation`, `changes`, `review` green; `test` / `test-windows` / `integration` still running at the time of writing and must be green before merge.

mlorentedev:
Follow-up on triage finding 1: the real half is now **#1545**.

Mechanism, measured after posting the triage: `compile-harness.sh --refresh` regenerates `harness/skills/` **from** the vault, so a change made only to the repo record is silently reverted at the next sync — the hazard is directional, and the reviewer was right that neither a human reading the diff nor CI can tell whether the vault half was done. `--check` is offline by design (`OK: no harness drift`, without opening the vault source), so it does not close it as written.

For this PR specifically, both halves are verified: the bodies are byte-identical modulo frontmatter, and `bash scripts/compile-harness.sh --check` exits 0.

mlorentedev:
## Review triage

Second round, after the spec and lesson commits re-triggered the reviewer.

| # | Finding | Disposition |
|---|---|---|
| 4 | **Verbose Error Message** — the refusal is long and *"includes multiple escape characters (`—`)"* | **Split: claim false, concern valid.** The em-dash claim is wrong and was checked rather than assumed — grepping the refusal string for `U+2014` returns **0**; there are no em-dashes in it. The verbosity half is right, though, and it undercut the change itself: rendered, the four exits ran together on one 260-character line, so the ordering this PR exists to establish was invisible. **Applied** — one exit per line, same order, review-preserving exit first. The guard is index-based and passes unchanged across the reflow. |

Rendered now:

```
review.md is stale: features.json, proposal.md changed after reviewed_sha 5a68096
  restore the contract files to reviewed_sha and record what changed as dispositions in verification.md (excluded from this check)
  re-run /adversarial-review against the current head
  declare `review: waived` with a reason in proposal.md
  pass --force-without-review
```

Findings 1-3 are dispositioned in the earlier triage comment; the real half of finding 1 is **#1545**.

**Verification after the change:** `go build`, `go vet`, `go test ./...`, `GOOS=windows go vet`, `GOOS=darwin go vet` all rc=0; `golangci-lint run` 0 issues.

mlorentedev:
## Review triage

Third round.

| # | Finding | Disposition |
|---|---|---|
| 5 | **Error message clarity** — the message lists four options but does not say the first preserves the review while the others discard it, so an operator might still pick an override on a valid review | **Applied, and it is the best finding on this PR.** It names the exact behaviour the change exists to influence: ordering alone does not tell anyone what the ordering *means*. The exits are now grouped under two labels, and the guard was extended to assert the labels bracket the exits they describe — mutation-proven by deleting a label and watching it go red. |
| 6 | **Routing rule clarity** — the new rule does not state whether a REAL Major in the contract set under FAIL requires a contract edit and then a re-review | **Declined — already stated, verbatim.** `SKILL.md:166-168`: *"So a Blocker or a REAL Major in the contract set means **fix, then re-review**: the next round is the mechanism, not an inconvenience."* Checked against the file rather than argued. |

Rendered now:

```
review.md is stale: features.json, proposal.md changed after reviewed_sha 5a68096
keeps the review:
  restore the contract files to reviewed_sha and record what changed as dispositions in verification.md (excluded from this check)
discards it — only if the review is genuinely no longer the right one:
  re-run /adversarial-review against the current head
  declare `review: waived` with a reason in proposal.md
  pass --force-without-review
```

**Verification:** `go build`, `go vet`, `go test ./...`, `GOOS=windows go vet`, `GOOS=darwin go vet` all rc=0. `golangci-lint run` initially flagged my own `QF1001` (De Morgan) in the new assertion — fixed, now **0 issues** on the pinned 2.12.2.

Findings 1-4 are dispositioned in the two earlier triage comments; #1545 carries the real half of finding 1.

mlorentedev:
## Review triage

Fourth and final round: **No major issues detected**, no security concerns, tests present, ticket compliance clean against both #1533 and #1516. **Nothing to apply — the queue is closed on this PR.**

### Summary across the four rounds

| Round | Finding | Outcome |
|---|---|---|
| 1 | Vault source not updated | Declined as a fact (separate repository, verified on `origin/master` by content); its real half filed as **#1545** |
| 1 | Test fragility — the guard greps a literal | Declined; anchoring on the literal is the point, and shared constants would make it vacuous |
| 2 | Verbose error message, *"includes em-dashes"* | Em-dash claim checked and false (0 occurrences); the verbosity half **applied** — one exit per line |
| 3 | The message does not say which exit preserves the review | **Applied**, and the strongest finding here — the exits are now labelled and the guard asserts the labels bracket them |
| 3 | Routing rule silent on a REAL Major under FAIL | Declined; `SKILL.md:166-168` states it verbatim |
| 4 | — | Clean |

Two of six applied, both improving the operator-facing behaviour this PR exists to change. The other four were checked against the file or the repository rather than argued.

**CodeRabbit** produced only a rate-limit notice across the whole PR — disclosed, not counted as a review.

**Final verification:** `go build`, `go vet`, `go test ./...`, `GOOS=windows go vet`, `GOOS=darwin go vet` all rc=0; `golangci-lint run` **0 issues** on the pinned 2.12.2; `bash scripts/compile-harness.sh --check` → `OK: no harness drift`.

**Merge precondition, not a CI check:** the reviewer reads the *deployed* skill. Both deployed copies still measure 2 defective forms / 0 routing-rule lines, so `scripts/compile-harness.sh --deploy` must run after this merges and before `dotf spec review HARNESS-111-review-base-and-reality` is launched.

mlorentedev:
## Review triage

Rounds 5 and 6: **No major issues detected**, no security concerns — byte-identical in substance to round 4. Nothing new to disposition; all six substantive findings are dispositioned in the comments above.

**Stopping the loop deliberately, and saying so rather than leaving it unexplained.** Posting a triage comment here re-triggers the workflows, which re-runs PR-Agent, which posts a fresh review and puts this PR back in `dotf pr triage-queue`. Measured four times this session. So a non-zero queue on this PR is now self-perpetuating and is not undispositioned work — it is the #906 behaviour, and I have added the measurement there.

Final state: every reviewer finding applied, ticketed, or declined with a reason; two applied. Ready to merge once the current check run lands green.

mlorentedev:
## Review triage — disclosure, posted after the merge

**This PR merged with an undispositioned review.** CodeRabbit came off its rate limit and posted its
first non-notice review of the session at 03:25:42, with 3 actionable inline findings. The merge
landed at 03:28:52, three minutes later. The findings were real, not noise — all three were
confirmed against the code:

| # | Severity | Finding | Status |
|---|---|---|---|
| 1 | Major | The **Major severity definition** still demanded a fix and re-review for *every* Major, contradicting the verdict list this spec's root 2 fixed to route a THEORETICAL Major to PASS WITH GAPS | Applied in #1549 |
| 2 | Major | f7's verification command discarded the first `grep -c` status, so the "zero defective forms" half was never asserted | Applied in #1549 |
| 3 | Minor | Lesson 275 described `review: waived` as asserting no review happened; `checkReviewGate` returns on the waiver *before* `FindReview`, so it asserts nothing | Applied in #1549 |

Finding 1 is the same class this PR exists to remove — an instruction contradicting the gate — one
section above where the PR was looking. The guard added for root 2 was blind to it because it only
read the verdict list; #1549 extends it to read both halves, mutation-proven.

Disclosing rather than staying silent: the doctrine allows proceeding on an unreviewed PR, not
proceeding quietly. Nothing here needed a revert — the three fixes are additive and land in #1549
before the HARNESS-111 archive review runs.

Issue #1537, opened 2026-09-06
HARNESS-120: the suggest hook names the persona and stops — nothing dispatches it, and its stated blocker is now closed


mlorentedev:
## What

`dotf harness suggest --from-hook` runs on `UserPromptSubmit` and resolves the prompt's matched pattern to the personas that own the matched skills. It prints them:

```
[persona] builder | reviewer | shipper  ← pattern: pattern-git-workflow
  skills: golang-pro, pr-review-triage, test, test-driven-development, …
  → each of these declares the matched skill; the work decides which
```

Then it stops. **Nothing launches the persona it just named.** A human or the main-thread agent decides, or more usually does not.

## Why this is worth reopening now

HARNESS-110 recorded the scope decision as *"v1 suggests; it does not launch. Launching is gated on #1434"*. **#1434 closed 2026-09-04 via #1471** — a named dispatch now resolves through the dispatch map to its true type. The stated blocker no longer exists and nothing captured that.

## The measurement that makes it urgent

Of **11 745** gate decisions, **11 168 (95.1%) are `no-role`** — the main thread, where no persona is in scope and the gate has nothing to enforce. Only 20 distinct sessions exist and **18 produced no persona decision at all**.

So the whole persona/gate apparatus governs a lane almost nothing travels. Suggesting a persona that never gets dispatched is the mechanism by which that happens: the suggestion is advisory, the dispatch is manual, and the manual step is skipped.

**This matters more than any `block` promotion.** Promoting a severity changes what happens on the 1.1% of calls that already resolve a persona. Dispatching the suggested persona changes which lane the work runs in at all.

## Hard constraint, measured and not negotiable

On `UserPromptSubmit`, **exit 2 blocks prompt processing and erases the prompt** — a data-loss mode strictly worse than the gate's. So this hook can only ever fail open, and dispatch cannot be implemented as "refuse until the right persona is running". It needs a mechanism that is not an exit code.

## Open questions for the design, not answered here

- Who dispatches — the hook itself, or the main-thread agent acting on the suggestion?
- What happens when the join is ambiguous? `code-complexity-and-refactor` resolves to `[builder, reviewer]`, and the recorded decision is that **ambiguity is a first-class output**: never rank, never tie-break.
- Should the main thread carry a persona at all? Today it carries none by construction, which is the 95%. That is a larger question than this ticket and may deserve its own.

## Acceptance criteria

[...]

PR #1546, opened 2026-09-06
feat(agent): dispatch the persona a task implies, on the tier it declares

### LINKED ISSUE #1538: HARNESS-121: personas declare a tier and model-map declares five pools, but nothing walks persona -> tier -> pool, so a persona can only ever run on Claude
## What

Two halves of cross-provider dispatch exist and are not joined.

**Half one — the persona already names a tier.** Every `AGENT.md` carries it:

| persona | `model:` |
|---|---|
| architect, curator | `top` |
| builder, shipper, planner, reviewer | `mid` |

**Half two — `model-map.json` already maps tier to five pools**, with fallback chains:

```
top → claude:opus · copilot:claude-opus-5 · gemini:gemini-3.1-pro-high
mid → claude:sonnet · copilot:gpt-5.6-sol · nan:deepseek-v4-flash · opencode:qwen3.6-plus
low → claude:haiku · copilot:gpt-5.4-mini · nan:qwen3.8-flash
```

**And the executor exists too.** `dotf agent run` *"dispatches one task to the first pool in a tier's chain that can serve it"*, with two backends: a subprocess running the harness binary non-interactively, and `hive delegate`.

**Nothing connects them.** `dotf agent run` dispatches *a task to a tier*. The gate governs *Claude subagents*. A persona's `model: top` is rendered into records and read by nothing that routes.

## The consequence

A persona is a Claude-only construct in practice. Dispatching `architect` runs Claude's `Agent` tool; there is no path by which `architect` runs on copilot, gemini, nan or opencode — even though the map says what model each of those would use for its tier, and the executor knows how to reach them.

That is the gap between what the orchestrator is *designed* to be (agnostic — the standing goal records the agnostic primitive as `dotf`, not Claude hooks) and what it *is* (one harness).

## Why file it now rather than discover it later

The persona records are merged and the tier field is in all six. The map is the declared SSOT. The executor shipped. Every piece is in place and the seam has never been walked, which is exactly the state where a design gets quietly re-invented by someone who did not know the halves existed.

## Open questions, deliberately not answered

- Does a persona dispatched to another pool still get gated? The gate normalises per-harness payloads into a neutral `ToolCall`, so in principle yes — but no non-claude harness has ever produced a gate decision (#1532 phase 1 measures this first).
- `dotf agent run` is one-shot and non-interactive. A persona is a multi-turn actor. Whether the seam is "dispatch a persona" or "dispatch a task carrying a persona's obligations" is the actual design question.
- `harnessFor` in `backends.go` deliberately holds argv in Go rather than in the map (*"a map, never a director"*, ADR-035 §4), and its own comment says **a third harness is the trigger to revisit**. Joining this seam is likely that trigger.

## Acceptance criteria

- [ ] A decision is recorded on whether a persona is dispatchable to a non-Claude pool at all.
- [ ] If yes: one persona is observed executing on a pool other than `claude`, chosen through its declared tier rather than hardcoded.
- [ ] The tier→pool resolution reads `model-map.json` rather than duplicating it.
- [ ] Gating behaviour for a cross-pool dispatch is stated, not left to be discovered.

Depends on #1532 phase 1 — a harness whose gate has never fired cannot be given a persona to enforce.

mlorentedev:
Closes #1538.
Refs #1537 — spec `specs/HARNESS-120/`.

## Why

`dotf harness suggest` derives the persona a prompt implies and stops. `dotf agent run` dispatches, but only if a human supplies the role and the tier by hand. Between them sat a manual step, and the measurement says it is not taken: of **11 745** gate decisions, **11 168 (95.1%)** are `no-role`, and **18 of 20** sessions produced no persona decision at all. The apparatus governed a lane almost nothing travelled, because entering it cost a deliberate act nobody performed.

Two fields told that story, and both were measured for the spec:

- **`Persona.Model`** — the tier each record declares for itself — was written at `persona.go:107` and read by **nothing**. `#1538`.
- **`Request.Role` reached no real backend.** Threaded from `--role` into `Request` at `dispatch.go:136`, and read only by the dry-run backend's echo (`dryrun.go:22`). `Subprocess` sends `--model` and the task; `Hive` sends `--model`, `--timeout`, `--prompt`. Neither sent the persona.

So `dotf agent run --role reviewer` dispatched a **generic** agent that happened to be logged as a reviewer. The specialization the six records carry — mandate, method, boundaries — stopped at the process boundary. **That is the half worth more than the routing:** a task could already be sent to another pool; it could not yet be sent to a *persona*.

## What

`dotf agent auto --task "<text>"`, chaining Suggest → ResolveOne → the persona's declared tier → a preamble carrying its record → `agent.Dispatch`.

```
$ dotf agent auto --task "open a ticket for the bitacora" --backend dry-run --timeout 60s
{
  "status": "dry_run", "tier": "mid", "role": "planner",
  "pool": "claude", "model": "sonnet",
  "resolution": { "role_from": "inferred", "tier_from": "inferred",
                  "pattern": "pattern-bitacora-tracking" }
}
```

The matched pattern is the one `proposal.md` predicted **before the code existed** — the criterion reproducible from the spec file rather than fitted to the result.

**The record travels.** Verified through a stub harness binary that prints its stdin, so the evidence is the bytes that crossed the process boundary rather than a Go value:

```
$ dotf agent auto --task "check this diff" --role reviewer --backend subprocess ...
ARGV: -p --model sonnet
--- STDIN AS RECEIVED ---
You are operating as the `reviewer` persona.
...
## Boundaries
You review; you do not edit, and you hold no write capability on purpose — a
reviewer who fixes what they find has stopped being independent of it. ...

===== TASK =====

check this diff
```

3219 bytes sent where the bare task is 15. The task arrives **last** and intact: a task read before the record has had its operating instruction arrive too late to shape how the work is read.

## Decisions worth disagreeing with

**`agent run` is deliberately NOT given the preamble — only `auto` is.** AC6's `--role reviewer` is `auto`'s flag, which AC5 settles by describing the same flag on the same command. `run` is the primitive that takes a route as given: its `--role` is a label, not a roster lookup, and making it one would refuse names it has always accepted — its own tests pass `--role r` against temp roots with no roster at all.

**The residual is real and stated rather than hidden: `dotf agent run --role reviewer` still dispatches a generic agent.** Pinned by `TestAgentRun_DoesNotComposeAPreamble` so the two commands cannot drift into each other unnoticed. Filed as follow-up rather than smuggled into this diff.

**`frontmatterBlock` became `splitRecord` returning both halves**, rather than the preamble getting its own body splitter. The duplicate had already been written and carried a literal `U+FEFF` where the `﻿` escape belonged — which Go rejects anywhere but byte 0, so it broke the build of the whole package, invisibly in the editor and in `git diff`. The reason to delete rather than patch it is the one its own comment stated: two ideas of where a record ends is how one of them starts being wrong. Same shape as `check-roster-consistency.py` reporting "no skills" in silence. Lesson 275.

**Composition happens at the command layer, before the walk.** `Dispatch` retries across pools; composing per attempt could send different bytes to the second pool than the first, making a comparison of the two answers a comparison of two different questions.

**`prepareDispatch` extracted from `run` before `auto` was written.** The identity gate and deny list are fail-closed (ADR-032 §8), and a second command re-deriving them is a second place they can be forgotten — where forgetting looks exactly like working.

## Contract change (additive, but a contract change)

`agent.Record` gains `role` and `resolution` — both `omitempty`. `resolution` reports `role_from` / `tier_from` (`inferred` | `dictated`) and the matched `pattern`. `tier: mid` is the same three characters whether a human typed it or a record declared it, and those want different scrutiny: the matcher is substring-based (`triggers.go:116`), so a task mentioning "docker" in passing resolves `shipper`. `run` emits no `resolution` because it derived nothing.

JSON readers keying on existing fields are unaffected; readers doing exact-shape comparison are not.

## Verification

`specs/HARNESS-120/verification.md` carries each criterion with the command's own output quoted verbatim, against the shipped records and shipped `model-map.json`.

- **AC1–AC6: closed.**
- **AC7 (one REAL dispatch): NOT satisfied, and recorded as blocked rather than left unticked.** This machine's `machine.json` declares no `machine.id`, so ADR-032 §8's identity gate denies every non-local pool. Declaring one is a decision about which pools this machine may spend quota on — the owner's, not the implementer's. **No `dry-run` quote is offered in its place:** AC7's whole content is that a real pool answered. Not new to `auto` — the same gate refuses `agent run` here and always has.

```
cd cli && go build ./... && go vet ./... && go test ./...   # all pass
GOOS=windows go vet ./...                                    # clean
golangci-lint run   # at the versions.conf pin → 0 issues
```

**Mutation-checked, because a passing test proves nothing until it is shown it can fail:**

| Mutation | Result |
|---|---|
| Send the bare task instead of the composed preamble | `TestAgentAuto_SendsThePersonasOwnRecordToTheBackend` FAILS |
| Make the ambiguous case return its first candidate | `TestAgentAuto_BothRefusalsDispatchNothingAndDiffer` + `TestResolveOneDispatchesOrRefuses` FAIL |

## For the reviewer

Sharpest places to push:

1. Is the `run`/`auto` split right, or is leaving `run --role` generic a defect this PR should have closed?
2. `TestAgentAuto_BothRefusalsDispatchNothingAndDiffer` asserts the two refusals *differ*, never their wording. Is that too weak?
3. `dispatchBackends` is a package-level var swapped by tests. Is a test seam on a dispatch path acceptable here, given `dryrun.go:13-15` argues against a `--backend` value existing for tests?


mlorentedev:
## Review triage

All 16 checks pass, 1 skipping. Rebased onto `39754ec`.

**Who actually reviewed, stated rather than implied.**

|---|---|---|
| **pr-agent** (`github-actions`) | Substantive: ticket-compliance analysis against #1537 and #1538, effort estimate, security assessment | **Yes** — names requirements, files and claims |
| **CodeRabbit** | *"Review limit reached — next included review available in 56 minutes."* | **No.** A notice *about* the review, not a review. This PR is unreviewed by CodeRabbit, and saying nothing would have been indistinguishable from a review that found nothing |

Triaged on pr-agent, which is the reviewer that produced output. Disclosure over silence.

### Findings

|---|---|---|---|
| 1 | pr-agent | **#1538 partially compliant** — no persona observed executing on a non-`claude` pool | **Ticketed, already disclosed.** This is AC7, and it is recorded as *blocked* in `verification.md` and `features.json` rather than left unticked. `machine.json` on the implementing box declares no `machine.id`, so ADR-032 §8's identity gate denies every non-local pool. pr-agent reaches the same conclusion independently and calls it "a machine configuration issue unrelated to this change" — correct. Tracked as **#1547** |
| 2 | pr-agent | **#1537 partially compliant** — no real prompt observed producing a dispatch whose gate decision names the persona | **Same root as #1547.** Requires a real dispatch, which the identity gate blocks. Not separately actionable |
| 3 | pr-agent | Effort to review: 4/5; contains tests; **no security concerns; no major issues detected** | **Noted.** No action |
| 4 | CI (`test`, run `34009285030`) | Red on the PR **merge commit** while every branch commit was green: `guard-lesson-numbers-unique.bats` caught two different `lesson-275` files — one from #1543, one from this branch | **Fixed** in `7b105e1`. Renumbered to 276 (file, `id:` frontmatter, `_index.md` row) and rebased. Worth noting the shape: neither tree was wrong alone, which is why only the merge commit failed |
| 5 | CI (`test`, run `34008756211`) | `check-md-escapes.sh` flagged the new lesson: literal `\n` followed by a line-start glyph inside Go string literals in a code example | **Fixed** by rewriting the example as a table with `<CR>`/`<LF>` placeholders. The guard was not weakened and no exclusion was added — a lesson about invisible characters is a poor place to argue for an exception to the check that finds them |

### Not done, deliberately

**The independent adversarial review required before `spec archive` has NOT been performed.** I implemented this change, and the doctrine is that the reviewer must not be the implementer — the independence is the entire value, and an adversarial review must not run on the model family that implemented the change. `dotf spec archive` will refuse without it, correctly. Flagging it here so "CI green + triage posted" is not mistaken for "reviewed".

### Follow-ups filed

- **#1547** — ADR-032 §8's identity gate is right, but nothing closes the loop it opens: `setup-linux.sh` seeds only `DOTFILES_REPO_DIR`, no `dotf` subcommand writes the `machine` block, and `dotf doctor` has **no check** for a missing `machine.id`. The only thing that reports an undeclared machine is a dispatch refusing at the moment you try to use it. `dotf agent run` has been refusing on this box all along and nobody noticed, because nothing asks
- **#1548** — `dotf agent run --role` still dispatches a generic agent. Deliberate boundary, pinned by `TestAgentRun_DoesNotComposeAPreamble`, filed so it stays revisitable rather than buried in a verification file

_Not re-running `dotf pr triage-queue` after this comment: posting under this heading re-triggers the workflows and re-queues the PR (#906, measured 4× today). Chasing a clean exit here is a loop, not work._

Issue #1545, opened 2026-09-06
HARNESS: nothing detects a skill record edited without its vault source, and --refresh silently reverts it


mlorentedev:
## What

A skill has two copies: the vault source at `00_meta/skills/<name>/SKILL.md` and the committed record
at `harness/skills/<name>/SKILL.md`. `compile-harness.sh --refresh` regenerates the record **from the
vault**. Nothing verifies the two agree.

The consequence is directional and silent: **a change made only to the repo record is reverted by the
next `--refresh`**, with no error and no diff to notice, because regenerating is exactly what refresh
is for. The PR passes CI, the change merges, and it disappears at the next sync.

`--refresh` is also the only path that reads the vault (`compile-harness.sh:725-734`, which errors if
the vault skills dir is absent). `--check` is offline by design and reported `OK: no harness drift`
against a record whose vault source it never opened, so it cannot close this gap as written.

## How it surfaced

PR-Agent raised it on #1543 as *"Vault Source Not Updated"* — that the diff touched only
`harness/skills/`. Wrong as a fact (the vault is a separate repository, so its commit cannot appear
in a dotfiles diff; it was pushed and verified on `origin/master` by content), but right about the
hazard: a reviewer looking at a dotfiles PR **cannot tell** whether the vault half was done. Neither
can CI. Today the only thing asserting parity is a manual verification command inside one spec.

## Why CI cannot simply be extended

The vault is a different repository and is not checked out in CI, so a naive "diff the two copies"
job cannot run there. That constraint is the interesting part of this ticket, not an excuse to skip
it — plausible shapes, to be chosen during design:

- Record a content digest of the vault source in the committed record's frontmatter at `--refresh`
  time, and have `--check` verify the record still matches the digest it was generated from. Offline,
  and it catches a record edited by hand.
- A pre-push or `dotf doctor` check that runs where the vault *does* exist, so drift is caught on the
  machine that can see both.

## Acceptance criteria

1. A record edited by hand, without the corresponding vault change, is detected — by CI, `dotf doctor`
   or `--check`, and the ticket states which and why.
2. The detection names the skill and the direction of the drift, not just "drift".
3. A committed test plants a hand-edited record and asserts the check goes red; a green check on a
   planted violation is the failure mode to avoid.

## Out of scope

Making CI able to read the vault, and any change to which copy is authoritative — the vault stays the
source, the record stays the record.

PR #1549, opened 2026-09-06
fix(harness): qualify the Major severity definition by reality, and make f7 assert

mlorentedev:
Refs #1533. Follow-up to #1543, which merged three minutes after CodeRabbit posted its first
non-notice review of the session and before those findings were dispositioned. All three are
applied here; each was confirmed against the code before applying, not taken on the reviewer's word.

## 1. The Major severity definition contradicted the verdict list (Major)

Root 2 of HARNESS-111 fixed the *verdict list* so a THEORETICAL Major routes to PASS WITH GAPS
instead of FAIL. The **severity definition** one section above it is a third statement of the same
rule, and it had drifted the same way — #1543 left it reading *"fix or spec update required, and a
re-review after it"* for every Major, unqualified. The routing paragraph #1543 added already scoped
the cost to a **REAL** Major; the definition above it did not.

A reviewer could follow either. That is exactly the failure mode HARNESS-111 exists to remove, one
section earlier than where it was found.

`guard-review-verdict-honours-reality.bats` could not see it: its six assertions only read the
verdict list. It now reads both halves — 7/7, and reverting the line turns assertion 6 red.

## 2. f7's verification command printed instead of asserting (Major)

```
for f in <two SKILL.md copies>; do grep -cE '<defective forms>' "$f"; grep -cE '<routing rules>' "$f"; done
```

`grep -c` prints the count and the loop's exit status is the *last* grep's, so the "zero defective
forms" half was never asserted — a later successful routing-rule match produced rc=0 regardless.
This is the class `.claude/CLAUDE.md` documents under *"a verification that reads a pipeline's exit
status is exactly where this bites"*, sitting in the evidence command for an acceptance criterion.

It now tests both counts per file and exits non-zero on either. Mutation-proven both ways:
reinstating `next steps (before archive)` → rc=1; deleting a routing-rule line → rc=1; clean → rc=0.

## 3. The lesson misdescribed `review: waived` (Minor)

Lesson 275 said the waiver *"asserts no review happened"*. It asserts nothing:
`checkReviewGate` (`cli/internal/spec/review.go:281-287`) returns `nil` on a waiver carrying a reason
**before** it calls `FindReview`, so a passing `review.md` in the folder is never read. Same wrong
exit, accurate mechanism.

## Verification

```
scripts/compile-harness.sh --check                      # rc=0, no harness drift
bats tests/guard-review-verdict-honours-reality.bats    # 7/7, mutation-proven
cd cli && go build ./... && go vet ./... && GOOS=windows go vet ./... && go test ./...   # all ok
<f7's new command>                                      # rc=0, defective=0 routing=4 in both copies
```

Both SKILL.md copies carry finding 1 — the committed record under `harness/skills/` and the vault
source at `00_meta/skills/` — because nothing yet detects the two diverging (#1545).

## Note on the contract set

This edits `features.json` and `tasks.md`, which are the contract set the archive gate measures.
That is legitimate *only* because HARNESS-111 has no review yet: the review runs after this merges.
Once it does, these three files are closed and further findings go to `verification.md`. That
ordering is the rule #1543 established, applied to itself.

### DIFF
--- a/docs/lessons/lesson-275-a-review-that-demands-contract-edits-invalidates-itself.md
+++ b/docs/lessons/lesson-275-a-review-that-demands-contract-edits-invalidates-itself.md
@@ -54,8 +54,10 @@ the spec archives, or the archived artifact is internally inconsistent. Editing 
 to correct a stale number creates a worse inconsistency than the one it removes.
 
 Two other exits were also wrong: `--force-without-review` discards a review that exists and passed,
-and `review: waived` asserts no review happened. Both are lies in the opposite direction from the
-stale number.
+and `review: waived` declares the requirement does not apply at all — `checkReviewGate` returns on a
+waiver with a reason *before* it calls `FindReview`, so a passing review sitting in the folder is
+never even read. Both walk past a verdict that was earned, in the opposite direction from the stale
+number.
 
 ## What to do instead
 
--- a/harness/skills/adversarial-review/SKILL.md
+++ b/harness/skills/adversarial-review/SKILL.md
@@ -150,7 +150,7 @@ For each acceptance criterion / scenario:
 Classify each finding:
 
 - **Blocker**: incorrect behavior, security/privacy issue, or spec violation that should stop archive.
-- **Major**: likely bug or significant gap; fix or spec update required, and a re-review after it.
+- **Major**: likely bug or significant gap. **Reality decides what it costs**: a REAL Major requires a fix and a re-review after it; a THEORETICAL or SPECULATIVE one is tracked with a disposition and can stand under PASS WITH GAPS. Read this with the reality axis below — severity alone never settles it.
 - **Minor**: clarity, maintainability, or low-risk gap; can follow up.
 - **Question / assumption**: needs human or author confirmation.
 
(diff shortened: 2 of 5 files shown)


mlorentedev:
## Review triage

|---|---|---|---|
| 1 | CodeRabbit | **"Review limit reached — next included review available in 43 minutes."** | **Not a finding — a notice that no review ran.** Disclosed, not counted as a pass. This PR merges unreviewed by CodeRabbit. |
| 2 | PR-Agent | #1533 "partially compliant" — `base_sha` resolution, base in the reviewer prompt, empty-diff test, backward compatibility "not addressed" | **Declined.** All four landed in #1535 and are `state: verified` in `features.json` (f1–f4). This is a follow-up PR to #1543, not the spec's implementation; asking it to re-implement roots 1 and 2 misreads a follow-up as the whole feature. Same class as the "#1543 as a ticket" misread PR-Agent produced on #1535 and #1531. |
| 3 | PR-Agent | No security concerns, no major issues | Noted. |

**Disclosure, not a claim of review.** CodeRabbit produced only rate-limit notices on this PR, so its content was never read by it. It did review the parent #1543 — three findings, all real — and those three are exactly what this PR applies. Merging with that stated rather than implied.

Not chasing a clean `dotf pr triage-queue` exit afterwards: posting this comment re-triggers the workflows, which re-runs PR-Agent and re-queues the PR (#906, measured 5x today). One post, then merge.

PR #1555, opened 2026-09-06
docs(lessons): a diagnosis repeated often enough reads as evidence

mlorentedev:
Refs #1409, #1469.

Documentation only — one lesson file and its index row. No code, no guards, no spec.

## Why

`#1409: bats red locally, green in CI: fixture isolation, not a regression` was carried across roughly **ten** archived handoffs until it sat in `MEMORY.md` — the always-loaded index — phrased as settled fact.

It was never bisected. I repeated it to my user as the known noise before verifying anything, which is what every session before me had done, because trusting the handoff is what a handoff is *for*.

The real cause took ten minutes to find and thirty seconds to confirm:

```
failures WITH    the dev dotf on PATH: 6
failures WITHOUT the dev dotf on PATH: 0
```

`install-dotf.sh` resolves the ambient `dotf` off `PATH`; a source build reports `dev`; the installer returns 0 **without installing**. The tests then fail on `[ -x "$DEST/dotf" ]` rather than on the status, which is why it looked like fixture damage.

And the test's own guard could not have worked. `VERSION="9.9.9"` targets the **version** gate, which sits four lines *after* the `dev` gate returns — not incomplete, **unreachable** in exactly the failing case.

## The rule

**A handoff claim that names a cause carries the command that established it, or is marked unverified.** The marked-unverified form is not the weaker one — it is the one that gets fixed, because it advertises the work nobody did. The unmarked form is the one that survives ten sessions.

## Scope note

The `install-dotf.sh` **fix** is not here — another session owns it (the `_dotf_current_version()` seam, a test asserting the `dev` branch fires and leaves `DEST` untouched, and the unread pipeline status on line 102). This PR is only the lesson.

Kept separate from **#1469** deliberately: fixing #1469 would make the six tests pass *by accident* while leaving the `dev` branch untested — turning a visible failure into an invisible one.

## A note on the sibling error

The lesson records a second instance from the same night, included with that session's explicit permission: it concluded "no guard exists" from a grep that never searched `tests/`, and retracted publicly. Same family — a non-observation treated as an observation.

Naming it is the point rather than the blame: in both cases **the checking step is what manufactured the confidence.** A check that cannot fail is worse than no check, because no check leaves you uncertain and a vacuous one leaves you sure.

### DIFF
--- a/docs/lessons/_index.md
+++ b/docs/lessons/_index.md
@@ -293,5 +293,6 @@ tags: [lessons, index, dotfiles]
 | [273 - The fix and the detector can be the same edit](lesson-273-the-fix-and-the-detector-can-be-the-same-edit.md) | 2026-09-05 |  |
 | [274 - A constant test double cannot reach a check that exists for change](lesson-274-a-constant-test-double-cannot-reach-a-check-that-exists-for-change.md) | 2026-09-05 |  |
 | [275 - A review that demands contract edits invalidates itself](lesson-275-a-review-that-demands-contract-edits-invalidates-itself.md) | 2026-09-05 |  |
+| [277 - A diagnosis repeated often enough becomes indistinguishable from evidence](lesson-277-a-diagnosis-repeated-often-enough-reads-as-evidence.md) | 2026-09-06 |  |
 
 - [2026-08-30 Intercepting Cobra Errors without Breaking SilenceErrors](2026-08-30-cobra-silence-errors-interception.md)
--- a/docs/lessons/lesson-277-a-diagnosis-repeated-often-enough-reads-as-evidence.md
+++ b/docs/lessons/lesson-277-a-diagnosis-repeated-often-enough-reads-as-evidence.md
@@ -1,0 +1,150 @@
+---
+id: lesson-277
+type: lesson
+status: active
+created: "2026-09-06"
+owner: manu
+tags: [lesson, handoff, diagnosis, testing, evidence, session-continuity]
+---
+
+# 277 — A diagnosis repeated often enough becomes indistinguishable from evidence
+
+## What happened
+
+`tests/install-dotf.bats` has been red locally and green in CI for months. A
+session at some point wrote this into its handoff:
+
+> **#1409 (BUG-771)** — bats red on clean `main` locally, green in CI: fixture
+> isolation, not a regression.
+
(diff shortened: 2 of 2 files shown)


mlorentedev:
## Review triage

All 12 checks pass.

|---|---|---|
| **pr-agent** (`github-actions`) | Ticket-compliance analysis, effort estimate, security assessment. *"No major issues detected. No security concerns identified."* | **Yes** |
| **CodeRabbit** | *"Review skipped — Auto incremental reviews are disabled on this repository."* | **No.** A notice about the review, not a review. **This PR is unreviewed by CodeRabbit** |

### Findings

|---|---|---|---|
| 1 | pr-agent | **#1409 partially compliant** — "Fix the bats test failure / make the suite green / implement a code fix for the root cause" listed as non-compliant | **DECLINED, with a reason: this reads `Refs` as `Closes`.** The PR body says the fix is explicitly out of scope and owned by another session (the `_dotf_current_version()` seam, a test asserting the `dev` branch fires, the unread pipeline status on line 102). A documentation PR that *cites* a ticket is not claiming to close it, and pr-agent has no way to tell the two apart — it treats every referenced issue as a contract the diff must satisfy |
| 2 | pr-agent | **#1469 partially compliant** — "implement the drift check in `dotf doctor`" listed as non-compliant | **DECLINED, same reason**, and #1469 is referenced here precisely to argue it should *not* be conflated with #1409: fixing #1469 would make the six tests pass **by accident** while leaving the `dev` branch untested |
| 3 | pr-agent | *"No relevant tests"* | **Correct and expected.** One lesson file and one index row. The guards that do apply were run: `guard-lesson-numbers-unique.bats` 4/4 and `check-md-escapes.bats` 13/13 |
| 4 | pr-agent | No major issues, no security concerns, effort 2/5 | **Noted.** No action |

### Note on the recurring shape

pr-agent's two findings here are a third instance of the pattern this very lesson documents in its corollary: **a correct mechanism producing a confident conclusion its inputs cannot support.** It is right that #1409 is unfixed; it is wrong that this PR claims to fix it. Nothing in a `Refs` line says which.

_Posting once. Not re-running `dotf pr triage-queue` afterwards: posting under this heading re-triggers the workflows and re-queues the PR (#906)._

PR #1556, opened 2026-09-06
fix(tests): stop two bats suites reading the developer's machine

### LINKED ISSUE #1409: TEST-007: the BUG-771 copilot-skills assertion fails on clean main, so the bats suite is permanently red

mlorentedev:
Makes the local bats suite green. It has been red on clean `main` for about ten sessions, carried forward in handoffs as *"fixture isolation, not a regression"* — a phrase repeated often enough that it read like a measurement. It is two tests reaching outside their fixture for a binary, and both bisect cleanly.

## TEST-007 (#1409) — the copilot assertion never made the absence true

`tests/skills-pipeline.bats` asserts copilot skills are **not** deployed when the copilot binary is absent. Its comment:

> `# No stub_copilot here -- this PATH has no copilot on it (true of this test suite's own environment already, ...)`

That is an assumption about the developer's machine, written down as a fact. It is false on any box that installed Copilot — here it arrives via nvm — so `requires_command` correctly declines to skip, the skills deploy, and the assertion fails. **The test was asserting that the gate fires on absence, against a PATH that still carried the binary: the opposite of its intent.**

```
failures in skills-pipeline.bats with copilot on PATH: 1
failures with the copilot directories removed:         0
```

`path_without_copilot()` now filters the entries holding it — filtering rather than replacing `PATH` wholesale, so every other tool the deploy needs stays reachable on machines that do not keep them in `/usr/bin` — and the test **asserts the absence was achieved** before relying on it. A fixture that fails to isolate now fails loudly instead of quietly re-testing the present-binary path and reporting a pass.

## A second instance, found while diagnosing the first

`install_dotf` resolved the ambient `dotf` inline. A source build answers `dev`, so it took the deliberate *"leave it in place"* branch and returned **0 without installing** — and six tests failed one line later on `[ -x "$DEST/dotf" ]`, on the status check passing.

```
failures in install-dotf.bats with a dev dotf on PATH: 6
failures without it:                                   0
```

`_dotf_current_version()` is now the single seam, with the `command_exists` guard folded into it so the function has exactly one line that depends on the machine. It had two — the lookup plus an `if command_exists dotf` wrapper — and the wrapper alone was enough to make a stubbed test read the real environment again.

### The guard that looks inert and is not

Removing `command_exists dotf || return 0` produces **identical** output — empty and silent. It survived mutation for that reason. But `setup-linux.sh` sources this under `set -euo pipefail`, and there:

```
WITH the guard:    survived, v=[]
WITHOUT it:        (no output — the shell aborted at the assignment)
```

`pipefail` makes `dotf version | grep | head` fail at the first element when dotf is absent, and `set -e` aborts setup — **on a fresh machine, the one box where `dotf` is guaranteed absent**. It now has a test that runs under those flags, which is what turns that surviving mutation into a caught one.

### Also covered for the first time: the `dev` branch itself

It sits ahead of the version check, so the fixture's `VERSION="9.9.9"` — commented *"no real dotf reports this -> idempotence stays off"* — could never disable it. The file assumed one gate and there were two in series. A deliberate, correct behaviour with no test read as six broken tests.

## Verification

Run on a machine carrying **both** a source-built `dotf` and Copilot, which is the configuration that broke it:

| | |
|---|---|
| `bats tests/*.bats` | **1572 passed, 0 failed**, exit 0 |
| same, with `dotf` removed from PATH | 0 failed |
| same, with `copilot` removed from PATH | 0 failed |
| `shellcheck --severity=error scripts/*.sh setup-linux.sh` | exit 0 |
| `bash -n` / `zsh -n` | clean |

**Five mutations, all caught:** delete the `dev` short-circuit; drop the `2>&1` (old binaries answer on stderr); remove `dev` from the regex; remove the absence guard (caught only by the new `set -euo pipefail` test); and revert the copilot test to assuming the absence.

## Why the tests are not just `PATH=`-cleared in `setup()`

That turns both suites green in one line and **hides** the branches that bit — the `dev` short-circuit and the `requires_command` gate — which is how a correct behaviour came to look like broken tests in the first place. The seam plus an explicit absence assertion covers them instead.

## SDD skip rationale

The Discipline Gate counts 84 LOC. The behavioural delta is three lines: extract the lookup into a function, call it, and fold `command_exists` inside. Everything else is comments recording the diagnosis, the dedent that follows removing an `if` wrapper, and test coverage for branches that had none. Measured on the production file: 26 changed non-comment lines, of which 14 are the same statements re-indented.

This is what `AGENTS.md` skips as a bug fix with an obvious cause and a mechanical refactor; it exceeds the line threshold on comment volume, not on design surface. No new contract, no new dependency, no public interface change — `install_dotf`'s behaviour is identical in every case, which is the point.

Closes #1409


mlorentedev:
## Review triage

|---|---|---|
| PR-Agent | **No major issues detected**, no security concerns, tests present; #1409 ticket compliance clean on all three points (test fixed, PATH no longer assumed, absence explicitly asserted) | Nothing to apply. |
| CodeRabbit | *"Review limit reached"* | **Not a review.** A notice about the review is not a review — disclosed rather than counted, so this PR carries one substantive reviewer, not two. |
| `spec-gate` | Passed after the `skip-sdd` label; rationale is in the PR body under `## SDD skip rationale` | The escape is used deliberately and the reasoning is measured, not asserted: 26 changed non-comment lines in the production file, 14 of them the same statements re-indented after removing an `if` wrapper. Overrule it and I will write a spec. |
| CI | `lint`, `lint-powershell`, `spec-gate`, `attestation`, `changes`, `test`, `integration`, `review`, `goreleaser` green; `test-windows` still running at the time of writing and must be green before merge |  |

### One caveat the reviewer could not see, stated rather than left implicit

The strongest evidence for this change is a **local** measurement, and CI structurally cannot reproduce it: the failures only appear on a machine that has a source-built `dotf` and Copilot installed, which no runner does. That is the whole defect — the suite was green in the one environment nobody reads and red in the one developers use.

So the check that matters here is not the green tick above; it is:

```
bats tests/*.bats            1572 passed, 0 failed, exit 0
  with dotf removed from PATH        0 failed
  with copilot removed from PATH     0 failed
```

run on the box that carries both. Five mutations are caught, including the one that only fails under `set -euo pipefail` — the flags `setup-linux.sh` actually runs with, and the reason the absence guard is load-bearing despite producing identical output without them.

mlorentedev:
## Review triage

Second round, dispositioning everything that arrived after the 10:16 table. Two findings
applied in `e22656e` and `4e4447e`; the rest declined with reasons.

|---|---|---|---|
| 1 | CodeRabbit (edited in place 00:01:27), Merge Risk 🟡 | "The installer may incorrectly preserve an invalid dotf installation." | **Applied — `e22656e`.** The conclusion is right and the stated mechanism is not, so the two are worth separating. An *empty* `_dotf_current_version` was never the problem: empty matches neither `dev` nor the pinned version, falls through, and installs. The real path is the `dev` gate. `dotf version 2>&1` merges stderr deliberately (BUG-070/#915 — old binaries answer the version there), and `dev` is a three-letter substring of ordinary paths, so a binary that cannot run prints `dotf: /home/dev/x: not found`, that matches, and the installer logs "source build; leaving it in place" and preserves an unusable binary on every subsequent setup. Reproduced on this branch before fixing. Now gated on `dotf version`'s exit status: non-zero means "nothing recognisable installed", which converges. |
| 2 | Found while verifying #1, same seam | A `dotf` that runs but prints nothing version-shaped aborts `setup-linux.sh`. | **Applied — `e22656e`.** `grep` matching nothing exits 1, `pipefail` promotes that to the pipeline's status, and `set -e` kills the caller at `_dotf_current="$(...)"` — and setup sources this under exactly those flags. Same class as the `command_exists` guard this PR already documents as load-bearing: that one covers the *absent* binary, and nothing covered the *present, unusable* one. The existing "unrecognisable output yields empty" test could not see it — it asserts the returned VALUE and runs without the flags, while production depends on the PROCESS surviving with them. Measured before the fix: exit 1, `survived:` never printed. |
| 3 | CodeRabbit, Merge Risk 🟡 | "The Copilot-absence test can fail for unrelated missing tools." | **Declined, reasoning recorded.** True in principle: dropping a PATH entry to hide `copilot` drops everything else in that directory. Two things make it the right trade. First, the failure direction is loud, not silent — the test asserts the deploy exits 0, so a missing tool turns it red rather than quietly green, which is the direction this repo requires of anything that reports absence. Second, it does not bite: `compile-harness.sh` shells out to `awk cat cp find git go grep head jq mkdir mv printf rm sed sort tr wc`, while the directory holding `copilot` here (`~/.nvm/versions/node/<v>/bin`) holds `bw copilot corepack hyperframes node npm npx obsidian opencode pi yarn yarnpkg` — no intersection. Filtering per-directory instead of replacing PATH wholesale is already the mitigation for the general case. Not ticketed: a hypothetical whose failure mode is loud does not justify another open issue. |
| 4 | CodeRabbit, Out of Scope Changes ⚠️ | Split the `install-dotf` changes out, or link an issue that covers them. | **Declined; already declared in the PR body.** Diagnosing #1409 surfaced a second instance of the same defect class — a bats suite resolving a binary from the developer's machine — in `install-dotf.bats`. They are one fix. Splitting them would land a test-only change whose subject is a bug fixed in the other half. |
| 5 | pr-agent, Standalone PR Review 23:50:44 | #1409 "partially compliant": no mutation test included in the PR. | **Declined — category error.** Mutation testing is a verification act performed *against* a branch, not code that ships; committing a deliberately broken variant is the opposite of the intent. Results are recorded instead. For the two changes above: dropping the `\|\| :` turns the survival assertion red, swallowing the exit status turns the `dev` assertion red, and disabling the PATH filter turns the copilot-absence assertion red (naming the binary it found). Each half of each fix is defended independently. |
| 6 | CI, `test` job FAILURE (job 101579967349) | `not ok 535 guard: no two lesson files share a number`, `not ok 536 guard: every lesson file is listed in _index.md`. | **Not caused by this diff.** `main` carries two `lesson-276-*` and two `lesson-277-*`, and the "Update branch" merge at 23:49:20Z (`d44eda2`) brought them onto this branch. This PR touches no file under `docs/lessons/`. Fixed by #1568 on `main`; this branch then needs one more update to pick the renumber up. A CI re-run alone will not clear it — it replays the same tree. |

**Verification, this session.** `bats tests/*.bats` before the "Update branch" merge:
**1573 pass, 0 fail**. On the current tree, after it: **1571 pass, 2 fail** — and the two are
`not ok 535` and `not ok 536`, row 6's imported guards, with nothing else changed. That is the
measurement behind the claim that row 6 is the whole of the red. Also `shellcheck` clean on
`scripts/install-dotf.sh`, `bash -n` and `zsh -n` clean, and `bats tests/install-dotf.bats` 13/13.

One judgement call worth surfacing rather than burying: a `dotf` on PATH that cannot run is now
**replaced**, not reported. Converging is consistent with what the function's contract already
said, and replacing a broken binary is the repair — but it does mean the installer will overwrite
something the machine's owner may have put there deliberately. Aborting loudly instead would be
defensible; it is a different decision, not an oversight.

mlorentedev:
## Review triage

Amendment. The table above was written against the 23:50 review; pr-agent re-reviewed at
03:25:48, two minutes before I posted it. The queue counted the PR as triaged because my
comment is newer — a proxy that happened to pass on a table which predates the content it
appears to dispose of. Correcting that here rather than leaving the newest review unread.

|---|---|---|---|
| 7 | pr-agent, Standalone PR Review 03:25:48 | #1409 now reads **"PR Code Verified"** (was "partially compliant"), naming the `run !` change and the explicit absence assertion. | **Noted.** Supersedes row 5; nothing to apply. |
| 8 | pr-agent, same review — "Potential Silent Failure" | `path_without_copilot` splits `$PATH` on `IFS=":"`; consecutive or trailing colons "may drop empty entries or produce a malformed PATH", which "could lead to other tools being unavailable in the test, potentially masking unrelated issues". | **Declined — one half is false, the other is intended.** *Malformed:* it cannot happen. The accumulator is `${_pwc_out:+$_pwc_out:}$_pwc_dir`, which only emits a separator when there is already something to separate. Exercised against `/usr/bin::/bin`, `:/usr/bin:/bin`, `/usr/bin:/bin:` and `/usr/bin:::/bin::` — every one returns exactly `/usr/bin:/bin`, with no leading, trailing or doubled colon. *Dropped empty entries:* real, and deliberate. A POSIX empty `PATH` entry means the current directory, so the function removes an implicit `.` from the fixture's PATH. That is the safer direction, not a regression. And the consequence claimed for it does not follow: the test asserts the deploy exits 0, so a genuinely missing tool turns it **red**, never silently green — the same reason row 3 was declined. |

One thing the review did not raise that a reader applying this repo's own prohibited-pattern
table would: `for _pwc_dir in $PATH` is unquoted parameter splitting, which **zsh does not
perform** — measured just now, 1 field under zsh against 2 under bash. It is not a live defect,
because bats is `#!/usr/bin/env bash` and this helper only ever runs there. Recording it so that
moving the helper into a `.sh` file later does not silently reintroduce it.

Also worth noting for anyone reading the checks: CodeRabbit's comment carries
`coveredCommitId: d44eda2b...`, which is the "Update branch" merge — it has **not** seen
`e22656e` or `4e4447e`. Its "Merge Risk 🟡 / installer may preserve an invalid dotf
installation" line is the pre-fix state, already dispositioned as rows 1 and 3 above, not a
fresh finding against the current head.

PR #1557, opened 2026-09-06
fix(cli): count a reviewer's in-place edit as new output in the triage queue

### LINKED ISSUE #1422: TOOL-019: dotf pr triage-queue compares timestamps, so a reviewer that edits its comment in place leaves findings behind a green queue

mlorentedev:
`dotf pr triage-queue` compared `createdAt`. Both reviewers here re-review by **growing an existing comment**, and `createdAt` does not move when they do — so a reviewer that re-reviewed after a push stayed behind the last triage and the queue reported exit 0 with findings unread.

That is the inverse of the hole the doctrine names. It says a non-zero exit is *"a queue you must read, never an empty one"*; this was a **zero exit that was not an empty queue**.

## Measured, on this repo, last night

```
#1543  coderabbitai   created 02:52:56   updated 03:28:29
```

The review that arrived in that edit carried three real findings. The PR merged at 03:28:52 — twenty-three seconds later.

## The ticket's blocker no longer exists

TOOL-019 concluded the fix was *"not reachable through the current data source"* and framed the ticket as a decision between **GraphQL** and a **content digest in the triage comment**. Neither is needed, and the reason is worth recording because it is not a disagreement with that analysis — it was correct when written.

It was true of `gh pr view --json comments`, which exposes `createdAt` and a timestamp-less `includesCreatedEdit` boolean and nothing else. But **CLI-071 moved this package to the REST API for unrelated reasons (#1454)**, and REST has always carried `updated_at`. The blocker was dissolved by a change that was not trying to dissolve it, and nobody revisited the ticket.

So: no GraphQL — the surface that hit a rate limit on 2026-08-31 while `gh api rate_limit` still read 5000/5000 — and no change to the triage comment format, which the reviewer registry declares and a second consumer reads.

## Two decisions inside the fix

**`Spoken()` is `max(CreatedAt, UpdatedAt)`, not `UpdatedAt`.** An unedited comment carries a zero `UpdatedAt`; reading that as "spoken at the epoch" would answer *nothing pending* for the entire queue. That is the one direction this package must never fail in, and every fixture predating the field would have triggered it.

**The asymmetry with `lastTriage` is deliberate, and now asserted rather than only commented.**

| | uses | if we get it wrong |
|---|---|---|
| reviewer output | `Spoken()` | missing an edit hides findings behind a green queue |
| triage record | `CreatedAt` | counting an edit lets a typo fix silently clear a queue, and hands anyone a one-keystroke dismissal with no reading |

Edits make the queue **louder, never quieter**. Both errors are possible; only one is safe.

## Verification

`go build`, `go vet`, `go test ./...`, `GOOS=windows go vet`, `GOOS=darwin go vet` all rc=0; `golangci-lint run` **0 issues** on the pinned 2.12.2.

**Four mutations, all caught**, including both subtle ones:

| mutation | caught by |
|---|---|
| `Spoken()` ignores edits (the original bug) | the in-place-edit table row |
| `Spoken()` returns `UpdatedAt` alone | the zero-`UpdatedAt` row |
| `lastTriage` counts triage edits too | the asymmetry row |
| the fetch layer drops the `updated_at` mapping | `TestFetchCarriesUpdatedAtFromTheWire` |

That last test exists because the domain tests construct `Comment` values directly: the field could be missing from the REST struct or unmapped in the conversion and every one of them would still pass. A stub that is the only thing ever tested proves nothing about the code it stands in for.

## Found while verifying this, and NOT fixed here

The queue is **structurally blind to CodeRabbit's substantive reviews**, which is a separate and arguably worse defect than the one this PR fixes. Its declared marker in `harness/review-attestation.json` is `"No actionable comments were generated in the recent review"` — the phrase for an **empty** review. Its real review on #1543 is 8,848 bytes, uses `## Walkthrough` and `> [!IMPORTANT]` callouts, and **does not contain that string**:

```
declared marker present in the substantive review: 0
```

So the queue never saw it, with or without this fix. That is filed separately rather than fixed here because the registry has two consumers — the attestation gate reads the same file — so changing a marker has blast radius this PR should not carry.

## SDD skip rationale

The Discipline Gate counts 59 LOC. The logic delta is 19 lines and most of those are mechanical: two struct fields, one field added to a conversion, a five-line `Spoken()`, and six call sites swapping `CreatedAt` for `Spoken()`. Everything else is comments recording the measurement and why each direction was chosen, plus test rows.

No new contract, no new dependency, no new API surface — the point of the fix is that it needs none. This is what `AGENTS.md` skips as a bug fix with an obvious cause; it exceeds the threshold on comment volume, not design surface.

**Note for the reviewer: this is my second consecutive `skip-sdd`** (the other is #1556). Both are genuine, and both trip the gate the same way — a small fix documented in this repo's own dense-commentary style. If the pattern is unwelcome, say so and I will write specs for both rather than keep taking the escape.

Closes #1422


mlorentedev:
Cross-reference: the second half of this defect is now **#1558**.

`dotf pr triage-queue` and the attestation gate share `harness/review-attestation.json`, and CodeRabbit's declared marker there is `"No actionable comments were generated in the recent review"` — the phrase for an **empty** review. Its substantive review on #1543 is 8,848 bytes, uses `## Walkthrough`, and does not contain that string, so neither consumer ever classified it as reviewer output.

This PR fixes the timestamp half. **For CodeRabbit specifically, that is necessary and not sufficient**: #1422 makes the queue miss re-reviews; #1558 makes it miss every substantive review, the first one included. On #1543 both fired at once.

Kept out of this PR deliberately — the registry has two consumers with **opposite safe directions**. Treating any CodeRabbit comment as review output makes the queue louder (safe for a queue) and makes the attestation gate credit a rate-limit notice as a review (unsafe for a gate). That needs deciding, not sliding in behind a timestamp fix.

mlorentedev:
## Review triage

|---|---|---|
| PR-Agent | **No major issues detected**, no security concerns | Nothing to apply. |
| CodeRabbit | *"Review limit reached"* | **Not a review.** Disclosed, not counted. |
| CI | all green except `test-windows`, still running at the time of writing and required before merge | |

### The disclosure above is exactly what #1558 makes unverifiable, and this PR is the wrong place to fix it

I classified CodeRabbit's comment by reading its body, which is what the doctrine requires. But `dotf pr triage-queue` would have reported the same thing whether that body was a notice or a full review with findings — its declared marker matches only the empty-review phrase. So the sentence above is true and the method behind it is not load-bearing.

Recording it here rather than quietly writing the usual line, because it is the same defect class this PR is about: a check that cannot fail leaves you sure rather than uncertain.

**Verification re-run at this head:** `go build`, `go vet`, `go test ./...`, `GOOS=windows go vet`, `GOOS=darwin go vet` all rc=0; `golangci-lint run` 0 issues on the pinned 2.12.2; four mutations caught.

mlorentedev:
## Review triage

Second round, dispositioning everything that arrived after the 17:49 table.

|---|---|---|---|
| 1 | CodeRabbit, comment edited in place 23:56:05 | **No review ran.** The body now reads "Review limit reached — next included review available in 53 minutes." | **Disclosed, not applied.** A notice *about* a review is not a review. The Go changes here have been reviewed by pr-agent only. Recording that explicitly, because green checks would otherwise imply a review that never happened. |
| 2 | pr-agent, Standalone PR Review 23:57:19 | No major issues, no security concerns, tests present; ticket-compliance section empty. | **Noted, nothing to apply.** |
| 3 | CI, `test` job FAILURE (job 101580807102) | `not ok 535 guard: no two lesson files share a number` and `not ok 536 guard: every lesson file is listed in _index.md`. | **Not caused by this diff.** `main` carries two `lesson-276-*` and two `lesson-277-*`, and the "Update branch" merge at 23:55:56Z (`c6de9dd`) brought them onto this branch, so the guard now fails against the branch itself rather than only against a computed merge commit. This PR touches four files, all under `cli/internal/prtriage/`. Fixed by #1568 on `main`; this branch then needs one more update to pick the renumber up. A CI re-run alone will not clear it — it replays the same tree. |

One note worth recording, because it is this PR's own subject matter. Row 1 arrived as an
**edit to an existing comment**: `created_at` stayed at 17:40:38 while `updated_at` moved to
23:56:05. The queue as it stands on `main` compares `created_at` only, so it did not surface
that edit at all — it flagged this PR on row 2's *new* comment, which happened to land a minute
later. Had pr-agent stayed quiet, the queue would have reported this PR clear while it carried
unread reviewer output. That is exactly the failure #1422 describes and this diff fixes.

No changes to the diff. Re-verified on the post-merge tree, this session:
`go build ./...`, `go vet ./...` and `GOOS=windows go vet ./...` clean, and
`go test -count=1 ./...` green across the module (`internal/prtriage` uncached, 0.066s).

mlorentedev:
## Review triage

Third round, after merging `main` (`961b8de`) now that #1568 has landed.

|---|---|---|---|
| 1 | CodeRabbit, edited in place 17:06:36 (`coveredCommitId: 2a01d6aa`) | **It actually reviewed this time** — Merge Risk 🔵 Low, 5/5 pre-merge checks passed. Its one note: "the REST fixture should also cover unedited timestamps to protect normal queue behavior… a bounded test-coverage risk". | **Declined — already covered, and not vacuously.** `cli/internal/prtriage/fetch_test.go:100-136` is exactly that fixture: a REST comment payload carrying `created_at` and **no** `updated_at`, driven through the real fetch path, asserting `At` equals the creation time. Proven live rather than asserted: mutating `Spoken()` to `return c.UpdatedAt` (dropping the max) fails that test at `fetch_test.go:134` with `At = 0001-01-01`, alongside `TestEvaluate/a_never-edited_comment_(zero_UpdatedAt)_still_counts_as_spoken`. The unedited path is the one this package must never get wrong, so it was covered first. |
| 2 | pr-agent, Standalone PR Review 17:04:35 | `#1422` **Partially compliant** — but with *zero* non-compliant requirements and zero items needing human verification. The "non-compliant" blocks are all against **#1543, #1454 and #1556**. | **Declined — known tooling artifact, not a finding.** pr-agent reads every `Refs #N` in the body as if it were `Closes #N` and grades the diff against those tickets. This PR closes #1422 only; #1543/#1454/#1556 are cross-references. Same artifact recorded on #1555 and #1564. Read on its own terms, row 2 says #1422 is fully satisfied. |
| 3 | pr-agent, Standalone PR Review 04:00:07 | Earlier run against the pre-merge head. | **Superseded** by row 2. |
| 4 | CI | All checks settled, none failing; `mergeStateStatus: CLEAN`. | **Nothing to do.** The `test` failure dispositioned in round 2 is gone: #1568 renumbered the colliding lessons to 279/280, and merging `main` in cleared it. `bats tests/guard-lesson-numbers-unique.bats` is 4/4 locally. |

**One correction to round 2.** There I disclosed that CodeRabbit had *not* reviewed this PR
("review limit reached") and that pr-agent was the only reviewer. That was true then and is no
longer: row 1 is a real review of the current head. The disclosure is withdrawn — this PR has
been reviewed by both.

Verified on the merged tree this session: `go build ./...`, `go vet ./...` and
`GOOS=windows go vet ./...` clean; `go test -count=1 ./...` green across the module.
No changes to the diff.

Issue #1562, opened 2026-09-06
agent auto: three Minor findings from HARNESS-120's passing review, tracked not fixed


mlorentedev:
From HARNESS-120's round-2 adversarial review (`nan/deepseek-v4-flash`, PASS, `reviewed_sha` `9d3a7ee`). All three are **REAL** and all three are **Minor / low priority**. Filed rather than applied, because the review passed and applying code changes after a PASS ships something the reviewer never saw. `specs/HARNESS-120/review.md` carries the full text.

## 1. The `--role` refusal cannot tell "not declared" from "declared but not invocable"

```
$ dotf agent auto --role hermes-nan --backend dry-run ...
Error: no persona named "hermes-nan" ...
        The roster declares: architect, builder, curator, planner, reviewer, shipper
```

`hermes-nan` **is** declared. It is `kind: autonomous` — an externally scheduled steward, not a phase a session can adopt — and `personaByName` filters on `Kind == "invocable"`, matching `ResolveRoles` at `roles.go:53`.

**The behaviour is right and should not change.** The message is what is wrong: it tells an operator who knows the record exists to go looking for a missing file. Two different states are being reported with one sentence, which is the defect class #1534 removed from the gate's allow reason.

Fix: distinguish the two cases, and pin the autonomous one with its own assertion. `TestAgentAuto_RefusesAPersonaTheRosterDoesNotDeclare` currently covers only the genuinely-absent path (`potato`); the `kind: autonomous` message is **UNTESTED**.

## 2. `resolvePersonaForTask` is ~48 lines against the repo's 40-line rule

`cli/internal/cmd/agent_auto.go`. Cyclomatic complexity is 9, within limits — the inflation is multi-line `fmt.Errorf` prose, not control flow.

Worth noting for whoever picks it up: the 40-line rule lives in `AGENTS.md` and **golangci does not enforce it**, so `golangci-lint run` reporting `0 issues` says nothing here. Extracting the two error branches is the obvious shape.

## 3. `taskDelimiter` collision is possible, and should be surfaced rather than gated

`taskDelimiter = "===== TASK ====="` in `preamble.go`. If a task text or a record body ever contained that exact string, the split point between persona and work becomes ambiguous.

[...]

PR #1564, opened 2026-09-06
docs(spec): HARNESS-120 round 2 review — PASS, and archive

mlorentedev:
Refs #1537, #1562, #1566. Follows #1546.

Spec artifacts and the archive move. No code.

## Why this exists separately

**#1546 merged while round 2 of the review was still running.** So `main` currently carries round 1's **FAIL** `review.md` and none of the round-2 artifacts. Anyone reading `specs/HARNESS-120/` in `main` today sees a failed review against a change that shipped, which is the opposite of what happened.

## What round 2 says

**PASS** — `nan/deepseek-v4-flash`, against `reviewed_sha` `9d3a7ee`.

A **different pool draw** from round 1's `agy/gemini-3.1-pro-high`. The launcher draws at random, so these are two independent reviewers, not one model reconsidering its own verdict.

Round 2 did its own work rather than reading mine:
- reproduced AC1–AC6 at the CLI against the shipped roster and `model-map.json`
- rebuilt and re-ran the Go layer, including `GOOS=windows go vet` and `golangci-lint` at the `versions.conf` pin
- **mutation-checked two of the branch's own paths** — making the ambiguous case return its first candidate, and sending the bare task instead of the composed preamble — confirming both tests fail when the behaviour is removed

Rubric: Verification **A**, Reliability **A**, Handoff-readiness **A**; Correctness B, Scope B, Maintainability B. No Blocker, no REAL Major. Its verdict states archiving is advisable, so this PR also performs the archive.

Round 1's two findings are both closed — the AC7 Blocker by a real dispatch (`claude:sonnet` off `chains.mid`, 12.7 s, `tier_from: inferred`), the Minor by the `ResolveChain` fix.

## Four Minors: one applied, three ticketed

**Applied** — AC1's quoted output was reformatted and missing `duration_ms` and `truncated`, while AC7 was quoted verbatim. Re-captured raw. Slightly worse than cosmetic: a reader could not tell which of two quotes had been transcribed.

**Ticketed as #1562** — the `--role hermes-nan` message (reports no such persona when the persona exists and is merely `kind: autonomous`), `resolvePersonaForTask` at ~48 lines against the 40-line rule, and a speculative `taskDelimiter` collision flagged surface-only.

They are ticketed **rather than fixed on purpose.** The review passed against a specific tree; applying code changes afterwards ships code no reviewer saw, under a PASS earned by different code. That includes the one I most wanted to fix.

## `tasks.md`'s review checkbox is deliberately left unticked

The archive gate measures staleness against exactly `proposal.md`, `tasks.md` and `features.json` — so *marking the review done is the act that invalidates it*. `docs/lessons/lesson-275-a-review-that-demands-contract-edits-invalidates-itself.md` documents this. I staged that edit before catching it, roughly an hour after reading that lesson, which is a fair measure of how natural the mistake is.

The review's own text closes the box in prose instead: *"the one unticked box is the adversarial review itself, which this run closes."*

## A defect found by predicting the wrong thing — #1566

This PR's first version warned that `dotf spec archive` would probably **refuse**, because #1546 was squash-merged and `9d3a7ee` is therefore not an ancestor of `main`.

It archived instead. The reason is worth more than the prediction was:

```go
git cat-file -e <reviewed_sha>^{commit}   // review.go:170
```

That asks whether the object **exists in this clone**, not whether it is reachable from `HEAD`. My clone still held the pre-squash objects from the deleted branch, so it passed. A fresh clone or CI would have refused with *"rewritten by a rebase?"* — when nothing was rebased.

Since this repo squash-merges, `reviewed_sha` is orphaned by the normal workflow on **every** spec, and the gate's answer depends on local garbage-collection state. Worse, the direction is inverted: permissive on the machine that just did the work and is most likely to run `archive`, restrictive everywhere else.

Filed as **#1566** with three candidate fixes rather than one, because making it strict would refuse every post-merge archive in a squash-merge repo — turning a silent inconsistency into a total block.


mlorentedev:
## Review triage

All 12 checks pass.

|---|---|---|
| **pr-agent** (`github-actions`) | Ticket-compliance analysis against #1537, #1562, #1566. *"No major issues detected. No security concerns identified."* Effort 1/5 | **Yes** |
| **CodeRabbit** | *"Reviews paused — this branch is under active development."* | **No.** A notice about the review, not a review. **This PR is unreviewed by CodeRabbit** |

Worth separating from the CI signal: this PR *carries* an independent adversarial review (`nan/deepseek-v4-flash`, **PASS**, `reviewed_sha` `9d3a7ee`) as its content. That review is the substantive one and it is not a CI reviewer. The table above is only about who reviewed **this diff**.

### Findings

|---|---|---|---|
| 1 | pr-agent | **#1537 partially compliant**, non-compliant requirements: *none* | **Noted.** The label is "partially" only because pr-agent grades every referenced ticket on the same scale; it lists nothing non-compliant. No action |
| 2 | pr-agent | **#1562 partially compliant** — "the findings are not fixed" | **DECLINED, with a reason, and the reviewer states it itself:** *"(but the ticket is for tracking, not fixing)"*. That is the whole point — the adversarial review passed against a specific tree, so applying its Minors now would ship code no reviewer saw, under a PASS earned by different code. Fixing them here would be the defect |
| 3 | pr-agent | **#1566 partially compliant** — "the bug is not fixed by this PR" | **DECLINED, same category error as #1562.** #1566 is the archive-staleness defect *discovered while doing this work* and filed from it. A PR that opens a ticket is not a PR that owes its fix — and #1566's body explicitly argues against the obvious fix, because making the check strict would refuse every post-merge archive in a squash-merge repo |
| 4 | pr-agent | *"No relevant tests"* | **Correct.** Spec artifacts and a directory move; no code. The Go layer and guards were run against the tree this archives and are recorded in `verification.md` |

### The recurring shape, third instance today

All three "non-compliant" findings are the same error: **pr-agent reads `Refs #N` as a commitment to close #N.** It has no way to distinguish a ticket a PR *cites* from one it *closes*, so every referenced issue becomes a contract the diff is graded against. The mechanism is sound and the conclusion is not supported by its input.

Filed nothing for it — it is pr-agent's own behaviour rather than this repo's, and the output is still useful once read with that in mind.

_Posting once. Not re-running `dotf pr triage-queue` afterwards: posting under this heading re-triggers the workflows and re-queues the PR (#906)._

Issue #1542, opened 2026-09-06
SDD: dotf spec init assigns an AREA-NNN another spec already owns — 9 collisions measured


mlorentedev:
## What

`dotf spec init <id>` accepts an `AREA-NNN` that another spec already owns. There is no uniqueness
check on the number: `grep -n "already exists\|collision\|IsExist" cli/internal/cmd/spec.go` returns
nothing, and creation is a plain directory write keyed on the full slug, so `HARNESS-111` and
`HARNESS-111-review-base-and-reality` are two different directories and neither command objects.

## Measured

Across 43 active and 198 archived specs, **9 AREA-NNN numbers are held by two unrelated specs**:

| Id | Spec A | Spec B |
|---|---|---|
| AI-022 | `hive-daemon-activation` | `pi-harness-slot` |
| AI-023 | `hive-auto-upgrade-timer` | `orca-ade-and-iac-doctrine` |
| CLI-025 | `dotf-mem-heal-and-session-start` | `secrets-render` |
| DOCS-013 | `agents-spec-subcommands` | `doc-path-guard` |
| HARNESS-027 | `HARNESS-027` | `adr025-hardening` |
| HARNESS-041 | `board-pickup-self-assign` | `ci-path-filtering` |
| HARNESS-111 | `HARNESS-111` (#1241) | `review-base-and-reality` (#1533) |
| MEMORY-001 | `cross-agent-session-bridge` | `mirror` |
| REFACTOR-012 | `entries-tagged-union` | `init-spec-issue-gate` |

Two exclusions, both deliberate and both evidence that the id has real meaning:

- `CLI-024-secrets-*` is **not** a collision — 18 sub-specs of one secrets epic, which is a
  legitimate use of a shared number and must keep working.
- `SDD-012` vs `SDD-012b` is **not** a collision either — it is the manual workaround. Someone hit
  this once and disambiguated by hand, which is the cheapest possible evidence that the problem is
  real and currently costs human attention.

## Why it matters

The id is the handle everything else uses. `dotf spec review <id>`, `dotf spec archive <id>` and the
spec-gate all resolve a spec by name, so a shared number makes "HARNESS-111" ambiguous in every
conversation, ticket reference and journal entry that does not spell the full slug. It also defeats
the point of an area-scoped counter: the number stops being an identifier and becomes decoration.

## Acceptance criteria

1. `dotf spec init` refuses an `AREA-NNN` already held by a spec in `specs/` **or** `specs/archive/`,
[...]

PR #1565, opened 2026-09-06
chore(specs): renumber colliding spec IDs and guard the number space

mlorentedev:
## Summary

`specs/HARNESS-111` (issue #1241) and `specs/archive/HARNESS-111-review-base-and-reality` (issue #1533) shared a number. Two sessions took the next free `HARNESS-NNN` from the same tree, neither saw the other's unpushed folder, and main accepted both. This is the lesson-number collision shape (`tests/guard-lesson-numbers-unique.bats`) one directory up.

- `HARNESS-111-review-base-and-reality` (the newer one) is renumbered to `HARNESS-112`, the next free number across `specs/` and `specs/archive/`. Every reference follows: its own frontmatter, headings and `features.json` ids, the review-base comments in `cli/internal/cmd/spec.go` and its two tests, `tests/guard-review-verdict-honours-reality.bats`, and the two mentions in the archived BUG-093 review and verification.
- Writing the guard surfaced two more pairs of the same shape, and the guard cannot be green on main with them in place, so they are renumbered here too: `HARNESS-027-adr025-hardening` (#457, beside bare `HARNESS-027`/#411) becomes `HARNESS-113`; `HARNESS-041-board-pickup-self-assign` (knowledge#140, beside active `HARNESS-041-ci-path-filtering`/#552) becomes `HARNESS-114`. Both are archived; their references were internal to their own folders. The `CHANGELOG.md` line that names HARNESS-027 with #457 is release-please history and is left alone.
- New guard `tests/guard-spec-ids-unique.bats`, next to the lessons guard, with two assertions: a bare-ID folder owns its number outright, and an active spec shares its ticket ID only with its own series. A series is folders sharing the ID and the first slug word, which is the eighteen `CLI-024-secrets-*` sub-specs; a third test asserts that exception is actually exercised. Mutations: a `specs/archive/HARNESS-112` bare folder fails the first, a `specs/HARNESS-114-something-else` fails the second.
- Six archive-only pairs with unrelated slugs (AI-022, AI-023, CLI-025, DOCS-013, MEMORY-001, REFACTOR-012) are outside what the guard asserts, because every future collision enters through an active spec and asserting over the archive would put the guard red on main until months of reviewed material is renumbered. They are recorded in #1563.
- `docs/lessons/lesson-268` gets a "Seen again" section for #1534 (the gate allow reason that was vacuously true) instead of a new lesson.

## SDD skip rationale

[...]

PR #1568, opened 2026-09-07
fix(lessons): main is red — renumber two colliding lessons to 279 and 280

mlorentedev:
Refs #1542. **`main` is currently red.** Documentation only — two renames, two `_index.md` rows, one postscript.

## The failure on `origin/main` right now

```
not ok 1 guard: no two lesson files share a number
#   lesson-276
#   lesson-277
not ok 2 guard: every lesson file is listed in _index.md
#   lesson-276-a-helper-written-to-mirror-a-parser-is-the-defect.md
#   lesson-277-a-diagnosis-repeated-often-enough-reads-as-evidence.md
```

Two files claim 276, two claim 277, and **both of mine are missing from the index.**

**The second failure is the one that matters.** Different filenames meant no conflict on the *files*, so whichever `_index.md` merged last simply won. Two lesson files are sitting in `main` with no index entry — which the guard's own message names exactly:

> the index is the only surface anything reads, so a lesson absent from it is a lesson nobody will find

A silent loss, not a noisy one. The number collision is the visible half; this is the half that would have cost something.

## The fix

`276-a-helper-written-to-mirror-a-parser` → **279**, `277-a-diagnosis-repeated-often-enough` → **280**. Filename, `id:` frontmatter, `#` heading, and the two restored index rows.

Renumbered **mine** rather than the others because the index already lists theirs — that is the coherent direction, not a claim about who was first. No file belonging to anyone else is touched.

## Nobody did anything wrong, and that is the point

**Both PRs were green when they merged.** Each was green against a `main` that did not yet contain the other. The guard fires on the *merge commit*, and a merge commit only exists once both have landed — so for two PRs merging inside the same window, **there is no moment at which either could have seen the collision.**

This is the **third and fourth instance today.** The first two were on lesson 275, between two sessions that had explicitly announced their numbers to each other in writing, ninety seconds apart — while writing up the lesson about lesson-number collisions.

Recorded as a postscript in 280 itself, because the generalisation is worth more than the incident:

> A check that runs at merge time cannot prevent a same-window collision. It can only report one afterwards. Detection and allocation are different jobs, and a guard is only the first.

Four instances in one evening, two of them while actively coordinating about the exact hazard, is not carelessness — it is what a net does when the thing it is asked to do is a lock's job. **#1542** is the same defect for `dotf spec init`'s `AREA-NNN` ids, and this should settle the shape of its AC1: allocation, not better detection.

## Verified

`guard-lesson-numbers-unique.bats` 4/4 and `check-md-escapes.bats` 13/13 on this branch.

### DIFF
--- a/docs/lessons/_index.md
+++ b/docs/lessons/_index.md
@@ -296,3 +296,5 @@ tags: [lessons, index, dotfiles]
 | [276 - Intercepting Cobra errors without breaking `SilenceErrors`](lesson-276-intercepting-cobra-errors-without-breaking-silenceerrors.md) | 2026-08-30 |  |
 | [277 - One invalid field rejects the whole file, and the deployed copy hides it until a deploy](lesson-277-one-invalid-field-rejects-the-whole-file-and-the-deployed-copy-hides-it.md) | 2026-09-05 |  |
 | [278 - A cap asserted in one unit while the payload overflowed in another](lesson-278-a-cap-asserted-in-one-unit-while-the-platform-counts-another.md) | 2026-09-05 |  |
+| [279 - A helper written to "mirror" an existing parser is the defect; the transcription bug is only how you find out](lesson-279-a-helper-written-to-mirror-a-parser-is-the-defect.md) | 2026-09-05 |  |
+| [280 - A diagnosis repeated often enough becomes indistinguishable from evidence](lesson-280-a-diagnosis-repeated-often-enough-reads-as-evidence.md) | 2026-09-06 |  |
--- a/docs/lessons/lesson-279-a-helper-written-to-mirror-a-parser-is-the-defect.md
+++ b/docs/lessons/lesson-279-a-helper-written-to-mirror-a-parser-is-the-defect.md
@@ -1,5 +1,5 @@
 ---
-id: lesson-276
+id: lesson-279
 type: lesson
 status: active
 created: "2026-09-05"
@@ -7,7 +7,7 @@ owner: manu
 tags: [lesson, go, parsing, duplication, unicode, build]
 ---
 
-# 276 — A helper written to "mirror" an existing parser is the defect; the transcription bug is only how you find out
+# 279 — A helper written to "mirror" an existing parser is the defect; the transcription bug is only how you find out
 
 ## What happened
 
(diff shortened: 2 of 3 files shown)


mlorentedev:
## Review triage

Reviewer output at `94dee5d`. **Disclosure first: CodeRabbit did not review this PR.** Its comment is
a *"Review limit reached — you've used all free OSS reviews for now"* notice, which is a message about
the review, not a review. pr-agent is the only reviewer that produced findings, so everything below
is its output.

|---|---|---|---|
| 1 | pr-agent | No major issues, no security concerns, effort 1/5 | Noted |
| 2 | pr-agent | #1542 items 1–3 non-compliant: no `dotf spec init` uniqueness check, no override flag, no tests | **Declined** — see below |
| 3 | pr-agent | Verify no other file references the old numbers `lesson-276` / `lesson-277` | **Applied** — verified, see below |
| 4 | pr-agent | "Confirm whether this is the implementation PR for #1542 or a separate CI fix" | Answered — this is the CI fix only; the spec-side work is #1565 |

### 2 — declined, with the reason

The body says `Refs #1542`, not `Closes #1542`. pr-agent's ticket-compliance section treats any
referenced number as a requirement the PR must satisfy, which is **#1466**, already filed. This PR
does not claim to implement the uniqueness check, and its own second compliance block — the one
pr-agent generated against what the PR actually says it does — reports `Non-compliant requirements:
(none)`. Declining rather than ticketing because the ticket exists.

### 3 — applied, and the result is a deliberate keep

`git grep -E 'lesson-276-a-helper|lesson-277-a-diagnosis' 94dee5d` returns exactly two hits, both in
`lesson-280`'s own postscript at lines 193–194, both inside a fenced block that quotes the guard's
failure output verbatim:

```
not ok 2 guard: every lesson file is listed in _index.md
#   lesson-276-a-helper-written-to-mirror-a-parser-is-the-defect.md
#   lesson-277-a-diagnosis-repeated-often-enough-reads-as-evidence.md
```

Those are not links and not stale references — they are the record of *what collided*, which is the
subject of the postscript. Renaming them would erase the evidence the lesson exists to carry. Nothing
else in the tree names the old numbers, so the renumber leaves no dangling reference.

The finding was worth running: this file has been renumbered twice (275 → 276 → 279), so a stale
self-reference was a live possibility rather than a theoretical one.

### State

`main` is red on `guard-lesson-numbers-unique.bats` (2/4) until this merges. At `94dee5d` all 11
completed checks are SUCCESS and none failed. Merging is the user's call.

PR #1585, opened 2026-09-08
docs(lessons): a readiness probe cheaper than its query reports ready too early

mlorentedev:
## What

Adds `docs/lessons/lesson-281` and its index row. Docs-only.

## Why

Filing this session's ticket batch hit a GitHub GraphQL **secondary** rate limit. Issue creation went through REST; the board writes were handed to a detached job that waits for capacity. Its readiness probe was `gh api graphql -f query='{viewer{login}}'` — one point, and being served — so the job logged `graphql available after 0s` and then failed immediately on the 1200-item project listing it was gating.

Three findings, each independently reusable:

1. **`gh api rate_limit` reported `remaining: 5000` on both counters while project queries were refused.** The secondary limit is not the counter that endpoint reports, so its `reset` timestamp is not a valid retry trigger.
2. **`gh` surfaces the refusal as `unknown owner type`.** `--owner X` resolves user-vs-org over GraphQL; when that resolution is throttled the error names a configuration cause. It sends you to check the owner and the token scopes, both of which were fine.
3. **The gap is a cost class, not a race.** A one-point query and a wide project listing are throttled apart, so polling the cheap probe faster never closes it.

The rule: a readiness probe must exercise the same path as the work it gates — same auth, same resolution steps, same cost class. The fix was replacing the probe with `gh project field-list` (same owner resolution, same project scope, small result), after which the job correctly reported unavailable and backed off.

There is also a note worth keeping in the worked example: the wide listing issued **to verify the writes** is what exhausted the budget for the follow-up job. On a point-priced API, verifying from what the write returned beats re-reading the collection.

## Verification

```
bats tests/guard-lesson-numbers-unique.bats   # 4/4 ok
bats tests/check-doc-paths.bats               # ok
bats tests/check-md-escapes.bats              # ok
```

pre-commit passed all six hooks on the commit.

## Scope

Docs-only, so outside the Discipline Gate. Not GitHub-specific in application: the same shape covers a `SELECT 1` in front of a report query, a TCP connect in front of a TLS handshake, or a token refresh in front of a scoped call.

### DIFF
--- a/docs/lessons/_index.md
+++ b/docs/lessons/_index.md
@@ -298,3 +298,4 @@ tags: [lessons, index, dotfiles]
 | [278 - A cap asserted in one unit while the payload overflowed in another](lesson-278-a-cap-asserted-in-one-unit-while-the-platform-counts-another.md) | 2026-09-05 |  |
 | [279 - A helper written to "mirror" an existing parser is the defect; the transcription bug is only how you find out](lesson-279-a-helper-written-to-mirror-a-parser-is-the-defect.md) | 2026-09-05 |  |
 | [280 - A diagnosis repeated often enough becomes indistinguishable from evidence](lesson-280-a-diagnosis-repeated-often-enough-reads-as-evidence.md) | 2026-09-06 |  |
+| [281 - A readiness probe cheaper than the work it gates reports ready while the work is still refused](lesson-281-a-cheap-readiness-probe-reports-ready-while-the-real-query-is-refused.md) | 2026-09-07 |  |
--- a/docs/lessons/lesson-281-a-cheap-readiness-probe-reports-ready-while-the-real-query-is-refused.md
+++ b/docs/lessons/lesson-281-a-cheap-readiness-probe-reports-ready-while-the-real-query-is-refused.md
@@ -1,0 +1,162 @@
+---
+id: lesson-281
+type: lesson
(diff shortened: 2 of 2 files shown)


mlorentedev:
## Review triage

|---|---|---|
| CI (12 checks) | all green — `lint`, `test`, `test-windows`, `integration`, `spec-gate`, `attestation`, `hygiene`, `changes`, `lint-powershell`, `add-to-project`, `review`, GitGuardian | nothing to apply |
| PR-Agent | reviewed: effort 1/5, no security concerns, **no major issues detected**, PR contains tests | no findings to dispose of |
| CodeRabbit | still in progress at triage time | not waited on, per the standing "one reviewer is enough" rule — PR-Agent produced substantive output and it is disclosed here as the reviewer of record |

Nothing applied, nothing ticketed, nothing declined — there were no findings. Recorded explicitly rather than left blank, because an unwritten triage is indistinguishable from nobody having looked.

Local evidence, this session:

```
bats tests/guard-lesson-numbers-unique.bats   # 1..4, all ok
bats tests/check-doc-paths.bats               # ok
bats tests/check-md-escapes.bats              # ok
dotf pr triage-queue                          # [OK] no reviewer output is awaiting a disposition
```

Docs-only change; handed over for the merge decision.

Issue #1586, opened 2026-09-08
HARNESS-135: Integrate herdr as the ADR-032 level-2 interactive agent backend


mlorentedev:
> Integrate `herdr` (https://github.com/herdrdev/herdr) as the **level-2 interactive executor
> backend** of the ADR-032 agent seam — the slot ADR-032 §2 assigned to Orca and that Orca has not
> been able to fill since 2026-08-09.

## Context

ADR-032 §2 defines a neutral executor seam — *"run role A on task T, isolated in W, return R"* —
with three implementations selected by probe:

| Implementation | When | Status today |
|---|---|---|
| `orca` | interactive machine, Orca present | **blocked** — #462, #649, #899 all OPEN since August |
| `subprocess` | the floor bootstrap guarantees | shipped (`claude -p`, `pi --print`) |
| `hive.delegate_task` | headless / CI / text-only offload | shipped |

The level-2 row has never executed anything. ADR-032 names why in its own text: *"the real limit is
that Orca is a GUI desktop application"* (#462), and #899 records that the Orca CLI wrapper is
unusable from a plain shell because the AppImage injects `--no-sandbox` into the electron-as-node
path. Both objections are properties of Orca's packaging, not of the level-2 design.

`herdr` does not have either property: single Rust binary, no Electron, headless background server,
CLI + local socket as the primary surface, attachable over SSH. It is therefore a candidate to fill
the slot ADR-032 reserved — which is the reason this is an integration ticket and not a
tool-evaluation ticket.

**Goal relevance.** This serves the standing *orchestrator agnostic binding* goal directly: it is
the executor half of the same arc as #1532 (proving the binding fires outside Claude). It does not
serve the `dotf` CLI migration drive, and it does not work against it either — see *Install
channel* below, which costs zero setup-script LOC and creates no new twin pair.

## What herdr provides (verified 2026-09-07)

Repository facts, read from the GitHub API rather than the landing page:

- Rust, Apache-2.0, created 2026-03-27, 36,225 stars, 2,657 forks, **300 open issues**.
- Latest release **v0.9.0, published 2026-09-07**. Seven releases in the last 30 days
  (`v0.8.2` 2026-08-19 → `v0.9.0` 2026-09-07, plus five `preview-*` tags).
- **v0.9.0 is a breaking release**: it removed the single-process `--no-session` mode; every TUI
  launch now attaches to a background server, and `server stop` ends the session.
[...]

PR #1587, opened 2026-09-08
fix(harness): new-ticket verified the board with a listing that truncates silently

mlorentedev:
> **This PR does not implement #1586 and does not close it.** It fixes a defect in the `new-ticket`
> skill that surfaced *while* #1586 was being filed — the skill's own verification step reported the
> freshly-created ticket as absent from the board. #1586 (herdr as the ADR-032 level-2 backend) is
> untouched here and none of its 15 acceptance criteria are in scope. Filed as a direct fix rather
> than a `BUG-*` ticket by explicit decision: the repo's meta-work rule takes the cheapest honest
> repair of review/board machinery over constructing more of it.

Regenerates the `new-ticket` skill record after the vault-side fix in `mlorentedev/knowledge@e65645e5`.

## The defect

`gh project item-list` returns items oldest-first and stops at `--limit` **without an error**. The `new-ticket` skill used it twice, with hardcoded limits, against a bitácora that now holds **3311 items** — so both call sites sampled a fraction of the board and answered confidently from it.

Both failure modes were observed in a single session on 2026-09-08:

| Call site | Limit | Sample | What it answered |
|---|---|---|---|
| Step 5, "confirm the fields landed" | 400 | 12% | `NOT FOUND ON BOARD` — for an item whose four fields had just been set successfully |
| Concurrency guard, board `ID` scan | 1000 | 30% | highest id `HARNESS-111`, while `HARNESS-120`…`134` already existed |

The second is the dangerous one: it would have proposed an id **23 numbers deep into occupied space**. Only the paginated REST title scan caught it — which is why no duplicate was actually filed.

This is the class the repo already names in `.claude/CLAUDE.md`: a read that comes back short or empty, and the short read gets used as a finding.

## The fix

- **Step 5 verifies by node id.** `$ITEM` is already in hand from step 4.2, so a single GraphQL node read is exact, cheap, and immune to board growth. Tested against #1586 before committing.
- **The live priority-distribution query is deleted, not repaired.** Reading the board whole to make it correct trips the GraphQL secondary rate limit outright (measured: `API rate limit already exceeded`), and it would burn the quota the create step needs a minute later. The rubric now carries a dated measured distribution — P2 ~88%, P3 ~7%, P1 ~5%, P0 zero — with an instruction to re-measure by hand and re-date it.
- **The concurrency guard names the paginated REST title scan as authoritative** and forbids `item-list` for the supplementary cross-repo board scan: *a truncated board scan is worse than none, because it answers.*
- One bullet added to **Common mistakes**.

`adversarial-review` carries a `generated_sha` correction only — its committed record was already stale against the vault on `main`, so `--refresh` was not idempotent. Body unchanged.

## Verification

```
./scripts/compile-harness.sh --refresh    # regenerates both records
./scripts/compile-harness.sh --refresh    # re-run: same two files, no further churn (changed=0)
```

Both new snippets were run verbatim before committing: the node-id verify returned #1586's five fields correctly; the whole-board read failed with the rate-limit error, which is what motivated deleting it rather than shipping it.

Pre-commit hooks green (secrets, dotfiles tests, bats names, doc paths, message format).

No production LOC — harness skill record only.

### DIFF
--- a/harness/skills/adversarial-review/SKILL.md
+++ b/harness/skills/adversarial-review/SKILL.md
@@ -1,7 +1,7 @@
 ---
 generated: true
 generated_from: 00_meta/skills/adversarial-review/SKILL.md
-generated_sha: 4696ee5e688ab476
+generated_sha: a43a700196f36a92
 id: adversarial-review-skill
 type: skill
 status: active
--- a/harness/skills/new-ticket/SKILL.md
+++ b/harness/skills/new-ticket/SKILL.md
@@ -1,7 +1,7 @@
 ---
 generated: true
 generated_from: 00_meta/skills/new-ticket/SKILL.md
-generated_sha: 1d1c81043049e6c9
+generated_sha: 76ce6474f5bf0aab
 id: new-ticket-skill
 type: skill
 status: active
@@ -119,15 +119,26 @@ Set the option variable for the *confirmed* Type/Priority (resolve `O_BUG`/`O_SP
 
 ### 5. Report
 
-Print the issue URL, the `AREA-NNN-slug` ID, the home repo, and the resulting board status. If self-assigned, confirm it moved to In Progress. Confirm the fields landed (the `ID` text field surfaces under the JSON key `iD`):
+Print the issue URL, the `AREA-NNN-slug` ID, the home repo, and the resulting board status. If self-assigned, confirm it moved to In Progress. Then confirm the fields landed — **by node id, never by listing the board**:
 
 ```bash
-gh project item-list "$PROJECT_NUM" --owner "$OWNER" --format json --limit 400 | python3 -c "
-import json,sys
-for i in json.load(sys.stdin)['items']:
-    if i.get('content',{}).get('number')==int('$NUM'):
-        print(i.get('status'), i.get('priority'), i.get('type'), i.get('iD'), '·', i['content']['title']); break"
+gh api graphql -f query='
+query($item: ID!) {
+  node(id: $item) {
+    ... on ProjectV2Item {
+      content { ... on Issue { number url } }
+      fieldValues(first: 20) { nodes {
+        ... on ProjectV2ItemFieldSingleSelectValue { name field { ... on ProjectV2SingleSelectField { name } } }
+        ... on ProjectV2ItemFieldTextValue  { text field { ... on ProjectV2Field { name } } }
+      } }
+    }
+  }
+}' -f item="$ITEM" --jq '.data.node
+  | {issue: .content.number,
+     fields: [.fieldValues.nodes[] | select(.field.name != null) | {(.field.name): (.name // .text)}]}'
 ```
+
+> **Never verify with `gh project item-list`.** It returns oldest-first and stops at `--limit`, silently — a ticket past the cut is reported as absent, with no error. Measured 2026-09-08: the bitácora holds **3311 items**, so the `--limit 400` this step used to carry sampled 12% of the board and answered `NOT FOUND ON BOARD` for an item whose four fields had just been set successfully. `$ITEM` is already in hand from step 4.2, and a node read is exact, cheap and immune to board growth. Same class as the `--paginate` split below and the empty-read failures in the repo's shell-compatibility table: the wrong answer arrives dressed as a finding.
 
 ## Non-interactive / autonomous mode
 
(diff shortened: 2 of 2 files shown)


mlorentedev:
## Review triage

**Checks:** all green on `4e62e58` — `lint`, `lint-powershell`, `hygiene`, `changes`, `attestation`, `spec-gate`, `review`, `integration`, `test`, `test-windows`, GitGuardian, `add-to-project`. Zero failures, zero cancelled.

**Reviewers:** Qodo pr-agent produced the substantive review. CodeRabbit was still `review in progress` at triage time and is not waited on (one reviewer is enough; disclosing which one actually reviewed).

### Dispositions

|---|---|---|---|
| 1 | Ticket compliance: linked to #1586, all 15 acceptance criteria unaddressed; *"clarification needed on whether this PR should be associated with a different ticket"* | pr-agent | **Applied.** The PR body now opens with an explicit statement that this does not implement or close #1586 — it fixes a `new-ticket` skill defect that surfaced *while* #1586 was being filed. The association pr-agent inferred was wrong, and the finding was the right way to catch that the PR named no ticket of its own. |
| 2 | Implied requirement set (the skill fix itself): 5 of 5 compliant, *"none identified"* non-compliant | pr-agent | **No action.** Confirms the diff matches what the description claims. |
| 3 | *"No relevant tests"* | pr-agent | **Declined, with reason.** The change is a Markdown skill record. Both new snippets were executed verbatim before committing — the node-id verify returned #1586's five board fields correctly, and the whole-board read failed with `API rate limit already exceeded`, which is precisely why that query was deleted rather than repaired. Executable evidence exists; a unit test does not apply to this artifact. |
| 4 | *"No security concerns"*, *"No major issues detected"* | pr-agent | **No action.** |

### Why no `BUG-*` ticket was filed for the defect

Deliberate, and stated so it is not read as an omission. The repo's amended incident→guard rule caps meta-work and prefers the cheapest honest repair of review/board machinery over constructing more of it. The defect was repaired at its source (vault `00_meta/skills/new-ticket/SKILL.md`, `mlorentedev/knowledge@e65645e5`) in the same session it was found; a ticket would have tracked work that was already done.

### Deploy state

`compile-harness.sh --deploy` has been run: the corrected skill is live on all five surfaces (`~/.claude`, `~/.config/opencode`, `~/.gemini` ×2, `~/.copilot`). Verified on the deployed copy — four `item-list` mentions remain, all of them prohibitions, and no executable `item-list` snippet survives. The deployed copy is briefly ahead of `main` until this merges.

mlorentedev:
## Review triage — round 2 (CodeRabbit)

CodeRabbit finished after the first triage was posted and left one inline finding. Pushing `8ce7038` reopens the window, so this dispositions it.

| # | Finding | Severity | Disposition |
|---|---|---|---|
| 1 | `SKILL.md:174` — *"Label the data as a sample, not the full board distribution"*: the percentages come from the first 1500 of 3311 items | 🟡 Minor | **Applied** in `8ce7038` (vault `knowledge@c3580e13`). The finding is right, and it matters more here than its severity suggests: the line led with *"The board's shape"* and qualified it afterwards, which is a partial read presented as a whole answer — the exact failure this PR exists to stop. Shipping it would have undercut the change. It now states the sample first (516 prioritised tickets among the oldest 1500 of 3311), and asks for order-of-magnitude calibration rather than distribution matching. |
| 2 | SkillSpector AS3 — *"Skill enumerates or reads other installed skills… cross-skill access is a privilege escalation"* | ⚠️ Tool warning | **Declined — false positive, verified.** The only cross-skill reference in the file is one Obsidian wiki link on line 256 (`Related skills: [[…/handoff/SKILL|handoff]], [[…/spec/SKILL|spec]]`). It is prose in a References section: nothing lists a skills directory, nothing reads another `SKILL.md`, and no command in the skill touches one. Checked by grepping every `skills/` and `SKILL.md` occurrence in the file. That line also predates this PR and is not in the diff. |

**Also on the branch:** `main` was merged in remotely (`ac40ead`, bringing #1585). My commit was rebased on top rather than force-pushed, so nothing of that merge was rewritten.

**State:** all checks were green on `4e62e58`; re-running on `8ce7038`. Deploy re-run and verified — the corrected wording is live on the deployed copy.

Nothing outstanding from either reviewer.

PR #1588, opened 2026-09-08
docs(lessons): a listing that truncates answers instead of failing

mlorentedev:
> **#1587 is already merged (`3b6a18c`).** This PR does not re-implement it — it is the knowledge follow-up the Definition of Done requires, and it adds no code.

Captures the defect fixed in #1587 as a lesson, per the Definition of Done's knowledge item.

`gh project item-list` returns items oldest-first and stops at `--limit` with no error, no truncation flag, and exit 0. The `new-ticket` skill used it twice with hardcoded limits against a **3311-item** board — sampling 12% and 30% — and both call sites answered confidently from what they had.

Filing one ticket surfaced both in the same session:

- step 5 printed `NOT FOUND ON BOARD` for an item whose four fields had just been set successfully, one command earlier, in the same script;
- an id scan at `--limit 1000` reported `HARNESS-111` as the highest existing id while `HARNESS-120`…`134` already existed — a proposal **23 numbers deep into occupied space**, caught only by the paginated REST title scan.

The lesson records why it survived (both limits were correct when written, and a truncating listing never announces that it stopped being right), what the fix was (verify by identity, not by search; delete a query you cannot make correct rather than growing its limit), and the general rule:

> If "broken" and "nothing found" print the same thing, the caller cannot tell them apart, and it will report the second one.

Same class as the empty-read rows in `.claude/CLAUDE.md`'s prohibited-pattern table and lesson 263 (`--paginate` splitting).

Docs-only. Lesson numbering and index guard green.

### DIFF
--- a/docs/lessons/_index.md
+++ b/docs/lessons/_index.md
@@ -299,3 +299,4 @@ tags: [lessons, index, dotfiles]
 | [279 - A helper written to "mirror" an existing parser is the defect; the transcription bug is only how you find out](lesson-279-a-helper-written-to-mirror-a-parser-is-the-defect.md) | 2026-09-05 |  |
 | [280 - A diagnosis repeated often enough becomes indistinguishable from evidence](lesson-280-a-diagnosis-repeated-often-enough-reads-as-evidence.md) | 2026-09-06 |  |
 | [281 - A readiness probe cheaper than the work it gates reports ready while the work is still refused](lesson-281-a-cheap-readiness-probe-reports-ready-while-the-real-query-is-refused.md) | 2026-09-07 |  |
+| [282 - A listing that truncates answers instead of failing, so the short read becomes the finding](lesson-282-a-listing-that-truncates-answers-instead-of-failing.md) | 2026-09-08 |  |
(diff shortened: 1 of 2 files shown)


mlorentedev:
## Review triage

**Checks:** all green on `5d83748`, zero failures — including the lesson numbering/index guard, which is the one that matters for this change.

**Reviewers:** Qodo pr-agent reviewed. CodeRabbit is still `review in progress`; not waited on, and disclosed rather than implied.

|---|---|---|---|
| 1 | Ticket compliance vs #1587: *"this PR is documentation only and does not fix the defect or regenerate the skill record"* | pr-agent | **No action — the finding is right and already satisfied elsewhere.** Both named requirements were delivered by #1587 itself, which merged as `3b6a18c` before this PR was opened. pr-agent resolved the `#1587` reference in the body as this PR's requirement source; #1587 is the *fix*, and this is its knowledge follow-up under Definition of Done §2 ("what was learned is written where it belongs, this session"). Body clarified so the next reader does not repeat the inference. |
| 2 | *"No relevant tests"* | pr-agent | **Declined, with reason.** A single Markdown lesson plus its index row. The repo's `docs/lessons` numbering-and-index guard is the applicable check and it passed; there is no unit under test. |
| 3 | *"No security concerns"*, *"No major issues detected"*, effort 1/5 | pr-agent | **No action.** |

Nothing outstanding.

Issue #1594, opened 2026-09-22
HARNESS-136: model limits in ai/pi/models.json drift unwatched from the provider catalog


mlorentedev:
## Problem

`ai/pi/models.json` hand-declares a context window and an output cap for every
model. The providers publish the real ones. Nothing compares the two, so the
declaration is free to drift and did: measured 2026-09-21, **all seven declared
models disagreed with the catalog**, in both directions at once.

```
nan/glm5.3-flash       1048576/65536   real 1000000/131072
nan/deepseek-v4-flash  1048576/65536   real 1000000/384000
nan/qwen3.8-flash       262144/16384   real  262144/131072
nan/qwen3.6             262144/16384   real  262144/65536
nan/mimo-v2.5          1048576/65536   real 1048576/131072
nan/gemma4              262144/16384   real  262144/32768
openrouter/deepseek/deepseek-chat
                        131072/8192    real  163840/16384
```

The two directions are different bugs. A window **above** the real limit makes
the provider reject a request this file invited — `glm5.3-flash` and
`deepseek-v4-flash` each claimed 2^20, which reads as "1M" and is 48576 tokens
more than the 1000000 granted. A cap **below** it forfeits paid-for capability
in silence: `qwen3.8-flash` shipped 16384 against a real 131072, eight times
less, on the model `qf` routes to.

The values survived review because they are plausible round powers of two and
because no check existed. This is the same class as HARNESS-067 (model *pin*
drift), one field over: that check asserts an id still resolves, and says
nothing about the limits declared beside it.

## What good looks like

- The seven declarations match what the provider publishes.
- A check reports any future disagreement, with severity following consequence:
  over-declaration fails (a request breaks), under-declaration warns (capability
  is lost quietly).
- The repository does not acquire a second hand-maintained copy of numbers
  somebody else publishes — which is what committing a catalog snapshot would be.

## Notes

Truth is verifiable from two independent sources that agree: the models.dev
catalog opencode itself routes on (`~/.cache/opencode/models.json`) and NaN's
published table at <https://nan.builders/docs/models>.

One trap worth recording: `qwen3.8-flash` is published by **both** `nan`
(262144 context) and `openrouter` (1000000). Any comparison keyed on the bare
model id reads the wrong row and reports an honest declaration as four times
too small.

PR #1595, opened 2026-09-22
fix(pi): align model limits with the provider catalog, and report future drift

mlorentedev:
Refs #1594

> Deliberately `Refs`, not `Closes`. The SDD lifecycle ends by archiving the
> spec in the PR that closes its issue, and `spec archive` requires a passing
> independent adversarial review first — reviewer != implementer. So this PR
> implements, a review follows, and a second PR archives the spec and closes
> the issue.
>
> (Worded without the keyword-plus-number adjacency on purpose: the gate's
> matcher cannot tell prose about closing from an intent to close, and this
> note tripped it once already.)

## What

`ai/pi/models.json` hand-declares a context window and an output cap per model
while the providers publish the real ones, and nothing compared the two. Measured
2026-09-21: **all seven declared models disagreed with the catalog at once.**

```
nan/glm5.3-flash       1048576/65536  -> 1000000/131072
nan/deepseek-v4-flash  1048576/65536  -> 1000000/384000
nan/qwen3.8-flash       262144/16384  ->  262144/131072
nan/qwen3.6             262144/16384  ->  262144/65536
nan/mimo-v2.5          1048576/65536  -> 1048576/131072
nan/gemma4              262144/16384  ->  262144/32768
openrouter/deepseek/deepseek-chat
                        131072/8192   ->  163840/16384
```

Two directions, two different failures. A window **above** the real limit makes
the provider reject a request this file invited — `glm5.3-flash` and
`deepseek-v4-flash` each claimed 2^20, which reads as "1M" and is 48576 tokens
more than the 1000000 granted. A cap **below** it forfeits paid-for capability in
silence: `qwen3.8-flash` shipped 16384 against a real 131072, eight times less, on
the model `qf` routes to.

The values survived because they are plausible round powers of two and nothing
looked. So this also adds the check: `dotf doctor` gains a **Model limit drift**
section comparing the declaration against opencode's models.dev cache — the same
catalog opencode itself routes on.

Severity follows consequence, as `checkModelPins` already does next door:
over-declaration **FAILS**, under-declaration **WARNS**. It never writes.

## Verification

- Values cross-checked against **two independent sources that agree**: the
  models.dev catalog and NaN's own table at <https://nan.builders/docs/models>,
  which states `qwen3.8-flash` at 262K/131K and `mimo-v2.5` at 1M/131K.
- **Proof by consequence, both directions.** The same binary reports
  `(1 checks, all ok)` against this branch and nine findings — 2 FAIL, 7 WARN —
  against `main`.
- **Mutation.** Four mutations, each killed by the test written for it and no
  other: drop the provider scoping, swap the two severities, drop the
  unpublished-limit guard, let an absent catalog PASS.
- `go build ./...`, `go vet ./...`, `GOOS=windows go vet ./...` clean;
  `go test -count=1 ./...` green; `golangci-lint run` at the pinned v2.12.2, 0
  issues; `bats tests/pi-config.bats` 18/18.
- All eight `features.json` commands executed, 8/8 pass.

## Two corrections worth reading

**Provider scoping is load-bearing.** `qwen3.8-flash` is published by both `nan`
(262144 context) and `openrouter` (1000000). The first pass matched on the bare
model id, read openrouter's row, and reported the honest `nan` declaration as
four times too small. `TestModelLimitsResolvesTheSameIdPerProvider` exists
because that happened, not because it might.

**AC1's own verification was vacuous, twice.** Written first as
`grep -qvE 'FAIL|WARN'` — which asks "is any line clean?", and half the lines in
a findings block are continuation text, so GNU grep answers yes and the guard
passes on a broken tree. It only looked right locally because `grep` in that
shell is **ugrep**, which answers the opposite: correct on the developer's
machine, vacuous in CI, with the divergence landing on the side that lets the
guard through. Negating the pipeline fixed that but not enough — asserting the
*absence* of findings still passes when the section never ran, which is what a
binary built before the check does. It now asserts the section **ran and reported
agreement**. Written up as lesson 283.

## Scope note

`glm5.2` is published by NaN and not declared here, and `glm5.3` appears in NaN's
docs but not in the catalog. Both left alone — this PR corrects what is declared
and does not change which models are offered.


mlorentedev:
## Review triage

|---|---|---|---|
| 1 | CodeRabbit, `up to 7418f` | Merge Risk 🔵 Low. One finding: **docstring coverage 78.57%** against an 80% threshold, 14 functions analysed across 3 files. | **Applied — `eb99a31`.** Seven declarations had no docstring. Closed with content rather than restatement: the four shape types exist because the upstream formats disagree — pi keys models in an **array**, models.dev keys them by id in a **map** — which is precisely why the comparison is a loop on one side and a map hit on the other, and why the lookup must be provider-scoped. The three test helpers now record why they are literals: marshalling the fixture from the production structs would let a field rename travel into the fixture, so every test would keep agreeing with a parser that had stopped reading the file. |
| 2 | CodeRabbit, same comment | "Review automatically paused — branch under active development." | **Noted, and disclosed.** The summary it produced describes the diff accurately and its Merge Risk verdict covers `7418f`, so this is a real review of the implementation, not a notice that none ran. What it has **not** seen is `eb99a31`, the docstring commit written in response to row 1. |
| 3 | pr-agent, Standalone PR Review | #1594 **Partially compliant** — with **zero** non-compliant requirements and zero items needing human verification. Three requirements listed compliant. No major issues, no security concerns, tests present. Effort 2/5. | **Nothing to apply.** Read on its own terms the ticket is fully satisfied; the "Partially" header with an empty non-compliant list is the known shape of this reviewer's output, recorded here rather than treated as a finding. |
| 4 | CI, `spec-gate` FAILURE | `SDD archive-on-merge violation: this PR closes an issue whose spec is still active (HARNESS-136, #1594)`. | **Applied — PR body, twice.** Correct and caught a real mistake. The lifecycle ends by archiving the spec in the PR that closes its issue, and `spec archive` requires a passing independent adversarial review first — reviewer ≠ implementer. So the keyword became `Refs`. The second failure was self-inflicted and worth recording: the explanatory note I added ended with the words "closes" and "#1594" adjacent, and the gate's matcher cannot tell prose *about* closing from an intent to close. `check-spec-gate.sh:202` documents that the gate has tripped over its own text twice before, so this is a third instance of a known trap, not a gate defect. Reworded; the matcher now reports no closing keyword. |

**Verification on the pushed head.** `go build ./...`, `go vet ./...` and
`GOOS=windows go vet ./...` clean; `go test -count=1 ./...` green;
`golangci-lint run` at the pinned v2.12.2, **0 issues**; `bats tests/pi-config.bats`
18/18 and `tests/guard-lesson-numbers-unique.bats` 4/4. All eight `features.json`
verification commands executed, 8/8 pass — run with GNU grep first on `PATH`,
which is the one that matters (see the note below).

**One thing worth flagging for the reviewer, because it bit this PR's own
evidence.** AC1's verification was vacuous twice before it was right. Written first
as `grep -qvE 'FAIL|WARN'` — which asks "is any line clean?", and half the lines in
a findings block are continuation text, so GNU grep answers yes and the guard passes
on a broken tree. It only looked correct locally because `grep` in that shell is
**ugrep**, which answers the opposite: correct on the developer's machine, vacuous
in CI, with the divergence landing on the side that lets the guard through.
Negating the whole pipeline fixed that but still passed against `main`, where the
check does not exist — no section, no findings, no failure. It now asserts the
section **ran and reported agreement**. Written up as lesson 283.

**Still owed, and deliberately not done here:** this PR does not archive the spec.
An independent adversarial review (reviewer ≠ implementer) runs first, then a
second PR archives HARNESS-136 and closes the issue.

Issue #1596, opened 2026-09-22
CLI-078: dotf secrets drift and reconcile — make the declared store layout and rotation policy enforceable


mlorentedev:
## Problem

Three questions about the secret store have no mechanical answer today, and all
three need the same data.

**1. Which secrets are due for rotation?** The registry declares `rotate: 90d` or
`180d` on 20 of 34 entries. `Entry.Rotate` is parsed at `registry.go:44` and
**read by no line of code**. It is dead metadata. The only live rotation signal is
`checks_pat.go`, which probes expiry for the 3 entries marked
`validate: github-token`.

**2. Which consumers are serving a stale value?** `dotf secrets sync ci` is a
one-way push, run per repo by hand. Measured 2026-09-21 via the Actions REST API
(names and dates only, no values):

| Repo | Secret | Last updated |
|---|---|---|
| dotfiles, kubelab, mlorente.dev | `BITACORA_PAT`, `NAN_API_KEY` | 2026-08-16 |
| dotfiles | `RELEASE_TOKEN` | 2026-06-27 |
| kubelab, mlorente.dev | `DOCKERHUB_TOKEN` | 2026-08-22 |
| pollex | `NAN_API_KEY` | **2026-09-06** |

`NAN_API_KEY` is declared for **13 CI consumers** and was synced to exactly one of
them. Nothing reports that.

**3. Is the store laid out the way the registry says?** The registry declares
item, field and folder for every entry, and nothing compares that declaration to
the vault. The result is visible drift: **8 of 34 entries declare no folder at
all** and sit unfoldered in the vault root, item naming follows four different
conventions at once (`zoho`, `openai-account`, `nan-api-key`, `x-twitter-api`),
and service grouping is inconsistent — `openai-account` holds ChatGPT's backup
and recovery codes together while `stripe-backup-code` sits outside
`stripe-api-key`.

All three are answerable from Bitwarden **metadata** — `revisionDate`, `folderId`,
field names — which is never a secret value.

## What good looks like

**`dotf secrets drift`** — one read, three reports:

- **age** — `revisionDate` (or a `rotated-at` field, see below) against `rotate:`
- **CI** — that timestamp against each `ci:` consumer's Actions `updated_at`
- **layout** — declared item/field/folder against the vault's actual shape

**`dotf secrets reconcile`** — prints the plan and changes nothing.
`--apply` executes. Idempotent: the second pass reports zero changes.

Both read metadata only. Neither ever prints or moves a secret value.

[...]

Issue #321, opened 2026-06-10
OPS-007: per-purpose GitHub token convention (one PAT for everything is an antipattern)


mlorentedev:
Surfaced 2026-06-09 during the OPS-002 rollout: today a single PAT serves every purpose (bitácora board writes, release-please, vault clones on agents, Hermes runtime, MCP…). One credential for everything means: any leak compromises everything, no per-purpose scoping (least privilege), rotation breaks N consumers at once, and failures are hard to attribute (the yt-metrics-cli `release-please` *Bad credentials* failure took debugging to trace to an expired shared token).

## Deliverable

A documented token convention (runbook + `sensitive/` naming) and migration of the existing consumers:

1. **Inventory** every place a GitHub token is consumed: repo Actions secrets (`BITACORA_PAT`, release tokens), agent boxes (`GITHUB_TOKEN_KNOWLEDGE` in Hermes `.env`), local age-encrypted secrets in `sensitive/`, MCP configs.
2. **Convention**: one fine-grained PAT per purpose, named `<PURPOSE>_PAT` (e.g. `BITACORA_PAT` = project+repo metadata only; `RELEASE_PAT` = contents+PR on the repos that release; `VAULT_CLONE_PAT` = read-only on knowledge), each with minimal scopes + explicit expiry, age-encrypted under `sensitive/` with one file per token.
3. **Rotation runbook**: per-token rotation steps + which consumers to update (extend `scripts/bitacora-rollout.sh`-style convergence where possible).
4. **Migrate** current consumers off the shared PAT; verify each workflow/agent still works.

## References
- OPS-002 (#258) rollout — `BITACORA_PAT` now seeded in 20 repos by `scripts/bitacora-rollout.sh` (re-running it converges a rotated value in one pass).
- yt-metrics-cli Release run 27247773369 (Bad credentials, 2026-06-10).

mlorentedev:
### Field evidence: the shared PAT silently broke board automation (2026-08-07/08)

Concrete failure of the "one PAT for everything" antipattern this ticket describes, with the failure mode being **silent board drift** rather than a visible error.

**What happened.** An interactive agent session on `kubelab` ran a few `gh project item-list` calls against the bitácora (1500 items each, heavily paginated) and exhausted the GraphQL quota. Minutes later, two issues were opened — and `add-to-project` failed on both:

```
Run actions/add-to-project@v2.0.0
##[error]Request failed due to following response errors:
[...]

PR #1597, opened 2026-09-22
feat(secrets): report where the Bitwarden store disagrees with the registry

mlorentedev:
Refs #1596, #586

> `Refs`, not a closing keyword: `spec archive` requires a passing independent
> adversarial review first (reviewer != implementer), so a later PR archives the
> spec and resolves the issue.

## What

`secrets/registry.yaml` has declared an item, a field and a folder for every
secret since ADR-028, and **nothing has ever compared that to the vault**. First
run against the live store — 185 items, 27 declared targets:

```
item-misfiled  dockerhub           is in No Folder, declared Dotfiles/apps
item-missing   github-cli-pat      declared but not in the vault (dormant declaration: …)
item-missing   github-release-pat  declared but not in the vault (dormant declaration: …)
item-missing   zoho                declared but not in the vault (dormant declaration: …)

27 declared target(s) across 16 item(s); 4 finding(s); 169 of 185 vault items unmanaged
```

Plus one the fix in the first commit removes from that list: **every declared
folder name was absent under that spelling.** The taxonomy said `apps`; the vault
holds `Dotfiles/apps`.

## Why each of those costs something

**The folder names.** `BWPut.ResolveFolder` matches by EXACT name and **creates**
on miss, and `secrets_set.go` hands it the declared value verbatim. The next
`dotf secrets set` provisioning an app-plane secret would have minted a second
folder called `apps` beside `Dotfiles/apps` and split the managed items across
both. The eleven items sitting correctly in `Dotfiles/apps` today got there by
hand.

**The missing items.** `GITHUB_PERSONAL_ACCESS_TOKEN` declared `github-cli-pat`,
never created. Invisible while the entry still read from age — and that age blob
holds a token GitHub answers **401** to, while `dotf secrets verify` reported OK
throughout, because resolving and working are different claims.

## The two decisions worth reviewing

**Dormant declarations are checked, and that is the point.** The registry writes a
`bw:` block for age-backed secrets too, as the migration target. The obvious
filter — `backend == bw` — is exactly why three absent items went unseen: a
dormant declaration resolves nothing, so it breaks nothing, right up until the
migration that needs it.

**No value can cross the projection, enforced by the type system.** `bw serve` has
no projection: `/list/object/items` answers with every item's complete plaintext.
So `itemWire` does not *name* `login.password`, a field's `value`, `card`,
`identity`, `sshKey` or `passwordHistory` — `encoding/json` discards what the
target struct does not name, so a password is dropped during decoding and never
exists as a Go value. There is no line to forget to redact and no filter that can
be reordered past. `notes` and `login.username` are the two exceptions, read and
reduced to booleans on the next line, named as such in the type's doc.

Read-only is structural: no `--fix`, no writer passed in, and a committed check
that greps the source (comments stripped) for any write call.

## Verification

- `go build` / `go vet` / `GOOS=windows go vet` clean; module green;
  `golangci-lint run` at the pinned **v2.12.2**, 0 issues
- All eight `features.json` commands executed, **8/8**
- Live run against the real store, four findings each independently confirmed
- **Five mutations, each killed** — including *a password reaching `ItemSummary`*

**Three of those five survived the first round.** The tests that kill them were
written in response, and the dormant-declaration hole is worth naming: the
original test handed `LayoutDrift` a struct with `Dormant: true` and passed
happily, while `BWDeclarations` — the walk that has to *discover* dormancy — was
never exercised at all. A fixture built past the code under test proves nothing
about it.

## Relationship to #586

#586 is the outcome — folder taxonomy, the GitHub 8-in-1 split (#321), naming
drift. This is the mechanism that measures distance from it, and the next slice
(`reconcile`) converges toward it. Neither replaces the other, and #586's first
acceptance criterion ("all managed items live under `Dotfiles/{apps,infra}`;
registry `bw.folder` matches") is now mechanically checkable for the first time.

## Not in this PR

Rotation age and CI staleness (both in #1596, both need this inventory first);
`reconcile`; and creating the three missing items — that is a vault write, and it
belongs to a converger rather than a hand-run command.


mlorentedev:
## Review triage

|---|---|---|---|
| 1 | CodeRabbit | **"Review limit reached — you've used all free OSS reviews for now."** | **No review ran, and this is a disclosure, not a finding.** A notice *about* a review is not a review: it names no file, no line and no claim. This PR has been reviewed by **pr-agent only**. Recording it explicitly because green checks would otherwise imply a second reviewer that never looked. |
| 2 | pr-agent, Standalone PR Review | **#1596 Partially compliant.** Seven requirements compliant — read-only `drift`, folder/item/field reporting, non-zero exit, dormant declarations, per-problem dedupe, and the value-free projection with its test. Three non-compliant: *rotation age*, *CI staleness*, *`reconcile`*. Zero items needing human verification. | **Declined — the three "non-compliant" items are this PR's stated scope boundary.** All three are named in `## Out of scope` in the PR body and in `proposal.md`, with the reason: rotation age and CI staleness both consume the inventory this PR builds, and `reconcile` writes to a credential store, which must not land in the same change as the report that produces its plan. pr-agent grades the diff against the whole issue rather than the slice, which is the known shape of its output here; read per-slice, #1596's first tranche is fully satisfied. |
| 3 | pr-agent, same review, ticket #586 | Partially compliant — the registry and its validation moved to `Dotfiles/apps` / `Dotfiles/infra`. | **Correct, and deliberately partial.** #586 is the *outcome* (taxonomy, the GitHub 8-in-1 split, naming drift); this PR is the *mechanism* that measures distance from it. The one AC it advances — "registry `bw.folder` matches" — is now mechanically checkable for the first time rather than asserted. |
| 4 | CI | `test` and `test-windows` still running at the time of writing; every settled check green, including the required `review-attestation` status. | **Nothing to do yet.** Locally on this head: `go build`, `go vet` and `GOOS=windows go vet` clean, `go test -count=1 ./...` green across the module, `golangci-lint run` at the pinned v2.12.2 with 0 issues, and all eight `features.json` commands green. |

**Worth the reviewer's attention, because it is where this PR was weakest.**
Five mutations now die; **three of them survived the first round**. The one worth
naming: the dormant-declaration test handed `LayoutDrift` a struct with
`Dormant: true` and passed happily, while `BWDeclarations` — the walk that has to
*discover* dormancy, and the function the bug was actually in — was never
exercised at all. A fixture built past the code under test proves nothing about
it. `TestBWDeclarationsIncludesDormantBlocksOnAgeBackedSecrets` now goes through
`ParseRegistry`.

**And one thing this PR asserts that only a human can really check.** The claim is
that no secret value can cross the projection, and the mechanism is that
`itemWire` does not *name* `login.password`, a field's `value`, `card`, `identity`,
`sshKey` or `passwordHistory` — so `encoding/json` discards them during decoding.
`TestDecodeItemsCannotCarryAValue` plants distinctive values and asserts none
survives a round-trip through `json.Marshal` of the result. If that argument has a
hole, it is the thing to find here.

**Still owed, deliberately not done here:** this PR does not archive the spec. An
independent adversarial review (reviewer != implementer) runs first, then a second
PR archives CLI-078 and resolves #1596.

PR #1600, opened 2026-09-22
feat(secrets): drift fixes, and converge the Bitwarden layout with dotf secrets reconcile

mlorentedev:
## Summary

This PR now carries **two specs**. #1601 (`dotf secrets reconcile`, CLI-080) was stacked on it and merged into this branch, so it lands on `main` with this PR.

### 1. CLI-078 `dotf secrets drift`: post-merge fixes and two review rounds

| Defect | Found by |
|---|---|
| unfoldered items read as filed in "No Folder" | live run + review round 1 |
| file-exposed secrets (8) never declared | live run + review round 1 (Blocker) |
| `folder: ""` read as "must be unfoldered" | live run |
| live canary would create a stray `apps` folder | review round 1 (Major) |
| AC8 exit code unpinned; registry id never printed | review round 1 |
| **duplicate item names judged by an arbitrary one** → new `item-ambiguous` finding | review round 2 (Blocker) |
| `field-missing` not deduped | review round 2 |

Both review rounds are committed verbatim. Dispositions are in `specs/CLI-078-secrets-layout-drift/verification.md`.

### 2. CLI-080 `dotf secrets reconcile` (from #1601, plus `retire`)

It converges the vault's layout toward the registry: plan by default, `--apply` explicit, and the command itself re-plans after applying and fails unless the result is empty. `bw.from` declares where a value comes from (the Terraform `moved` idea). `retire` removes the source once the copy holds **the same value**, compared in memory. Ambiguous items and unknown finding kinds block. It also adds `dotf secrets sync ci [SECRET_NAME...]`, which scopes an upload to the named secrets.

**Applied to the live vault (authorized):**
- `dockerhub` moved to `Dotfiles/apps`
- `github-cli-pat` and `github-release-pat` created from the `GitHub` item, then their source fields retired. **Each token is now in exactly one place.**
- Both tokens verified by consequence: HTTP 200
- `RELEASE_TOKEN` pushed to this repo's Actions secrets; `release-please` went from `Bad credentials` to only the account rate limit
- The `from:` records were deleted from the registry after they were satisfied

Also lesson 284: a mutation "killed" by a compile error proves nothing.

## Merge order and what stays broken until then

`main` still maps `GITHUB_PERSONAL_ACCESS_TOKEN`/`RELEASE_TOKEN` to the revoked age blob. Local `dotf secrets run` keeps reading it until this merges, the setup is deployed, and `~/.local/bin/dotf` is rebuilt.

## Pending

[...]

PR #1601, opened 2026-09-22
feat(secrets): converge the Bitwarden layout toward the registry with dotf secrets reconcile

mlorentedev:
## Summary

Adds `dotf secrets reconcile`, which converges the Bitwarden store's **layout** toward `secrets/registry.yaml`. It prints a plan by default; `--apply` performs it. First use: split the two GitHub tokens out of the shared `GitHub` item, which unblocks `release-please` (failing on `main` with `Bad credentials`).

**Stacked on #1600**: this plans from the inventory #1600 fixes. The base is `fix/secrets-drift-inventory`, so the diff shows only this change. It gets rebased onto `main` once #1600 merges.

## What it does

| Operation | When |
|---|---|
| `create-folder` | a declared folder no folder carries |
| `move-item` | an item outside its declared folder (never for an item declared with no folder) |
| `create-item` / `add-field` | absent, **and** the declaration names a source in the new `bw.from` |

- **It never makes up a value.** An absent item with no `bw.from` is **blocked** and the plan names the remedy. If the declaration is dormant (age-backed), it's **deferred** to `migrate` and blocks nothing.
- **It never overwrites a value.** Changing a value is the job of `set`/`rotate`.
- **Values never reach the output.** Copies happen in memory. Tests plant a distinctive value and assert it appears in neither the plan, the report nor the errors.
- **Apply runs in two phases.** It reads every source before the first write, so an empty source leaves the store untouched.
- **Idempotence is checked by the command itself.** `--apply` syncs, re-plans and **fails unless the second plan is empty**.

`bw.from` works like Terraform's `moved {}`: a migration record that is declared, reviewed and applied by the tool. Once satisfied, reconcile reports it as removable.

## First use

`GITHUB_PERSONAL_ACCESS_TOKEN` and `RELEASE_TOKEN` flip to `backend: bw` and each declares its source. Live plan (read-only) against the real vault:

```
~ move-item      dockerhub                (no folder) -> Dotfiles/apps
+ create-item    github-cli-pat           field "GITHUB_PERSONAL_ACCESS_TOKEN" in Dotfiles/apps, copied from GitHub/"Personal Access Token"
+ create-item    github-release-pat       field "RELEASE_TOKEN" in Dotfiles/apps, copied from GitHub/"release-token"
- deferred       zoho                     … -> dotf secrets migrate ZOHO_APP_PASSWORDS
- deferred       zoho                     … -> dotf secrets migrate ZOHO_RECOVERY_CODE

Plan: 3 to apply, 0 blocked, 2 deferred.
```

## Order of operations

1. Review this PR.
2. `dotf secrets reconcile --apply` **from this branch, before merging**. Every operation is additive, and the registry is read from the invoking worktree.
3. Probe both PATs by consequence (HTTP 200).
4. `dotf secrets sync` for `ci:mlorentedev/dotfiles`, so the repo secret `RELEASE_TOKEN` gets the live token and `release-please` recovers.
5. Merge.

## Test plan

- [x] `go build`, `go vet`, `GOOS=windows go vet`, `go test ./...`; pinned `golangci-lint`: 0 issues
- [x] `features.json` AC1–AC8: 8/8 pass
- [x] 16 mutations: 15 killed, 1 equivalent (reasoning in `verification.md`)
- [x] Idempotence proven against the daemon fake: plan → apply → plan = empty. The fake now lists full items and hides new folders until sync, the way the real daemon does.
- [x] Live plan against the real vault (above)
- [ ] AC9: live `--apply`, then clean re-plan, PAT probes and CI sync (steps 2–4)

## Out of scope

- Renaming existing fields to the env-var name: needs its own transition design (separate slice).
- Deleting the source fields from `GitHub`: that comes after the consumers are verified.
- `rotated-at` rotation age: the decision is recorded in the proposal.

Refs #1596


mlorentedev:
## Review triage

|---|---|---|
| CodeRabbit | **"Review skipped"**: auto-review is disabled for bases other than `main` (this PR is stacked on #1600). A notice, **not a review** | disclosed. It gets reviewed once it's rebased onto `main` after #1600 merges |
| pr-agent | no security concerns, no major issues; 8 compliant items | nothing to apply |
| pr-agent compliance, #1596 | rotation age and CI staleness not addressed | **declined, as scope**: out of scope in `proposal.md`. The rotation-age decision (`rotated-at`, `revisionDate` as a lower bound) is recorded there for that slice |
| pr-agent, human verification | AC9: live apply and the CI sync for `RELEASE_TOKEN` | **pending the operator**: steps 2–4 in the PR body |

Reviewed by pr-agent only. CodeRabbit has not reviewed. The adversarial review runs after AC9, before the spec archives.

PR #1602, opened 2026-09-22
fix(doctor): fail on an unreadable model declaration, and stop counting nothing as a match

### LINKED ISSUE #1594: HARNESS-136: model limits in ai/pi/models.json drift unwatched from the provider catalog

mlorentedev:
## Summary

Applies the findings of the HARNESS-136 adversarial review, round 1 (`FAIL`, `nan/qwen3.8-flash`), to the `dotf doctor` model-limit check merged in #1595. The review is committed verbatim, and its full disposition is in `specs/HARNESS-136-model-limit-drift/verification.md`.

| Finding | Fix |
|---|---|
| **Major (REAL):** an `ai/pi/models.json` that exists but can't be read WARNed, so doctor exited 0 on a declaration nobody had read (AC7's unreadable half) | now FAILs, like an unparseable one; test with a mode-000 file |
| **Major:** a catalog row with no published limit counted as compared, and printed "1 models match" | only a model with at least one comparable limit counts; otherwise SKIP |
| Minor: the scoping example named openrouter (which keys `qwen/qwen3.8-flash`) | corrected with the real collisions, measured against the cache |
| Minor: "no checkout → SKIP" wasn't pinned | test |
| Minor: `checkModelLimits` was 84 lines at CC≈17 | split into `readDeclaration` / `readCatalog` / `compareModel` |
| Minor: `features.json` all `pending` | every command executed, result recorded |
| Speculative: empty `HOME` | declined: a package-wide convention, not introduced here |

The repo-vs-cache asymmetry is now explicit where each file is read. The declaration is repo content, so failing to read it is a **FAIL**. The catalog is a per-machine cache, so failing to read it is a **WARN**.

## Test plan

- [x] 3 new tests, each **red with its fix reverted** (3 mutations, 3 killed)
- [x] `go build`, `go vet`, `GOOS=windows go vet`, `go test ./...`; pinned `golangci-lint`: 0 issues
- [x] `features.json` 8/8 passing
- [x] Adversarial review, round 2 (`nan/deepseek-v4-flash`): **PASS**. Both round-1 Majors confirmed by the reviewer's own mutations. The six Minors and one Question it left are ticketed as #1604 rather than applied after the verdict

## Spec

Archives `HARNESS-136-model-limit-drift` after the passing independent review, so this PR closes it.

Closes #1594


mlorentedev:
## Review triage

|---|---|---|
| pr-agent | #1594 "PR Code Verified"; no security concerns, no major issues | nothing to apply |
| Adversarial HARNESS-136 r1 | FAIL (2 Majors, 5 Minors) | all applied except one speculative Minor (declined with a reason) |
| Adversarial HARNESS-136 r2 | **PASS** (`nan/deepseek-v4-flash`) | its 6 Minors + 1 Question are ticketed in #1604 rather than applied after the verdict |
| CodeRabbit | not reviewed (quota) | disclosed |

Note: the archive commit (`68de48e`, spec archived + `Closes #1594`) is local. The pre-push spec-gate reads this PR over GraphQL and the account quota is exhausted, so it gets pushed after the reset. It will not be pushed with `--no-verify`.

mlorentedev:
## Review triage

Supersedes the earlier triage (pr-agent re-reviewed after the archive commit `68de48e`).

|---|---|---|
| pr-agent | #1594 **fully compliant**; no security concerns, no major issues | nothing to apply |
| pr-agent, #1604 | 5 items "non-compliant" | **declined, as scope**: #1604 is the follow-up ticket for the gaps the PASS review left. They are deliberately not in this PR, because applying them after the verdict would ship unreviewed code. pr-agent reads #1604 as a requirement of this PR because the PR mentions it |
| CI | all green, both check-runs and commit statuses (`review-attestation`, CodeRabbit) | — |
| Adversarial HARNESS-136 r2 | PASS | spec archived in this PR; closes #1594 |

PR #1612, opened 2026-09-23
fix(doctor): gate PAT resolution on the readability probe, not GET /status

### LINKED ISSUE #1611: BUG: dotf doctor's PAT expiry check reads bw serve items right after GET /status, the #988 trigger
## Symptom

`dotf doctor` reports, **intermittently**, `[PAT expiry] WARN <pat> is declared but did not resolve (... bw serve returned no parseable envelope (HTTP 500, 21 bytes): invalid character 'I' ...)`. Over three consecutive runs against an unchanged, unlocked vault, the section produced 0, 1 and 2 such warnings. In the same session, `dotf secrets verify` passed 35/35 on every one of 4 runs.

## Root cause

This is #988 (BUG-082) at a call site that its fix missed. `GET /status` poisons `bw serve`'s item-read path for a short window: every `GET /object/item/<id>` inside it returns HTTP 500. This is upstream bitwarden/clients#20951, still present in CLI 2026.5.0. #988 fixed the CLI's read path: `SelectBWBackend` probes with `GET /list/object/folders` (`probeReadable`) and never calls `/status`.

Doctor never adopted that. `resolvePATToken` (`cli/internal/doctor/checks_pat.go`) gates each bw-backed PAT on `sys.BWServeStatus()`, which calls `GET /status`, and then resolves the item straight away. Every read it authorises lands inside the poison window. The WARN's advice ("`dotf secrets verify` isolates the cause") is misleading here: verify never calls `/status`, so it always passes.

`BWFallbackReader.Field` (`cli/internal/secrets/bwserve.go`) has the identical status-then-read shape. No production caller uses it; only its own tests do.

## Measured (2026-09-23, CLI 2026.5.0, unlocked daemon, status codes only)

| sequence | result |
|---|---|
| `GET /object/item` ×150, polled every 0.2s | 150/150 200 |
| `GET /status` then `GET /object/item`, ×40 | **27/40 HTTP 500** |
| same, with 0.1s between them | 60/60 200 |
| `GET /status` then `GET /list/object/items?search=` | 20/20 200 |
| after a 500, retry after 0.15s | 16/16 200 |

The window measured here is < 0.1s, against about 0.5s in #988's measurement. It is time-based, so moving the status call out of the loop does not fix it.

## Fix

- Gate doctor's PAT resolution on the same readability probe the CLI uses (`/list/object/folders`), not on `/status`.
- Delete `BWFallbackReader`, which is dead code carrying the same defect.
- Guard: a test fails if the PAT check consults the status seam before resolving.

Refs #988

mlorentedev:
## Summary

`dotf doctor`'s PAT expiry check gated each bw-backed PAT on `GET /status` and then resolved it straight away. That is the #988 trigger: a status call poisons `bw serve`'s item reads for a short window (bitwarden/clients#20951, still present in CLI 2026.5.0). #988's fix moved the CLI's read path to the `/list/object/folders` probe (`SelectBWBackend` → `probeReadable`), but doctor never adopted it. As a result `[PAT expiry]` warned intermittently with `HTTP 500, 21 bytes: invalid character 'I'` against a vault that `dotf secrets verify` read cleanly every time.

- `secrets.BWServeClient.Readable()` exports the existing probe as three string states (`absent` / `refused` / `ready`). The probe itself is unchanged.
- `doctor.System.BWServeReadable` is a new seam. `resolvePATToken` gates on it and never calls `BWServeStatus`. A refusal SKIPs with the daemon's own reason rather than asserting "locked".
- `BWFallbackReader` is removed: it had the identical status-then-read shape and no production caller. Only its own three tests used it.

The three other `BWServeStatus` callers in doctor (daemon section, launchability, cache age) do not read an item afterwards, so they are left alone.

## Evidence

Live, against the unlocked daemon, recording status codes only:

| | before | after |
|---|---|---|
| `GET /status` then `GET /object/item` (curl, ×40) | 27/40 HTTP 500 | — |
| `[PAT expiry]` HTTP 500 warnings over repeated `dotf doctor` runs | 1, 0, 1, 2 across 4 runs | **0 across 5 runs**; `(3 checks, all ok)` |

- Guard: `TestCheckPATExpiry_NeverAsksStatusBeforeReading` fails (`t.Fatal`) if the check touches the status seam. It was confirmed red against the unfixed check, on a tree that builds.
- `TestReadable_AnswersWithoutCallingStatus` counts `/status` hits through the existing `countingHandler` fake.
- Mutation, on trees that build (lesson 284), 3/3 killed: `Readable` calls `Status()` first; refused mapped to ready; absent mapped to refused.
- `go build`, `go vet`, `GOOS=windows go vet`, `go test ./...` all green. `golangci-lint` v2.12.2 (pinned): 0 issues.

## SDD skip rationale

This is a bug fix whose root cause is already characterised and documented in #988 and #1611. It adds no new contract: `Readable` exports an existing, tested probe under the same policy `SelectBWBackend` already follows, and the rest is one call site plus deletion of dead code. The production diff is +46/−34. A spec would restate #988's measurements.

Fixes #1611
Refs #988


mlorentedev:
## Review triage

Head `0b20f6b`. CI all green (including both Windows legs and spec-gate with `skip-sdd`).

|---|---|---|
| coderabbitai | Full review of all 8 files: "No actionable comments", merge risk minimal | Nothing to apply |
| pr-agent | #1611 compliant on all three requirements | Nothing to apply |
| pr-agent | #988 "non-compliant": capture the raw body | **Declined.** #988 was closed on the opposite decision, still in the code: `BWServeClient.call` reports status code and byte count and never the body, because a 200 that failed to parse can still carry vault material. |
| pr-agent | #988 "non-compliant": bounded retry with backoff on 5xx | **Declined.** #988 found the deterministic cause (`/status` poisons item reads), and the fix is never to trigger it. Retrying would mask that trigger for the next caller that reintroduces it; the guard test makes the reintroduction fail instead. |
| pr-agent | #988 "non-compliant": verify should tell absent from failed | **Declined, already true.** `verify` prints `N ok, N missing, N failed` and classifies each row, as the live run in #1611 shows. |

The three #988 items are that closed issue's original proposals; this PR only `Refs` it.

PR #1614, opened 2026-09-23
chore(secrets): refresh the DR escrow after the reconcile items were created

mlorentedev:
## Summary

Refreshes the age-encrypted Bitwarden DR escrow after `dotf secrets reconcile` created `github-cli-pat` and `github-release-pat` (#1600). `dotf doctor` reported "DR escrow no longer describes the vault: 2 item(s) added". This is the same routine as #1599.

- `sensitive/dr/bitwarden-export.age`: a new export, age-encrypted (header `age-encryption.org/v1`).
- `sensitive/dr/escrow-manifest.json`: `count` 185 → 187, plus a new `max_revision` and `digest`. Metadata only; no values.

Written by `dotf secrets backup` in the main checkout. It was moved here unchanged (`cmp` identical), and the checkout was restored.

## Verification

After the backup, before the move, `dotf doctor`'s `[Disaster recovery]` section no longer reported the stale-escrow WARN. Once this merges, `setup-linux.sh` refreshes the deploy mirror, which clears the transient `Repo↔deploy-dir drift` on these two files.

PR #1621, opened 2026-09-23
fix(secrets): converge copy+retire in one apply, and check what drift used to skip

mlorentedev:
## Summary

#1600 merged with two FAIL adversarial reviews still undispositioned: CLI-080 round 1 (`agy/gemini-3.1-pro-high`) and CLI-078 round 3 (`nan/deepseek-v4-flash`), both committed verbatim in `specs/*/review.md`. This PR dispositions every finding in each spec's `verification.md`. Re-reviews follow, one at a time.

### CLI-080 `reconcile`
- **Blocker (REAL): a declaration that copies and retires failed the first `--apply`.** `--apply` now runs up to two passes (`applyUntilConverged`). A retire is planned only once its copy exists, and it compares the copy as the store holds it after a sync. So a pass can unlock a retire but never perform one, and that invariant is kept. The second pass admits only retires: anything else left after a pass (a store ignoring writes, a flap) fails where it appears. The bound is in the loop header (see lesson 286 for why).
- **Minor, graded REAL here: `ResolveFolder` created a folder without syncing.** The daemon does not list a folder created since its last sync, so the next resolve of the same name made a **second folder with the same name**. The test double already modelled that cache, and the new subtest was red with two `Dotfiles/infra` folders.

### CLI-078 `drift`
- **Major (REAL): a declaration naming no field was counted and never checked.** It is now a `field-missing` finding, and `hasField("")` returns false, like the reader. Live, drift went from 1 finding to 2, exposing `AGE-SECRET-KEY-PERSONAL`. The registry now declares `field: notes` (verified value-free: the item is a secure note), and drift is back to 1 (zoho, deferred). `verify` 35/35.
- **Major (THEORETICAL): AC7 covered one decode error path.** Both are covered now, and the review's mutation M13 is killed.
- **Minor: AC4 "matches `fieldFromItem` exactly" was false for empty values, and its test never ran the reader.** `TestLayoutDriftAgreesWithTheReaderOnEveryItemShape` runs one raw item through both sides, across 13 shapes. The two designed divergences (an empty password or custom-field value, which the projection never decodes) are declared in the table. AC4 is reworded.
- **Minor: the misfiled dedupe was keyed on the item.** Fixed at the source: the registry refuses one item declared in two folders.
- **Minor: `LayoutDrift` was 71 lines at CC≈18.** It is now a `driftWalk` with one method per question, and `LayoutDrift` is 10 lines.
- **Minor: stale "reconcile = CLI-078" comments, unexplained counts, and `apps/<item>` in `docs/secrets-inventory.md`.** All fixed.

### Contract changes
Contract edits, which invalidate `reviewed_sha` on purpose:
- CLI-078 AC4 and AC5; CLI-080 AC10;
- `features.json` f4/f5 (CLI-078) and f10 (CLI-080) now run the new tests.

### Lesson 286
While verifying this, a mutation that deleted the loop's bound (it was in the body back then) turned `cmd.test` into a 9 GB process. The OOM killer took the terminal host (Orca) four times. The rule: bound loops in the header, and run mutation tests in a memory-capped `systemd-run --user --scope`, one at a time, restoring from `git checkout HEAD`.

## Evidence
- `go build`, `go vet`, `GOOS=windows go vet`, `go test ./...` all green (run in a 3 GB-capped scope). `golangci-lint` v2.12.2 (pinned): 0 issues.
- `features.json`: CLI-078 8/8; CLI-080 9/10. **f9 is the live `reconcile` plan and could not run: the `bw serve` daemon is down** (connection refused on :8087, most likely lost with the OOM events). It needs `dotf secrets unlock` by the operator, and it is re-checked before either spec archives.
- Every new test was red before its fix, on a tree that builds.
- Mutations, 15 killed out of 15, each under a memory cap:
  - the review's own survivors (M13, the inverted notes dispatch, `HasNotes: true`);
  - the empty-field flip;
  - the three pass-loop mutants and the folder sync;
  - the folder rule, and the detail text.

Refs #1596. The issue stays open for rotation, and both specs archive only after a PASS re-review.


mlorentedev:
## Review triage

Head `5fd4de6` (rebased onto `main` after #1614/#1616/#1619; the only conflict was `docs/lessons/_index.md`, resolved by keeping both 285 and 286). CI all green.

|---|---|---|
| pr-agent | Compliant: two-pass convergence, field-less declaration, `ResolveFolder` sync, the new tests, and the spec verification updates | Nothing to apply |
| pr-agent | "Non-compliant": the bw `folder:` guard, the `drift`/`reconcile` implementations and the folder taxonomy are not in this PR | **Declined, out of scope by design.** Those are #1596's earlier deliverables: #1597 and #1600 shipped them and `checkBWFolder` enforces the taxonomy. This PR only dispositions the two FAIL reviews #1600 merged with. |
| pr-agent | Verify: live convergence (CLI-080 AC9) | **Verified in-session**, after the operator's `dotf secrets unlock`: `go run ./cmd/dotf secrets reconcile` from this branch → `Plan: 0 to apply, 0 blocked, 2 deferred` (zoho, deferred to `migrate`). That is CLI-080 f9, so the features now pass 18/18. |
| pr-agent | Verify: the live registry complies with `checkBWFolder` | **Verified in-session.** The live `secrets/registry.yaml` parses clean with this branch's validator, including the new one-folder-per-item rule. `dotf secrets verify` from the branch → 35/35. |

Independent adversarial re-review in flight: CLI-080 round 2 on `nan/glm5.3-flash` (a different family from round 1's `agy/gemini-3.1-pro-high`). CLI-078 round 4 follows it, not in parallel (#1107).

